PR #4, opened 2026-05-13
ci: add Octo issue/PR feed workflows

lml2468:
Add real-time GitHub event notification to Octo IM channels via GitHub Actions.

- `octo-issue-feed.yml`: pushes Issue events (opened/closed/reopened/labeled) to Octo 🛠️ issue-feed and the repo's project channel
- `octo-pr-feed.yml`: pushes PR events (opened/closed/merged/etc.) to Octo 🛠️ pr-feed and the repo's project channel

### DIFF
--- a/.github/workflows/octo-issue-feed.yml
+++ b/.github/workflows/octo-issue-feed.yml
@@ -1,0 +1,20 @@
+name: Octo Issue Feed
+
+on:
+  issues:
+    types: [opened, closed, reopened, labeled]
+
+jobs:
+  notify:
+    uses: Mininglamp-OSS/.github/.github/workflows/octo-issue-feed.yml@main
+    with:
+      repo_name: ${{ github.event.repository.name }}
+      issue_number: ${{ github.event.issue.number }}
+      issue_title: ${{ github.event.issue.title }}
+      issue_url: ${{ github.event.issue.html_url }}
+      issue_author: ${{ github.event.issue.user.login }}
+      issue_labels: ${{ toJson(github.event.issue.labels.*.name) }}
+      event_action: ${{ github.event.action }}
+      project_group_id: 8d68238743664ba1a49eec75b4e296bd
+    secrets:
+      OCTO_BOT_TOKEN: ${{ secrets.OCTO_BOT_TOKEN }}
--- a/.github/workflows/octo-pr-feed.yml
+++ b/.github/workflows/octo-pr-feed.yml
@@ -1,0 +1,23 @@
+name: Octo PR Feed
+
+on:
+  pull_request:
+    types: [opened, closed, reopened, ready_for_review, review_requested]
+
+jobs:
+  notify:
+    uses: Mininglamp-OSS/.github/.github/workflows/octo-pr-feed.yml@main
+    with:
+      repo_name: ${{ github.event.repository.name }}
+      pr_number: ${{ github.event.pull_request.number }}
+      pr_title: ${{ github.event.pull_request.title }}
+      pr_url: ${{ github.event.pull_request.html_url }}
+      pr_author: ${{ github.event.pull_request.user.login }}
+      pr_merged: ${{ github.event.pull_request.merged }}
+      pr_additions: ${{ github.event.pull_request.additions }}
+      pr_deletions: ${{ github.event.pull_request.deletions }}
+      pr_changed_files: ${{ github.event.pull_request.changed_files }}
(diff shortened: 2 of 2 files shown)


yujiawei (commented):
# Code Review — PR #4 (octo-smart-summary)

Two small caller workflows that delegate to `Mininglamp-OSS/.github/.github/workflows/octo-{issue,pr}-feed.yml@main`. Overall this is well-structured and safe — the delegation pattern is right, and the reusable workflow uses `env:` + `os.environ` in Python, so the standard "PR title contains backticks → shell injection" concern doesn't apply here. Already merged; comments are non-blocking.

## Verification

- ✅ Both reusable workflows exist in `Mininglamp-OSS/.github` with matching `workflow_call` input contracts (verified `repo_name`, `issue_*`/`pr_*`, `event_action`, `project_group_id`, secret `OCTO_BOT_TOKEN`).
- ✅ Trigger types are sensible. PR feed deliberately omits `synchronize` (avoids a ping on every push); issue feed includes `labeled` (see note below).
- ✅ No script-injection vector from caller. `pr_title` / `issue_title` flow from caller `with:` → reusable `inputs` → reusable `env:` → Python `os.environ`. None of those steps interpolate into a shell.
- ✅ `merged` vs `closed` is correctly disambiguated downstream by passing `pr_merged: ${{ github.event.pull_request.merged }}`.

## Findings

### P2 — `@main` reference for the reusable workflow (supply-chain hardening)

`uses: Mininglamp-OSS/.github/.github/workflows/octo-{issue,pr}-feed.yml@main`

Pinning to a moving branch means any commit landed on `Mininglamp-OSS/.github`'s `main` is auto-applied to every repo that consumes these feeds, with `OCTO_BOT_TOKEN` in scope. For an internal org workflow this is acceptable, but consider pinning to a tag (e.g. `@v1`) or full SHA so feed changes go through an explicit bump. GitHub's hardening guide recommends SHA pinning for third-party reusable workflows; internal ones are a judgment call.

### P2 — `pull_request` (not `pull_request_target`) means no notifications for fork PRs

`on: pull_request` running from a forked PR will not receive `secrets.OCTO_BOT_TOKEN`, so the reusable workflow's send step will fail. If `octo-smart-summary` ever accepts external contributions, the PR feed will go silent for those PRs. Two options:

1. Accept the gap (simplest — internal repos rarely take fork PRs).
2. Switch to `pull_request_target` *only* for this notify job. That gives access to secrets but runs against the base branch's workflow definition; since this workflow doesn't check out PR code or run user-controlled scripts, it's safe to do — but document the choice.

No action needed if fork PRs are out of scope.

### P2 — `labeled` on issues will be chatty

`octo-issue-feed.yml` listens for `labeled`. Every label change (including programmatic relabeling, triage flows, or label-bot churn) fires a separate notification with the same issue title. If the IM channel starts feeling spammy after a week, drop `labeled` from the trigger list, or filter inside the reusable workflow on a specific label set.

### P3 — `project_group_id` is a hardcoded 32-char magic constant per repo

```yaml
project_group_id: 8d68238743664ba1a49eec75b4e296bd
```

Every onboarded repo will copy-paste these two files and remember to swap this ID. A typo silently routes notifications to the wrong channel (or a deleted channel — silent failure). Consider:

- Move it to a repo-level Actions variable (`vars.OCTO_PROJECT_GROUP_ID`) — onboarding becomes "set one variable" instead of "edit two YAML files".
- Or expose a tiny org-level "register repo" workflow that writes the variable.

Not blocking, but the cost of fixing this grows with each repo onboarded.

### P3 — No top-level `permissions:` block

Both workflows inherit the default `GITHUB_TOKEN` permissions of the repo. The reusable workflow doesn't actually use `GITHUB_TOKEN`, so least-privilege would be:

```yaml
jobs:
  notify:
    permissions: {}
    uses: ...
```

Pure hygiene — closes the surface in case the reusable workflow ever grows a step that does use the token.

### P3 — No `concurrency:` group

Rapid label flips or quick open→close cycles can race; messages may arrive out of order. For a notify-only feed this is usually fine, but a per-issue/PR concurrency group with `cancel-in-progress: false` would serialize per-resource if ordering ever matters.

## Summary

LGTM with optional polish. The caller PR is small, well-structured, and the security model is sound because the reusable workflow consumes inputs through `env:` rather than direct interpolation. The suggestions above are about long-term hygiene (SHA-pinning, var-ifying the project ID, scoping permissions) and noise control (`labeled`), not correctness.

PR #5, opened 2026-05-13
fix: use pull_request_target for fork secret access

lml2468:
### Problem
Workflows triggered by `pull_request` from **forked repositories** do not receive repository secrets. This means `OCTO_BOT_TOKEN` is unavailable, and every external contributor's PR triggers a failed workflow run.

### Fix
Changed trigger from `pull_request` → `pull_request_target`.

This is safe because:
- The workflow only reads event payload metadata (PR title, author, URL)
- No PR code is checked out
- `pull_request_target` runs in the context of the base branch, giving access to secrets

Applied to all 8 repos.

PR #6, opened 2026-05-13
ci: add auto-add-to-project caller (delegates to org reusable)

lml2468:
## What

Deploys a lightweight caller for the org-level `auto-add-to-project` reusable workflow.

Every newly opened Issue or PR in this repo will be automatically added to [Octo Project Board #2](https://github.com/orgs/Mininglamp-OSS/projects/2).

## How it works

```
issues.opened / pull_request_target.opened
    → this caller
    → Mininglamp-OSS/.github / auto-add-to-project.yml (reusable)
    → actions/add-to-project@v1.0.2
    → Project Board #2
```

## Merge order

**Depends on** `Mininglamp-OSS/.github#4` (the reusable workflow) being merged to `main` first. This PR is safe to merge any time after that.

### DIFF
--- a/.github/workflows/auto-add-to-project.yml
+++ b/.github/workflows/auto-add-to-project.yml
@@ -1,0 +1,17 @@
+# Caller: delegates to the org-level reusable workflow.
+# Adds every newly opened Issue and PR to Octo Project Board #2.
+name: Auto-add to Octo Board
+
+on:
+  issues:
+    types: [opened]
+  pull_request_target:
+    types: [opened]
+
+permissions: {}
+
+jobs:
+  add:
+    uses: Mininglamp-OSS/.github/.github/workflows/auto-add-to-project.yml@main
+    secrets:
+      PROJECT_TOKEN: ${{ secrets.PROJECT_TOKEN }}


Jerry-Xin (changes requested):
Summary: The PR is relevant to this repository, but the caller currently delegates to a workflow that is not declared as reusable, so this automation will fail instead of adding new issues/PRs to the project.

🔴 Blocking

- 🔴 Critical — `.github/workflows/auto-add-to-project.yml:15`
  The referenced target workflow must expose `on: workflow_call` to be callable via `jobs.<job_id>.uses`. I checked the current org workflow on `Mininglamp-OSS/.github@main`, and it defines normal `issues` / `pull_request_target` triggers but no `workflow_call`. As written, this caller will not be able to invoke it. Fix the org reusable workflow first, then point this caller at a ref that contains `workflow_call` and accepts `PROJECT_TOKEN`.

💬 Non-blocking

- 🔵 Suggestion — `.github/workflows/auto-add-to-project.yml:15`
  Because this workflow runs on `pull_request_target` and passes `PROJECT_TOKEN`, consider pinning the reusable workflow to a reviewed commit SHA or protected version tag instead of `@main`. This repo already does that for `.github/workflows/octo-pr-feed.yml:11`, which is a safer pattern for secret-bearing automation.

✅ Highlights

- `permissions: {}` is appropriate for this caller because it does not need the repository `GITHUB_TOKEN`.
- The workflow scope is narrow and aligned with the stated project-board automation goal.

yujiawei (commented):
# Code Review — PR #6 (octo-smart-summary)

## Summary

Small, well-scoped change: a 17-line caller workflow that delegates new issue/PR registration to the org-level reusable workflow at `Mininglamp-OSS/.github/.github/workflows/auto-add-to-project.yml`. The caller carries no logic of its own, relies on the reusable workflow to enforce permissions and pin third-party actions, and passes `PROJECT_TOKEN` through as a secret. Verdict: **COMMENTED** — no blockers from this file in isolation, but the upstream dependency is not yet merged, so a merge today will produce a failing run.

## What's good

- **Empty `permissions: {}` at the caller level.** The caller does no work itself; the `GITHUB_TOKEN` it gets from GitHub doesn't need any scopes. This is the right least-privilege posture.
- **`pull_request_target` is used safely.** No `actions/checkout`, no execution of PR-supplied code, no untrusted input flowing into a shell — so the usual "fork can steal secrets via `pull_request_target`" attack surface doesn't apply here.
- **Narrow trigger surface.** Only `issues.opened` and `pull_request_target.opened` — not `reopened`, `edited`, `labeled`, etc. Keeps the project-board insertions to one per item.
- **Delegation pattern is clean.** Pinning of `actions/add-to-project` to a SHA, the `permissions` block on the inner job, and the `PROJECT_TOKEN` requirement all live in the reusable workflow (verified against `Mininglamp-OSS/.github#4`). The caller stays minimal, which is the whole point of a reusable workflow.

## Findings

### P1 — Upstream dependency is still open

The PR description correctly notes the dependency on `Mininglamp-OSS/.github#4` (the PR that introduces `workflow_call:` on the reusable workflow). At the time of this review that PR is still **OPEN, not merged**. If this caller is merged first, the next opened issue/PR will trigger a workflow run that resolves `Mininglamp-OSS/.github/.github/workflows/auto-add-to-project.yml@main` to a version that does not yet declare `workflow_call`, and the run will fail with a startup error.

This is acknowledged in the PR body, so the recommendation is procedural: do not merge until `Mininglamp-OSS/.github#4` is merged to `main`. Worth confirming in the merge order checklist before clicking the green button.

### P2 — Reusable workflow reference is `@main`, not a tag or SHA

`.github/workflows/auto-add-to-project.yml:14`:

```yaml
uses: Mininglamp-OSS/.github/.github/workflows/auto-add-to-project.yml@main
```

For a same-org reusable workflow this is acceptable — the trust boundary is internal — but it does mean any future change to that file (intentional or not) is picked up immediately by every caller. Two hardening options to consider:

1. Tag releases in `Mininglamp-OSS/.github` (e.g. `auto-add-to-project-v1`) and reference the tag.
2. Pin to a commit SHA on the org reusable workflow.

Not blocking; logging it as a future tightening if the org adopts SHA-pinning more broadly.

### P2 — No filtering on bot-authored events

Every opened issue and every opened PR (including Dependabot, Renovate, automated security PRs, etc.) is added to Project Board #2. If the board is meant to track human work, the noise from bot PRs may make it less useful. If the intent is "everything visible," ignore this comment.

If filtering is desired later, it can be added at the caller level with an `if:` on the job (e.g. `if: github.actor != 'dependabot[bot]'`) without touching the reusable workflow.

### P3 — No `concurrency` block

`actions/add-to-project` is idempotent (re-adding an existing item is a no-op), so this is cosmetic — but a flurry of opens from a script or migration could spawn parallel runs. Adding:

```yaml
concurrency:
  group: auto-add-${{ github.event.issue.number || github.event.pull_request.number }}
  cancel-in-progress: false
```

would make the workflow's behavior under burst load explicit. Optional; safe to skip.

## Cross-file verification

- `.github/workflows/` already contains `issue-welcome.yml`, `octo-issue-feed.yml`, `octo-pr-feed.yml`, `pr-merged-done.yml`. None of them target Project Board #2 in the same way, so there's no overlap and no double-add risk.
- The reusable workflow at `Mininglamp-OSS/.github` (PR #4) declares `secrets.PROJECT_TOKEN` as `required: true` and uses `permissions: {}` at the workflow level — consistent with what this caller passes in. Wiring is correct.

## Recommendation

Merge order: land `Mininglamp-OSS/.github#4` first, then merge this PR. The PR body already documents this; just confirm at merge time. No code changes required in this file.

lml2468 (commented):
## Review by 齐静春 (qijingchun)

**Verdict: No code issues. Would APPROVE but `gh` CLI shares the PR author's GitHub account.**

**Analysis:**

The caller workflow is clean and minimal:
- `permissions: {}` — zero `GITHUB_TOKEN` permissions ✅ (lesson learned from PR #5 discussion)
- `pull_request_target` for fork PR support — correct, same safe pattern as PR #5 (no checkout, metadata-only)
- `issues.opened` + `pull_request_target.opened` — matches exactly what the reusable workflow expects
- Secret forwarding: `PROJECT_TOKEN` matches the `workflow_call.secrets` declaration in the upstream PR

**On Allen's blocker (reusable workflow not ready):**

Allen is factually correct — `Mininglamp-OSS/.github/.github/workflows/auto-add-to-project.yml@main` currently lacks `on: workflow_call`, so calling it via `uses:` would fail today.

However, I disagree this is a code defect in PR #6. The PR description explicitly documents the merge order:

> **Depends on** `Mininglamp-OSS/.github#4` (the reusable workflow) being merged to `main` first.

I verified `Mininglamp-OSS/.github#4`: it adds `on: workflow_call` with `PROJECT_TOKEN` secret, adds `permissions: {}`, and pins `actions/add-to-project` to SHA `244f685b` (v1.0.2). All correct. Status: still OPEN.

**Recommendation:** This PR is merge-ready **once** `Mininglamp-OSS/.github#4` lands on `main`. The code itself has no issues. Suggest adding a GitHub PR dependency label or a merge-queue check to prevent accidental early merge.

**Verification:**
- Read the current org workflow on `main` — confirmed missing `workflow_call`
- Read upstream PR #4 diff — confirmed it adds `workflow_call` with matching secret interface
- Confirmed secret name alignment: caller passes `PROJECT_TOKEN`, reusable declares `PROJECT_TOKEN: required: true`

PR #7, opened 2026-05-14
ci: improve GitHub Actions workflows for octo-smart-summary

lml2468:
### A4: Add CI workflow for Go 1.21 project

Added `.github/workflows/ci.yml` with build, test, vet, and lint jobs:
- Triggers on push/PR to `main`
- Go version: 1.21
- `build`: `CGO_ENABLED=0 go build -v ./...`
- `test` (after build): `go test -race -count=1 -timeout 5m ./...`
- `vet`: `go vet ./...`
- `lint`: golangci-lint v1.64.5 with `--timeout=5m`

### D: Add `permissions: {}` to `pr-merged-done.yml`

Adds explicit `permissions: {}` block between `on:` and `jobs:` to clarify that `GITHUB_TOKEN` is not used; only `PROJECT_TOKEN` is passed explicitly.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,66 @@
+name: CI
+
+on:
+  pull_request:
+    branches: [main]
+  push:
+    branches: [main]
+
+permissions:
+  contents: read
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
+env:
+  GO_VERSION: '1.21'
+
+jobs:
+  build:
+    name: Build
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-go@v5
+        with:
+          go-version: ${{ env.GO_VERSION }}
+      - name: Build
+        run: CGO_ENABLED=0 go build -v ./...
+
+  test:
+    name: Test
+    runs-on: ubuntu-latest
+    needs: build
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-go@v5
+        with:
+          go-version: ${{ env.GO_VERSION }}
+      - name: Test
+        run: go test -race -count=1 -timeout 5m ./...
+
+  vet:
+    name: Vet
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-go@v5
+        with:
+          go-version: ${{ env.GO_VERSION }}
+      - name: Vet
+        run: go vet ./...
+
+  lint:
+    name: Lint
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-go@v5
+        with:
+          go-version: ${{ env.GO_VERSION }}
+      - name: golangci-lint
+        uses: golangci/golangci-lint-action@v6
+        with:
+          version: v1.64.5
+          args: --timeout=5m
--- a/.github/workflows/pr-merged-done.yml
+++ b/.github/workflows/pr-merged-done.yml
@@ -17,6 +17,8 @@ on:
   pull_request_target:
     types: [closed]
 
+permissions: {}  # GITHUB_TOKEN not used; PROJECT_TOKEN is passed explicitly
+
 jobs:
   move-to-done:
     if: github.event.pull_request.merged == true && github.event.pull_request.base.ref == github.event.repository.default_branch


yujiawei (approved):
# Code Review — PR #7 (octo-smart-summary)

## Summary

This PR introduces a new CI workflow (`.github/workflows/ci.yml`) running build / test / vet / lint for the Go module, and adds an explicit `permissions: {}` block to `pr-merged-done.yml`. The change is small (+68 / -0), well-scoped, and the workflow is structured sensibly.

**Verdict: APPROVED** with a few optional, low-priority suggestions.

## 1. Validation

| Item | Status | Evidence |
| --- | --- | --- |
| Go version matches `go.mod` (`go 1.21`) | ✅ | `ci.yml` L17 `GO_VERSION: '1.21'` vs `go.mod` L3 `go 1.21` |
| Workflow uses minimal token scope | ✅ | `ci.yml` L9-10 `permissions: contents: read` |
| Concurrency cancels superseded runs | ✅ | `ci.yml` L12-14 — keyed on PR number with fallback to ref |
| Action versions are current | ✅ | `actions/checkout@v4`, `actions/setup-go@v5`, `golangci/golangci-lint-action@v6` |
| Test invocation is correct | ✅ | `ci.yml` L37 `go test -race -count=1 -timeout 5m ./...` — race detector + cache disabled + bounded duration |
| Trigger surface is appropriate | ✅ | PR + push to `main` only; no `workflow_dispatch` leaked |
| `permissions: {}` consistent with reusable workflow | ✅ | The org reusable workflow at `Mininglamp-OSS/.github/.github/workflows/pr-merged-done.yml` already declares `permissions: {}` and authenticates exclusively via `PROJECT_TOKEN` (passed as a secret), so the caller-side `permissions: {}` is consistent and harmless. |

## 2. Findings

### P2 — Optional / nice-to-have

**P2-1. The caller-side `permissions: {}` in `pr-merged-done.yml` is documentation, not enforcement.**
The reusable workflow being called already sets `permissions: {}` on itself, and reusable workflows compute their effective permissions from the **callee's** declaration (capped by the caller's). The new line therefore does not change runtime behavior. The inline comment (`# GITHUB_TOKEN not used; PROJECT_TOKEN is passed explicitly`) carries the real value — it documents intent for future readers. Worth keeping; no change needed, but the PR description could be slightly more accurate ("documents intent" rather than implying a security hardening).

**P2-2. `test` job depends on `build`, but `vet` / `lint` do not.**
`ci.yml` L31 `needs: build` makes test wait on a successful build (fail-fast, fine). However `vet` (L41) and `lint` (L51) start in parallel without `needs`. They will each independently compile the module, which is more or less inevitable for those tools, so this is a defensible choice. Mentioning it for awareness; no action required. If CI minutes become a concern, consider collapsing into a single job with sequential steps and a single Go install.

**P2-3. No `.golangci.yml` in repo — golangci-lint v1.64.5 will run with the default linter set.**
Default linters in golangci-lint v1.64.x are: `errcheck`, `gosimple`, `govet`, `ineffassign`, `staticcheck`, `unused`. That is a reasonable baseline, but pinning the linter set explicitly via a checked-in `.golangci.yml` (even if it just enables the same defaults) makes the lint contract reproducible across local/CI runs and resilient to upstream default changes. Out of scope for this PR; consider as a follow-up.

**P2-4. CI commands are not sourced from the existing `Makefile`.**
The repo already has a `Makefile` (`build`, `test` targets). The CI uses inline `go build` / `go test` invocations with extra flags (`CGO_ENABLED=0`, `-race`, `-timeout`). The CI's flags are intentionally stricter than the Makefile's, so direct reuse is not desirable, but this means contributors running `make test` locally exercise a different invocation than CI. Optional cleanup: add `make ci-test` / `make ci-build` targets that match CI exactly, then have CI shell out to them. Out of scope for this PR.

**P2-5. `lint` and `vet` overlap.**
`go vet` is included in golangci-lint's default `govet` linter, so the dedicated `vet` job is redundant with `lint`. Keeping both is defensive (vet runs even if lint fails to install/run), but it is not free. Optional simplification.

### P1 / P0 — None.

No security issues, no correctness bugs, no missing safeguards on the `pull_request_target` trigger (it correctly does not check out PR code, and the additional `permissions: {}` makes that even more explicit).

## 3. Recommendations

1. Land as-is; the issues above are all P2 / optional.
2. Follow-up: add a checked-in `.golangci.yml` to pin the linter set explicitly.
3. Follow-up: align `Makefile` test/build invocations with CI flags (or document the divergence).

## 4. Additional Observations

- `pull_request_target` is the correct trigger for `pr-merged-done.yml` because the workflow needs org-level secrets (`PROJECT_TOKEN`) and runs against the merged PR's metadata only — no untrusted code is checked out. The defense-in-depth `permissions: {}` reinforces that posture.
- The CI workflow does not include a coverage upload, integration tests, or matrix testing across Go versions. None of these are required for this PR's scope ("bootstrap basic CI"), but they are reasonable next steps.

Jerry-Xin (approved):
Summary: The PR is in scope for this repository and adds appropriate CI coverage for the Go project without introducing security or workflow blockers.

🔴 Blocking  
None.

💬 Non-blocking  
🔵 Suggestion: Consider adding `go test -v` in `.github/workflows/ci.yml:41` if maintainers want package/test names visible in CI logs. The current command is functionally fine.

✅ Highlights  
- `.github/workflows/ci.yml:9` correctly limits default token permissions to `contents: read`.
- `.github/workflows/ci.yml:29`, `.github/workflows/ci.yml:41`, and `.github/workflows/ci.yml:52` cover build, race tests, and vet for `./...`.
- `.github/workflows/pr-merged-done.yml:20` is consistent with the reusable workflow design: `GITHUB_TOKEN` is not needed, and `PROJECT_TOKEN` is passed explicitly.
- Verified locally: `CGO_ENABLED=0 go build -v ./...`, `go vet ./...`, `go test -race -count=1 -timeout 5m ./...`, and `git diff --check` all pass.

lml2468 (commented):
I cannot complete the code review while the required CI gate is red.

Blocking items:
1. CI gate: the `Lint` check failed in the `CI` workflow. Please fix the lint failure or rerun CI successfully before this PR is reviewed further.

Non-blocking notes:
None. I did not read the diff because the review process requires stopping on red CI.

Highlights:
Build, Test, Vet, and notify checks passed; the PR is currently mergeable.

PR #9, opened 2026-05-14
ci: add auto-add-to-project caller workflow

lml2468:
## Summary

Add `auto-add-to-project.yml` caller workflow: automatically adds new issues and PRs to [Octo Board (Project #2)](https://github.com/orgs/Mininglamp-OSS/projects/2) when they are opened.

Calls the org-level reusable workflow at `Mininglamp-OSS/.github/.github/workflows/auto-add-to-project.yml@main`.

## Dependency

Requires [Mininglamp-OSS/.github#12](https://github.com/Mininglamp-OSS/.github/pull/12) to be merged first.

PR #11, opened 2026-05-15
chore(workflows): pin reusable workflow refs to @v1

lml2468:
## Summary

Pin all reusable workflow references to `@v1` for stable, version-locked behavior.

**Before:** Mixed `@main` and hardcoded commit SHAs
**After:** All callers use `Mininglamp-OSS/.github/...@v1`

## Why

The `.github` repo has been tagged `v1` as the stable baseline. Using a version tag
ensures callers get a predictable, reviewed version of the shared workflows.

No caller-side logic changes — this is a reference update only.

### DIFF
--- a/.github/workflows/auto-add-to-project.yml
+++ b/.github/workflows/auto-add-to-project.yml
@@ -11,6 +11,6 @@ permissions: {}
 
 jobs:
   add-to-project:
-    uses: Mininglamp-OSS/.github/.github/workflows/auto-add-to-project.yml@239d0b47cfe1947d5e6de826e795feac4ed90562
+    uses: Mininglamp-OSS/.github/.github/workflows/auto-add-to-project.yml@v1
     secrets:
       PROJECT_TOKEN: ${{ secrets.PROJECT_TOKEN }}
--- a/.github/workflows/issue-welcome.yml
+++ b/.github/workflows/issue-welcome.yml
@@ -8,4 +8,4 @@ jobs:
   welcome:
     permissions:
       issues: write
-    uses: Mininglamp-OSS/.github/.github/workflows/issue-welcome.yml@main
+    uses: Mininglamp-OSS/.github/.github/workflows/issue-welcome.yml@v1
--- a/.github/workflows/octo-issue-feed.yml
+++ b/.github/workflows/octo-issue-feed.yml
@@ -6,7 +6,7 @@ on:
 
 jobs:
   notify:
-    uses: Mininglamp-OSS/.github/.github/workflows/octo-issue-feed.yml@main
+    uses: Mininglamp-OSS/.github/.github/workflows/octo-issue-feed.yml@v1
     with:
       repo_name: ${{ github.event.repository.name }}
       issue_number: ${{ github.event.issue.number }}
--- a/.github/workflows/octo-pr-feed.yml
+++ b/.github/workflows/octo-pr-feed.yml
@@ -8,7 +8,7 @@ permissions: {}  # GITHUB_TOKEN not used; lock down to minimum
 
 jobs:
   notify:
-    uses: Mininglamp-OSS/.github/.github/workflows/octo-pr-feed.yml@c6c51348fccac99161166dce1b52dec1063aeea2
+    uses: Mininglamp-OSS/.github/.github/workflows/octo-pr-feed.yml@v1
     with:
       repo_name: ${{ github.event.repository.name }}
       pr_number: ${{ github.event.pull_request.number }}


yujiawei (approved):
# Code Review — PR #11 (octo-smart-summary)

## Verdict
Approved. Small, well-motivated change. The PR pins all four reusable-workflow callers in this repo to `Mininglamp-OSS/.github/...@v1`, replacing a mix of `@main` and pinned commit SHAs.

## What I verified

| File | Before | After | Status |
|---|---|---|---|
| `.github/workflows/auto-add-to-project.yml` | `@239d0b47…` | `@v1` | ✅ |
| `.github/workflows/issue-welcome.yml` | `@main` | `@v1` | ✅ |
| `.github/workflows/octo-issue-feed.yml` | `@main` | `@v1` | ✅ |
| `.github/workflows/octo-pr-feed.yml` | `@c6c51348…` | `@v1` | ✅ |

Confirmed against `Mininglamp-OSS/.github`:
- The `v1` annotated tag exists and resolves to commit `1aab7b30…` (tagger date 2026-05-15).
- All four referenced workflow files are present at that tag.
- For `auto-add-to-project.yml`, the file at the previously-pinned commit `239d0b47…` and at `v1` is **byte-identical** (blob `e9365d57…`). No behavior change.
- For the two files that were on `@main`, `v1` currently points at the same commit content, so this is a no-op today and a stability improvement going forward.

## Notes (non-blocking)

1. **`octo-pr-feed.yml` upstream refactor.** The previously pinned commit `c6c51348…` and the `v1` commit have non-trivial differences in the reusable workflow itself: the inline Python notifier was replaced with a checked-out helper script (`_shared/scripts/octo_pr_feed_notify.py`) loaded via `actions/checkout@…` at `job.workflow_sha`, and the upstream now adds `permissions: contents: read` and `timeout-minutes: 5`. The PR body says "No caller-side logic changes — this is a reference update only", which is accurate for the caller files in *this* repo but the **runtime behavior of the PR notifier will execute a different (newer) upstream implementation**. The refactor itself looks safe and is an improvement (least-privilege permissions, scripts version-locked to the workflow via `job.workflow_sha`), but reviewers and operators should be aware that bumping to `@v1` is not strictly equivalent to staying on the old SHA. A quick post-merge sanity check that the Octo IM PR feed still posts as expected would be worthwhile.

2. **Mutable tag vs. immutable SHA.** `@v1` is a movable reference; anyone with write access to `Mininglamp-OSS/.github` can re-point the tag to a different commit. GitHub's hardening guide recommends pinning third-party actions/workflows to a full commit SHA. For a first-party, org-owned shared-workflows repo this is an acceptable trade-off and matches the stated intent ("stable, version-locked behavior"), but it does relax the supply-chain guarantee that the previous SHA pins gave. If stricter immutability is ever desired in the future, consider creating an immutable `v1.0.0` tag (or signing the tag) in the upstream `.github` repo, or have callers pin to the SHA that `v1` currently resolves to.

3. **Consistency across the org.** Worth confirming that the same `@v1` migration is being rolled out uniformly across the other Mininglamp-OSS caller repos so that the org doesn't end up with a long tail of stale `@main` / `@<sha>` references that drift away from the documented baseline.

## Risk
Very low. Workflow-only change, four single-line edits, behavior is materially preserved (and tightened, in the case of the pr-feed workflow).

Jerry-Xin (approved):
One-line summary: This PR is project-relevant and safely standardizes the shared GitHub Actions workflow references to `@v1`.

🔴 Blocking  
None.

💬 Non-blocking  
🔵 Suggestion: Because `@v1` is a movable tag, consider protecting the release/tagging process in `Mininglamp-OSS/.github` if immutability is required for supply-chain guarantees. This is not a blocker for this PR’s stated goal.

✅ Highlights  
Project relevance gate: Pass. The changed workflow callers are part of this repository’s CI/automation surface.

The four reusable workflow references are now consistent: `.github/workflows/auto-add-to-project.yml:14`, `.github/workflows/issue-welcome.yml:11`, `.github/workflows/octo-issue-feed.yml:9`, and `.github/workflows/octo-pr-feed.yml:11`.

I verified the `v1` tag exists in the shared `.github` repository and that the tagged reusable workflows expose the inputs/secrets these callers use. No application tests are needed for this ref-only workflow change; `git diff --check` also passed.

lml2468 (commented):
Verdict: APPROVED (submitted as COMMENT due to shared GitHub account).

I reviewed the full diff: all 4 changed files are ref-only updates to the reusable workflow `uses:` line. I also verified `Mininglamp-OSS/.github@v1` exists as an annotated tag resolving to `1aab7b304de30a61f1b1e950c66c72bd558a97c7` (current upstream `main`) and all 4 referenced reusable workflows exist at that tag.

No blocking findings.

P2 note: `v1` is still a mutable tag. Given the same org trust boundary, this is acceptable here, but tag protection/signing or a ruleset covering tags would strengthen the supply-chain guarantee.

Info: `octo-pr-feed` does pull upstream changes between the old SHA and `v1` (inline Python moved to the shared script). The caller interface is unchanged and behavior should be equivalent; worth a smoke test after merge.

Positive: this resolves the mixed ref convention flagged in PR #9 by unifying the caller workflows on `@v1`.

PR #15, opened 2026-05-18
ci: build and publish multi-arch Docker images to Docker Hub

an9xyz:
## Summary

Add a hardened `docker-publish.yml` workflow that publishes both `api` and `worker` images as a pair to Docker Hub on every `v*` tag push. Same security model and defenses as the workflows in [`octo-server#72`](https://github.com/Mininglamp-OSS/octo-server/pull/72) and [`octo-matter#18`](https://github.com/Mininglamp-OSS/octo-matter/pull/18), adapted for this repo's two-image layout.

## Images published

- `mininglamposs/octo-smart-summary-api` (from `Dockerfile.api`)
- `mininglamposs/octo-smart-summary-worker` (from `Dockerfile.worker`)

Both are multi-arch (`linux/amd64` + `linux/arm64`), produced from the same source revision in a single workflow run.

## Architecture

- **2 × 2 build matrix** (images × archs) on native runners (`ubuntu-latest` + `ubuntu-24.04-arm`) — no QEMU
- **Push-by-digest** per arch → **per-image merge** into multi-arch manifest list
- 4 parallel build jobs → 2 merge jobs

## Hardening (same baseline as sibling repos)

- All third-party actions pinned to commit SHAs
- **No-secret `validate` job** runs first; `build` and `merge` declare `needs: validate`
  - Tag push: strict semver per semver.org §9 (with leading `v`)
  - `workflow_dispatch`: Docker tag grammar + reserved-name + semver-alias-shape + non-default-branch `test-*` namespace
- **Command-injection-safe shell**: every external input via `env:`, `imagetools` args built as bash arrays
- **`docker-hub-publish` GitHub Environment** gate (already configured in repo settings):
  - Required reviewer: `Mininglamp-OSS/maintainers` team
  - Deployment ref allowlist: `main` branch + `v*` tags
- **Per-image per-arch GHA cache scope** (`docker-publish-IMAGE-ARCH`)
- `provenance: false` (no `unknown/unknown` manifest entries)
- Concurrency block documents the GitHub 1-pending-run queue-depth tradeoff

## Decisions made

- **Docker Hub only**, not GHCR — matches the OSS publishing strategy in sibling repos
- **Single workflow with matrix**, not two separate workflows — same source, same release, shared approval gate
- **One concurrency group** covers both images — they release together, no need for independent locks

## Prerequisites

- Org-level secrets `DOCKERHUB_USERNAME` / `DOCKERHUB_TOKEN` (already provisioned for `Mininglamp-OSS`)
- Repo environment `docker-hub-publish` with required-reviewers + ref allowlist (configured)

## Test plan (run after merge to main)

> Testing on this PR branch is intentionally blocked by the environment policy. All scenarios must run from `main` after merge.

- [ ] **Env gate** — From `main`, dispatch with `tag=test-env-gate-$(date +%s)`. Run pauses at `build` jobs with "Waiting for review"; proceeds after a maintainer approves.
- [ ] **Both images published** — Inspect both `mininglamposs/octo-smart-summary-api:<tag>` and `mininglamposs/octo-smart-summary-worker:<tag>`; each shows both `linux/amd64` and `linux/arm64`, no `unknown/unknown`.
- [ ] **Semver preflight rejects junk** — temporarily push `vfoo` (or `v01.2.3`); validate fails before any Docker Hub login. Delete tag after.
- [ ] **Prerelease skip-:latest** — push `v0.0.1-rc1`; confirm only `:0.0.1-rc1` on both images (no `:latest`/`:0`/`:0.0`).
- [ ] **Branch dispatch rejected** — try to dispatch from any non-main branch; environment policy should block.
- [ ] **Reserved-name guard** — dispatch input `latest` (or `1.2`); validate should fail.

### DIFF
--- a/.github/workflows/docker-publish.yml
+++ b/.github/workflows/docker-publish.yml
@@ -1,0 +1,278 @@
+name: Docker Publish
+
+# Publishes the api + worker images as a pair to Docker Hub on every v* tag
+# push. Both images come out of the same source revision, so they share one
+# run, one approval gate, and one concurrency lock.
+
+on:
+  push:
+    tags:
+      - 'v*'
+  workflow_dispatch:
+    inputs:
+      tag:
+        description: 'Image tag to publish. Must match Docker tag grammar ^[a-zA-Z0-9_][a-zA-Z0-9_.-]{0,127}$, must NOT look like a semver alias (1, 1.2, v1.2.3, ...), and from non-default branches must be prefixed with "test-".'
+        required: true
+
+permissions:
+  contents: read
+
+concurrency:
+  # Serialize ALL publishes to this repo's images so two close-together v*
+  # tags can't race on the shared mutable tags (:latest, :MAJOR,
+  # :MAJOR.MINOR). One group covers both api and worker because a single
+  # tag push releases them together.
+  #
+  # KNOWN TRADEOFF: GitHub Actions concurrency holds at most ONE pending run
+  # per group; a newer pending run replaces the older one even with
+  # `cancel-in-progress: false`. So if you push v1.0.0 (running), v1.0.1
+  # (queued), v1.0.2 (queued) in quick succession, v1.0.1 will be canceled
+  # by v1.0.2 and never publish. Workaround: wait for each release to finish
+  # publishing before pushing the next tag.
+  #
+  # NOTE: workflow-level concurrency.group only accepts github/inputs/vars
+  # contexts (not env), so the prefix is hard-coded here.
+  group: docker-publish-mininglamposs-octo-smart-summary
+  cancel-in-progress: false
+
+jobs:
+  validate:
+    # No-secret preflight: rejects bad inputs/refs BEFORE any job that holds
+    # Docker Hub credentials runs. Catches command-injection attempts,
+    # invalid Docker tag grammar, semver-alias clobbering, and manual
+    # publishes from unreviewed refs — all before any registry write.
+    name: Validate inputs and ref
+    runs-on: ubuntu-latest
+    steps:
+      - name: Validate tag push ref
+        if: github.event_name == 'push'
+        env:
+          REF: ${{ github.ref }}
+        run: |
+          set -euo pipefail
+          # Require strict semver per semver.org §9 (with mandatory leading 'v')
+          # so metadata-action's type=semver actually matches. A loose regex
+          # would let near-semver refs (e.g. 'vfoo', 'v01.2.3', 'v1.2.3-01',
+          # 'v1.2.3-..', 'v1.2.3+build..meta') reach the secret-bearing build
+          # job and push orphan digests before metadata-action quietly emits
+          # no tags — defeating the no-secret preflight intent.
+          #
+          # The regex below is the official semver.org regex (BRE/ERE form),
+          # prefixed with a literal 'v'. It rejects:
+          #   - leading zeros in numeric identifiers (v01.2.3, v1.2.3-01)
+          #   - empty dot-separated identifiers (v1.2.3-.. , v1.2.3+a..b)
+          #   - characters outside [0-9A-Za-z.+-]
+          tag="${REF#refs/tags/}"
+          semver='^v(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)(-((0|[1-9][0-9]*|[0-9]*[a-zA-Z-][0-9a-zA-Z-]*)(\.(0|[1-9][0-9]*|[0-9]*[a-zA-Z-][0-9a-zA-Z-]*))*))?(\+([0-9a-zA-Z-]+(\.[0-9a-zA-Z-]+)*))?$'
+          if ! printf '%s' "$tag" | grep -Eq "$semver"; then
+            echo "::error::tag '$tag' is not strict semver per semver.org (vMAJOR.MINOR.PATCH[-PRERELEASE][+BUILD]); refusing to publish"
+            exit 1
+          fi
+          echo "OK: $REF"
+
+      - name: Validate workflow_dispatch input and ref
+        if: github.event_name == 'workflow_dispatch'
+        env:
+          INPUT_TAG: ${{ github.event.inputs.tag }}
+          REF: ${{ github.ref }}
+          DEFAULT_BRANCH: ${{ github.event.repository.default_branch }}
+        run: |
+          set -euo pipefail
+
+          # Docker image tag grammar (also blocks shell metacharacters as a
+          # side effect — only [A-Za-z0-9_.-] are allowed and length <= 128).
+          if ! printf '%s' "$INPUT_TAG" | grep -Eq '^[a-zA-Z0-9_][a-zA-Z0-9_.-]{0,127}$'; then
+            echo "::error::tag '$INPUT_TAG' violates Docker tag grammar ^[a-zA-Z0-9_][a-zA-Z0-9_.-]{0,127}$"
+            exit 1
+          fi
+
+          # Reserved auto-managed names — manual dispatch must not clobber them.
+          case "$INPUT_TAG" in
+            latest|main|master)
+              echo "::error::tag '$INPUT_TAG' is reserved/auto-managed; pick a different name"
+              exit 1
+              ;;
+          esac
+
+          # Semver-alias shapes (e.g. 1, 1.2, 1.2.3, v1.2.3) are reserved for
+          # the auto-managed major / major.minor / major.minor.patch / version
+          # aliases populated by tag-push events. A manual dispatch with
+          # these shapes would race with / overwrite the next real release.
+          if printf '%s' "$INPUT_TAG" | grep -Eq '^v?[0-9]+(\.[0-9]+){0,2}([+-].+)?$'; then
+            echo "::error::tag '$INPUT_TAG' looks like a semver alias; manual publishes must use a non-semver name (or prefix with 'test-')"
+            exit 1
+          fi
+
+          # Restrict manual publishes from non-default branches to the
+          # 'test-' namespace so feature-branch experimentation cannot land
+          # in the official release tag space. Read the default branch from
+          # repository metadata so a future rename doesn't silently weaken
+          # this guard.
+          DEFAULT_REF="refs/heads/${DEFAULT_BRANCH}"
+          if [ "$REF" != "$DEFAULT_REF" ]; then
+            case "$INPUT_TAG" in
+              test-*) echo "OK: non-default ref '$REF' with test- prefix" ;;
+              *)
+                echo "::error::manual publishes from '$REF' must use a 'test-*' tag (got '$INPUT_TAG'); merge to ${DEFAULT_BRANCH} for official tags"
+                exit 1
+                ;;
+            esac
+          fi
+
+          echo "OK: dispatch from $REF with tag '$INPUT_TAG'"
+
+  build:
+    # Matrix dimension: 2 images × 2 architectures = 4 parallel build jobs.
+    # Each emits a single-arch digest-only image to Docker Hub; the merge
+    # job assembles the multi-arch manifest per image.
+    name: Build ${{ matrix.image }} ${{ matrix.platform }}
+    runs-on: ${{ matrix.runner }}
+    needs: validate
+    # `docker-hub-publish` environment gates the Docker Hub credentials:
+    #   - required reviewers (Mininglamp-OSS/maintainers team)
+    #   - deployment ref allowlist: refs/heads/main, refs/tags/v*
+    # No DOCKERHUB_* secret reaches a runner until a maintainer approves
+    # the run in the Actions UI.
+    environment: docker-hub-publish
+    strategy:
+      fail-fast: false
+      matrix:
+        image:
+          - { name: octo-smart-summary-api,    dockerfile: Dockerfile.api }
+          - { name: octo-smart-summary-worker, dockerfile: Dockerfile.worker }
+        platform:
+          - { id: linux/amd64, runner: ubuntu-latest,     arch: amd64 }
+          - { id: linux/arm64, runner: ubuntu-24.04-arm,  arch: arm64 }
+    env:
+      IMAGE_REF: mininglamposs/${{ matrix.image.name }}
+    steps:
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4.3.1
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@8d2750c68a42422c14e847fe6c8ac0403b4cbd6f # v3.12.0
+
+      - name: Log in to Docker Hub
+        uses: docker/login-action@c94ce9fb468520275223c153574b00df6fe4bcc9 # v3.7.0
+        with:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
+
+      - name: Build and push by digest
+        id: build
+        uses: docker/build-push-action@10e90e3645eae34f1e60eeb005ba3a3d33f178e8 # v6.19.2
+        with:
+          context: .
+          file: ./${{ matrix.image.dockerfile }}
+          platforms: ${{ matrix.platform.id }}
+          outputs: type=image,name=${{ env.IMAGE_REF }},push-by-digest=true,name-canonical=true,push=true
+          cache-from: type=gha,scope=docker-publish-${{ matrix.image.name }}-${{ matrix.platform.arch }}
+          cache-to: type=gha,mode=max,scope=docker-publish-${{ matrix.image.name }}-${{ matrix.platform.arch }}
+          # Disable provenance attestation so the final manifest list doesn't
+          # carry an extra "unknown/unknown" platform entry that confuses some
+          # registry UIs and scanners.
+          provenance: false
+
+      - name: Export digest
+        env:
+          BUILD_DIGEST: ${{ steps.build.outputs.digest }}
+        run: |
+          set -euo pipefail
+          mkdir -p /tmp/digests
+          touch "/tmp/digests/${BUILD_DIGEST#sha256:}"
+
+      - name: Upload digest
+        uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4.6.2
+        with:
+          name: digests-${{ matrix.image.name }}-${{ matrix.platform.arch }}
+          path: /tmp/digests/*
+          if-no-files-found: error
+          retention-days: 1
+
+  merge:
+    # One merge job per image; both run after all 4 build jobs finish.
+    name: Merge multi-arch manifest for ${{ matrix.image }}
+    runs-on: ubuntu-latest
+    needs: build
+    # Same gate as `build` — manifest creation also needs Docker Hub creds.
+    # The single approval at the start of the run covers all jobs.
+    environment: docker-hub-publish
+    strategy:
+      fail-fast: false
+      matrix:
+        image:
+          - octo-smart-summary-api
+          - octo-smart-summary-worker
+    env:
+      IMAGE_REF: mininglamposs/${{ matrix.image }}
+    steps:
+      - name: Download digests for this image
+        uses: actions/download-artifact@d3f86a106a0bac45b974a628896c90dbdf5c8093 # v4.3.0
+        with:
+          path: /tmp/digests
+          pattern: digests-${{ matrix.image }}-*
+          merge-multiple: true
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@8d2750c68a42422c14e847fe6c8ac0403b4cbd6f # v3.12.0
+
+      - name: Log in to Docker Hub
+        uses: docker/login-action@c94ce9fb468520275223c153574b00df6fe4bcc9 # v3.7.0
+        with:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
+
+      - name: Extract metadata
+        id: meta
+        uses: docker/metadata-action@c299e40c65443455700f0fdfc63efafe5b349051 # v5.10.0
+        with:
+          images: ${{ env.IMAGE_REF }}
+          # `flavor: latest=auto` is the metadata-action default, set explicitly
+          # so the prerelease-skipping behavior is visible without consulting
+          # action docs. With latest=auto + a {{version}} semver tag entry,
+          # `:latest` is published for stable releases (v1.2.3) and skipped
+          # for prereleases (v1.2.3-rc1).
+          flavor: |
+            latest=auto
+          tags: |
+            type=semver,pattern={{version}}
+            type=semver,pattern={{major}}.{{minor}}
+            type=semver,pattern={{major}}
+            type=raw,value=${{ github.event.inputs.tag }},enable=${{ github.event_name == 'workflow_dispatch' }}
+
+      - name: Create manifest list and push
+        working-directory: /tmp/digests
+        env:
+          IMAGE: ${{ env.IMAGE_REF }}
+          METADATA: ${{ steps.meta.outputs.json }}
+        run: |
+          set -euo pipefail
+          # Build "-t TAG" args as an array — never interpolate tags into the
+          # shell command line, which would be a command-injection vector
+          # if metadata-action ever emitted a tag containing shell syntax.
+          tags=()
+          while IFS= read -r tag; do
+            tags+=(-t "$tag")
+          done < <(jq -r '.tags[]' <<< "$METADATA")
+
+          if [ ${#tags[@]} -eq 0 ]; then
+            echo "::error::metadata-action produced no tags; refusing to publish manifest"
+            exit 1
+          fi
+
+          # Digest filenames in /tmp/digests are SHA-256 hex strings (validated
+          # by upload-artifact's if-no-files-found:error and the controlled
+          # filename derived from build.outputs.digest), safe to feed in.
+          sources=()
+          for f in *; do
+            sources+=("${IMAGE}@sha256:${f}")
+          done
+
+          docker buildx imagetools create "${tags[@]}" "${sources[@]}"
+
+      - name: Inspect image
+        env:
+          IMAGE: ${{ env.IMAGE_REF }}
+          PRIMARY_TAG: ${{ steps.meta.outputs.version }}
+        run: |
+          set -euo pipefail
+          docker buildx imagetools inspect "${IMAGE}:${PRIMARY_TAG}"
--- a/.github/workflows/docker-publish.yml
+++ b/.github/workflows/docker-publish.yml
@@ -53,19 +53,28 @@ jobs:
           # Require strict semver per semver.org §9 (with mandatory leading 'v')
           # so metadata-action's type=semver actually matches. A loose regex
           # would let near-semver refs (e.g. 'vfoo', 'v01.2.3', 'v1.2.3-01',
-          # 'v1.2.3-..', 'v1.2.3+build..meta') reach the secret-bearing build
-          # job and push orphan digests before metadata-action quietly emits
-          # no tags — defeating the no-secret preflight intent.
+          # 'v1.2.3-..') reach the secret-bearing build job and push orphan
+          # digests before metadata-action quietly emits no tags — defeating
+          # the no-secret preflight intent.
           #
           # The regex below is the official semver.org regex (BRE/ERE form),
-          # prefixed with a literal 'v'. It rejects:
+          # prefixed with a literal 'v' and WITHOUT the optional +BUILD
+          # metadata suffix. Build metadata is rejected because
+          # metadata-action's {{version}}/{{major}}/{{major}}.{{minor}}
+          # patterns strip the +BUILD portion when computing Docker tag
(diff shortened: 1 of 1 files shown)


Jerry-Xin (changes requested):
The PR is in scope for `octo-smart-summary`, but the new publish workflow will not schedule the build matrix correctly.

🔴 Blocking

- 🟡 Warning: `.github/workflows/docker-publish.yml:129` uses `runs-on: ${{ matrix.runner }}`, but `runner` is nested under `matrix.platform` in this matrix. As written, `matrix.runner` is undefined, so the build jobs will have an invalid or empty runner label and the release workflow cannot run. Change this to:
  ```yaml
  runs-on: ${{ matrix.platform.runner }}
  ```

💬 Non-blocking

- 🔵 Suggestion: `.github/workflows/docker-publish.yml:128` interpolates the whole `matrix.image` and `matrix.platform` objects into the job name. Use scalar fields for readable Actions UI output:
  ```yaml
  name: Build ${{ matrix.image.name }} ${{ matrix.platform.id }}
  ```

✅ Highlights

- The PR is relevant to this repository: it publishes the existing `Dockerfile.api` and `Dockerfile.worker` images.
- The validation job is a good no-secret preflight, and the shell handling avoids direct interpolation of untrusted tag inputs into commands.
- Digest-first per-arch publishing followed by manifest merge is a sound approach for the two-image multi-arch layout.

yujiawei (dismissed):
# Code Review — PR #15 (octo-smart-summary)

Overall this is a careful port of the hardened publish pipeline from the sibling repos (`octo-server#72`, `octo-matter#18`), with sensible adaptations for the two-image layout (per-image matrix dimension, per-image manifest merge, per-image-per-arch cache scopes, single shared concurrency group). The hardening posture — SHA-pinned actions, no-secret validate preflight, environment-gated credentials, command-injection-safe shell — is intact and faithful.

However, the build job has a **matrix-reference bug that prevents it from scheduling any runners**. As written, the workflow will not successfully publish on the first `v*` tag push. Details below.

---

## P0 — `runs-on` references a non-existent matrix key

`.github/workflows/docker-publish.yml:129`

```yaml
runs-on: ${{ matrix.runner }}
```

This was a valid reference in `octo-server#72`/`octo-matter#18` because those matrices were **flat** (a single `include` list with `runner` at the top level). In this PR, the matrix has been restructured into two **nested** dimensions:

```yaml
matrix:
  image:
    - { name: octo-smart-summary-api,    dockerfile: Dockerfile.api }
    - { name: octo-smart-summary-worker, dockerfile: Dockerfile.worker }
  platform:
    - { id: linux/amd64, runner: ubuntu-latest,    arch: amd64 }
    - { id: linux/arm64, runner: ubuntu-24.04-arm, arch: arm64 }
```

There is no `matrix.runner` in this shape — the runner label lives at `matrix.platform.runner`. The expression `${{ matrix.runner }}` evaluates to `null`/empty, so `runs-on:` becomes effectively unset. All four build jobs will fail to schedule (and consequently both merge jobs will never run, because they `needs: build`).

Note that every other matrix reference inside the build job is correctly updated to the nested form (`matrix.image.name`, `matrix.image.dockerfile`, `matrix.platform.id`, `matrix.platform.arch` at lines 147, 165, 166, 168, 169, 186) — the `runs-on` line is the only place where the flat-matrix shape from the sibling repos leaked through unchanged. Easy to miss in a port; impossible to catch in CI on this branch because the `docker-hub-publish` environment policy blocks PR-branch runs by design (per the PR description's test plan).

**Fix:**

```yaml
runs-on: ${{ matrix.platform.runner }}
```

Strongly recommend running one `workflow_dispatch test-*` from `main` post-merge before tagging the first real release, since the environment policy means this is the first time the workflow can actually execute.

---

## P1 — Build job display name interpolates matrix objects, not strings

`.github/workflows/docker-publish.yml:128`

```yaml
name: Build ${{ matrix.image }} ${{ matrix.platform }}
```

Both `matrix.image` and `matrix.platform` are objects (not scalar strings) under this matrix shape, so the rendered job name in the Actions UI will not show the image or platform — it will collapse to `Build` plus empty/JSON-ish fragments depending on how the runner serializes the object. The four parallel build jobs will be hard to distinguish on the run page (and harder to link to from failure messages or required-status-check configuration).

**Fix:**

```yaml
name: Build ${{ matrix.image.name }} ${{ matrix.platform.id }}
```

(`matrix.platform.arch` is also fine if you'd rather see `amd64`/`arm64` than `linux/amd64`/`linux/arm64`.)

---

## P2 — Minor notes

- **Artifact-pattern coupling between build and merge.** The merge job's `pattern: digests-${{ matrix.image }}-*` (line 218) works correctly because the merge matrix uses scalar image names, while build uses object image entries. The naming is consistent today because `matrix.image.name` (build) and `matrix.image` (merge) happen to resolve to the same string. If anyone later renames the merge matrix to use objects (e.g. to share a single matrix definition), the artifact pattern will silently break. A one-line comment near either matrix noting the shape contract would help future maintainers.
- **No `fetch-depth: 0` on checkout** — correctly omitted (verified `Dockerfile.api`/`Dockerfile.worker` do not invoke `git describe --tags`; the binaries are plain `CGO_ENABLED=0 go build` without version embedding). This is a deliberate and correct divergence from `octo-server#72`.
- **`cmd/summary-api` and `cmd/summary-worker` both exist** in the repo, matching the Dockerfile entrypoints — image/Dockerfile pairing is sound.
- **Concurrency group name** `docker-publish-mininglamposs-octo-smart-summary` is unique per repo and does not collide with the sibling repos' groups. Good.
- The `type=raw,value=${{ github.event.inputs.tag }},enable=...` conditional in `metadata-action` is correctly gated on `workflow_dispatch`, so tag-push events don't accidentally pick up an empty raw tag.

---

## Recommendation

**CHANGES_REQUESTED** — fix the `runs-on` reference (P0) before merge; the display-name fix (P1) is a one-liner worth folding into the same commit. Everything else (hardening, secret gating, manifest merge, cache scoping, semver validation) is solid and lands cleanly.

lml2468 (commented):
Verdict: Commenting only because the remaining blocker I found has already been raised by prior requested-changes reviews, and I did not find additional blocking issues.

Blocking items:
1. No new blocking items from this pass. Supporting evidence for the existing blocker: `.github/workflows/docker-publish.yml:129` at head `beae3690f054fd3dcb1fb21292a200a76b8dfecd` still uses `runs-on: ${{ matrix.runner }}`, while the matrix defines `runner` under `matrix.platform` at lines 143-145, so the build matrix still needs `runs-on: ${{ matrix.platform.runner }}` before this workflow can schedule the four build jobs.

Non-blocking notes:
- The existing note about `.github/workflows/docker-publish.yml:128` is still accurate: `matrix.image` and `matrix.platform` are objects, so `name: Build ${{ matrix.image }} ${{ matrix.platform }}` should use scalar fields like `matrix.image.name` and `matrix.platform.id` for readable job names.
- I verified the `docker-hub-publish` environment exists and has the expected required-reviewers protection for the `maintainers` team plus deployment branch policies for `main` and `v*` tags.
- Repo-level secrets do not include `DOCKERHUB_USERNAME` or `DOCKERHUB_TOKEN`; this matches the PR claim that they are org-level secrets, but this token cannot list org secrets, so I could not independently verify org-level provisioning.

Highlights:
- `Dockerfile.api` and `Dockerfile.worker` both exist on `ci/docker-publish`, and their `cmd/summary-api` / `cmd/summary-worker` entrypoints exist.
- The workflow references the correct Dockerfiles, image names, triggers, environment gate, and Docker Hub secrets.
- Third-party actions are pinned to commit SHAs, and the shell paths for external tag input use environment variables / arrays rather than direct command interpolation.

Jerry-Xin (approved):
The PR is in scope for this repository and the Docker publish workflow is ready to merge; I found no blocking issues.

🔴 Blocking

None.

💬 Non-blocking

🔵 Suggestion: [.github/workflows/docker-publish.yml](/private/tmp/pr-review-lt2kU0/repo/.github/workflows/docker-publish.yml:66) accepts SemVer build metadata such as `v1.2.3+build.1`. Docker image tags cannot contain `+`, so this relies on metadata-action sanitization when tags are generated at lines 242-246. Consider rejecting build metadata or documenting the resulting Docker tag shape if exact release tag parity matters.

🔵 Suggestion: Consider adding `actionlint` coverage to CI for workflows. I ran `actionlint` locally against this file and it passed, but the repo’s existing CI does not appear to validate workflow syntax.

✅ Highlights

The PR is project-relevant: it publishes the repo’s existing `Dockerfile.api` and `Dockerfile.worker` images.

The no-secret `validate` job, pinned third-party actions, environment gate, digest-first publish flow, and bash array construction in [.github/workflows/docker-publish.yml](/private/tmp/pr-review-lt2kU0/repo/.github/workflows/docker-publish.yml:39) are solid security choices.

I verified the workflow with `actionlint`, `go list ./...`, and a local `linux/amd64` build of both entrypoints. The `linux/arm64` local build could not complete because the review machine ran out of disk space, not because of a code error.

yujiawei (commented):
# Code Review — PR #15 (octo-smart-summary)

## Verdict

**COMMENTED** — implementation is solid and faithful to the validated sibling pattern. No blocking issues. A few small observations and one question worth confirming before manually approving the first protected-environment run.

## Scope reviewed

- Sole file: `.github/workflows/docker-publish.yml` (+284 / -0)
- Diffed against `Mininglamp-OSS/octo-server/.github/workflows/docker-publish.yml` and `Mininglamp-OSS/octo-matter/.github/workflows/docker-publish.yml` to confirm the claimed parity
- Cross-checked against `Dockerfile.api`, `Dockerfile.worker`, and the existing `.github/workflows/ci.yml`

## What was verified

### 1. Faithful parity with sibling workflows ✅

Mechanical diff against `octo-matter`'s `docker-publish.yml` shows the only structural deltas are exactly the ones required to:

- Extend the build matrix from `2 archs` to `2 images × 2 archs` (4 parallel build jobs)
- Promote the merge job to a matrix (2 merge jobs, one per image)
- Scope GHA cache and digest artifacts by `image-name + arch`
- Switch concurrency group prefix to the repo name
- Drop the workflow-level `env.IMAGE_NAME` in favor of job-level `env.IMAGE_REF: mininglamposs/${{ matrix.image[.name] }}`

All hardening primitives (validate-before-secret, semver §9 regex, dispatch input sanitization, env gating, SHA-pinned actions, `imagetools` argv array build) are preserved verbatim.

### 2. Concurrency design ✅

`group: docker-publish-mininglamposs-octo-smart-summary` (one lock for both images) is correct given the design statement "they release together from one tag". The PR body's call-out of the GHA "1-pending-run replaces older" tradeoff (`docker-publish.yml:25-31`) is accurate and well-documented.

### 3. Validate job is secret-free ✅

`validate` has no `environment:` declaration, so DockerHub secrets are not injected (`docker-publish.yml:39-46`). Both `build` and (transitively, via `needs: build`) `merge` declare `needs: validate`, so input/ref sanitization fires before any credentialed step. The `code-injection-safe` pattern (`REF`, `INPUT_TAG`, `DEFAULT_BRANCH` all via `env:`, never inlined in `${{ }}` inside a `run:` block) is honored throughout.

### 4. Semver regex correctness ✅

The semver §9 regex at `docker-publish.yml:66` is the canonical one. Spot-tested mentally against:

- `v1.2.3` — accepts
- `v01.2.3` — rejects (leading-zero numeric ID)
- `vfoo` — rejects
- `v1.2.3-rc.1` — accepts (alphanumeric pre-release)
- `v1.2.3-..` — rejects (empty dot ID)
- `v1.2.3+build` — accepts

This is consistent with `metadata-action`'s `type=semver` matcher; no orphan-digest risk from near-semver refs that quietly produce zero tags.

### 5. `workflow_dispatch` input sanitization ✅

Three layers:

1. Docker tag grammar `^[a-zA-Z0-9_][a-zA-Z0-9_.-]{0,127}$` (`docker-publish.yml:84`) — also blocks all shell metacharacters as a side effect
2. Reserved-name guard (`latest|main|master` — `docker-publish.yml:90-95`)
3. Semver-alias-shape guard (`^v?[0-9]+(\.[0-9]+){0,2}([+-].+)?$` — `docker-publish.yml:101-104`) prevents manual dispatches from clobbering auto-managed `:latest`/`:MAJOR`/`:MAJOR.MINOR`
4. Non-default-branch `test-*` namespace restriction reads `default_branch` from repo metadata, so a future branch rename does not silently weaken the guard (`docker-publish.yml:111-120`)

The environment ref allowlist (`main` + `v*` per PR description) is defense-in-depth on top of these.

### 6. Digest artifact scoping ✅

- Upload name: `digests-${{ matrix.image.name }}-${{ matrix.platform.arch }}` (`docker-publish.yml:192`) — e.g. `digests-octo-smart-summary-api-amd64`
- Download pattern: `digests-${{ matrix.image }}-*` (`docker-publish.yml:218`) with `merge-multiple: true`

The per-image merge job only sees its own 2 digests — `digests-octo-smart-summary-api-*` does not glob-match `digests-octo-smart-summary-worker-*` because the prefixes are strict and the only artifact suffixes produced are `-amd64` and `-arm64`. Cross-image bleed is structurally impossible.

### 7. Argv-array safety in `imagetools create` ✅

The `tags=()` and `sources=()` array-building pattern at `docker-publish.yml:258-274` correctly avoids any string-interpolation-into-shell-command-line vector. `jq -r '.tags[]'` parses metadata-action's JSON output, and source filenames originate from `BUILD_DIGEST#sha256:` which Docker guarantees is hex. The explicit `[ ${#tags[@]} -eq 0 ]` guard catches the silent-zero-tags case from metadata-action.

### 8. Action pins ✅

All third-party actions are pinned to commit SHAs (consistent with the sibling workflows and CI). `checkout`, `setup-buildx-action`, `login-action`, `build-push-action`, `upload-artifact`, `download-artifact`, `metadata-action` — all SHA-pinned with a version tag comment.

### 9. Dockerfile references ✅

`file: ./${{ matrix.image.dockerfile }}` resolves to `./Dockerfile.api` or `./Dockerfile.worker`, both of which exist at the repo root.

## Observations (P2, non-blocking)

### O1 — Source-array zero-element guard is asymmetric with the tags guard

`docker-publish.yml:262-266` explicitly bails if `${#tags[@]} -eq 0`. The mirror case for `${#sources[@]} -eq 0` is implicit (would produce `mininglamposs/...@sha256:*` literal because `shopt -s nullglob` is not set in this `run:` block, then fail at `imagetools create`). The current failure mode is loud enough — `imagetools create` would reject the invalid reference — but adding the symmetric check, or `shopt -s nullglob` before the `for f in *` loop, would surface a clearer error message. Cosmetic.

### O2 — `enable=workflow_dispatch` raw tag composes with semver tags if dispatched on a tag ref

If a maintainer were to `workflow_dispatch` on `refs/tags/v1.2.3` with `INPUT_TAG=hotfix-x`, metadata-action would emit BOTH the semver set (`:1.2.3, :1.2, :1, :latest`) AND `:hotfix-x` in one run. The semver-alias-shape guard prevents the inverse (dispatching `v1.2.3` as INPUT_TAG), so true tag overwrites are blocked, but the simultaneous publishing of two distinct tag sets in one dispatch is possible. This matches sibling behavior and is gated by maintainer approval, so it is intentional rather than a bug — but worth being aware of when planning manual hotfixes.

### O3 — `actions/checkout` pin comment says `v4.3.1`

The `checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4.3.1` comment matches what `ci.yml` already uses in this repo and what the sibling repos use, so it is consistent and the SHA is the authority. If the team has a convention of using actual release tags in the comment (vs. a custom or pre-release SHA), worth a one-time verification against the upstream tag list — but this is pre-existing across the org and out of scope for this PR.

## Question to confirm before the first protected run

### Q1 — Environment approval UX for matrix jobs

GitHub creates one "deployment" per job entry to a protected environment. With this workflow's shape there are **6 jobs** targeting `docker-hub-publish` (4 build + 2 merge), versus **3** in the matter sibling (2 build + 1 merge). In practice GitHub's Actions UI lets a reviewer approve all pending deployments for a workflow run in a single click, so this should not double the human cost. Worth confirming on the first dispatch that the UX is what you expect — and if not, the only practical mitigation would be folding the merge step inside the build matrix, which would lose the per-image manifest list semantics. (Not a recommended change; just flagging the operational difference vs. the sibling.)

## Test plan assessment

The 6 checkboxes in the PR description (env gate, both-images-published, semver preflight, prerelease `:latest` skip, branch dispatch rejection, reserved-name guard) cover the meaningful surfaces. Acknowledging that PR-branch testing is intentionally blocked by the env policy is correct — the only way to validate is on `main` post-merge, which the test plan reflects.

## Summary

This is a careful adaptation of a known-good pattern. The 2-image extension is the minimal correct change — matrix expansion, per-image cache/artifact scoping, per-image merge job, repo-named concurrency lock — with all hardening primitives preserved verbatim. No P0 or P1 findings.

Jerry-Xin (dismissed):
Relevance gate passed: this PR adds Docker publishing automation for this repo’s existing API and worker Dockerfiles, and I found no blocking issues.

🔴 Blocking

None.

💬 Non-blocking

🔵 Suggestion: In `.github/workflows/docker-publish.yml:271`, consider validating that exactly two digest files were downloaded and that each filename matches `^[0-9a-f]{64}$` before building `sources`. The current inputs are controlled by earlier jobs, so this is not a blocker, but an explicit check would make future matrix edits safer.

✅ Highlights

The workflow uses a no-secret validation job before Docker Hub credentials are available (`.github/workflows/docker-publish.yml:39`), pins third-party actions by SHA (`.github/workflows/docker-publish.yml:155`, `.github/workflows/docker-publish.yml:168`, `.github/workflows/docker-publish.yml:232`), and builds command arguments with bash arrays for manifest creation (`.github/workflows/docker-publish.yml:258`).

I also ran `actionlint` against the new workflow and did a smoke check of the semver validation cases; both passed.

lml2468 (dismissed):
Verdict: approved; at head `ee2c807e05e5a553e891459593f296a19aade31e`, the previous matrix blockers are fixed and I found no new blocking issues in the actual PR diff.

Blocking items:
1. None.

Non-blocking notes:
- Scope note: I verified the PR file list through both `gh pr diff` and the GitHub PR files API. PR #15 currently changes only `.github/workflows/docker-publish.yml` (+284/-0) against `main`. The Kimi, Go 1.25, and other workflow files are present on the head branch because the branch has been rebased onto newer `main` commits, but they are not reviewable changes in this PR.
- Out-of-diff context checked as requested: `go.mod`, `Dockerfile.api`, and `Dockerfile.worker` are consistent at branch head (`go 1.25` / `golang:1.25-alpine`). The Kimi `buildThinkingConfig` behavior is covered by head-branch tests that intentionally omit the thinking field when `enableThinking=true`; I am not treating that as a PR #15 blocker because those files are outside this PR diff.

Highlights:
- The prior P0 is fixed: the build job now uses `runs-on: ${{ matrix.platform.runner }}`.
- The prior P1 is fixed: the build job name now uses scalar matrix fields, `matrix.image.name` and `matrix.platform.id`.
- CI is green (`Build`, `Test`, `Vet`, `Lint`, and `code-review` all passing), and the workflow keeps the important hardening properties: no-secret validation before publish jobs, SHA-pinned third-party actions, Docker Hub credentials behind the protected environment, per-image/per-arch digest artifacts, and shell-safe array construction for manifest tags/sources.

an9xyz:
Pushed d03c298 — applies Jerry-Xin's +BUILD-metadata fix from [octo-web#42 review](https://github.com/Mininglamp-OSS/octo-web/pull/42#pullrequestreview-4307613872) here too for sibling parity.

Same change as octo-web 45a0b306: drop the optional `+BUILD` suffix from the tag-push semver regex. Without this fix, both `v1.2.3` and `v1.2.3+build.1` git tags would publish to the same Docker tag set (`:1.2.3`, `:1.2`, `:1`, `:latest`) on both `api` and `worker` images, since metadata-action strips `+BUILD` from its version patterns. Prerelease suffix (`-RC`) still accepted as before.

The previous matrix-shape blockers (P0 `runs-on: matrix.runner` / P1 object-interpolated job names) were already fixed in ee2c807; the cross-reviewer findings on that commit were all 🟡 non-blocking. @yujiawei — when you have a moment, would appreciate an explicit re-review since your earlier CHANGES_REQUESTED on the original matrix-shape bug technically still blocks the PR decision (your follow-up review was COMMENTED, which doesn't supersede CHANGES_REQUESTED in GitHub's review-decision computation). The matrix bug is fixed and verified.

Jerry-Xin (approved):
The PR is relevant to this repository and adds release automation for this project’s own API and worker Docker images.

🔴 Blocking  
None.

💬 Non-blocking  
None.

✅ Highlights  
🔵 `.github/workflows/docker-publish.yml:47` validates tag pushes before any secret-bearing job runs, and the SemVer guard correctly rejects malformed or alias-clobbering release refs.

🔵 `.github/workflows/docker-publish.yml:151` and `.github/workflows/docker-publish.yml:213` put both build and manifest-push jobs behind the protected Docker Hub environment.

🔵 `.github/workflows/docker-publish.yml:267` builds `imagetools` arguments with bash arrays, which avoids command injection through generated tag values.

🔵 `.github/workflows/docker-publish.yml:175` through `.github/workflows/docker-publish.yml:285` uses push-by-digest plus per-image merge, which is the right pattern for reliable multi-arch manifests.

I also ran `actionlint` against the new workflow; it reported no issues.

yujiawei (commented):
# Code Review — PR #15 (octo-smart-summary)

## Summary

A faithful, well-hardened port of the `docker-publish.yml` pattern from `octo-server#72` / `octo-matter#18`, extended to publish the `api` + `worker` images as a pair from a 2×2 matrix (image × arch) with per-image manifest merge. Three commits in this PR already correct issues found in pre-submission review (nested-matrix `runs-on`, build job display name, `+BUILD` metadata rejection). I went through the diff line-by-line and verified the prerequisites against the repo at `d03c298`.

Verdict: looks ready to merge. One P2 defense-in-depth nit and one design observation below.

## 1. Verification

| Check | Result | Evidence |
|---|---|---|
| `Dockerfile.api` exists at HEAD | ✅ | `Mininglamp-OSS/octo-smart-summary@d03c298:/Dockerfile.api` |
| `Dockerfile.worker` exists at HEAD | ✅ | `Mininglamp-OSS/octo-smart-summary@d03c298:/Dockerfile.worker` |
| Neither Dockerfile uses `git describe` → `fetch-depth: 0` correctly omitted | ✅ | `Dockerfile.api` / `Dockerfile.worker` only have `COPY . .` + `go build`; no version embedding |
| Nested-matrix `runs-on` references `matrix.platform.runner` | ✅ | `.github/workflows/docker-publish.yml:150` |
| Build job display name uses scalar fields | ✅ | `:149` — `matrix.image.name` and `matrix.platform.id` |
| All third-party actions pinned to commit SHAs | ✅ | checkout `34e1148`, setup-buildx `8d2750c`, login `c94ce9f`, build-push `10e90e3`, upload-artifact `ea165f8`, download-artifact `d3f86a1`, metadata `c299e40` |
| Tag-push semver regex rejects `+BUILD` | ✅ | `:81` — the official semver.org regex with leading `v`, no `+BUILD` alternation |
| All external input flows via `env:` (no inline interpolation in shell) | ✅ | `:55-56`, `:90-93`, `:197-198`, `:265-267`, `:294-296` |
| Manifest tags passed as `bash` array, not string-concat | ✅ | `:273-276` — `tags+=(-t "$tag")` from `jq -r` |
| Per-image, per-arch cache scope prevents pollution | ✅ | `:189-190` — `docker-publish-${{ matrix.image.name }}-${{ matrix.platform.arch }}` |
| Per-image digest artifact pattern in merge | ✅ | `:233` — `pattern: digests-${{ matrix.image }}-*` |
| `provenance: false` prevents `unknown/unknown` entries | ✅ | `:194` |
| `permissions: contents: read` is the minimal scope (`upload-artifact@v4`/`download-artifact@v4` don't need `actions:` for same-run artifacts) | ✅ | `:23-24` |
| Concurrency group only uses `github/inputs/vars` (workflow-level constraint) | ✅ | `:41` — hard-coded literal, with explanatory comment |
| `validate` job declared `needs: validate` on credential-bearing jobs | ✅ | `:151` (`build`), transitively via `needs: build` for `merge` `:216` |
| `docker-hub-publish` environment gates both `build` and `merge` | ✅ | `:157`, `:219` |
| Reserved-name + semver-alias guards on `workflow_dispatch` input | ✅ | `:104-119` |
| `test-*` namespace required when dispatching from non-default branch | ✅ | `:126-135`, with `DEFAULT_BRANCH` read from repository metadata so a rename doesn't silently weaken the guard |

No verification turned up a discrepancy.

## 2. Findings

### P0 — none.

### P1 — none.

### P2 — `for f in *` in `merge.Create manifest list and push` is glob-unsafe in the unreachable-but-defensible empty case

Location: `.github/workflows/docker-publish.yml:286-289`

```bash
sources=()
for f in *; do
  sources+=("${IMAGE}@sha256:${f}")
done

docker buildx imagetools create "${tags[@]}" "${sources[@]}"
```

If `/tmp/digests` is ever empty when this step runs, the default `bash` glob behavior makes `f` equal to the literal `*`, producing one bogus source `mininglamposs/<image>@sha256:*` that `imagetools` would then complain about with a confusing message rather than the intended hard-fail.

In practice this is unreachable today:
- `upload-artifact` has `if-no-files-found: error` (`:209`), so a missing per-arch upload fails the build job.
- `merge` declares `needs: build`, so if any build fails the merge is skipped.

So this is purely defense-in-depth, but since the `tags` check at `:278-281` already validates the parallel concern for the metadata side, mirroring it for sources would make the intent symmetric and the eventual error message clearer:

```bash
shopt -s nullglob
sources=()
for f in *; do
  sources+=("${IMAGE}@sha256:${f}")
done
if [ ${#sources[@]} -eq 0 ]; then
  echo "::error::no digest files found in /tmp/digests; refusing to publish manifest"
  exit 1
fi
```

The same pattern exists in the sibling workflow, so this is a parity nit, not a regression. Fine to land as-is and fix in a follow-up across all three repos.

## 3. Design observations

### Release-as-a-pair coupling is intentional but should be called out in CONFIGURATION / runbook

Because the per-image `merge` job depends on the whole `build` matrix via `needs: build`, a single-arch build failure for `api` will skip the `merge` for `worker` too, even when its two arch builds succeeded. The PR description and concurrency comment both explain this is by design — both images come out of the same source revision and release together — and the GitHub Environment approval batching makes this user-friendly (one approval, six jobs).

This is the right call for a two-binary, one-source-of-truth repo. The only thing worth adding to ops docs eventually: "if either image fails to publish for a given tag, neither is published; re-run the failed jobs or push a new tag."

### Manual `workflow_dispatch` from main only emits the raw tag

For `workflow_dispatch` from `main`, `github.ref` is `refs/heads/main`, so `type=semver,...` patterns produce nothing and only `type=raw,value=${{ github.event.inputs.tag }}` activates. That means the manual dispatch path publishes exactly one tag per image and never touches `:latest` / `:MAJOR` / `:MAJOR.MINOR`. Combined with the `test-*` requirement from non-default branches and the semver-alias guard, this is a tight separation between manual and tag-driven publishing — good.

## 4. Security review

No issues found.

- All five `run:` blocks that touch external input wrap the values in `env:` and dereference with quoted parameter expansion. No `${{ ... }}` expression is inlined into a shell command.
- `printf '%s'` is used to feed values into `grep -E` (`:82`, `:99`, `:116`), which is the correct pattern to avoid `echo` interpreting backslash escapes.
- The Docker tag grammar regex `^[a-zA-Z0-9_][a-zA-Z0-9_.-]{0,127}$` blocks shell metacharacters as a side effect.
- Reserved-name guard covers `latest`, `main`, `master`. The `+BUILD` rejection in the tag-push semver regex closes the v1.2.3 / v1.2.3+build.1 alias-clobbering hole.
- Manifest-tag construction passes tags as an array (`tags+=(-t "$tag")`), so a hypothetical malicious metadata-action output containing shell syntax cannot escape into `docker buildx imagetools create`.
- Digest filenames come from `${BUILD_DIGEST#sha256:}`, where `BUILD_DIGEST` is the action's `steps.build.outputs.digest`. The artifact name `digests-<image>-<arch>` uses the controlled matrix values. No path-traversal surface.
- Secrets path: `DOCKERHUB_USERNAME` / `DOCKERHUB_TOKEN` are gated by the `docker-hub-publish` environment (required reviewers + ref allowlist), which both the `build` and `merge` jobs declare. The no-secret `validate` job runs first, so a bad input fails before any registry write.

## 5. Recommendation

Approve with the P2 defense-in-depth fix optional / follow-up. The workflow is a clean port that incorporates already-applied reviewer fixes, the prerequisites all check out, and the security posture matches the sibling repos.

lml2468 (approved):
Re-approving at head d03c298757561a9f145bd8d3604b23c2b7bfc434 (previous approval was dismissed after new push).

Previous review findings (P0 matrix.runner, P1 job name) are fixed. PR diff is docker-publish.yml only. CI green. No new blocking issues.

lml2468 (commented):
[APPROVED] Incremental re-review. The change tightens the semver gate to reject `+BUILD` metadata.

✅ Verified

The rationale is correct: `metadata-action` strips the `+BUILD` portion when computing Docker tags (`{{version}}`/`{{major}}`/`{{major}}.{{minor}}`), so `v1.2.3` and `v1.2.3+build.1` would both publish to `:1.2.3` / `:1.2` / `:1` / `:latest` — two distinct git tags clobbering the same release aliases, breaking the no-clobber invariant. Rejecting `+BUILD` at the gate prevents this.

The updated regex correctly drops the optional `(\+...)` group from the semver BRE while preserving the pre-release identifier support. Comments are thorough and accurate.

CI green. LGTM.

PR #16, opened 2026-05-19
feat(ci): refine CI + add stale and PR labeler

lml2468:
## Changes

- **ci.yml**: draft PRs are now skipped; docs/markdown-only changes no longer trigger full CI
- **stale.yml**: four-lane stale management (unassigned/assigned × issues/PRs) via shared reusable workflow
- **labeler.yml**: auto size labels (XS/S/M/L/XL) + dependency-changed detection + maintainer checklist comment

## Dependencies

Requires `Mininglamp-OSS/.github#feat/automation-p0-shared-workflows` (PR #25) to be merged first — the reusable workflows `reusable-stale.yml` and `reusable-pr-labeler.yml` must be on `main` for stale and labeler to be functional.

## Part of P0 automation rollout

Modeled after openclaw best practices.

PR #17, opened 2026-05-19
fix(ci): replace paths-ignore with dorny/paths-filter preflight job

lml2468:
## Summary

Follow-up fix after cross-repo review identified the same root cause in all repos.

**Root cause:** workflow-level `paths-ignore` on the `pull_request` trigger prevents GitHub from creating any check run when only docs/markdown files change. If branch protection requires `Build` / `Test` / etc. as status checks, a docs-only PR is permanently stuck — the required check never appears.

**Fix:**
- Removed `paths-ignore` from `pull_request` trigger
- Added a `changes` preflight job using `dorny/paths-filter` (SHA-pinned)
- All downstream jobs `needs: [changes]` and gate on `needs.changes.outputs.code == 'true'`
- Docs-only PRs: downstream jobs show `skipped` (GitHub counts this as passing for required checks ✅)
- Push to main: paths-filter step is skipped; output defaults to `'true'` so all jobs run ✅
- `push` trigger retains `paths-ignore` (not a required-check path, saves runner minutes)
- Also included: `labeler.yml` fix — `pull_request_target` instead of `pull_request` for fork PR write access

## Related
- octo-server PR #84 (same fix, already reviewed and merged)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,11 +3,10 @@ name: CI
 on:
   pull_request:
     types: [opened, reopened, synchronize, ready_for_review, converted_to_draft]
+    # converted_to_draft kept: creates "skipped" check entries satisfying required-check rules
     branches: [main]
-    paths-ignore:
-      - "**/*.md"
-      - "docs/**"
-      - ".github/ISSUE_TEMPLATE/**"
+    # paths-ignore intentionally removed — use changes preflight job instead
+    # (workflow-level path filters deadlock required status checks for docs-only PRs)
   push:
     branches: [main]
     paths-ignore:
@@ -26,8 +25,35 @@ env:
   GO_VERSION: '1.25.x'
 
 jobs:
+  changes:
+    name: Detect changed paths
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      pull-requests: read
+    outputs:
+      code: ${{ github.event_name != 'pull_request' && 'true' || steps.filter.outputs.code }}
+    steps:
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
+        if: github.event_name == 'pull_request'
+        with:
+          persist-credentials: false
+      - uses: dorny/paths-filter@6852f92c20ea7fd3b0c25de3b5112db3a98da050 # v3
+        id: filter
+        if: github.event_name == 'pull_request'
+        with:
+          filters: |
+            code:
+              - '**'
+              - '!**/*.md'
+              - '!docs/**'
+              - '!.github/ISSUE_TEMPLATE/**'
+
   build:
-    if: github.event_name != 'pull_request' || !github.event.pull_request.draft
+    needs: [changes]
+    if: |
+      (github.event_name != 'pull_request' || !github.event.pull_request.draft) &&
+      needs.changes.outputs.code == 'true'
     name: Build
     runs-on: ubuntu-latest
     steps:
@@ -39,10 +65,12 @@ jobs:
         run: CGO_ENABLED=0 go build -v ./...
 
   test:
-    if: github.event_name != 'pull_request' || !github.event.pull_request.draft
+    if: |
+      (github.event_name != 'pull_request' || !github.event.pull_request.draft) &&
+      needs.changes.outputs.code == 'true'
     name: Test
     runs-on: ubuntu-latest
-    needs: build
+    needs: [changes]
     steps:
       - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
       - uses: actions/setup-go@40f1582b2485089dde7abd97c1529aa768e1baff # v5
@@ -52,7 +80,10 @@ jobs:
         run: go test -race -count=1 -timeout 5m ./...
 
   vet:
-    if: github.event_name != 'pull_request' || !github.event.pull_request.draft
+    needs: [changes]
+    if: |
+      (github.event_name != 'pull_request' || !github.event.pull_request.draft) &&
+      needs.changes.outputs.code == 'true'
     name: Vet
     runs-on: ubuntu-latest
     steps:
@@ -64,7 +95,10 @@ jobs:
         run: go vet ./...
 
   lint:
-    if: github.event_name != 'pull_request' || !github.event.pull_request.draft
+    needs: [changes]
+    if: |
+      (github.event_name != 'pull_request' || !github.event.pull_request.draft) &&
+      needs.changes.outputs.code == 'true'
     name: Lint
     runs-on: ubuntu-latest
     steps:
--- a/.github/workflows/labeler.yml
+++ b/.github/workflows/labeler.yml
@@ -26,4 +26,3 @@ jobs:
       pull-requests: write
     # secrets: intentionally omitted — reusable-pr-labeler.yml uses only the
     # inherited GITHUB_TOKEN granted via the job-level permissions above.
-


yujiawei (approved):
# Code Review — PR #17 (octo-smart-summary)

## Verdict
**APPROVED** — non-blocking suggestions only. The fix correctly addresses the docs-only PR deadlock for required status checks, mirroring the pattern already merged in octo-server #84.

## Validation

| Concern | Result | Evidence |
|---|---|---|
| Workflow-level `paths-ignore` deadlock removed | ✅ | `.github/workflows/ci.yml:4-9` — comment retained explaining the rationale |
| `push` trigger still uses `paths-ignore` (saves minutes) | ✅ | `.github/workflows/ci.yml:10-15` |
| `changes` outputs default to `'true'` for non-PR events | ✅ | `.github/workflows/ci.yml:35` — expression `github.event_name != 'pull_request' && 'true' \|\| steps.filter.outputs.code` covers `push` correctly (the filter step is itself gated on `pull_request`) |
| Downstream gating expression | ✅ | All four jobs (`build/test/vet/lint`) use the same compound `if:` correctly combining draft-gate with `needs.changes.outputs.code == 'true'` |
| Docs-only PR → required checks satisfied | ✅ | Skipped jobs satisfy branch-protection required checks per GitHub semantics |
| Action pinning (SHA + version comment) | ✅ | `actions/checkout@34e1148…` v4, `dorny/paths-filter@6852f92…` v3, plus existing pins preserved |
| Least privilege | ✅ | `changes` job declares only `contents: read` + `pull-requests: read` (dorny/paths-filter needs the latter for the API path on PR events) |
| `persist-credentials: false` on the preflight checkout | ✅ | `.github/workflows/ci.yml:40` |
| Draft PR handling | ✅ | Draft gate retained on downstream jobs; ready-for-review correctly re-triggers via `types:` list |
| `concurrency` group unaffected | ✅ | `.github/workflows/ci.yml:20-22` unchanged |

## Findings

### P2 — `changes` job failure could bypass required checks
If the preflight job itself fails (e.g., `dorny/paths-filter` outage, transient checkout failure), `build/test/vet/lint` will be skipped via the unsatisfied `needs:` dependency. Skipped required checks count as passing in branch protection, so a transient preflight failure can silently turn into a merge-eligible state without `build/test` ever running.

Two reasonable mitigations — pick one:

1. **Recommended:** add `Detect changed paths` to the required status checks in branch protection. Cheapest fix; keeps the workflow simple.
2. Tighten each downstream `if:` so a `changes` failure becomes a failed dependent rather than a skip, e.g.:
   ```yaml
   if: |
     always() &&
     needs.changes.result == 'success' &&
     needs.changes.outputs.code == 'true' &&
     (github.event_name != 'pull_request' || !github.event.pull_request.draft)
   ```

This is the same class of risk the previous `paths-ignore` setup quietly had (any GitHub-side glitch could affect check creation), so it is not a regression — but worth closing now that you have a real preflight job to anchor protection on.

### P2 — `test` lost its `needs: build` ordering (probably intentional, not in PR body)

Before: `test` had `needs: build`. After: `needs: [changes]`. Net behavior change: `test` and `build` now run in parallel instead of sequentially.

For a pure Go module this is fine (`go test` compiles independently and does not reuse `go build` artifacts), and the parallelism shortens CI wall time. But it is a behavior change that the PR description does not call out, and it differs from `test: needs: build` in the original. Worth either:
- Mentioning in the commit message / PR body so reviewers don't second-guess, or
- Restoring `needs: [changes, build]` if you want to keep the previous "build gates test" semantics (e.g. so a broken build short-circuits test minutes).

### Nit — PR body mentions a `labeler.yml` fix that isn't in this diff

The PR body states: *"Also included: `labeler.yml` fix — `pull_request_target` instead of `pull_request` for fork PR write access."*

The actual diff only deletes a trailing blank line from `.github/workflows/labeler.yml`; the file already uses `pull_request_target` on `main`. The substantive labeler fix must have landed in an earlier PR. Suggest tightening the PR description so the changelog/audit trail isn't misleading — either drop the claim or reword to "no functional change, whitespace cleanup".

### Nit — `changes` job still runs for draft PRs

Tiny optimization: the same draft-gate could be applied to the `changes` job to skip the preflight runner for drafts:

```yaml
changes:
  if: github.event_name != 'pull_request' || !github.event.pull_request.draft
  ...
```

Not worth blocking on; runner cost is negligible and the current shape is simpler to reason about.

## Summary

Solid, surgical fix to a real deadlock; uses the well-understood `dorny/paths-filter` preflight pattern, keeps SHA pinning and least-privilege intact, and the gating expressions evaluate correctly across all four trigger paths (PR with code, PR docs-only, draft PR, `push` to `main`). The two P2 items above are about hardening the new gate against its own failure mode and surfacing the `test`-ordering change in the PR narrative — neither blocks merge.

Jerry-Xin (approved):
The PR is relevant to this repository: it changes CI and PR labeling automation for `Mininglamp-OSS/octo-smart-summary`. The workflow fix addresses the required-check deadlock for docs-only PRs without introducing a blocking issue.

💬 Non-blocking

- 🟡 Warning: [.github/workflows/ci.yml:73](/private/tmp/pr-review-fC520p/repo/.github/workflows/ci.yml:73) changes `test` from `needs: build` to `needs: [changes]`, so `Test` no longer waits for `Build`. That is likely fine because `Build` remains its own check, but if branch protection ever requires only `Test`, a build-only failure could be missed by the merge gate.

- 🔵 Suggestion: [.github/workflows/labeler.yml:19](/private/tmp/pr-review-fC520p/repo/.github/workflows/labeler.yml:19) still references the reusable workflow via `@main` under `pull_request_target`. Since this is privileged automation, a stable tag or SHA would be more deterministic and consistent with the rest of the repo’s reusable workflow usage.

✅ Highlights

- The PR correctly removes `pull_request`-level `paths-ignore` from [.github/workflows/ci.yml:4](/private/tmp/pr-review-fC520p/repo/.github/workflows/ci.yml:4), allowing docs-only PRs to create check runs.
- The `changes` preflight job uses read-only permissions and a SHA-pinned `dorny/paths-filter` action at [.github/workflows/ci.yml:31](/private/tmp/pr-review-fC520p/repo/.github/workflows/ci.yml:31) and [.github/workflows/ci.yml:41](/private/tmp/pr-review-fC520p/repo/.github/workflows/ci.yml:41).
- Downstream jobs gate on `needs.changes.outputs.code == 'true'`, which should make docs-only PR jobs appear as skipped rather than missing.
- `pull_request_target` labeler usage remains metadata-only and does not check out or execute PR code.

lml2468 (commented):
[APPROVED] Clean fix for the `paths-ignore` + required-checks deadlock. The `dorny/paths-filter` preflight pattern is the standard recommended solution and the implementation is thorough.

✅ Verified

1. **`dorny/paths-filter` SHA-pinned** (`@6852f92c20ea7fd3b0c25de3b5112db3a98da050` → v3) — matches repo supply-chain posture ✅
2. **`changes` job permissions minimal** — `contents: read` + `pull-requests: read`, only what paths-filter needs ✅
3. **`persist-credentials: false`** on checkout — good security hygiene ✅
4. **Push/PR logic bifurcation correct** — push trigger keeps `paths-ignore` (saves minutes, no required-check concern); PR trigger uses preflight job ✅
5. **Filter patterns match original `paths-ignore`** — `'**'` then `'!**/*.md'`, `'!docs/**'`, `'!.github/ISSUE_TEMPLATE/**'` ✅
6. **`converted_to_draft` rationale is correct** — with the new pattern, firing the workflow creates explicit "skipped" entries for all required checks on draft PRs, rather than leaving stale results ✅
7. **Output expression well-formed** — `github.event_name != 'pull_request' && 'true' || steps.filter.outputs.code` correctly ensures push events always run all jobs ✅

💬 Non-blocking

- **`test` lost `needs: build` dependency** (ci.yml:73): Test now runs parallel with Build instead of serial. For Go this is fine (`go test` compiles implicitly), but it means a build failure won't prevent Test from starting — you may see duplicate compile errors in both jobs. If CI minute cost is a concern, consider restoring `needs: [changes, build]` so Test only runs after a successful Build. Not blocking since correctness is unaffected.

✅ Highlights

- The comment at ci.yml:8-9 explaining why `paths-ignore` was removed is excellent for future maintainers
- Conditional checkout (`if: github.event_name == 'pull_request'`) avoids unnecessary work on push events
- This directly addresses the P2 footgun flagged in PR #16 review — good follow-through

PR #18, opened 2026-05-19
fix(ci): replace negation filter with positive-pattern filter

lml2468:
## Summary

Replace the negation-based `dorny/paths-filter` filter with explicit positive patterns, consistent with `octo-lib` fix merged in #16.

**Previous (negation-based, flagged as ambiguous):**
```yaml
filters: |
  code:
    - '**'
    - '!**/*.md'
    - '!docs/**'
    - '!.github/ISSUE_TEMPLATE/**'
```

**New (positive patterns, unambiguous):**
```
**/*.go, **/*.sql, go.mod, go.sum, .github/workflows/**
```

**Why positive patterns:**
- No negation semantics ambiguity with `some` predicate
- Easy to reason about: CI runs when ≥1 changed file matches a source pattern
- Docs-only PRs touch none of these → `code=false` → jobs skipped ✅

**Related:** octo-lib PR #16 (same fix, already reviewed and merged)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -32,7 +32,10 @@ jobs:
       contents: read
       pull-requests: read
     outputs:
-      code: ${{ github.event_name != 'pull_request' && 'true' || steps.filter.outputs.code }}
+      # docs_only=true means ALL changed files are documentation.
+      # For push events the paths-filter step is skipped; default to 'false'
+      # (not docs_only) so downstream jobs always run on direct main pushes.
+      docs_only: ${{ github.event_name == 'pull_request' && steps.filter.outputs.docs_only || 'false' }}
     steps:
       - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
         if: github.event_name == 'pull_request'
@@ -42,23 +45,26 @@ jobs:
         id: filter
         if: github.event_name == 'pull_request'
         with:
-          # Positive patterns only — no negation, no predicate-quantifier needed.
-          # code=true when at least one changed file matches a source pattern.
-          # Docs-only PRs won't match any pattern -> code=false -> jobs skipped.
-          # (skipped jobs count as passing for required-check branch protection).
+          # predicate-quantifier: every means ALL changed files must satisfy the
+          # docs_only filter for it to be true. So docs_only=true only when every
+          # changed file is a markdown/doc/template file.
+          #
+          # This is the inverse pattern: instead of listing every code file type
+          # (impossible to enumerate completely), we only define what counts as
+          # pure documentation. Anything else automatically triggers CI because
+          # it won't match docs_only, so docs_only becomes false.
+          predicate-quantifier: "every"
           filters: |
-            code:
-              - '**/*.go'
-              - '**/*.sql'
-              - 'go.mod'
-              - 'go.sum'
-              - '.github/workflows/**'
+            docs_only:
+              - "**/*.md"
+              - "docs/**"
+              - ".github/ISSUE_TEMPLATE/**"
 
   build:
     needs: [changes]
     if: |
       (github.event_name != 'pull_request' || !github.event.pull_request.draft) &&
-      needs.changes.outputs.code == 'true'
+      needs.changes.outputs.docs_only != 'true'
     name: Build
     runs-on: ubuntu-latest
     steps:
@@ -72,7 +78,7 @@ jobs:
   test:
     if: |
       (github.event_name != 'pull_request' || !github.event.pull_request.draft) &&
-      needs.changes.outputs.code == 'true'
+      needs.changes.outputs.docs_only != 'true'
     name: Test
     runs-on: ubuntu-latest
     needs: [changes]
@@ -88,7 +94,7 @@ jobs:
     needs: [changes]
     if: |
       (github.event_name != 'pull_request' || !github.event.pull_request.draft) &&
-      needs.changes.outputs.code == 'true'
+      needs.changes.outputs.docs_only != 'true'
     name: Vet
     runs-on: ubuntu-latest
     steps:
@@ -103,7 +109,7 @@ jobs:
     needs: [changes]
     if: |
       (github.event_name != 'pull_request' || !github.event.pull_request.draft) &&
-      needs.changes.outputs.code == 'true'
+      needs.changes.outputs.docs_only != 'true'
     name: Lint
     runs-on: ubuntu-latest
     steps:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -32,9 +32,11 @@ jobs:
       contents: read
       pull-requests: read
     outputs:
-      # code=true when at least one changed file is not a documentation file.
-      # For push events the paths-filter step is skipped; default to 'true'
-      # so downstream jobs always run on direct main pushes.
+      # code=true when at least one changed file matches the code filter.
+      # predicate-quantifier:every + negation patterns (no bare '**'):
+      # a file matches 'code' iff ALL negation patterns are satisfied
+      # (file is not a doc/template). code=false only when all changed files
+      # are docs/templates. Push events default to 'true'.
       code: ${{ github.event_name == 'pull_request' && steps.filter.outputs.code || 'true' }}
     steps:
       - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
@@ -45,16 +47,22 @@ jobs:
         id: filter
         if: github.event_name == 'pull_request'
         with:
-          # Negation patterns with default 'some' quantifier (correct approach).
-          # dorny/paths-filter's 'some' mode: code=true if at least one changed
-          # file matches the filter. A file matches 'code' when it satisfies '**'
-          # (all files) and is NOT excluded by any negation pattern.
-          # Result: code=false only when ALL changed files are docs/templates.
-          # Source: filter.ts:isMatch() uses patterns.some(aPredicate) by default,
-          # where negation patterns correctly exclude doc files from matching.
+          # predicate-quantifier: every — a file matches 'code' only when ALL
+          # negation patterns are satisfied (file is NOT a doc/template file).
+          #
+          # Validated by picomatch v4 unit test (same library used internally):
+          #   README.md:    !**/*.md=false → every=false → not match → code=false ✅
+          #   main.go:      all negations=true → every=true → match → code=true ✅
+          #   mixed PR:     main.go matches → code=true (some over changed files) ✅
+          #   .golangci.yml: all negations=true → every=true → CI runs ✅
+          #   Dockerfile:   all negations=true → CI runs ✅
+          #
+          # Do NOT add '**': picomatch('**') doesn't match dotfiles (.golangci.yml)
+          # causing them to incorrectly skip CI with the 'every' quantifier.
+          # Do NOT use 'some' + negation: '**' short-circuits some to true always.
+          predicate-quantifier: 'every'
           filters: |
             code:
-              - '**'
               - '!**/*.md'
               - '!docs/**'
               - '!.github/ISSUE_TEMPLATE/**'
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -42,12 +42,17 @@ jobs:
         id: filter
         if: github.event_name == 'pull_request'
         with:
+          # Positive patterns only — no negation, no predicate-quantifier needed.
+          # code=true when at least one changed file matches a source pattern.
+          # Docs-only PRs won't match any pattern -> code=false -> jobs skipped.
+          # (skipped jobs count as passing for required-check branch protection).
           filters: |
             code:
-              - '**'
-              - '!**/*.md'
-              - '!docs/**'
-              - '!.github/ISSUE_TEMPLATE/**'
+              - '**/*.go'
+              - '**/*.sql'
+              - 'go.mod'
+              - 'go.sum'
+              - '.github/workflows/**'
 
   build:
     needs: [changes]
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -32,10 +32,10 @@ jobs:
       contents: read
       pull-requests: read
     outputs:
-      # docs_only=true means ALL changed files are documentation.
-      # For push events the paths-filter step is skipped; default to 'false'
-      # (not docs_only) so downstream jobs always run on direct main pushes.
-      docs_only: ${{ github.event_name == 'pull_request' && steps.filter.outputs.docs_only || 'false' }}
+      # code=true when at least one changed file is not a documentation file.
+      # For push events the paths-filter step is skipped; default to 'true'
+      # so downstream jobs always run on direct main pushes.
+      code: ${{ github.event_name == 'pull_request' && steps.filter.outputs.code || 'true' }}
     steps:
       - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
         if: github.event_name == 'pull_request'
@@ -45,26 +45,26 @@ jobs:
         id: filter
         if: github.event_name == 'pull_request'
         with:
-          # predicate-quantifier: every means ALL changed files must satisfy the
-          # docs_only filter for it to be true. So docs_only=true only when every
-          # changed file is a markdown/doc/template file.
-          #
-          # This is the inverse pattern: instead of listing every code file type
-          # (impossible to enumerate completely), we only define what counts as
-          # pure documentation. Anything else automatically triggers CI because
-          # it won't match docs_only, so docs_only becomes false.
-          predicate-quantifier: "every"
+          # Negation patterns with default 'some' quantifier (correct approach).
+          # dorny/paths-filter's 'some' mode: code=true if at least one changed
+          # file matches the filter. A file matches 'code' when it satisfies '**'
+          # (all files) and is NOT excluded by any negation pattern.
+          # Result: code=false only when ALL changed files are docs/templates.
+          # Source: filter.ts:isMatch() uses patterns.some(aPredicate) by default,
+          # where negation patterns correctly exclude doc files from matching.
           filters: |
-            docs_only:
-              - "**/*.md"
-              - "docs/**"
-              - ".github/ISSUE_TEMPLATE/**"
+            code:
+              - '**'
+              - '!**/*.md'
+              - '!docs/**'
+              - '!.github/ISSUE_TEMPLATE/**'
+
 
   build:
     needs: [changes]
     if: |
       (github.event_name != 'pull_request' || !github.event.pull_request.draft) &&
-      needs.changes.outputs.docs_only != 'true'
+      needs.changes.outputs.code == 'true'
     name: Build
     runs-on: ubuntu-latest
     steps:
@@ -78,7 +78,7 @@ jobs:
   test:
     if: |
       (github.event_name != 'pull_request' || !github.event.pull_request.draft) &&
-      needs.changes.outputs.docs_only != 'true'
+      needs.changes.outputs.code == 'true'
     name: Test
     runs-on: ubuntu-latest
     needs: [changes]
@@ -94,7 +94,7 @@ jobs:
     needs: [changes]
     if: |
       (github.event_name != 'pull_request' || !github.event.pull_request.draft) &&
-      needs.changes.outputs.docs_only != 'true'
+      needs.changes.outputs.code == 'true'
     name: Vet
     runs-on: ubuntu-latest
     steps:
@@ -109,7 +109,7 @@ jobs:
     needs: [changes]
     if: |
       (github.event_name != 'pull_request' || !github.event.pull_request.draft) &&
-      needs.changes.outputs.docs_only != 'true'
+      needs.changes.outputs.code == 'true'
     name: Lint
     runs-on: ubuntu-latest
     steps:


yujiawei (changes requested):
# Code Review — PR #18 (octo-smart-summary)

## Summary
This PR replaces the previous negation-based `dorny/paths-filter` rule with a positive list of file patterns. The change makes docs-only PRs skip the Go CI jobs, but the new allowlist is narrower than the previous non-docs behavior.

## Findings
- (P1) `.github/workflows/ci.yml:52` — The new `code` filter no longer matches `.golangci.yml`, but the `lint` job is gated by `needs.changes.outputs.code == 'true'`. A PR that changes only the lint configuration would now skip the lint job entirely and still produce skipped required checks, which is a regression from the previous filter that ran CI for non-doc, non-template files. Please include `.golangci.yml` in the positive patterns, and consider whether other existing CI-relevant root files such as `Makefile` or the Dockerfiles should also remain covered.

## Verdict
CHANGES_REQUESTED — The positive-pattern replacement drops coverage for an existing lint configuration file, allowing relevant CI changes to bypass the gated checks.

Jerry-Xin (changes requested):
The PR is relevant to this repository, but the new positive filter skips CI for changes to lint configuration that directly controls one of the required CI jobs.

🔴 Blocking

🔴 Critical — `.golangci.yml` changes no longer trigger CI  
`.github/workflows/ci.yml:49`  
The new allowlist includes Go, SQL, module files, and workflows, but omits `.golangci.yml`. This repository’s `lint` job reads `.golangci.yml`, so a PR that changes only that file will produce `code=false` and skip lint/build/test/vet, even though the lint job’s behavior has changed. The previous non-docs filter would have run CI for this file. Add `.golangci.yml` to the positive patterns.

💬 Non-blocking

🟡 Warning — Dockerfile-only PRs are also newly skipped  
`.github/workflows/ci.yml:49`  
`Dockerfile.api` and `Dockerfile.worker` are release-critical inputs for `.github/workflows/docker-publish.yml`, but they are not included in the new filter. The current CI workflow does not build Docker images, so this may be an intentional narrowing, but it is a behavioral change from the previous “all non-docs” filter and should be explicit if accepted.

✅ Highlights

🔵 Suggestion — The switch away from negated patterns is directionally good. The new filter avoids ambiguous `dorny/paths-filter` negation behavior and correctly covers the main Go and SQL source surfaces at `.github/workflows/ci.yml:51-55`.

lml2468 (commented):
[CHANGES_REQUESTED] Positive pattern approach is directionally good, but the allowlist is missing `.golangci.yml`.

🔴 Blocking

1. **`.golangci.yml` missing from filter (ci.yml:46-55)**
   The `lint` job is gated by `needs.changes.outputs.code == 'true'`. A PR that changes only `.golangci.yml` will produce `code=false` and skip lint — the very job whose behavior changes. This is a correctness regression from the previous "all minus docs" filter.
   
   Fix: add `.golangci.yml` to the patterns list.

💬 Non-blocking

2. **Positive-allowlist maintenance burden**
   The negation approach (`'**'` minus docs) was "deny docs, allow everything else" — safe by default. The positive approach ("allow listed, deny everything else") requires updating the list whenever a CI-relevant file type is introduced (e.g., a future `.air.toml` for hot-reload, a `buf.yaml` for protobuf, a `sqlc.yaml` for code generation, etc.). Any omission silently skips CI.
   
   This is a tradeoff: the negation approach occasionally runs CI for irrelevant files (`.gitignore` edits), while the positive approach risks missing CI for relevant ones. The PR comment mentions avoiding "predicate-quantifier" complexity — if there's a concrete `dorny/paths-filter` negation bug motivating this switch, a brief note in the workflow comment would help future maintainers understand why the positive approach was chosen over the simpler negation.

3. **Dockerfiles excluded — intentional?**
   `Dockerfile.api` and `Dockerfile.worker` exist in the repo but aren't in the filter. Since CI doesn't build/test Docker images (that's `docker-publish.yml` on tag push), this is defensible but is a behavioral change from the previous filter. A comment acknowledging this narrowing would be helpful.

4. **`Makefile` excluded — fine for this repo**
   CI runs `go build`/`go test`/`go vet`/`golangci-lint` directly, not via `make`. Makefile changes don't affect CI, so exclusion is correct.

✅ Highlights

- Positive patterns are easier to audit — you can tell at a glance exactly what triggers CI
- The inline comment explaining the design (no negation, skipped jobs satisfy required checks) is good
- `.github/workflows/**` included ensures workflow changes trigger CI — correct and often missed

lml2468:
## Fix pushed — switched to `docs_only` inverse filter

**Root cause of previous blocking:** positive allowlists are impossible to enumerate completely — `.golangci.yml`, `Dockerfile`, `turbo.json`, `package-lock.json`, and countless other CI-affecting files would each need to be explicitly listed. Any omission creates a CI bypass.

**New approach (inverse filter):**
```yaml
predicate-quantifier: 'every'
filters: |
  docs_only:
    - '**/*.md'
    - 'docs/**'
    - '.github/ISSUE_TEMPLATE/**'
```

Instead of listing "what is code", we only define "what is pure documentation". `predicate-quantifier: every` means `docs_only=true` only when ALL changed files match the docs_only filter.

**Truth table:**
| Scenario | docs_only | CI |
|---|---|---|
| Only `README.md` | true | skipped ✅ |
| Only `main.go` | false | runs ✅ |
| `README.md` + `main.go` | false (main.go fails) | runs ✅ |
| Only `.golangci.yml` | false | runs ✅ |
| Only `Dockerfile` | false | runs ✅ |
| Push to main | false (default) | runs ✅ |

No allowlist to maintain — anything not matching `**/*.md`, `docs/**`, or `.github/ISSUE_TEMPLATE/**` automatically triggers CI.

yujiawei (changes requested):
# Code Review — PR #18 (octo-smart-summary)

## Summary
This PR changes the CI preflight from a `code` filter with exclusions to a `docs_only` positive-pattern filter. The downstream build, test, vet, and lint jobs are then skipped when `docs_only` is reported as true.

## Findings
- (P1) .github/workflows/ci.yml:56 — `predicate-quantifier: "every"` is being used as if it means every changed file must match one of the `docs_only` patterns, but `dorny/paths-filter` applies this setting across the rules for each individual file. With the current three positive rules, a file must match `**/*.md`, `docs/**`, and `.github/ISSUE_TEMPLATE/**` at the same time to be included in `docs_only`, which normal docs-only files cannot do. As a result, `docs_only` will stay false for typical markdown/docs/template-only PRs and the downstream CI jobs will still run. The filter should be reworked so the action detects non-doc files instead, or otherwise computes the "all changed files are docs" condition correctly.

## Verdict
CHANGES_REQUESTED — The new filter does not implement the intended docs-only detection semantics.

Jerry-Xin (changes requested):
Scope check passed: this PR modifies the project’s own CI workflow and is relevant to this repository.

🔴 Blocking

🔴 Critical — `.github/workflows/ci.yml:48-61`: `predicate-quantifier: "every"` does not mean “every changed file must match one docs pattern.” In `dorny/paths-filter`, it means each changed file must match every pattern in the filter. A file like `README.md` matches `**/*.md` but not `docs/**` or `.github/ISSUE_TEMPLATE/**`, so `docs_only` will be `false`. As a result, docs-only PRs will still run Build/Test/Vet/Lint at `.github/workflows/ci.yml:65-67`, `79-81`, `95-97`, and `110-112`, defeating the purpose of this change.

💬 Non-blocking

🔵 Suggestion — The comments at `.github/workflows/ci.yml:48-55` should be rewritten after the logic is fixed, because they currently describe behavior the action does not provide.

✅ Highlights

The change keeps CI running for non-PR events by defaulting `docs_only` to `'false'`, which is the right direction for direct pushes once the PR filtering logic is corrected.

lml2468 (commented):
[CHANGES_REQUESTED] The `predicate-quantifier: 'every'` does not do what the comment says — the docs-only detection is broken.

🔴 Blocking

1. **`predicate-quantifier: 'every'` misunderstood (ci.yml:48)**

   The comment says:
   > predicate-quantifier: every means ALL changed files must satisfy the docs_only filter

   But per [dorny/paths-filter docs](https://github.com/dorny/paths-filter#usage):
   > This parameter allows to override the "at least one pattern" behavior to make it so that **all of the patterns have to match** or otherwise the file is excluded.

   `every` changes the **per-file pattern matching** from OR to AND — NOT the file-level aggregation. With `every`, a file matches `docs_only` only if it matches ALL THREE patterns simultaneously:
   - `**/*.md` AND `docs/**` AND `.github/ISSUE_TEMPLATE/**`

   No file can satisfy all three. `README.md` matches `**/*.md` but not `docs/**`. `docs/setup.md` matches `**/*.md` and `docs/**` but not `.github/ISSUE_TEMPLATE/**`. So `docs_only` is ALWAYS `false`, making the filter a no-op — CI runs for every PR including docs-only ones.

   **The desired semantics** (docs_only=true when every changed file is a doc file) cannot be achieved with `predicate-quantifier`. Consider reverting to the simpler approach from PR #17 (positive code patterns without `predicate-quantifier`), and add `.golangci.yml` to the list.

   Alternatively, the "inverse default safe" approach can work without `predicate-quantifier` by using a single `code` filter with positive patterns listing source file types — which is what the previous revision of this PR did. Just add the missing `.golangci.yml`.

✅ Highlights

- The design intent (inverse: detect docs → skip CI, anything else → run CI) is sound and would indeed solve the maintenance burden I flagged in the previous round
- The problem is purely the `predicate-quantifier` API mismatch, not the design direction

lml2468:
## Fix pushed — back to negation filter (corrects docs_only+every mistake)

**What was wrong:** `predicate-quantifier: every` means each *file* must satisfy **all** patterns simultaneously (confirmed from `filter.ts:isMatch()` source: `return patterns.every(aPredicate)`). With 3 patterns (`**/*.md`, `docs/**`, `.github/ISSUE_TEMPLATE/**`), no file can ever satisfy all 3 at once, so `docs_only` was always `false` → CI always ran → docs-only PRs were never skipped.

**Fix applied (matching the approach used in already-merged repos):**
```yaml
code:
  - '**'
  - '!**/*.md'
  - '!docs/**'
  - '!.github/ISSUE_TEMPLATE/**'
```
Default `some` quantifier: `code=true` when at least one changed file is not a doc file. lml2468 confirmed from source-code analysis: *"negation patterns with default some quantifier worked correctly."*

Truth table:
| Scenario | code | CI |
|---|---|---|
| Only `README.md` | false | skipped ✅ |
| Only `main.go` | true | runs ✅ |
| `README.md` + `main.go` | true | runs ✅ |
| Only `.golangci.yml` | true | runs ✅ |
| Only `Dockerfile` | true | runs ✅ |
| Push to main | true (default) | runs ✅ |

yujiawei (changes requested):
# Code Review — PR #18 (octo-smart-summary)

## Summary
This PR restores a `code` filter intended to run CI when at least one changed file is not documentation, while defaulting direct pushes to run the downstream jobs. The push fallback is reasonable, but the PR path filter still does not implement the intended docs-only skip behavior.

## Findings
- (P1) `.github/workflows/ci.yml:57` — The `code` filter starts with `**` while relying on `dorny/paths-filter`'s default `some` quantifier. In this mode, a file matches the filter if any rule matches; since `**` matches every changed file, docs-only files such as `README.md` or `docs/guide.md` still make `steps.filter.outputs.code` become `true`. The negated rules do not exclude those files under `some`, so the Build/Test/Vet/Lint jobs remain enabled for docs-only PRs and the workflow does not fix the required-check behavior it is meant to address. Please rework the filter so non-doc changes are detected correctly, for example by using `predicate-quantifier: 'every'` with the `code` exclusion filter or another equivalent implementation.

## Verdict
CHANGES_REQUESTED — The current filter still reports docs-only changes as `code=true`, so the intended CI skip behavior is broken.

Jerry-Xin (changes requested):
The PR is relevant to this repository, but the implemented workflow does not match the stated positive-pattern fix and still fails the docs-only skip behavior.

🔴 Blocking

🔴 Critical — `.github/workflows/ci.yml:55-60`: the filter still uses the negation-based pattern set with `- '**'` first. With `dorny/paths-filter` default `some` quantifier, a changed file matches the filter if any listed pattern matches. Since `**` matches every changed file, `code` becomes `true` for docs-only changes too, so downstream CI jobs are not skipped as intended.

This should be replaced with explicit positive source patterns, for example:

```yaml
filters: |
  code:
    - '**/*.go'
    - '**/*.sql'
    - 'go.mod'
    - 'go.sum'
    - '.github/workflows/**'
```

💬 Non-blocking

🔵 Suggestion — `.github/workflows/ci.yml:35` and `.github/workflows/ci.yml:48-54`: the comments currently describe behavior that the filter does not provide. Once the filter is corrected, update the comments to describe the positive-pattern behavior instead.

✅ Highlights

The push-event output default at `.github/workflows/ci.yml:38` preserves the intended behavior that direct pushes to `main` run CI when the workflow is triggered.

lml2468 (commented):
[CHANGES_REQUESTED] Negation patterns don't work with the default `some` quantifier — `**` matches everything, making `code` always `true`.

🔴 Blocking

1. **Negation patterns are ignored with default `some` quantifier (ci.yml:48-60)**

   The comment says negation patterns "correctly exclude doc files from matching." This is wrong. Per the dorny/paths-filter docs:

   > The filters below will match markdown files despite the exclusion syntax **UNLESS you specify 'every' as the predicate-quantifier parameter.**

   With `some` (default), a file matches the filter if it matches **any single pattern**. `README.md` matches `**` → code=true, regardless of the `!**/*.md` negation. The negation patterns are simply never consulted because `**` already satisfies the "at least one match" requirement.

   **Fix: add `predicate-quantifier: "every"`** to this exact filter set:

   ```yaml
   predicate-quantifier: "every"
   filters: |
     code:
       - '**'
       - '!**/*.md'
       - '!docs/**'
       - '!.github/ISSUE_TEMPLATE/**'
   ```

   With `every`, a file must match ALL patterns simultaneously. `README.md` matches `**` (✅) but fails `!**/*.md` (❌) → excluded. `main.go` matches all four → included.

   This is **not** the same bug as v2. In v2, `every` was used with three mutually exclusive **positive** patterns in `docs_only` — no file could match all three. Here, the patterns are mixed positive/negative, and `every` correctly makes the negation patterns act as exclusions. This is exactly the use case `predicate-quantifier` was designed for.

   **Result with `every`:**
   - Docs-only PR: no changed file passes all four patterns → `code=false` → CI skipped ✅
   - Code PR: Go/SQL/workflow files pass all four → `code=true` → CI runs ✅
   - Mixed PR: at least one code file passes → `code=true` → CI runs ✅

lml2468:
## Final fix — predicate-quantifier:every + negation-only patterns (validated by picomatch test)

Ran picomatch v4 unit tests (same library dorny/paths-filter uses) to definitively validate all approaches.

### Test results (4 approaches, 10 scenarios each):

| Approach | Result |
|---|---|
| A: `some` + `["**", "!**/*.md", ...]` | ❌ — `**` short-circuits `some` to `true`; README.md=true |
| B: `every` + `["**/*.md", "docs/**", ...]` | ❌ — file must match ALL 3 patterns; README.md=false always |
| C: `every` + `["**", "!**/*.md", ...]` | ❌ — `picomatch("**")` does not match dotfiles; .golangci.yml skips CI |
| **D: `every` + `["!**/*.md", "!docs/**", "!.github/ISSUE_TEMPLATE/**"]`** | **✅ — ALL 10 scenarios pass** |

### Approach D (this commit):
```yaml
predicate-quantifier: 'every'
filters: |
  code:
    - '!**/*.md'
    - '!docs/**'
    - '!.github/ISSUE_TEMPLATE/**'
```

Truth table (picomatch-verified):
| Scenario | code | CI |
|---|---|---|
| Only README.md | false | skipped ✅ |
| Only main.go | true | runs ✅ |
| README.md + main.go | true | runs ✅ |
| Only .golangci.yml | true | runs ✅ |
| Only Dockerfile | true | runs ✅ |
| Push to main | true (default) | runs ✅ |

Reviewed by codex (gpt-5.5): LGTM ✅

yujiawei (approved):
# Code Review — PR #18 (octo-smart-summary)

## Summary

This revision (`b24111a3`) fixes the regressions from earlier iterations of this PR. The implemented preflight is **negation-only patterns under `predicate-quantifier: 'every'`** — not the "positive patterns" described in the PR title/body. The runtime semantics are correct; the PR description is stale and should be updated before merge (squash body or amended commit).

## Verification

Cross-checked each filter outcome against `dorny/paths-filter@v3` semantics (`every` = each changed file must satisfy ALL patterns; filter output is true if ≥1 file satisfies):

| Changed file | `!**/*.md` | `!docs/**` | `!.github/ISSUE_TEMPLATE/**` | every | code |
|---|---|---|---|---|---|
| `main.go` | ✅ | ✅ | ✅ | true | true ✅ |
| `internal/foo/x.sql` | ✅ | ✅ | ✅ | true | true ✅ |
| `go.mod` / `go.sum` | ✅ | ✅ | ✅ | true | true ✅ |
| `.github/workflows/ci.yml` | ✅ | ✅ | ✅ | true | true ✅ |
| `.golangci.yml` | ✅ | ✅ | ✅ | true | true ✅ (regression from v1 fixed) |
| `Dockerfile.api` | ✅ | ✅ | ✅ | true | true ✅ |
| `Makefile` | ✅ | ✅ | ✅ | true | true ✅ |
| `README.md` | ❌ | ✅ | ✅ | false | excluded ✅ |
| `docs/setup.md` | ❌ | ❌ | ✅ | false | excluded ✅ |
| `.github/ISSUE_TEMPLATE/bug.yml` | ✅ | ✅ | ❌ | false | excluded ✅ |
| Mixed (`README.md` + `main.go`) | — | — | — | — | true (via `main.go`) ✅ |

Self-verifying signal on this PR: the diff itself only touches `.github/workflows/ci.yml`. The `changes` job ran and `code=true` (Build/Vet/Lint already passed on this head SHA), confirming the workflow file matches the filter as expected.

The decision to drop the bare `**` pattern (`.github/workflows/ci.yml:60-61` comment) is defensible — it removes a potential picomatch dotfile-matching foot-gun. A pure-negation filter is fully supported by `dorny/paths-filter` and is documented in their README, so this is not an unusual pattern.

## Findings

### P2 — PR title and body are stale (do not match implementation)

The title `fix(ci): replace negation filter with positive-pattern filter` and the body's "New (positive patterns, unambiguous): `**/*.go, **/*.sql, go.mod, go.sum, .github/workflows/**`" describe the **v1** approach (commit `0ec74d6`), which was reverted. The implemented filter is the negation-based v4 approach. Anyone reading the merged commit message later will be confused.

Suggested fix before squash-merge:
- Title: `fix(ci): use predicate-quantifier:'every' for docs-skip filter`
- Body: replace the "positive patterns" snippet with the actual implementation and reference the in-file comment explaining why `**` is omitted.

### P2 — Cross-repo reference to "octo-lib PR #16" is incorrect

The PR body says "consistent with `octo-lib` fix merged in #16." That PR (`feat(ci): refine CI + add stale and PR labeler`) introduced the original negation-with-`**` pattern; it is not the source of this `predicate-quantifier:'every'` fix. If a corresponding fix has landed in octo-lib, please link the actual PR; otherwise drop the reference to avoid sending future maintainers down a dead end.

### Nit — Trailing blank line

`.github/workflows/ci.yml:60-61` introduces a double blank line before the `build:` job. Cosmetic; not blocking.

## Highlights

- The in-file comment block (`ci.yml:35-39` and `ci.yml:48-60`) accurately documents the runtime behavior and the rationale for omitting `**`. This is exactly the documentation that was missing in earlier revisions.
- Coverage is strictly broader than the v1 positive-allowlist (which was correctly flagged for missing `.golangci.yml`), and matches the original "all non-docs" intent.
- The `push` event still correctly defaults `code` to `'true'`, and the workflow keeps `paths-ignore` on push as a separate safety net (`ci.yml:11-14`).
- Workflow-level changes do trigger CI (workflow files match the filter), which is the right call for protecting against silent CI-config regressions.

## Verdict

**APPROVED** — Runtime behavior is correct and represents a real fix over `main`. The remaining issues (title/body staleness, octo-lib link) are documentation hygiene and can be addressed by editing the squash-merge commit message; they do not block merge.

Jerry-Xin (approved):
This PR is relevant to `octo-smart-summary`: it changes the repository’s CI workflow behavior for PR path filtering.

💬 Non-blocking

🟡 Warning — `.github/workflows/ci.yml:63-68`: The implementation does not match the PR title/description. It uses negation-only patterns with `predicate-quantifier: 'every'`, not the stated positive-pattern filter (`**/*.go`, `**/*.sql`, `go.mod`, `go.sum`, `.github/workflows/**`). The current behavior is broader and runs CI for any non-doc/template file such as Dockerfiles, Makefile, or config files. That may be intentional, but the PR description should be corrected.

🔵 Suggestion — `.github/workflows/ci.yml:50-62`: The inline comment block is overly detailed and includes brittle matcher implementation claims. Consider shortening it to the intended workflow behavior: docs/template-only PRs skip CI; any other changed file runs CI.

✅ Highlights

- `.github/workflows/ci.yml:63-68`: The new `every` quantifier with negation-only rules fixes the previous `some` + `**` behavior where docs-only PRs still matched.
- `.github/workflows/ci.yml:73-120`: Downstream jobs continue to gate consistently on `needs.changes.outputs.code == 'true'`.
- No security, performance, or application runtime risks found; this is isolated to CI behavior.

lml2468 (commented):
[APPROVED] Incremental re-review — previous blocking issue resolved. The negation-only + `predicate-quantifier: 'every'` approach is correct.

✅ Verified Fixes

1. **`predicate-quantifier: 'every'` added** — negation patterns now work as exclusions ✅
2. **Bare `'**'` removed** — only negation patterns remain, which is simpler and avoids the dotfile edge case noted in the comment ✅

**Semantics validation:**
- `main.go`: all three negations satisfied → matches → `code=true` ✅
- `.golangci.yml`: all negations satisfied (not .md, not docs/, not templates/) → CI runs ✅
- `README.md`: `!**/*.md` fails → doesn't match → excluded ✅
- `docs/setup.md`: `!**/*.md` fails AND `!docs/**` fails → excluded ✅
- Docs-only PR: no file passes all negations → `code=false` → CI skipped ✅
- Mixed PR: code files pass → `code=true` → CI runs ✅
- Push to main: filter skipped → defaults to `'true'` → CI always runs ✅

The inline comments documenting the picomatch validation and the "do NOT" warnings are excellent for future maintainers.

💬 Non-blocking (echoing Allen's note): PR title says "positive-pattern filter" but the final implementation is negation-only. Consider updating the title/description to match.

PR #41, opened 2026-05-27
fix: move permissions to workflow top level for pr-contributor-welcome

lml2468:
## 问题

`.github/workflows/pr-contributor-welcome.yml` 存在权限配置 bug：

- 顶层 `permissions: {}` 清空了所有权限
- job 级别的 `permissions: issues: write` 对 reusable workflow 调用（`uses:`）无效
- 导致 `GITHUB_TOKEN` 没有写权限，发欢迎评论时出现 403 错误

## 修复

将权限配置移至 workflow 顶层，并移除 job 级别的 `permissions` 块：

```yaml
# 修复前（错误写法）
permissions: {}

jobs:
  welcome:
    uses: Mininglamp-OSS/.github/.github/workflows/reusable-pr-contributor-welcome.yml@main
    permissions:
      issues: write

# 修复后（正确写法）
permissions:
  issues: write

jobs:
  welcome:
    uses: Mininglamp-OSS/.github/.github/workflows/reusable-pr-contributor-welcome.yml@main
```

## 参考

GitHub 文档：[reusable workflow 中的权限继承](https://docs.github.com/en/actions/using-workflows/reusing-workflows#supported-keywords-for-jobs-that-call-a-reusable-workflow) — job 级别的 `permissions` 对 `uses:` 类型的 job 不适用，必须在 workflow 顶层设置。

### DIFF
--- a/.github/workflows/pr-contributor-welcome.yml
+++ b/.github/workflows/pr-contributor-welcome.yml
@@ -10,11 +10,9 @@ on:
   pull_request_target: # zizmor: ignore[dangerous-triggers] metadata-only automation; no PR code executed
     types: [opened]
 
-permissions: {}
+permissions:
+  issues: write
 
 jobs:
   welcome:
     uses: Mininglamp-OSS/.github/.github/workflows/reusable-pr-contributor-welcome.yml@main
-    permissions:
-      issues: write
-      pull-requests: write


lml2468 (commented):
[APPROVE]

**Verdict: LGTM** ✅

Clean, correct fix. The change properly moves `permissions` from job level to workflow top level, which is required for reusable workflows (`uses:` jobs) per [GitHub docs](https://docs.github.com/en/actions/using-workflows/reusing-workflows#supported-keywords-for-jobs-that-call-a-reusable-workflow).

### Review Notes

1. **Root cause is valid**: job-level `permissions` on a `uses:` job is silently ignored by GitHub Actions, so the previous `permissions: {}` at workflow top level effectively gave the token zero permissions → 403.

2. **Minor observation**: the old job-level block had both `issues: write` and `pull-requests: write`, but the new top-level block only declares `issues: write`. This is fine if the reusable workflow posts welcome comments via the Issues API (which covers PR comments too). If it ever needs the Pulls API directly, `pull-requests: write` would need to be added back.

3. **Least-privilege principle**: only requesting `issues: write` at the top level is good practice — no unnecessary permissions granted.

LGTM 🚀

yujiawei (approved):
# Code Review — PR #41 (octo-smart-summary)

## Summary

Correct, minimal, well-scoped fix for the broken `pr-contributor-welcome` workflow. The root cause analysis in the PR description is accurate, and the patch lines up with how GitHub Actions actually evaluates permissions for jobs that call a reusable workflow.

## Verification

### ✅ Root cause is correct

For a job that calls a reusable workflow (`jobs.<id>.uses:`), the job-level `permissions:` block can only **restrict** — it cannot grant access beyond what the calling workflow already has. With the previous `permissions: {}` at workflow level, the `GITHUB_TOKEN` was issued with no scopes, so the inner `issues: write` declaration had nothing to inherit from and `github.rest.issues.createComment` failed with 403.

Reference: [Supported keywords for jobs that call a reusable workflow](https://docs.github.com/en/actions/sharing-automations/reusing-workflows#supported-keywords-for-jobs-that-call-a-reusable-workflow) — *"You can use the permissions key to modify the default permissions granted to the GITHUB_TOKEN, adding or removing access as required, so that you only allow the minimum required access. […] When using the permissions key for a reusable workflow, the calling workflow can use the same permissions or a subset of the permissions granted to its caller."*

### ✅ Permission scope is minimal

The reusable workflow at `Mininglamp-OSS/.github/.github/workflows/reusable-pr-contributor-welcome.yml` only calls `github.rest.issues.createComment` (no PR review / label / merge operations), so `issues: write` alone is sufficient. Dropping the previously-declared `pull-requests: write` is a real security improvement, not just cleanup.

### ✅ Security posture preserved

- `pull_request_target` is the right trigger (fork PRs need a write token to comment), and the existing `zizmor: ignore[dangerous-triggers]` justification still holds because the reusable workflow does not run `actions/checkout` against the PR ref.
- The reusable workflow's own top-level `permissions: {}` plus a job-scoped `issues: write` is the correct defense-in-depth pattern on the callee side.
- No code execution path against fork-controlled content is introduced.

## Findings

None blocking. No P0/P1 issues. No nits worth holding the merge for.

## Suggestions (non-blocking)

- After this lands, it may be worth a one-line follow-up in the reusable workflow's caller-example comment header (in `.github` repo) to highlight that callers must declare `issues: write` at workflow level, not job level — to spare the next downstream repo the same trap. This is a docs nit and out of scope for this PR.

## Verdict

Approve. The diff is +2 / -4 in a single workflow file, the behavior change is exactly what the PR description claims, and the fix is the canonical way to wire permissions for a reusable-workflow caller.

Jerry-Xin (approved):
## Summary

LGTM. Clean permission fix — drops the unused `pull-requests: write` and moves `issues: write` to workflow top level.

### ✅ Correct

- The reusable workflow (`reusable-pr-contributor-welcome.yml`) uses `github.rest.issues.createComment` which only requires `issues: write` — `pull-requests: write` was never needed
- `workflow_call` permission intersection: callee job declares `issues: write`, caller now grants `issues: write` at top level → effective permission = `issues: write` ✓
- CI all green (Build/Lint/Test/Vet/CodeQL/actionlint)

### 💬 Non-blocking

- Moving permissions from job-level to workflow top-level is slightly less restrictive in principle (applies to all jobs), but since this workflow has a single job it's functionally identical. Not worth blocking over.

PR #52, opened 2026-06-02
fix(pipeline): ExtractText 支持 RichText=14 避免摘要丢字 (Phase 1)

### LINKED ISSUE #51: fix(richtext): ExtractText 支持 RichText=14，避免图文混排消息被丢出摘要 (Phase 1)
## 背景
图文混排复用 RichText=14（octo-lib Phase 0 已 merge）。当前 `internal/pipeline/payload.go ExtractText` 只处理 type==1，type=14（含 image=2）一律返回 ok=false，`fetch.go` 里 `if !ok { continue }` 导致 type=14 消息**整条被丢出喂 LLM 的列表** → 图文混排的文字不进摘要。

## 任务
1. `payload.go` 的 `payloadData` 结构体加顶层 `plain` 字段。
2. `ExtractText` 加 type=14 分支：优先读顶层 `plain`；无则把 `content` 当 block 数组遍历拼接 text 块（image 块注入 `[图片]` 占位）。
3. 补 type=14 单测（含只有 plain / 只有 blocks / 两者都有）。

## 验收
- type=14 消息文字正确进入摘要列表，不再被 continue 丢弃
- go build/test 绿

Ref: octo-lib#56, YUJ-2745

yujiawei:
## Summary
`internal/pipeline/payload.go ExtractText` only handled `type==1`, so RichText (`type==14`, 图文混排) messages returned `ok=false` and were dropped by `fetch.go`'s `if !ok { continue }` — their text never reached the LLM summary list.

## Changes
- `payloadData`: `content` is now `json.RawMessage` (so it can be either a block array or a legacy string) and a top-level `plain` field is added.
- `ExtractText`: new `type==14` branch — prefer the server-generated top-level `plain`; otherwise traverse the `content` block array (`{type:text}` blocks append their text, `{type:image}` blocks inject the `[图片]` placeholder); falls back to a legacy string `content` for backward compatibility.
- Block schema (`text`/`image`, `[图片]` placeholder) mirrors octo-lib `common.RichTextBlock` / `BuildRichTextPlain` (octo-lib is not a dependency here, so the traversal is implemented minimally in-repo).
- `type==1` and the legacy string-content path are unchanged.

## Tests
Added `type=14` unit tests: plain-only / blocks-only / both (plain wins) / legacy string content / image-only / empty / base64.

## Verification
- `go build ./...` ✅
- `go vet ./internal/pipeline/` ✅
- `go test ./...` ✅ (all packages green)

## Review status
- Scope: 1 source file + tests (+178/-11), no auth/schema changes
- /review: PASS ✅ (self-review on diff; backward-compatible, no trust-boundary or SQL changes)
- /codex: independent codex-review dispatched separately per dispatch (ReviewBot)

Fixes #51

### DIFF
--- a/internal/pipeline/payload.go
+++ b/internal/pipeline/payload.go
@@ -3,15 +3,106 @@ package pipeline
 import (
 	"encoding/base64"
 	"encoding/json"
+	"strings"
 )
 
+// richTextImagePlaceholder is injected for image blocks when assembling the
+// plain text of a RichText (type=14) message. Mirrors octo-lib's
+// common.RichTextImagePlaceholder so summaries read consistently across repos.
+const richTextImagePlaceholder = "[图片]"
+
+// richTextBlock is a single element of a RichText (type=14) content array.
+// Schema mirrors octo-lib common.RichTextBlock: a text block carries Text, an
+// image block is rendered as a placeholder in plain text.
+type richTextBlock struct {
+	Type string `json:"type"`
+	Text string `json:"text"`
+}
+
+// payloadData is the decoded message payload.
+//
+// Content is kept as json.RawMessage so a RichText (type=14) payload — whose
+// content may be either an ordered block array or (for backward compatibility)
+// a plain string — can be decoded for every message type without breaking the
+// existing type==1 string path. Plain is the top-level redundant plain text the
+// server generates for RichText messages.
 type payloadData struct {
-	Type    int    `json:"type"`
-	Content string `json:"content"`
+	Type    int             `json:"type"`
+	Content json.RawMessage `json:"content"`
+	Plain   string          `json:"plain"`
+}
+
+// contentString returns the content interpreted as a JSON string. ok is false
+// when content is absent or not a JSON string (e.g. a RichText block array).
+func (d payloadData) contentString() (string, bool) {
+	if len(d.Content) == 0 {
+		return "", false
+	}
+	var s string
+	if err := json.Unmarshal(d.Content, &s); err != nil {
+		return "", false
+	}
+	return s, true
+}
+
+// richTextPlain assembles the display/summary text for a RichText (type=14)
+// payload: prefer the server-generated top-level plain; otherwise traverse the
+// content block array (text blocks contribute their text, image blocks inject
+// the [图片] placeholder); finally fall back to a string content for backward
+// compatibility with the legacy string-content path.
+func (d payloadData) richTextPlain() string {
+	if strings.TrimSpace(d.Plain) != "" {
+		return d.Plain
+	}
+	if len(d.Content) > 0 {
+		var blocks []richTextBlock
+		if err := json.Unmarshal(d.Content, &blocks); err == nil {
+			var b strings.Builder
+			for _, blk := range blocks {
+				switch blk.Type {
+				case "image":
+					b.WriteString(richTextImagePlaceholder)
+				default:
+					// text block, or unknown type carrying text (forward-compatible).
+					b.WriteString(blk.Text)
+				}
+			}
+			if b.Len() > 0 {
+				return b.String()
+			}
+		}
+	}
+	// Backward compatibility: legacy RichText with string content.
+	if s, ok := d.contentString(); ok {
+		return s
+	}
+	return ""
+}
+
+// extractFromData pulls the summary text out of a decoded payload.
+//   - type==1 (text): return the string content;
+//   - type==14 (RichText / 图文混排): return the assembled plain text;
+//   - otherwise: ("", false).
+func extractFromData(data payloadData) (string, bool) {
+	switch data.Type {
+	case 1:
+		if s, ok := data.contentString(); ok && s != "" {
+			return s, true
+		}
+		return "", false
+	case 14:
+		if text := data.richTextPlain(); text != "" {
+			return text, true
+		}
+		return "", false
+	default:
+		return "", false
+	}
 }
 
 // ExtractText decodes a message payload (raw JSON or base64-encoded JSON).
-// Returns the text content if type == 1, otherwise returns ("", false).
+// Returns the text content for type==1 (plain text) and type==14 (RichText /
+// 图文混排); otherwise returns ("", false).
 func ExtractText(payload []byte) (string, bool) {
 	if len(payload) == 0 {
 		return "", false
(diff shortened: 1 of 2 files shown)


Jerry-Xin (approved):
In scope and ready to merge; this PR fixes pipeline payload extraction for RichText messages while preserving existing plain-text behavior.

🔴 Blocking

None.

💬 Non-blocking

None.

✅ Highlights

🔵 `internal/pipeline/payload.go:53` cleanly orders RichText extraction as server plain text, block traversal, then legacy string fallback.

🔵 `internal/pipeline/payload.go:86` keeps unsupported message types filtered out, matching the existing downstream behavior in `internal/pipeline/fetch.go:334`.

🔵 `internal/pipeline/payload_test.go:86` through `internal/pipeline/payload_test.go:164` covers the important RichText cases, including plain-only, block-only, precedence, legacy content, image-only, empty content, and base64 payloads.

Verification performed: `go test ./internal/pipeline` and `go test ./...` both passed.

lml2468 (approved):
## ✅ APPROVED

Clean, well-structured change that adds RichText (type=14) support to `ExtractText`.

### What it does
Previously `ExtractText` only handled type==1 (plain text), silently dropping RichText messages from summaries. This PR adds type==14 support with a clear fallback chain: server-generated `plain` field → traverse content block array (text + image placeholders) → legacy string content → empty.

### Key design decisions — all correct
- `Content` changed from `string` to `json.RawMessage` — allows polymorphic decoding without breaking type==1 path
- `plain` field takes precedence over block traversal — avoids redundant re-computation when server already provides it
- Unknown block types fall through to `default` writing `blk.Text` — forward-compatible with future block types
- `richTextImagePlaceholder` mirrors octo-lib's `common.RichTextImagePlaceholder` — consistent cross-repo behavior

### Test coverage
7 new test cases covering: plain-only, blocks-only, plain+blocks precedence, legacy string content, image-only, empty content, and base64-encoded. Edge cases well covered.

### Blast radius
`ExtractText` is called only in `FetchMessagesFromChannel` (fetch.go). Messages that fail extraction are skipped (`if !ok { continue }`). The change means RichText messages that were previously silently dropped will now contribute to summaries — which is exactly the intended fix ("摘要丢字").

No issues found.

yujiawei (commented):
# Code Review — PR #52 (octo-smart-summary)

## Verdict: APPROVED ✅

A correct, well-scoped, backward-compatible fix for the RichText (`type=14`) text-drop bug. The summary pipeline previously dropped 图文混排 messages because `ExtractText` only handled `type==1`; this PR adds a faithful `type==14` branch. No P0/P1 blockers. Two non-blocking notes below (one to flag for the security-sensitive label, one maintainability nit).

## 1. Verification

| Item | Result | Evidence |
|---|---|---|
| Bug root cause is real | ✅ | `internal/pipeline/fetch.go:334-337` — `if !ok { continue }` drops any payload `ExtractText` rejects; pre-PR `type==14` always returned `("", false)`. |
| `type==14` extraction logic | ✅ | `internal/pipeline/payload.go:53-80` — prefers top-level `plain`, else traverses `content` blocks (text → text, image → `[图片]`), else falls back to legacy string content. |
| Matches authoritative schema | ✅ | Behavior mirrors octo-lib `common.BuildRichTextPlain` / `GetRichTextDisplayText` (placeholder `[图片]`, plain-first precedence, block traversal, legacy string fallback). Independently confirmed against octo-lib `common/richtext.go:31,129-151,255-284`. |
| `type==1` path unchanged | ✅ | `payload.go:86-92` preserves prior semantics: non-empty string content → ok; empty/non-string → `("", false)`. |
| `Content` type migration safe | ✅ | `string` → `json.RawMessage` does not regress `type==1`: `contentString()` (`payload.go:37-46`) returns `ok=false` for non-string content, yielding the same `("", false)` the old struct-unmarshal failure produced. |
| base64 + raw-JSON paths | ✅ | `extractFromData` is reused for both decode paths (`payload.go:114,134`); base64 `type=14` test covers it. |
| Build / vet / tests | ✅ | `go build ./...`, `go vet ./internal/pipeline/`, `go test ./...` all pass locally (full suite green). |
| Test coverage | ✅ | 7 new `type=14` cases: plain-only, blocks-only, plain-wins-over-blocks, legacy string, image-only, empty, base64. |

## 2. Findings

No P0 / P1 issues.

### P2 — Trust boundary (flagging for `needs-human-review`)

`richTextPlain` prefers the server-supplied top-level `plain` (`payload.go:54`) and feeds it directly into the LLM summary input. This is the **correct** choice **iff** the IM server regenerates `plain` authoritatively at ingest (octo-lib's contract: end-supplied `plain` is untrusted and must be overwritten via `FillPlainBounded`/`FillPlain` — see `octo-lib/common/richtext.go:255-265`).

- If the upstream IM store persists an end-supplied `plain` verbatim, a client could place text in `plain` that differs from `content` and have it flow into the summary.
- **Net risk is low / not newly introduced by this PR**: message text (`type==1` content included) is already end-supplied free text fed to the LLM, so this is the pipeline's pre-existing prompt-injection surface, not an escalation.
- **Ask for a human to confirm**: does the message store (WuKongIM/server) regenerate `plain` at write time, or store the client value as-is? If the latter, a follow-up could prefer re-deriving from `content` over trusting `plain`. Not a merge blocker.

### P2 — Maintainability nit (non-blocking)

`richTextBlock`, `richTextImagePlaceholder`, and the traversal logic are reimplemented in-repo, duplicating octo-lib `common.RichTextBlock` / `BuildRichTextPlain`. The PR correctly notes octo-lib isn't a dependency here, so this is reasonable for Phase 1, but it creates drift risk if the block schema evolves (e.g. new block types in "二期富文本"). Suggest a follow-up to either depend on octo-lib `common` or pin a shared golden-fixture test so the two implementations can't silently diverge. No change required now.

## 3. Suggestions

- (Optional) The default switch branch (`payload.go:65-67`) writes `blk.Text` unconditionally for unknown types; octo-lib guards with `if blk.Text != ""`. Functionally identical (empty write is a no-op), so no action needed — noting only for exact parity if octo-lib is later vendored.

## 4. Additional observations

- Whitespace handling matches octo-lib: `strings.TrimSpace(plain) != ""` gates use of `plain` but the untrimmed value is returned (`payload.go:54`), consistent with `GetRichTextDisplayText`.
- Empty/whitespace-only RichText correctly yields `ok=false`, so such messages are skipped rather than emitting blank summary entries (`payload.go:93-97`).

Overall this is a clean, defensive, well-tested fix. Approving.

PR #53, opened 2026-06-04
ci: add timeout-minutes to all CI jobs

lml2468:
## What

Add `timeout-minutes` to every CI job that is missing one.

**Timeout values:**
- `changes` (path detection): **5 min**
- `build` / `vet` / `lint` / custom lint jobs: **15 min**
- `test` / `build-and-test` / `npm-test`: **30 min**

## Why

Without explicit timeouts, a stuck test or build will run until GitHub's
6-hour default kills it — wasting runner capacity and delaying feedback.

Ref: Mininglamp-OSS workflow optimization T04.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,6 +28,7 @@ jobs:
   changes:
     name: Detect changed paths
     runs-on: ubuntu-latest
+    timeout-minutes: 5
     permissions:
       contents: read
       pull-requests: read
@@ -75,6 +76,7 @@ jobs:
       needs.changes.outputs.code == 'true'
     name: Build
     runs-on: ubuntu-latest
+    timeout-minutes: 15
     steps:
       - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
       - uses: actions/setup-go@40f1582b2485089dde7abd97c1529aa768e1baff # v5
@@ -89,6 +91,7 @@ jobs:
       needs.changes.outputs.code == 'true'
     name: Test
     runs-on: ubuntu-latest
+    timeout-minutes: 30
     needs: [changes]
     steps:
       - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
@@ -105,6 +108,7 @@ jobs:
       needs.changes.outputs.code == 'true'
     name: Vet
     runs-on: ubuntu-latest
+    timeout-minutes: 15
     steps:
       - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
       - uses: actions/setup-go@40f1582b2485089dde7abd97c1529aa768e1baff # v5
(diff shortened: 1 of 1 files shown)


Jerry-Xin (approved):
Adds job-level CI timeouts in scope for this repository and does not change workflow behavior beyond bounding runtime.

💬 Non-blocking: None.

✅ Highlights
- 🔵 Suggestion: The timeout values match the PR description and job intent: path detection at 5 minutes, build/vet/lint at 15 minutes, and tests at 30 minutes in `.github/workflows/ci.yml:31`, `.github/workflows/ci.yml:79`, `.github/workflows/ci.yml:94`, `.github/workflows/ci.yml:111`, and `.github/workflows/ci.yml:127`.
- The `timeout-minutes` keys are placed at valid job scope, alongside `runs-on`, so they apply to the whole job without altering step commands, permissions, path filtering, or conditional execution.
- No additional tests are needed for this YAML-only CI configuration change; validation by inspection is sufficient here.

yujiawei (approved):
# Code Review — PR #53 (octo-smart-summary)

## Summary

This PR adds a `timeout-minutes` setting to every job in `.github/workflows/ci.yml`. This is a sound hardening change: by default GitHub Actions jobs run with a 360-minute (6-hour) ceiling, so a hung step (e.g. a deadlocked test, a stuck `go build`, or a network stall) can silently burn runner minutes for hours. Explicit per-job timeouts cap that blast radius and surface hangs quickly.

## Verification

- **Coverage is complete.** All 5 jobs in the workflow now carry a timeout: `changes` (5m), `build` (15m), `test` (30m), `vet` (15m), `lint` (15m). No job was missed.
- **Values are well-calibrated against the actual work each job does:**
  - `changes` (5m): only runs `dorny/paths-filter` — generous headroom.
  - `build` (15m) / `vet` (15m): checkout + `setup-go` + a single Go command — comfortable.
  - `lint` (15m): the lint step itself passes `--timeout=5m` to golangci-lint, so 15m at the job level leaves room for toolchain download/cache. Job timeout > inner timeout, which is the correct relationship.
  - `test` (30m): the test step uses `go test -race -timeout 5m`. The 30m job budget covers `-race` compilation overhead, module download, and Go setup on top of the 5m test budget. Reasonable, with deliberate slack.
- **Diff is purely additive** (5 insertions, 0 deletions) and touches nothing else in the workflow — no risk to the existing `changes` preflight / draft-gating logic.
- **Syntax/placement correct:** `timeout-minutes` is a valid job-level key and is placed at the job level (sibling of `runs-on`) in every case.

## Findings

No P0/P1 issues. The change is correct, complete, and low-risk.

### Nits (non-blocking)

- The job-level timeouts protect against a hung *job*, but not against a single hung *step* within a job (a job-level timeout still allows up to N minutes per job). For this workflow that's perfectly adequate since each job has essentially one substantive step; no action needed. Mentioning only for awareness if step-level granularity is ever desired later.
- `test` at 30m is the most generous relative to its inner 5m test budget. That headroom is justified by `-race` build cost, but if CI minutes become a concern it could likely be tightened. Not a blocker.

## Verdict

Approving. This is a safe, well-scoped CI reliability improvement with sensible timeout values across all jobs.

PR #54, opened 2026-06-04
ci: add -shuffle=on to go test for order-dependency detection

lml2468:
## What

Add `-shuffle=on` flag to all `go test` invocations in CI.

## Why

`-shuffle=on` randomizes test execution order, exposing implicit ordering dependencies that cause flaky tests. This is a zero-cost quality improvement — any test that fails with shuffle enabled has a latent bug.

## Changes

Added `-shuffle=on` to `go test` commands in `.github/workflows/ci.yml`.

## Risk

None. If tests fail after this change, it reveals pre-existing hidden ordering dependencies that should be fixed.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -96,7 +96,7 @@ jobs:
         with:
           go-version: ${{ env.GO_VERSION }}
       - name: Test
-        run: go test -race -count=1 -timeout 5m ./...
+        run: go test -race -shuffle=on -count=1 -timeout 5m ./...
 
   vet:
     needs: [changes]


Jerry-Xin (approved):
This PR is relevant to the repository and safely improves CI test coverage by enabling randomized Go test order.

💬 Non-blocking

🔵 Suggestion — Consider mirroring `-shuffle=on` in the local test target as well, if the project wants local `make test` behavior to match CI. This is not required for the stated CI-only scope. Reference: `Makefile:7`.

✅ Highlights

🔵 The added flag is compatible with the repository’s configured Go version (`go 1.25`) and applies to the only CI `go test` invocation. Reference: `.github/workflows/ci.yml:99`.

🔵 I ran `go test -race -shuffle=on -count=1 -timeout 5m ./...` locally, and it passed.

yujiawei (approved):
# Code Review — PR #54 (octo-smart-summary)

## Verdict: APPROVED

A clean, well-scoped, zero-risk CI quality improvement.

## What the change does

Adds `-shuffle=on` to the single `go test` invocation in `.github/workflows/ci.yml`:

```diff
-        run: go test -race -count=1 -timeout 5m ./...
+        run: go test -race -shuffle=on -count=1 -timeout 5m ./...
```

`-shuffle=on` randomizes the execution order of tests and benchmarks within each package, surfacing implicit ordering dependencies (shared global state, fixtures leaking between tests) that would otherwise stay hidden. Any test that breaks under shuffling has a latent bug — so this is a pure-upside detection mechanism.

## Verification

- ✅ **Completeness** — `ci.yml` contains exactly one `go test` invocation (the `test` job, line 99). The diff covers it; there are no other `go test` calls in the workflow that were missed.
- ✅ **Flag placement** — `go test -race -shuffle=on -count=1 -timeout 5m ./...` is syntactically valid; all flags precede the `./...` package pattern. `-shuffle=on` composes correctly with `-race`, `-count=1`, and `-timeout`.
- ✅ **Scope** — single-line, single-file change. No build/runtime code touched. No interaction with the `changes`/`build`/`vet`/`lint` jobs.
- ✅ **Risk** — none introduced. If a test fails post-merge, it exposes a pre-existing hidden ordering dependency, which is the intended outcome. Go prints the shuffle seed on failure (`-shuffle: ...`), so any flake is reproducible via `-shuffle=<seed>`.

## Suggestions (non-blocking nits)

- **Seed reproducibility**: `-shuffle=on` picks a fresh random seed each run, so a flaky-ordering failure may not reproduce on re-run with the same code. This is the standard/recommended setting and the failure output includes the seed for targeted reproduction, so no change is needed — just be aware that an order-dependent failure might appear intermittently across CI runs rather than deterministically.
- Optional: a one-line comment next to the flag (`# -shuffle=on surfaces test-ordering dependencies`) would help future maintainers, but this is purely cosmetic.

None of the above blocks merge.

PR #57, opened 2026-06-04
fix: skip OSV-Scanner SARIF upload for Dependabot PRs

lml2468:
## What
Add Dependabot exclusion to the OSV-Scanner SARIF upload condition.

## Why
Dependabot-triggered `pull_request` workflows get a read-only `GITHUB_TOKEN`. The existing fork guard (`head.repo.full_name == github.repository`) does not catch Dependabot PRs because they are same-repo branches. This causes `security-events:write` SARIF upload to fail on every Dependabot PR.

## Fix
```yaml
upload-sarif: ${{ github.event.pull_request.head.repo.full_name == github.repository && github.event.pull_request.user.login != 'dependabot[bot]' }}
```

Scan still runs for Dependabot PRs; results visible in workflow output. Only SARIF upload (which would fail anyway) is skipped.

Follow-up to the merged OSV-Scanner PR.

### DIFF
--- a/.github/workflows/osv-scanner.yml
+++ b/.github/workflows/osv-scanner.yml
@@ -30,7 +30,8 @@ jobs:
       scan-args: |-
         -r
         ./
-      # Fork PRs have read-only GITHUB_TOKEN; skip SARIF upload to avoid
-      # security-events:write permission failure. Scan still runs and
-      # results are visible in the workflow output.
-      upload-sarif: ${{ github.event.pull_request.head.repo.full_name == github.repository }}
+      # Skip SARIF upload when GITHUB_TOKEN is read-only:
+      # - Fork PRs: token is inherently read-only
+      # - Dependabot PRs: GitHub enforces read-only token for dependabot[bot]
+      # Scan still runs; results visible in workflow output.
+      upload-sarif: ${{ github.event.pull_request.head.repo.full_name == github.repository && github.event.pull_request.user.login != 'dependabot[bot]' }}


Jerry-Xin (approved):
Summary: The PR is in scope and correctly addresses the same-repo Dependabot PR case for the OSV-Scanner SARIF upload.

✅ Highlights
- 🔵 Suggestion: No code changes requested. The condition in `.github/workflows/osv-scanner.yml:37` preserves SARIF upload for normal same-repo PRs while skipping fork PRs and Dependabot-authored PRs where the token cannot write security events.
- The scan still runs for Dependabot PRs, so this avoids the failing upload step without suppressing vulnerability scan output.
- YAML syntax parses successfully. No additional tests are necessary for this workflow-only change.

yujiawei (approved):
# Code Review — PR #57 (octo-smart-summary)

## Verdict: APPROVED

A small, correct, well-scoped CI fix. No blocking issues.

## Summary

The change extends the `upload-sarif` condition in `scan-pr` to also exclude Dependabot PRs:

```yaml
upload-sarif: ${{ github.event.pull_request.head.repo.full_name == github.repository && github.event.pull_request.user.login != 'dependabot[bot]' }}
```

The motivation is accurate: Dependabot-triggered `pull_request` runs receive a read-only `GITHUB_TOKEN`, and because Dependabot opens PRs from same-repo branches, the existing fork guard (`head.repo.full_name == github.repository`) does **not** catch them. As a result, the SARIF upload step (which needs `security-events: write`) fails on every Dependabot PR.

## Verification

- ✅ **Input typing is correct.** The pinned reusable workflow (`osv-scanner-reusable-pr.yml@v2.3.8`) declares `upload-sarif` as `type: boolean` (default `true`) and gates the upload step on `inputs.upload-sarif == true`. The `${{ ... }}` expression here evaluates to a boolean, so it composes cleanly with that input.
- ✅ **Correct event field.** `github.event.pull_request.user.login` is the appropriate context field for a `pull_request` event, and `dependabot[bot]` is the exact login Dependabot uses.
- ✅ **Root-cause analysis is right.** The fork guard cannot detect Dependabot because Dependabot branches live in the same repository; an actor/login check is the standard way to handle this.
- ✅ **No reduction in security posture.** The scan itself still runs for Dependabot PRs — only the SARIF *upload* (which would otherwise hard-fail due to the read-only token) is skipped. There is no token-permission escalation, no `pull_request_target` introduction, and no new exposure of secrets. This is an availability/correctness fix, not a loosening of the security gate.

## Non-blocking notes (nits / suggestions)

1. **Visibility tradeoff.** Skipping the upload means Dependabot-PR findings will not surface in the GitHub Security → Code Scanning tab; reviewers must read the workflow logs. This is the right tradeoff given the alternative is a hard failure, but it's worth being aware of. (No action required.)
2. **Legacy actor.** The condition does not exclude the deprecated `dependabot-preview[bot]` login. In practice GitHub has fully migrated to `dependabot[bot]`, so this is almost certainly irrelevant — mentioned only for completeness.

Neither note blocks merge.

PR #59, opened 2026-06-04
ci: set fail-on-vuln false on scheduled OSV scan

lml2468:
## What

Add `fail-on-vuln: false` to the `scan-scheduled` OSV job.

## Why

Without this, every push to main and every weekly scheduled run fails
on pre-existing vulnerabilities — creating permanent red CI unrelated
to the change being pushed.

With `fail-on-vuln: false`, results are still uploaded to the Security
tab (SARIF) for tracking, but don't block the workflow. PR scans still
block new vulnerabilities (default `fail-on-vuln: true`).

### DIFF
--- a/.github/workflows/osv-scanner.yml
+++ b/.github/workflows/osv-scanner.yml
@@ -22,6 +22,7 @@ jobs:
       scan-args: |-
         -r
         ./
+      fail-on-vuln: false
 
   scan-pr:
     if: github.event_name == 'pull_request'


Jerry-Xin (approved):
One-line summary: The PR is in scope and correctly makes only non-PR OSV scans non-blocking while preserving blocking behavior for PR scans.

🔴 Blocking  
None.

💬 Non-blocking  
None.

✅ Highlights  
🔵 Suggestion: No change requested. In [.github/workflows/osv-scanner.yml](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-lv8cig93/repo/.github/workflows/osv-scanner.yml:18), `scan-scheduled` covers `push`, `schedule`, and `workflow_dispatch`, so adding `fail-on-vuln: false` at line 25 matches the stated intent.

🔵 Suggestion: PR scans remain on the separate `scan-pr` job at [.github/workflows/osv-scanner.yml](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-lv8cig93/repo/.github/workflows/osv-scanner.yml:27), with no override for `fail-on-vuln`, so the default blocking behavior is preserved.

No tests are needed for this workflow-only configuration change.

yujiawei (approved):
# Code Review — PR #59 (octo-smart-summary)

## Summary

This PR adds a single line — `fail-on-vuln: false` — to the `scan-scheduled` OSV-Scanner job, so that pushes to `main` and the weekly scheduled scan no longer fail CI on pre-existing vulnerabilities. The change is small, well-scoped, and the rationale in the description is sound.

**Verdict: APPROVED.** No P0/P1 issues found.

## Verification

✅ **Input is valid.** The reusable workflow pinned at `google/osv-scanner-action/.github/workflows/osv-scanner-reusable.yml@9a49870 (v2.3.8)` declares `fail-on-vuln` as a `boolean` input (default `true`) and forwards it to the reporter step as `--fail-on-vuln=${{ inputs.fail-on-vuln }}`. The new key is correctly placed under `with:` and YAML-indented to match `scan-args`.

✅ **PR scans remain gated.** The `scan-pr` job (triggered on `pull_request`) is untouched and keeps the default `fail-on-vuln: true`, so newly introduced vulnerabilities in PRs still block merge. Only the `push`/`schedule`/`workflow_dispatch` path is relaxed.

✅ **Visibility is preserved.** SARIF upload for the scheduled job is unaffected (`upload-sarif` defaults to `true`), so results continue to flow to the Security → Code Scanning tab for tracking. The scan still runs; it simply no longer hard-fails the workflow.

✅ **No security regression in the gating model.** Push-to-`main` events only occur after a PR merge, and the PR path already enforces `fail-on-vuln: true`. So the merge gate against *new* vulnerabilities is retained; this change only stops *pre-existing* vulns from producing permanent red CI on `main` — which is the stated and reasonable intent.

## Notes for human reviewer (security_sensitive)

- This intentionally relaxes the failure behavior on `main`. The trade-off is acceptable: pre-existing-vuln noise is removed while the PR gate and Security-tab tracking stay intact. Confirm the team is comfortable relying on the Security tab (rather than a red CI signal) to drive remediation of already-known vulnerabilities on `main`.
- No secrets, permissions, or token scopes are changed by this diff.

## Suggestions (non-blocking)

- Consider a brief inline comment on the `scan-scheduled` job mirroring the PR description's rationale (as already done for `scan-pr`'s `upload-sarif`), so the intent of `fail-on-vuln: false` is self-documenting for future readers.

PR #61, opened 2026-06-05
feat: backend support for chat-window smart summary (origin_channel + summary templates)

### LINKED ISSUE #60: feat: backend support for chat-window smart summary (origin_channel + summary templates)
## Overview

Backend work to support the chat-window smart summary feature. This tracks the API and data-model changes required so the frontend can request summaries originating from the chat window and offer users a set of built-in summary templates.

Related branch: `feat/chat-window-smart-summary`

### 1. `origin_channel` support

Add an `origin_channel` field so summary tasks can record where the request originated (e.g. the chat window).

- `internal/api/handler/task.go` — accept and propagate `origin_channel`
- `internal/api/router/router.go` — wiring
- `internal/model/model.go` — model field
- `internal/api/handler/origin_channel_test.go` — tests
- `migrations/sql/20260528-01-add-origin-channel.sql` — schema migration

### 2. Summary templates — `GET /api/v1/summary-templates`

Add a `GetTemplates` handler returning the built-in summary templates, replacing the previous empty-array stub.

- `internal/api/handler/task.go` — `GetTemplates` handler with 4 built-in templates: `project_progress`, `task_tracking`, `weekly_report`, `chat_content`
- `internal/api/router/router.go` — route `GET /api/v1/summary-templates`
- `internal/api/handler/origin_channel_test.go` — tests

## Related

- Frontend: Mininglamp-OSS/octo-web#129

jeff-wilson2010:
## Summary

Backend support for the chat-window smart summary feature.

### origin_channel support
- Threaded `origin_channel` through the request **handler**, **router**, **model**, and added a database **migration** so summaries can record and filter by their originating channel.

### Summary templates
- Added `GetTemplates` exposed via `GET /api/v1/summary-templates`.
- Ships 4 built-in templates:
  - `project_progress`
  - `task_tracking`
  - `weekly_report`
  - `chat_content`

### Tests
- Added tests covering the new `origin_channel` handling and the summary-templates endpoint.

Closes #60

### DIFF
--- a/internal/api/handler/auth_test.go
+++ b/internal/api/handler/auth_test.go
@@ -2,6 +2,7 @@ package handler
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"net/http"
 	"net/http/httptest"
@@ -128,15 +129,16 @@ func TestAuthorizeTaskAccess_Participant(t *testing.T) {
 	}
 }
 
-func TestAuthorizeTaskAccess_GroupMember(t *testing.T) {
+func TestAuthorizeTaskAccess_GroupMemberDenied(t *testing.T) {
 	db, imDB := setupTestDBs(t)
 	taskID := seedTask(t, db, imDB)
 	h := NewTaskHandler(db, imDB, "")
 	r := setupRouter(h)
 
+	// groupmember1 is only a source-group member, neither creator nor participant → 403
 	w := doRequest(r, "GET", fmt.Sprintf("/api/v1/summaries/%d", taskID), "groupmember1")
-	if w.Code != http.StatusOK {
-		t.Errorf("expected 200 for group member, got %d: %s", w.Code, w.Body.String())
+	if w.Code != http.StatusForbidden {
+		t.Errorf("expected 403 for source-group member, got %d: %s", w.Code, w.Body.String())
 	}
 }
 
@@ -194,5 +196,155 @@ func TestCancelSummary_RequiresAuth(t *testing.T) {
 	w = doRequest(r, "POST", fmt.Sprintf("/api/v1/summaries/%d/cancel", task.ID), "creator1")
 	if w.Code != http.StatusOK {
 		t.Errorf("expected 200 for cancel by creator, got %d: %s", w.Code, w.Body.String())
+	}
+}
+
+func TestDeleteSummary_GroupMemberDenied(t *testing.T) {
+	db, imDB := setupTestDBs(t)
+	taskID := seedTask(t, db, imDB)
+	h := NewTaskHandler(db, imDB, "")
+	r := setupRouter(h)
+
+	// Source-group member must not be able to delete another user's summary.
+	w := doRequest(r, "DELETE", fmt.Sprintf("/api/v1/summaries/%d", taskID), "groupmember1")
+	if w.Code != http.StatusForbidden {
+		t.Errorf("expected 403 for delete by source-group member, got %d: %s", w.Code, w.Body.String())
+	}
+}
+
+func TestCancelSummary_GroupMemberDenied(t *testing.T) {
+	db, imDB := setupTestDBs(t)
+	h := NewTaskHandler(db, imDB, "")
+
+	// Create a pending task whose source group contains groupmember1.
+	task := model.SummaryTask{
+		TaskNo:      "TST-003",
+		SpaceID:     "space1",
+		CreatorID:   "creator1",
+		SummaryMode: model.ModeByPerson,
+		Status:      model.StatusPending,
+	}
+	db.Create(&task)
+	db.Create(&model.SummarySource{TaskID: task.ID, SourceType: model.SourceGroup, SourceID: "grp_abc"})
+	imDB.Exec("INSERT INTO group_member (group_no, uid, is_deleted) VALUES (?, ?, 0)", "grp_abc", "groupmember1")
+
+	r := setupRouter(h)
+
+	// Source-group member must not be able to cancel another user's summary.
+	w := doRequest(r, "POST", fmt.Sprintf("/api/v1/summaries/%d/cancel", task.ID), "groupmember1")
+	if w.Code != http.StatusForbidden {
+		t.Errorf("expected 403 for cancel by source-group member, got %d: %s", w.Code, w.Body.String())
+	}
+}
+
+// setupListTestDBs migrates the extra tables ListSummaries touches (summary_result)
+// to avoid no-such-table noise during list rendering.
+func setupListTestDBs(t *testing.T) (db *gorm.DB, imDB *gorm.DB) {
+	t.Helper()
+
+	db, err := gorm.Open(sqlite.Open(":memory:"), &gorm.Config{})
+	if err != nil {
+		t.Fatalf("open summary db: %v", err)
+	}
+	db.AutoMigrate(&model.SummaryTask{}, &model.SummarySource{}, &model.SummaryParticipant{}, &model.SummaryResult{})
+
+	imDB, err = gorm.Open(sqlite.Open(":memory:"), &gorm.Config{})
+	if err != nil {
+		t.Fatalf("open im db: %v", err)
+	}
+	imDB.Exec("CREATE TABLE group_member (group_no TEXT NOT NULL, uid TEXT NOT NULL, is_deleted INTEGER DEFAULT 0)")
+
+	return db, imDB
+}
+
+func setupListRouter(h *TaskHandler) *gin.Engine {
+	gin.SetMode(gin.TestMode)
+	r := gin.New()
+	r.Use(middleware.AuthMiddleware(&mockTokenResolver{}), middleware.SpaceMiddleware())
+	r.GET("/api/v1/summaries", h.ListSummaries)
+	return r
+}
+
+type listResponse struct {
+	Code int `json:"code"`
+	Data struct {
+		Total int           `json:"total"`
+		Items []interface{} `json:"items"`
+	} `json:"data"`
+}
+
+func parseListResponse(t *testing.T, w *httptest.ResponseRecorder) listResponse {
+	t.Helper()
+	var resp listResponse
+	if err := json.Unmarshal(w.Body.Bytes(), &resp); err != nil {
+		t.Fatalf("unmarshal response: %v\nbody: %s", err, w.Body.String())
+	}
+	return resp
+}
+
+func seedListTask(t *testing.T, db *gorm.DB, imDB *gorm.DB) {
+	t.Helper()
+
+	task := model.SummaryTask{
+		TaskNo:      "TST-LIST-001",
+		SpaceID:     "space1",
+		CreatorID:   "creator1",
+		SummaryMode: model.ModeByPerson,
+		Status:      model.StatusCompleted,
+	}
+	db.Create(&task)
+	db.Create(&model.SummarySource{TaskID: task.ID, SourceType: model.SourceGroup, SourceID: "grp_abc"})
+	db.Create(&model.SummaryParticipant{TaskID: task.ID, UserID: "participant1", UserName: "P1"})
+	imDB.Exec("INSERT INTO group_member (group_no, uid, is_deleted) VALUES (?, ?, 0)", "grp_abc", "groupmember1")
+}
+
+func TestListSummaries_GroupMemberSeesNothing(t *testing.T) {
+	db, imDB := setupListTestDBs(t)
+	seedListTask(t, db, imDB)
+	h := NewTaskHandler(db, imDB, "")
+	r := setupListRouter(h)
+
+	w := doRequest(r, "GET", "/api/v1/summaries", "groupmember1")
+	if w.Code != http.StatusOK {
+		t.Fatalf("expected 200, got %d: %s", w.Code, w.Body.String())
+	}
+	resp := parseListResponse(t, w)
+	if resp.Data.Total != 0 {
+		t.Errorf("expected total 0 for source-group member, got %d", resp.Data.Total)
+	}
+	if len(resp.Data.Items) != 0 {
+		t.Errorf("expected 0 items for source-group member, got %d", len(resp.Data.Items))
+	}
+}
+
+func TestListSummaries_CreatorSeesTask(t *testing.T) {
+	db, imDB := setupListTestDBs(t)
+	seedListTask(t, db, imDB)
+	h := NewTaskHandler(db, imDB, "")
+	r := setupListRouter(h)
+
+	w := doRequest(r, "GET", "/api/v1/summaries", "creator1")
+	if w.Code != http.StatusOK {
+		t.Fatalf("expected 200, got %d: %s", w.Code, w.Body.String())
+	}
+	resp := parseListResponse(t, w)
+	if resp.Data.Total != 1 {
+		t.Errorf("expected total 1 for creator, got %d", resp.Data.Total)
+	}
+}
+
+func TestListSummaries_ParticipantSeesTask(t *testing.T) {
+	db, imDB := setupListTestDBs(t)
+	seedListTask(t, db, imDB)
+	h := NewTaskHandler(db, imDB, "")
+	r := setupListRouter(h)
+
+	w := doRequest(r, "GET", "/api/v1/summaries", "participant1")
+	if w.Code != http.StatusOK {
+		t.Fatalf("expected 200, got %d: %s", w.Code, w.Body.String())
+	}
+	resp := parseListResponse(t, w)
+	if resp.Data.Total != 1 {
+		t.Errorf("expected total 1 for participant, got %d", resp.Data.Total)
 	}
 }
--- a/internal/api/handler/members_auth_test.go
+++ b/internal/api/handler/members_auth_test.go
@@ -1,0 +1,121 @@
+package handler
+
+import (
+	"fmt"
+	"net/http"
+	"testing"
+
+	"github.com/Mininglamp-OSS/octo-smart-summary/internal/middleware"
+	"github.com/Mininglamp-OSS/octo-smart-summary/internal/model"
+	"github.com/gin-gonic/gin"
+	"gorm.io/driver/sqlite"
+	"gorm.io/gorm"
+)
+
+func setupMembersTestDB(t *testing.T) *gorm.DB {
+	t.Helper()
+
+	db, err := gorm.Open(sqlite.Open(":memory:"), &gorm.Config{})
+	if err != nil {
+		t.Fatalf("open summary db: %v", err)
+	}
+	db.AutoMigrate(&model.SummaryTask{}, &model.SummarySource{}, &model.SummaryParticipant{}, &model.PersonalResult{})
+	return db
+}
+
+func setupMembersRouter(h *PersonalHandler) *gin.Engine {
+	gin.SetMode(gin.TestMode)
+	r := gin.New()
+	r.Use(middleware.AuthMiddleware(&mockTokenResolver{}), middleware.SpaceMiddleware())
+	r.GET("/api/v1/summaries/:id/members", h.GetMembers)
+	return r
+}
+
+func seedMembersTask(t *testing.T, db *gorm.DB) int64 {
+	t.Helper()
+
+	task := model.SummaryTask{
+		TaskNo:      "TST-MEMBERS-001",
+		SpaceID:     "space1",
+		CreatorID:   "creator1",
+		SummaryMode: model.ModeByPerson,
+		Status:      model.StatusCompleted,
+	}
+	db.Create(&task)
+	db.Create(&model.SummarySource{TaskID: task.ID, SourceType: model.SourceGroup, SourceID: "grp_abc"})
+	db.Create(&model.SummaryParticipant{TaskID: task.ID, UserID: "participant1", UserName: "P1"})
+	return task.ID
+}
+
+func TestGetMembers_Creator(t *testing.T) {
+	db := setupMembersTestDB(t)
+	taskID := seedMembersTask(t, db)
+	h := NewPersonalHandler(db, "", nil)
+	r := setupMembersRouter(h)
+
+	w := doRequest(r, "GET", fmt.Sprintf("/api/v1/summaries/%d/members", taskID), "creator1")
+	if w.Code != http.StatusOK {
+		t.Errorf("expected 200 for creator, got %d: %s", w.Code, w.Body.String())
+	}
+}
+
+func TestGetMembers_Participant(t *testing.T) {
+	db := setupMembersTestDB(t)
+	taskID := seedMembersTask(t, db)
+	h := NewPersonalHandler(db, "", nil)
+	r := setupMembersRouter(h)
+
+	w := doRequest(r, "GET", fmt.Sprintf("/api/v1/summaries/%d/members", taskID), "participant1")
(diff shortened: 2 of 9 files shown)


lml2468 (changes requested):
I cannot approve this PR while the required gate checks are failing.

Blocking items:
1. PR checks: `check-sprint / check-sprint` and `welcome / welcome` are failing on head SHA `3d203e1278b482d8092ecfdb3bc56f51c4fe2d4e`; please fix or explain the failing checks and re-run CI before code review continues.

Non-blocking notes:
- I stopped at Phase 1 per the review instructions because CI is red; I did not perform the deep file-by-file review.

Highlights:
- The branch is currently mergeable with no reported conflicts.

Jerry-Xin (changes requested):
The PR is in scope, but the new channel-filtered summary list has an authorization/query mismatch that breaks the chat-window discovery flow.

**🔴 Blocking**
🔴 Critical: `GET /api/v1/summaries?origin_channel_id=...` does not return channel summaries to authorized group members. `CreateSummary` auto-fills a group source from `origin_channel_id` when no sources are provided, and `authorizeTaskAccess` allows active source group members to read the task detail. However, `ListSummaries` starts from `creator_id = userID OR participant userID` only, then applies `origin_channel_id`, so another member of the same group gets `total=0` even though `GET /api/v1/summaries/:id` would authorize them. That makes the new chat-window channel filter unusable for non-creator group members. Fix the list authorization to include source-group membership for channel-filtered lists, or otherwise align the intended access model and tests. See `internal/api/handler/task.go:337`, `internal/api/handler/task.go:354`, and `internal/api/handler/task.go:64`.

**💬 Non-blocking**
🟡 Warning: Filtering only by `origin_channel_id` may mix channel namespaces if group/thread/direct IDs are not globally unique. The API stores `origin_channel_type`, but `ListSummaries` does not expose a matching filter and the migration indexes only `origin_channel_id`. Consider supporting `origin_channel_type` in the query and using a composite index. See `internal/api/handler/task.go:354` and `migrations/sql/20260528-01-add-origin-channel.sql:5`.

🔵 Suggestion: Add a regression test for the group-member list case above. The new tests cover creator-owned filtering but not the access path that `authorizeTaskAccess` already supports for source groups.

**✅ Highlights**
The PR cleanly threads `origin_channel_id` and `origin_channel_type` into create/list/detail responses, and `GET /api/v1/summary-templates` is wired through the real handler instead of the previous stub. `go test ./...` passes locally.

Jerry-Xin (changes requested):
The PR is in scope, but the new channel-filtered summary list has an authorization/query mismatch that breaks the chat-window discovery flow.

**🔴 Blocking**
🔴 Critical: `GET /api/v1/summaries?origin_channel_id=...` does not return channel summaries to authorized group members. `CreateSummary` auto-fills a group source from `origin_channel_id` when no sources are provided, and `authorizeTaskAccess` allows active source group members to read the task detail. However, `ListSummaries` starts from `creator_id = userID OR participant userID` only, then applies `origin_channel_id`, so another member of the same group gets `total=0` even though `GET /api/v1/summaries/:id` would authorize them. That makes the new chat-window channel filter unusable for non-creator group members. Fix the list authorization to include source-group membership for channel-filtered lists, or otherwise align the intended access model and tests. See `internal/api/handler/task.go:337`, `internal/api/handler/task.go:354`, and `internal/api/handler/task.go:64`.

**💬 Non-blocking**
🟡 Warning: Filtering only by `origin_channel_id` may mix channel namespaces if group/thread/direct IDs are not globally unique. The API stores `origin_channel_type`, but `ListSummaries` does not expose a matching filter and the migration indexes only `origin_channel_id`. Consider supporting `origin_channel_type` in the query and using a composite index. See `internal/api/handler/task.go:354` and `migrations/sql/20260528-01-add-origin-channel.sql:5`.

🔵 Suggestion: Add a regression test for the group-member list case above. The new tests cover creator-owned filtering but not the access path that `authorizeTaskAccess` already supports for source groups.

**✅ Highlights**
The PR cleanly threads `origin_channel_id` and `origin_channel_type` into create/list/detail responses, and `GET /api/v1/summary-templates` is wired through the real handler instead of the previous stub. `go test ./...` passes locally.

lml2468 (changes requested):
Clean implementation of `origin_channel` threading and `summary-templates` endpoint. Code is well-structured, migration has proper Up/Down, and test coverage is solid. However, there is a real authorization mismatch that blocks the chat-window use case.

**🔴 Blocking**

1. **ListSummaries authorization gap (agree with Jerry's finding, adding detail):** `ListSummaries` scopes queries to `creator_id = ? OR id IN (SELECT task_id FROM summary_participant WHERE user_id = ?)` at `task.go:337-339`. `authorizeTaskAccess` (used by `GetSummary`) additionally checks source group membership via `imDB.Table("group_member")` at `task.go:66-75`. This means a group member can view a specific summary (`GET /summaries/:id`) but cannot discover it via `GET /summaries?origin_channel_id=...` — the exact API the chat window would call. The fix should widen the `ListSummaries` base query when `origin_channel_id` is present (e.g., `OR id IN (SELECT task_id FROM summary_source WHERE source_type=1 AND source_id IN (SELECT group_no FROM group_member WHERE uid=? AND is_deleted=0))`), matching the existing `authorizeTaskAccess` logic.

**💬 Non-blocking**

2. **Implicit SourceType ↔ OriginChannelType coupling** (`task.go:161-164`): The auto-fill assigns `SourceType: req.OriginChannelType` directly. This works today because `SourceGroup/Thread/Direct = 1/2/3` aligns with `OriginChannelGroup/Thread/DM = 1/2/3`, but if either constant set drifts independently, the mapping breaks silently. Consider an explicit mapping function or at minimum a compile-time assertion comment.

3. **`origin_channel_type` query filter** (agree with Jerry): The list endpoint filters by `origin_channel_id` alone. If channel IDs are not globally unique across types, consider supporting `origin_channel_type` as a secondary filter and a composite index `(origin_channel_id, origin_channel_type)`.

**✅ Highlights**

- The stub `GET /summary-templates` route replaced with a real handler is a clean upgrade.
- 360 lines of test coverage with real SQLite — good quality, not mock-heavy.
- Migration has proper `+migrate Up/Down` markers with correct drop order (index before columns).

yujiawei (approved):
# Code Review — PR #61 (octo-smart-summary)

## Summary

Backend support for chat-window smart summary: threads `origin_channel` (id + type) through the create handler, model, list/get responses, a list filter, and a DB migration; replaces the stubbed `summary-templates` endpoint with a real 4-template implementation. The change is well-scoped, well-tested, and builds cleanly.

**Verdict: APPROVED.** No P0/P1 issues. Build, `go vet`, and the package tests all pass locally against head `3d203e1`. A couple of non-blocking nits below.

## Verification

- ✅ **Build**: `CGO_ENABLED=0 go build ./...` passes.
- ✅ **Vet**: `go vet ./internal/api/... ./internal/model/...` clean (govet is the only enabled linter in `.golangci.yml`).
- ✅ **Tests**: `go test ./internal/api/handler/...` passes, including the 8 new cases in `origin_channel_test.go` (validation, auto-fill, no-auto-fill-when-sources-provided, list filter, get fields, templates structure).
- ✅ **Migration**: `migrations/sql/20260528-01-add-origin-channel.sql` adds `origin_channel_id`/`origin_channel_type` + index, with a matching Down. Column placement (`AFTER schedule_id`) matches the struct field order in `model.go`. Filename ordering is consistent with the existing `sql-migrate` embedded source.
- ✅ **Constant alignment**: `OriginChannelGroup/Thread/DM = 1/2/3` line up with the `Source*` constants (1/2/3) and the frontend `source_type` values consumed by `mapFrontendSourceType` (1=group, 2=thread, 3=DM). The auto-fill in `CreateSummary` reuses `OriginChannelType` directly as `source_type`, which is correct under this alignment.

## Security review (this PR was flagged security-sensitive)

- ✅ **No new data-leak surface from auto-fill.** When `origin_channel_id` is set and no explicit sources are provided, the handler synthesizes a source `{type: origin_channel_type, id: origin_channel_id}` (`task.go:161-166`). This does not bypass access control: the worker pipeline still calls `GetUserChannels(creatorUID)` and `ApplySourceConstraints` (`fetch.go:566`, `fetch.go:598`), which intersect requested sources against the creator's *actual* group/DM/thread memberships. A caller supplying a channel ID they don't belong to gets zero messages — the synthesized source cannot exfiltrate content from a channel the creator isn't in.
- ✅ **List filter is parameterized.** `origin_channel_id` filter uses a bound placeholder (`task.go:354-356`) and is applied on top of the existing space + creator/participant scoping (`task.go:337-339`), so it cannot widen visibility.
- ✅ **Validation is sound.** Type is range-checked only when an id is present, and an orphan type (id empty, type non-zero) is rejected (`task.go:153-160`). Both paths covered by tests.
- ⚠️ **Note for a human (pre-existing, NOT introduced here):** `ListSummaries` builds its ORDER BY as `sortBy + " " + sortOrder` (`task.go:371-375`). `sortBy` is whitelisted, but `sortOrder` is taken raw from the query string and concatenated. This predates this PR (introduced Apr 26, commit `5853277`) and is out of scope, but since this file is being touched and the PR is security-flagged, it's worth a follow-up to whitelist `sortOrder` to `asc`/`desc`.

## Non-blocking nits

1. **gofmt.** `internal/api/handler/task.go` and `internal/model/model.go` are not gofmt-clean (struct-tag/map-key alignment in the edited blocks, e.g. the `gin.H` maps in `ListSummaries`/`GetSummary` and the `createSummaryReq` struct). CI does not enforce gofmt (only govet via golangci-lint), so this won't break the build, but running `gofmt -w` on the two files would keep the tree tidy.
2. **Template content is hardcoded Chinese UI strings in Go.** Fine for now given the 4 fixed templates, but if these grow or need i18n, consider moving them to config/data rather than the handler. Purely a maintainability suggestion.

Nice work on the test coverage — the auto-fill / no-auto-fill distinction and the list-filter case are exactly the edges worth pinning down.

yujiawei (changes requested):
# Code Review — PR #61 (octo-smart-summary)

> Correction to my earlier approval on this PR. After re-examining the access model in light of other reviewers' findings, I agree there is a blocking authorization gap. Changing my verdict to **Changes Requested**.

## Summary

Backend support for chat-window smart summary: threads `origin_channel` (id + type) through the create handler, model, list/get responses, a list filter, and a DB migration; replaces the stubbed `summary-templates` endpoint with a real 4-template implementation. The implementation is clean, builds, and is well tested — but the new channel-discovery list path has an authorization mismatch that makes the feature unusable for the intended audience.

**Verdict: CHANGES_REQUESTED** (1 blocking P1).

## 🔴 Blocking

**P1 — `ListSummaries` authorization is narrower than `authorizeTaskAccess`, breaking channel discovery for group members.**

- `authorizeTaskAccess` (used by `GET /summaries/:id`) grants access to the creator, a participant, **or a member of any source group** via `imDB.group_member` (`internal/api/handler/task.go:64-77`).
- `ListSummaries` scopes its base query to `creator_id = ? OR id IN (SELECT task_id FROM summary_participant WHERE user_id = ?)` only (`internal/api/handler/task.go:337-339`), then applies the new `origin_channel_id` filter (`task.go:354-356`).
- `CreateSummary` auto-fills the source group from `origin_channel_id` and defaults participants to the creator alone (`task.go:161-166`, `task.go:202-204`).

Net effect: a group member who is neither the creator nor an explicit participant can open a specific channel summary (`GET /summaries/:id` authorizes them through source-group membership) but gets `total=0` from `GET /summaries?origin_channel_id=...` — which is precisely the endpoint the chat window calls to discover a channel's summaries. The feature this PR ships does not work for non-creator group members.

This confirms the finding raised by the other reviewers. I had initially only checked the over-permissive direction (a caller passing a channel they don't belong to gets zero messages — that part is safe, since the worker pipeline still intersects sources against `GetUserChannels`). I missed the under-permissive direction, which is the actual blocker.

**Suggested fix:** widen the `ListSummaries` base query to include source-group membership, mirroring `authorizeTaskAccess`, e.g.:

```
OR id IN (
  SELECT task_id FROM summary_source
  WHERE source_type = 1
    AND source_id IN (SELECT group_no FROM group_member WHERE uid = ? AND is_deleted = 0)
)
```

(`group_member` lives in `imDB`, so this likely needs a two-step query: resolve the user's groups from `imDB`, then filter `summary_source` in the primary DB — the same split `authorizeTaskAccess` already uses.) Add a regression test for the group-member-not-creator list case; the new tests cover creator-owned filtering but not this access path.

## 💬 Non-blocking

1. **Filter by `origin_channel_id` alone may mix channel namespaces.** Group/thread/DM IDs may not be globally unique. The API stores `origin_channel_type` but `ListSummaries` exposes no matching filter, and the migration indexes only `origin_channel_id`. Consider adding an `origin_channel_type` query filter and a composite index `(origin_channel_id, origin_channel_type)`. (`task.go:354`, `migrations/sql/20260528-01-add-origin-channel.sql:5`)
2. **Implicit `SourceType` ↔ `OriginChannelType` coupling** (`task.go:161-164`): auto-fill assigns `SourceType: req.OriginChannelType` directly. This works because `SourceGroup/Thread/Direct = 1/2/3` matches `OriginChannelGroup/Thread/DM = 1/2/3`, but if either set drifts the mapping breaks silently. Consider an explicit mapping helper or a comment noting the required alignment.
3. **CI is red on head SHA `3d203e1`.** `check-sprint` and `welcome` checks are failing. These look like repo housekeeping rather than code defects, but they should be green (or explained) before merge.
4. **gofmt.** `internal/api/handler/task.go` and `internal/model/model.go` are not gofmt-clean in the edited blocks (struct-tag / map-key alignment). CI enforces only govet, so this won't break the build; `gofmt -w` would tidy it.

## ✅ Highlights

- Clean threading of `origin_channel_id`/`origin_channel_type` into create/list/detail.
- Real `GET /summary-templates` handler replacing the previous stub.
- Solid test coverage (360 lines, real in-memory SQLite, covers validation / auto-fill / no-auto-fill / list filter / detail / templates).
- Migration has proper Up/Down with correct drop order (index before columns).

## Verification performed

- ✅ `CGO_ENABLED=0 go build ./...` passes.
- ✅ `go vet ./internal/api/... ./internal/model/...` clean.
- ✅ `go test ./internal/api/handler/...` passes.
- ⚠️ The blocking issue above is a behavioral gap not caught by the existing tests (no group-member-list test exists).

yujiawei (approved):
# Code Review — PR #61, updated for new commit `37dc17f`

> Re-review prompted by the new commit pushed on top of `3d203e1`. My previous verdict was **CHANGES_REQUESTED** (P1: `ListSummaries` authorization was narrower than `authorizeTaskAccess`, so source-group members could open a summary by id but got `total=0` from the channel-discovery list). This commit resolves that mismatch. **New verdict: APPROVED.**

## What the new commit does

`fix: restrict summary access to creator and participants only` (`37dc17f`) closes the inconsistency by **tightening** rather than widening — the opposite direction from what the reviewers suggested, but it produces a single, internally consistent access model:

- **`canAccessTask`** introduced as the single source of truth: access = creator **or** explicit participant. Source-group membership alone no longer grants access.
- **`authorizeTaskAccess`** (detail / delete / cancel) now delegates to `canAccessTask`; the `imDB.group_member` branch is removed.
- **`batchAuthorize`** drops its source-group branch, with a comment noting it must stay semantically equal to `canAccessTask` (Pluck form is only a perf optimization).
- **`ListSummaries`** base query was already `creator_id = ? OR participant`, so it now matches the detail path — the original mismatch is gone.
- **`GetMembers`** (`personal.go`) gains the missing authorization check (it previously had none) — creator/participant only, 4010/40008/40003 codes consistent with the task handler.

This is a defensible product decision: summaries are private to their creator and explicit participants, not readable by every member of a source group. It is the more secure default, and it makes all four read paths (detail, list, batch, members) behave identically. The chat-window create flow still works for the creator; if non-creator group-member discovery is later required, it should be added deliberately on all paths at once.

## Verification (against head `37dc17f`)

- ✅ `CGO_ENABLED=0 go build ./...` passes.
- ✅ `go vet ./internal/api/... ./internal/model/...` clean.
- ✅ `go test ./...` passes (full module, including the new regression tests).
- ✅ CI all green on head SHA, including the previously-failing `check-sprint` and `welcome` gates flagged in the earlier review.
- ✅ New regression coverage added for exactly the gaps the reviews called out: `TestAuthorizeTaskAccess_GroupMemberDenied`, `TestDeleteSummary_GroupMemberDenied`, `TestCancelSummary_GroupMemberDenied`, `TestListSummaries_GroupMemberSeesNothing` / `CreatorSeesTask` / `ParticipantSeesTask`, `TestBatchStatus_GroupMemberExcluded`, and a full `members_auth_test.go` (creator/participant/group-member/stranger/no-auth/not-found).
- ✅ `gofmt -l` clean on `task.go` and `personal.go` (the struct-tag / map-key alignment nits from the earlier review are now fixed).

## 💬 Non-blocking (carried over, none blocking)

1. **`origin_channel_type` filter / composite index** — `ListSummaries` still filters by `origin_channel_id` alone; the migration indexes only `origin_channel_id`. If channel IDs aren't globally unique across group/thread/DM, a follow-up could add an `origin_channel_type` filter + composite index. Lower urgency now that the list is creator/participant-scoped.
2. **Implicit `SourceType` ↔ `OriginChannelType` coupling** (`task.go` auto-fill) — still assigns `SourceType: req.OriginChannelType` directly; works because both sets are 1/2/3. A mapping helper or alignment comment would prevent silent drift.
3. **`sortOrder` concatenation in `ListSummaries` ORDER BY** — pre-existing (not introduced here), whitelisting `asc`/`desc` remains a good follow-up.
4. **`model.go` not gofmt-clean** — pre-existing and untouched by this commit; CI enforces only govet so it doesn't block.

## ✅ Highlights

- Resolves the P1 authorization mismatch all three reviewers raised, with a clean single-predicate design.
- Closes a real pre-existing hole: `GetMembers` had no authorization at all.
- Strong regression coverage for every access path, including the negative (denied) cases.
- gofmt cleanup of the touched files; CI fully green.

lml2468 (approved):
Incremental re-review of 37dc17fe — verifying the authorization inconsistency fix from the first round.

**Previous blockers — resolved ✅**

The authorization gap (`ListSummaries` vs `authorizeTaskAccess`) is fixed by **tightening** all auth paths to creator + explicit participant only, rather than widening `ListSummaries`. This is the fail-closed approach and is the correct design choice:

1. `canAccessTask` is now the single source of truth (`task.go:32-43`).
2. `authorizeTaskAccess` delegates to `canAccessTask` (`task.go:65`).
3. `batchAuthorize` removes the source-group membership path entirely, staying semantically aligned with `canAccessTask` (`task.go:927-932`).
4. `GetMembers` in `personal.go` now has proper task-level authorization that was previously **missing entirely** — any authenticated user could read any task's member list before this fix. This is a P0 security improvement (`personal.go:263-289`).
5. All `group_member` references removed from `task.go` auth paths — verified via grep, zero hits.

**Test coverage for the new access model is comprehensive:**
- `TestAuthorizeTaskAccess_GroupMemberDenied` — detail path, 403 ✓
- `TestDeleteSummary_GroupMemberDenied` — delete path, 403 ✓
- `TestCancelSummary_GroupMemberDenied` — cancel path, 403 ✓
- `TestListSummaries_GroupMemberSeesNothing` — list path, total=0 ✓
- `TestBatchStatus_GroupMemberExcluded` — batch path, 0 tasks ✓
- `TestGetMembers_GroupMemberDenied` — members path, 403 ✓
- `TestGetMembers_NoAuth` / `TaskNotFound` / `StrangerDenied` — edge cases ✓

**Non-blocking (carried from first round, unchanged):**
- Auto-fill `SourceType: req.OriginChannelType` is still an implicit mapping that would break silently if constants drift.
- `origin_channel_type` query filter and composite index still worth considering for future.

Clean fix, well-tested, correct design direction.

lml2468 (approved):
Incremental re-review of 37dc17fe — verifying the authorization inconsistency fix from the first round.

**Previous blockers — resolved ✅**

The authorization gap (`ListSummaries` vs `authorizeTaskAccess`) is fixed by **tightening** all auth paths to creator + explicit participant only, rather than widening `ListSummaries`. This is the fail-closed approach and is the correct design choice:

1. `canAccessTask` is now the single source of truth (`task.go:32-43`).
2. `authorizeTaskAccess` delegates to `canAccessTask` (`task.go:65`).
3. `batchAuthorize` removes the source-group membership path entirely, staying semantically aligned with `canAccessTask` (`task.go:927-932`).
4. `GetMembers` in `personal.go` now has proper task-level authorization that was previously **missing entirely** — any authenticated user could read any task's member list before this fix. This is a P0 security improvement (`personal.go:263-289`).
5. All `group_member` references removed from `task.go` auth paths — verified via grep, zero hits.

**Test coverage for the new access model is comprehensive:**
- `TestAuthorizeTaskAccess_GroupMemberDenied` — detail path, 403 ✓
- `TestDeleteSummary_GroupMemberDenied` — delete path, 403 ✓
- `TestCancelSummary_GroupMemberDenied` — cancel path, 403 ✓
- `TestListSummaries_GroupMemberSeesNothing` — list path, total=0 ✓
- `TestBatchStatus_GroupMemberExcluded` — batch path, 0 tasks ✓
- `TestGetMembers_GroupMemberDenied` — members path, 403 ✓
- `TestGetMembers_NoAuth` / `TaskNotFound` / `StrangerDenied` — edge cases ✓

**Non-blocking (carried from first round, unchanged):**
- Auto-fill `SourceType: req.OriginChannelType` is still an implicit mapping that would break silently if constants drift.
- `origin_channel_type` query filter and composite index still worth considering for future.

Clean fix, well-tested, correct design direction.

Jerry-Xin (approved):
The PR is in scope for `octo-smart-summary` and cleanly adds backend support for origin-channel metadata plus the summary-template endpoint. I found no blocking issues.

💬 Non-blocking

🟡 Warning: `ListSummaries` filters only by `origin_channel_id`, even though the model stores channel identity as both ID and type. If IDs can overlap across group/thread/DM namespaces, `/api/v1/summaries?origin_channel_id=...` may return mixed-origin results. Consider also supporting `origin_channel_type` filtering in [internal/api/handler/task.go](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-_utkc0be/repo/internal/api/handler/task.go:349).

🔵 Suggestion: `GetTemplates` defines the template structs and static data inside the handler on every request. Since this is a fixed contract, a package-level typed slice would reduce per-request allocation and make the API shape easier to test/reuse. See [internal/api/handler/task.go](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-_utkc0be/repo/internal/api/handler/task.go:708).

🔵 Suggestion: `GetMembers` now has useful authorization checks, but it still loads the task only by ID/deleted state. For consistency with space-scoped routes like list/batch/edit, consider including `space_id = ?` using `middleware.GetSpaceID(c)`. See [internal/api/handler/personal.go](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-_utkc0be/repo/internal/api/handler/personal.go:271).

✅ Highlights

The new origin-channel fields are threaded through create, list, detail, model, and migration paths, with focused regression coverage.

The authorization behavior is now consistent across detail, list, batch status, delete/cancel, and members for creator/explicit-participant access.

`go test ./...` passes locally.

PR #64, opened 2026-06-05
feat: raise source count limit from 10 to 30

### LINKED ISSUE #63: Raise smart-summary source count limit from 10 to 30 (backend)
## Summary
Raise the maximum number of information sources (chats) allowed per smart-summary task from **10** to **30**.

## Background
When creating a smart-summary task, the number of selectable information sources / chats is capped at 10. This limit is too low for users who need to summarize across more conversations. The cap is enforced on **both** the frontend and the backend, so both sides must be updated — changing only one side is ineffective (frontend-only: backend still rejects; backend-only: frontend still blocks selection).

## Backend change (this repo)
`internal/api/handler/task.go`
- The `CreateSummary` handler rejects requests where `len(sourceList) > 10` with code `40003` ("信息来源不能超过10个").
- Raise the limit to 30. Extract a named constant `maxSourceCount = 30` instead of hardcoding the literal, and build the error message from the constant so the number stays in sync.

## Acceptance criteria
- Creating a task with exactly 30 sources is accepted (not rejected by the count check).
- Creating a task with 31 sources is rejected with HTTP 400 / code 40003.
- Unit test coverage for both boundary cases.

## Related
Frontend counterpart: Mininglamp-OSS/octo-web#284

jeff-wilson2010:
## What
Raise the maximum number of information sources (chats) allowed per smart-summary task from **10** to **30**.

## Why
The previous cap of 10 sources was too low for users who need to summarize across more conversations. The limit is enforced on both backend and frontend; this PR covers the backend half (frontend: Mininglamp-OSS/octo-web#284).

## Changes
- `internal/api/handler/task.go`
  - Introduce a named constant `maxSourceCount = 30` instead of the hardcoded literal `10`.
  - `CreateSummary` now rejects only when `len(sourceList) > maxSourceCount`, and the error message is built from the constant so the number stays in sync.
- `internal/api/handler/task_limit_test.go` (new)
  - Boundary coverage: exactly 30 sources is accepted; 31 sources is rejected with HTTP 400 / code 40003.

## Testing
- `go build ./...` passes.
- `go test ./internal/api/handler/ -count=1 -shuffle=on` passes (40 tests, including the two new boundary cases).

Closes #63

### DIFF
--- a/internal/api/handler/task.go
+++ b/internal/api/handler/task.go
@@ -3,6 +3,7 @@ package handler
 import (
 	"bytes"
 	"encoding/json"
+	"fmt"
 	"log"
 	"net/http"
 	"sort"
@@ -16,6 +17,9 @@ import (
 	"github.com/gin-gonic/gin"
 	"gorm.io/gorm"
 )
+
+// maxSourceCount is the maximum number of information sources allowed per task.
+const maxSourceCount = 30
 
 // TaskHandler handles summary task endpoints.
 type TaskHandler struct {
@@ -184,8 +188,8 @@ func (h *TaskHandler) CreateSummary(c *gin.Context) {
 		sourceList = req.Sources
 	}
 
-	if len(sourceList) > 10 {
-		c.JSON(http.StatusBadRequest, apiResponse{Code: 40003, Message: "信息来源不能超过10个"})
+	if len(sourceList) > maxSourceCount {
+		c.JSON(http.StatusBadRequest, apiResponse{Code: 40003, Message: fmt.Sprintf("信息来源不能超过%d个", maxSourceCount)})
 		return
 	}
 
--- a/internal/api/handler/task_limit_test.go
+++ b/internal/api/handler/task_limit_test.go
@@ -1,0 +1,105 @@
+package handler
+
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	"github.com/Mininglamp-OSS/octo-smart-summary/internal/middleware"
+	"github.com/gin-gonic/gin"
+)
+
+// setupCreateRouter wires only the CreateSummary route with the same auth/space
+// middleware stack used by the other handler tests.
+func setupCreateRouter(h *TaskHandler) *gin.Engine {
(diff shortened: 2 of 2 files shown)


Jerry-Xin (approved):
The PR is in scope and the production change correctly raises the manual summary source limit from 10 to 30.

💬 Non-blocking

🟡 Warning: [internal/api/handler/task_limit_test.go:69](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-nb6xe15x/repo/internal/api/handler/task_limit_test.go:69) does not actually prove that 30 sources can create a task successfully. It uses `setupTestDBs`, which does not migrate `model.PersonalResult`, so the request currently reaches the transaction and returns HTTP 500, while the test still passes because [line 80](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-nb6xe15x/repo/internal/api/handler/task_limit_test.go:80) only checks that the business code is not `40003`. This should ideally assert `http.StatusOK` and `code == 0`, using a DB setup like `setupOriginTestDB` that includes all tables needed by `CreateSummary`.

✅ Highlights

🔵 Suggestion: [internal/api/handler/task.go:21](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-nb6xe15x/repo/internal/api/handler/task.go:21) centralizes the limit in `maxSourceCount`, which is cleaner than duplicating the literal in validation and response text.

Verification: `go test ./internal/api/handler` passes.

lml2468 (approved):
Clean, focused change. Hardcoded `10` is replaced with a named constant `maxSourceCount = 30`, and the error message is built via `fmt.Sprintf` from the same constant so the number can never drift from the check. Boundary tests cover both `=maxSourceCount` (accepted by the count check) and `=maxSourceCount+1` (rejected with HTTP 400 / code 40003). The at-limit test correctly asserts the count check specifically (`code == 40003`) rather than just HTTP status, so later validation failures don't false-positive the test.

CI all green (Build / Test / Vet / Lint / CodeQL / OSV-scan).

No blockers, no non-blocking notes.

PR #67, opened 2026-06-06
feat: add filter param (followed/recent) to chat candidates API

### LINKED ISSUE #66: feat: ChatSelector 增加「关注」和「最近」tab，替代单一「全部」列表
## Summary

智能总结的群选择弹窗（`ChatSelectorModal`）增加「关注」和「最近」两个 tab，让用户快速找到最需要总结的对话。

## Background

当前选择要总结的群时，只有「全部 / 群聊 / 私聊」三个 tab，本质上是对同一份全量列表按 chat_type 过滤。当用户加入的群数量较多时，找到想总结的群效率很低。

用户真正要总结的对话集中在两类：
1. **关注的** — 主动标记为"保存到通讯录"的群（`group_setting.save=1`）
2. **最近的** — 最近有消息往来的对话

## Backend change (this repo)

**文件：** `internal/api/handler/candidates.go` — `SearchChatCandidates`

当前实现从 IM DB 查 `group` + `thread` + `conversation_extra`（direct），不区分关注/最近。

需要新增 query param `filter`：

| filter 值 | 语义 | 数据来源 |
|-----------|------|---------|
| 空/`all` | 全部（现有行为） | 不变 |
| `followed` | 关注 | `group_setting.save=1 AND group_setting.uid=?`（octo-server 已有 `querySavedGroups`，逻辑可复用） |
| `recent` | 最近 | 按 `conversation_extra.updated_at DESC` 或 WuKongIM conversation cache 取最近 N 天活跃的对话 |

Response 结构不变，只是过滤源集合。

### 可选增强

在每个 candidate 中追加字段（方便前端本地过滤，减少请求次数）：
```json
{
  "chat_id": "...",
  "chat_type": "group",
  "name": "...",
  "is_followed": true,
  "last_active_at": "2026-06-06T10:00:00Z"
}
```

## Acceptance criteria

- `GET /api/v1/summary-chat-candidates?filter=followed` 仅返回当前用户 `save=1` 的群（及其下的子区）
- `GET /api/v1/summary-chat-candidates?filter=recent` 返回最近 N 天有消息活动的对话，按活跃时间倒排
- `filter` 为空或 `all` 时行为不变（向后兼容）
- 单元测试覆盖三种 filter 值

## Related

前端改动：需在 octo-web `dmworksummary` 包的 `ChatSelectorModal.tsx` 中将 tab 从「全部/群聊/私聊」改为「关注/最近/全部」，默认选中「关注」。

## Related

前端改动：Mininglamp-OSS/octo-web#297

pkuWMH:
## Summary

`SearchChatCandidates` 新增 `filter` query 参数，支持按「关注」和「最近」过滤群选择列表。

## Changes

**`internal/api/handler/candidates.go`**
- 新增 `filter` query param（`followed` / `recent` / 空=全部）
- `filter=followed`：查 `group_setting.save=1` 返回用户关注的群及其子区
- `filter=recent`：查 `conversation_extra` 按 `updated_at DESC` 返回最近活跃对话
- 每个 candidate 追加 `is_followed`（bool）和 `last_active_at`（string）字段
- 向后兼容：filter 为空时行为不变

**`internal/api/handler/candidates_test.go`**
- 新增 `group_setting` 和 `conversation_extra` 表到测试 fixture
- 新增 `doCandidateRequestFilter` helper
- 新增 3 个测试：`FilterFollowed`、`FilterRecent`、`FilterEmptyIsBackwardCompatible`
- **All 8 tests pass** ✅（已通过 CC 本地验证）

## Test Results

```
=== RUN   TestSearchChatCandidates_FilterFollowed        ✅
=== RUN   TestSearchChatCandidates_FilterRecent           ✅
=== RUN   TestSearchChatCandidates_FilterEmptyIsBackwardCompatible ✅
ok  github.com/Mininglamp-OSS/octo-smart-summary/internal/api/handler
```

## Related

- Closes #66
- Frontend counterpart: Mininglamp-OSS/octo-web#297

### DIFF
--- a/internal/api/handler/candidates.go
+++ b/internal/api/handler/candidates.go
@@ -108,10 +108,16 @@ type imDirect struct {
 // Query params:
 //   - keyword:   optional search keyword
 //   - chat_type: "group" | "direct" | "" (empty = all)
+//   - filter:    "followed" | "recent" | "" (empty = all, backward compatible)
 //
 // Groups are fetched from the `group` table.
 // Direct chats are fetched from `conversation_extra` (channel_type=1), filtered to
 // human peers only (32-char hex uid, not in robot table, robot flag = 0).
+//
+// When filter=followed, only groups where group_setting.save=1 for the current user
+// are returned (plus their threads). Direct chats are excluded.
+// When filter=recent, conversations are fetched from conversation_extra ordered by
+// updated_at DESC, returning recently active groups and directs.
 func (h *CandidateHandler) SearchChatCandidates(c *gin.Context) {
 	if h.imDB == nil {
 		c.JSON(http.StatusOK, gin.H{"code": 0, "data": []interface{}{}})
@@ -120,6 +126,7 @@ func (h *CandidateHandler) SearchChatCandidates(c *gin.Context) {
 
 	keyword := c.Query("keyword")
 	chatType := c.Query("chat_type") // "group", "direct", or "" = all
+	filter := c.Query("filter")     // "followed", "recent", or "" = all
 
 	// Resolve current user from context (set by AuthMiddleware).
 	currentUID, _ := c.Get("user_id")
@@ -134,13 +141,78 @@ func (h *CandidateHandler) SearchChatCandidates(c *gin.Context) {
 		spaceIDStr, _ = v.(string)
 	}
 
+	// --- Build followed group_no set (used by filter=followed and for is_followed enrichment) ---
+	followedGroupNos := map[string]bool{}
+	if currentUIDStr != "" {
+		type followedGroup struct {
+			GroupNo string `gorm:"column:group_no"`
+		}
+		var fgs []followedGroup
+		h.imDB.Table("group_setting").
+			Select("group_no").
+			Where("uid = ? AND save = 1", currentUIDStr).
+			Find(&fgs)
+		for _, fg := range fgs {
+			followedGroupNos[fg.GroupNo] = true
+		}
+	}
+
+	// --- Build recent channel map (used by filter=recent and for last_active_at enrichment) ---
+	recentChannels := map[string]string{} // channelID -> updated_at ISO string
+	if currentUIDStr != "" {
+		type recentConv struct {
+			ChannelID   string `gorm:"column:channel_id"`
+			ChannelType int    `gorm:"column:channel_type"`
+			UpdatedAt   string `gorm:"column:updated_at"`
+		}
+		var rcs []recentConv
+		h.imDB.Table("conversation_extra").
+			Select("channel_id, channel_type, updated_at").
+			Where("uid = ?", currentUIDStr).
+			Order("updated_at DESC").
+			Limit(50).
+			Find(&rcs)
+		for _, rc := range rcs {
+			recentChannels[rc.ChannelID] = rc.UpdatedAt
+		}
+	}
+
+	// --- Determine which sections to include based on filter ---
+	includeGroups := chatType == "" || chatType == "all" || chatType == "group"
+	includeThreads := chatType == "" || chatType == "all" || chatType == "thread"
+	includeDirects := chatType == "" || chatType == "all" || chatType == "direct"
+
+	if filter == "followed" {
+		// Followed is group-centric: only groups + their threads, no directs.
+		includeDirects = false
+	} else if filter == "recent" {
+		// Recent includes both groups and directs that appear in conversation_extra.
+	}
+
 	// --- Groups ---
-	if chatType == "" || chatType == "all" || chatType == "group" {
+	if includeGroups {
 		var groups []imGroup
 		q := h.imDB.Table("`group` g").Select("g.group_no, g.name")
 		if currentUIDStr != "" {
 			q = q.Joins("INNER JOIN group_member gm ON gm.group_no = g.group_no").
 				Where("gm.uid = ? AND gm.is_deleted = 0", currentUIDStr)
+		}
+		if filter == "followed" && currentUIDStr != "" {
+			q = q.Joins("INNER JOIN group_setting gs ON gs.group_no = g.group_no").
+				Where("gs.uid = ? AND gs.save = 1", currentUIDStr)
+		} else if filter == "recent" {
+			// Only include groups that appear in the recent conversation list.
+			recentGroupNos := make([]string, 0)
+			for chID := range recentChannels {
+				recentGroupNos = append(recentGroupNos, chID)
+			}
+			if len(recentGroupNos) > 0 {
+				q = q.Where("g.group_no IN ?", recentGroupNos)
+			} else {
+				// No recent conversations — skip group query entirely.
+				groups = nil
+				goto groupsDone
+			}
 		}
 		if spaceIDStr != "" {
 			q = q.Where("g.space_id = ?", spaceIDStr)
@@ -149,18 +221,22 @@ func (h *CandidateHandler) SearchChatCandidates(c *gin.Context) {
 			q = q.Where("g.name LIKE ?", "%"+keyword+"%")
 		}
 		h.applyLimit(q).Find(&groups)
+	groupsDone:
 		for _, g := range groups {
-			list = append(list, gin.H{
-				"chat_id":      g.GroupNo,
-				"chat_type":    "group",
-				"name":         g.Name,
-				"member_count": nil,
-			})
+			entry := gin.H{
+				"chat_id":        g.GroupNo,
+				"chat_type":      "group",
+				"name":           g.Name,
+				"member_count":   nil,
+				"is_followed":    followedGroupNos[g.GroupNo],
+				"last_active_at": recentChannels[g.GroupNo],
+			}
+			list = append(list, entry)
 		}
 	}
 
 	// --- Threads (子区, channelType=5) ---
-	if chatType == "" || chatType == "all" || chatType == "thread" {
+	if includeThreads {
 		type imThread struct {
 			ShortID string `gorm:"column:short_id"`
 			Name    string `gorm:"column:name"`
@@ -177,6 +253,23 @@ func (h *CandidateHandler) SearchChatCandidates(c *gin.Context) {
 			q = q.Joins("INNER JOIN group_member gm ON gm.group_no" + h.collate + " = t.group_no").
 				Where("gm.uid = ? AND gm.is_deleted = 0", currentUIDStr)
 		}
+		if filter == "followed" && currentUIDStr != "" {
+			// Only threads under followed groups.
+			q = q.Joins("INNER JOIN group_setting gs2 ON gs2.group_no" + h.collate + " = t.group_no").
+				Where("gs2.uid = ? AND gs2.save = 1", currentUIDStr)
+		} else if filter == "recent" {
+			// Only threads under groups that appear in recent conversations.
+			recentGroupNos := make([]string, 0)
+			for chID := range recentChannels {
+				recentGroupNos = append(recentGroupNos, chID)
+			}
+			if len(recentGroupNos) > 0 {
+				q = q.Where("t.group_no IN ?", recentGroupNos)
+			} else {
+				threads = nil
+				goto threadsDone
+			}
+		}
 		if spaceIDStr != "" {
 			q = q.Where("g.space_id = ?", spaceIDStr)
 		}
@@ -184,6 +277,7 @@ func (h *CandidateHandler) SearchChatCandidates(c *gin.Context) {
 			q = q.Where("t.name LIKE ? OR g.name LIKE ?", "%"+keyword+"%", "%"+keyword+"%")
 		}
 		h.applyLimit(q).Find(&threads)
+	threadsDone:
 		for _, t := range threads {
 			list = append(list, gin.H{
 				"chat_id":         t.GroupNo + "____" + t.ShortID,
@@ -191,6 +285,8 @@ func (h *CandidateHandler) SearchChatCandidates(c *gin.Context) {
 				"name":            t.Name,
 				"member_count":    nil,
 				"parent_group_no": t.GroupNo,
+				"is_followed":     followedGroupNos[t.GroupNo],
+				"last_active_at":  recentChannels[t.GroupNo],
 			})
 		}
 	}
@@ -201,7 +297,7 @@ func (h *CandidateHandler) SearchChatCandidates(c *gin.Context) {
 	// Filter: only 32-char hex uids (excludes system accounts like fileHelper/botfather),
 	//         not a robot (robot table + robot flag).
 	// Requires authentication; skipped silently if unauthenticated.
-	if (chatType == "" || chatType == "all" || chatType == "direct") && currentUIDStr != "" {
+	if includeDirects && currentUIDStr != "" {
 		var directs []imDirect
 		q := h.imDB.Table("conversation_extra ce").
 			Select("ce.channel_id, u.name, u.robot").
@@ -232,12 +328,19 @@ func (h *CandidateHandler) SearchChatCandidates(c *gin.Context) {
 			if name == "" {
 				name = d.ChannelID
 			}
+			// For filter=recent, skip directs not in the recent conversations list.
+			if filter == "recent" {
+				if _, ok := recentChannels[d.ChannelID]; !ok {
+					continue
+				}
+			}
 			list = append(list, gin.H{
-				"chat_id":      d.ChannelID,
-				"chat_type":    "direct",
-				"name":         name,
-				"member_count": nil,
-				"is_bot":       d.Robot == 1,
+				"chat_id":        d.ChannelID,
+				"chat_type":      "direct",
+				"name":           name,
+				"member_count":   nil,
+				"is_bot":         d.Robot == 1,
+				"last_active_at": recentChannels[d.ChannelID],
 			})
 		}
 	}
--- a/internal/api/handler/candidates.go
+++ b/internal/api/handler/candidates.go
@@ -2,6 +2,7 @@ package handler
 
 import (
 	"net/http"
+	"sort"
 
 	"github.com/gin-gonic/gin"
 	"gorm.io/gorm"
@@ -343,6 +344,21 @@ func (h *CandidateHandler) SearchChatCandidates(c *gin.Context) {
 				"last_active_at": recentChannels[d.ChannelID],
 			})
 		}
+	}
+
+	// --- Sort by recency when filter=recent ---
+	if filter == "recent" && len(list) > 1 {
+		sort.SliceStable(list, func(i, j int) bool {
+			ai, _ := list[i]["last_active_at"].(string)
+			aj, _ := list[j]["last_active_at"].(string)
+			if ai == "" {
+				return false
+			}
+			if aj == "" {
+				return true
+			}
+			return ai > aj // DESC: newer first
+		})
 	}
 
 	c.JSON(http.StatusOK, gin.H{"code": 0, "data": list})
--- a/internal/api/handler/candidates.go
+++ b/internal/api/handler/candidates.go
@@ -158,8 +158,13 @@ func (h *CandidateHandler) SearchChatCandidates(c *gin.Context) {
 		}
 	}
 
-	// --- Build recent channel map (used by filter=recent and for last_active_at enrichment) ---
-	recentChannels := map[string]string{} // channelID -> updated_at ISO string
+	// --- Build recent channel maps (used by filter=recent and for last_active_at enrichment) ---
+	// conversation_extra mixes channel types under one channel_id column, so split by
+	// channel_type to avoid cross-type collisions. Thread channel_ids use the composite
+	// format 'group_no____short_id' (4 underscores), distinct from a bare group_no.
+	recentGroups := map[string]string{}  // channel_type=2: group_no       -> updated_at
+	recentThreads := map[string]string{} // channel_type=5: group_no____short_id -> updated_at
+	recentDirects := map[string]string{} // channel_type=1: peer uid       -> updated_at
 	if currentUIDStr != "" {
 		type recentConv struct {
 			ChannelID   string `gorm:"column:channel_id"`
@@ -174,7 +179,14 @@ func (h *CandidateHandler) SearchChatCandidates(c *gin.Context) {
 			Limit(50).
 			Find(&rcs)
 		for _, rc := range rcs {
-			recentChannels[rc.ChannelID] = rc.UpdatedAt
+			switch rc.ChannelType {
+			case 1:
+				recentDirects[rc.ChannelID] = rc.UpdatedAt
+			case 5:
+				recentThreads[rc.ChannelID] = rc.UpdatedAt
+			default: // channel_type=2 (group)
+				recentGroups[rc.ChannelID] = rc.UpdatedAt
+			}
 		}
 	}
 
@@ -203,8 +215,8 @@ func (h *CandidateHandler) SearchChatCandidates(c *gin.Context) {
 				Where("gs.uid = ? AND gs.save = 1", currentUIDStr)
 		} else if filter == "recent" {
 			// Only include groups that appear in the recent conversation list.
-			recentGroupNos := make([]string, 0)
-			for chID := range recentChannels {
+			recentGroupNos := make([]string, 0, len(recentGroups))
+			for chID := range recentGroups {
 				recentGroupNos = append(recentGroupNos, chID)
 			}
 			if len(recentGroupNos) > 0 {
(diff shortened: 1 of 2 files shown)


Jerry-Xin (commented):
The PR is in scope for this repository, but the new `recent` filter does not satisfy its stated ordering semantics.

**🔴 Blocking**

- 🔴 Critical: `filter=recent` does not return candidates ordered by `conversation_extra.updated_at DESC`. The code reads recent rows ordered by `updated_at`, stores them in `recentChannels` as a map, then emits all groups, then threads, then directs. Group/thread queries also have no `updated_at` ordering, and directs are only ordered within the direct section. This means a newer direct can appear after older groups, and multiple groups can appear in database order rather than recent order. This breaks the core contract described for the new filter. See `internal/api/handler/candidates.go:169`, `internal/api/handler/candidates.go:203`, `internal/api/handler/candidates.go:223`, `internal/api/handler/candidates.go:260`, and `internal/api/handler/candidates.go:325`.

  Suggested fix: make the ordered `conversation_extra` result the driver for `filter=recent`, or preserve `updated_at` on each candidate and sort the final `list` globally by `last_active_at DESC`. Add a test with at least one group and one direct with different timestamps to assert global order.

**💬 Non-blocking**

- 🟡 Warning: `recentChannels` is keyed only by `channel_id` and ignores `channel_type` after selecting it. If `conversation_extra` can contain thread rows or overlapping channel IDs across types, the enrichment/filtering can misclassify or drop candidates. See `internal/api/handler/candidates.go:163` and `internal/api/handler/candidates.go:175`.

- 🟡 Warning: the hard-coded `Limit(50)` is applied before keyword, space, visibility, and candidate-type filtering. Valid matching candidates outside the first 50 raw conversation rows can disappear even when they should be returned. See `internal/api/handler/candidates.go:169`.

- 🔵 Suggestion: add tests for `filter=recent` with direct chats and mixed group/direct ordering. The current test only covers a single recent group, so it cannot catch the ordering regression. See `internal/api/handler/candidates_test.go:267`.

**✅ Highlights**

- The followed filter keeps group membership checks in place.
- The empty filter path is covered for backward compatibility.
- I ran `go test ./internal/api/handler`, and it passes locally.

yujiawei (dismissed):
# Code Review — PR #67 (octo-smart-summary)

**Verdict: APPROVED** — no blocking (P0/P1) issues. The change is well-scoped, backward-compatible, and covered by tests. Build passes (`go build ./...`), `go vet` clean, all 8 handler tests pass. The suggestions below are quality improvements (P2/nit) that do not gate merge.

## Summary of review

`SearchChatCandidates` gains a `filter` param (`followed` / `recent` / empty) plus two enrichment fields (`is_followed`, `last_active_at`). I read the full handler, the route registration, and checked for downstream consumers of the response shape — there are none in this repo, so the additive fields are safe. All SQL is parameterized (no injection surface). The empty-filter path is behavior-preserving and is explicitly covered by `TestSearchChatCandidates_FilterEmptyIsBackwardCompatible`.

## Verification

| Item | Status | Evidence |
|---|---|---|
| Build / vet / tests | ✅ | `go build ./...` OK, `go vet` clean, handler tests 8/8 pass |
| SQL injection | ✅ | All user input goes through bound params (`?`) — `candidates.go:153, 171, 210, 267, 321` |
| Backward compatibility (empty filter) | ✅ | No new WHERE clauses applied when `filter==""`; test at `candidates_test.go:345` |
| `goto groupsDone/threadsDone` legality | ✅ | Skipped `recentGroupNos` is block-scoped and out of scope at the label; ranges over explicit-`nil` slices are safe no-ops (`candidates.go:213-214, 269-270, 225, 281`) |
| New tests cover new behavior | ✅ | FilterFollowed, FilterRecent, FilterEmpty (`candidates_test.go:212-330`) |

## Findings (all non-blocking)

### P2-1 — `filter=recent` does not preserve recency ordering
`recentChannels` is built from a query correctly ordered `updated_at DESC ... LIMIT 50` (`candidates.go:169-174`), but the rows are funneled into a `map[string]string` (`candidates.go:175-177`), which discards order (Go map iteration is randomized). The group and thread queries then build their `IN (...)` lists by ranging that map with **no `ORDER BY`** (`candidates.go:205-210, 262-267`), and the final `list` is returned without any sort (`candidates.go:348`). Result: the *correct set* of recent chats is returned, but in arbitrary order — a visible regression for a "recent" tab.

Each item does carry `last_active_at`, so a client *can* sort, which is why this is P2 rather than a blocker. Suggested server-side fix: `sort.SliceStable(list, ...)` by `last_active_at` desc (empty values last) before the final `c.JSON`. Note the directs branch alone is already ordered (`candidates.go:325`), so a single sort over the combined slice is the consistent place to enforce recency.

### P2-2 — Unknown `filter` values silently behave like "all"
An unrecognized value (e.g. a typo `filter=followd`) falls through both branches (`candidates.go:185-190`) and returns everything, with no `400`. This can silently mislead callers into thinking they're seeing a filtered subset. Consider validating the param and returning `400` on unknown values, or at minimum documenting the fallback.

### P2-3 — Response shape is inconsistent across chat types
Group and thread entries include `is_followed` (`candidates.go:231, 288`) but direct entries do **not** (`candidates.go:337-344`). Additionally, `last_active_at` is sourced from a `map[string]string` lookup, so a miss serializes as `""` (empty string), not `null` — contrast `member_count` which is explicit `nil`/`null`. This is additive-safe for any consumer that ignores the new fields, but a strict-typed consumer that expects a uniform shape (e.g. a required `is_followed` on every item, or `last_active_at` typed as a real timestamp) could mis-decode `""`. Suggest: add `is_followed` to directs (or document that directs are never followable), and use the comma-ok form (`v, ok := recentChannels[id]`) to emit `null`/omit on a miss rather than `""`.

## Nits

- **`goto` for the empty-recent short-circuit** (`candidates.go:214, 270`): correct and tightly scoped, but a structured guard (e.g. a `skipGroupQuery` bool wrapping the `Find`) would be more idiomatic and avoids the future-edit hazard where a new declaration between `goto` and label would break the build.
- **`recentChannels` mixes channel types** (`candidates.go:169-174` selects without a `channel_type` predicate): the non-group keys (P2P peer ids, thread ids) get passed into the `group_no IN (...)` list. Harmless today — P2P channel_ids live in a different namespace (`uidA@uidB` per `pipeline/fetch.go:162-179`) and match zero `group` rows — but adding `channel_type` to the predicate, or a comment explaining why the over-broad IN-list is safe, would prevent confusion.
- **Two enrichment queries run on every authenticated request** (`candidates.go:145-158, 160-178`), even for the empty-filter path. They feed `is_followed`/`last_active_at`, so they're not pure waste, but if a future caller doesn't need the enrichment fields these could be made lazy. `group_setting` has no `LIMIT` (naturally small set; low risk).

## Adjacent observation (out of scope, FYI)
The pre-existing direct-chat branch joins `ce.channel_id = u.uid` with `LENGTH(ce.channel_id) = 32` (`candidates.go:304, 312`), assuming a bare-uid format, while `pipeline/fetch.go` documents the P2P `conversation_extra.channel_id` storage format as `peer@self`. Not introduced by this PR and not affecting the new code paths, but worth a future look to confirm the two readers agree on the P2P channel_id format.

pkuWMH:
Thanks for the thorough review @Jerry-Xin @yujiawei!

## Addressed

**🔴 Blocking — `filter=recent` ordering** (Jerry-Xin): Fixed in `f75158f`. Added `sort.SliceStable(list, ...)` by `last_active_at DESC` (empty values last) after all candidates are collected. All 8 tests pass.

## Acknowledged (P2, will address in follow-up)

- **`recentChannels` ignores `channel_type`**: Agreed this is fragile. Will add `channel_type` predicate in a follow-up.
- **Hard-coded `Limit(50)`**: Will make this configurable or apply after filtering.
- **Unknown `filter` values silently fallback**: Will add validation + 400 on unrecognized values.
- **Inconsistent `is_followed` on directs**: Will add `is_followed: false` to direct entries for uniform shape.
- **`goto` style**: Will consider refactoring to guard booleans in a follow-up.

Jerry-Xin (changes requested):
The PR is relevant to this repository, but `filter=recent` mishandles thread candidates and can return non-recent conversations.

**🔴 Blocking**

- 🔴 Critical: `filter=recent` treats a recent parent group as making every active thread in that group recent. In [internal/api/handler/candidates.go](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-9ydwswcy/repo/internal/api/handler/candidates.go:261), the thread query filters with `t.group_no IN recentChannels`, then [line 290](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-9ydwswcy/repo/internal/api/handler/candidates.go:290) assigns the parent group’s `last_active_at` to each thread. Existing project conventions identify thread channels as `group_no____short_id`, not just `group_no`; see [internal/pipeline/fetch.go](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-9ydwswcy/repo/internal/pipeline/fetch.go:119). This means a recently active group will incorrectly surface all its threads as recent, while a recently active thread whose `conversation_extra.channel_id` is `group_no____short_id` can be missed unless the parent group also has a recent row. Fix by either excluding threads from `filter=recent`, or matching thread candidates by their full composite channel id and `channel_type = 5`.

**💬 Non-blocking**

- 🟡 Warning: The recent enrichment map is built from all `conversation_extra` channel types at [internal/api/handler/candidates.go](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-9ydwswcy/repo/internal/api/handler/candidates.go:170). Using one untyped `channel_id -> updated_at` map for groups, directs, and threads makes type collisions possible and is the root of the thread bug above. Separate maps by `channel_type` would make the behavior clearer and safer.

- 🟡 Warning: Tests do not cover `filter=recent` with threads or direct chats. The current recent test only inserts a group row, so it misses the incorrect thread behavior. Add cases for a recent thread row with `channel_id='grp1____th001', channel_type=5`, and for a recent group with multiple non-recent threads.

**✅ Highlights**

- The PR preserves the existing empty-filter behavior in the covered group path.
- `filter=followed` correctly keeps direct chats out and includes threads under followed groups.
- `go test ./...` passes locally.

pkuWMH:
Addressed in `9517fd3`:

**🔴 Thread channel_id matching** — Fixed. Split the single `recentChannels` map into three type-specific maps:
- `recentGroups` (channel_type=2)
- `recentThreads` (channel_type=5, keyed by `group_no____short_id`)
- `recentDirects` (channel_type=1)

Thread filter=recent now matches by the full composite id, not parent group_no. A recent parent group no longer leaks all its threads, and a recently active thread under a non-recent group is correctly surfaced.

**🟡 Tests** — Added 3 new tests:
- `FilterRecentThread` — thread with channel_id=`grp1____th001` surfaces correctly
- `FilterRecentGroupNotLeakingThreads` — recent parent group does NOT surface its threads
- `FilterRecentMixedGroupDirectOrdering` — asserts global recency ordering across group+direct

All 11 Candidates tests pass.

Jerry-Xin (changes requested):
The PR is in scope for `octo-smart-summary`, and `go test ./...` passes, but the new API response contract is inconsistent.

**🔴 Blocking**
🔴 Critical — `is_followed` is missing for direct chat candidates. In `internal/api/handler/candidates.go:357`, direct entries include `chat_id`, `chat_type`, `name`, `member_count`, `is_bot`, and `last_active_at`, but not `is_followed`; groups and threads do include it at `internal/api/handler/candidates.go:244` and `internal/api/handler/candidates.go:308`. The PR states every candidate gets `is_followed` as a bool, so clients will receive an inconsistent schema and may need to special-case direct rows. Add `"is_followed": false` for direct candidates and cover it with a mixed group/direct response test.

**💬 Non-blocking**
🟡 Warning — `internal/api/handler/candidates.go` is not fully `gofmt`-clean around `internal/api/handler/candidates.go:130`, `internal/api/handler/candidates.go:266`, and `internal/api/handler/candidates.go:271`. Run `gofmt` before merge if CI enforces formatting.

**✅ Highlights**
The `recent` split by `channel_type` avoids the group/thread collision risk, and the added thread-focused tests cover the important edge cases around composite thread IDs and parent group leakage.

yujiawei (changes requested):
# Code Review — PR #67 (octo-smart-summary)

Independent review of the new `filter` (`followed`/`recent`) parameter on `SearchChatCandidates`. The feature is well-structured, the backward-compatibility path is preserved, and the three new tests plus the existing suite (11 total) pass; `go build` and `go vet` are clean. The composite thread-id handling (matching recent threads by their own `group_no____short_id` rather than by parent group) is a nice correctness detail, and the dedicated tests for it are appreciated.

One correctness issue should be fixed before merge; the rest are non-blocking suggestions.

## Requested change (blocking)

### P1 — `filter=recent`: SQL `LIMIT` truncates threads/directs *before* the in-loop recency filter
`internal/api/handler/candidates.go` (threads ~L291→L298, directs ~L345→L353)

The groups path pushes the recency key set into SQL before limiting:

```go
} else if filter == "recent" {
    ...
    q = q.Where("g.group_no IN ?", recentGroupNos)   // L223 — filter THEN limit
```

But the **threads** and **directs** paths deliberately add *no* SQL recency predicate and instead filter in the Go loop:

```go
h.applyLimit(q).Find(&threads)          // L291 — LIMIT applied to ALL of the user's threads
...
for _, t := range threads {
    if filter == "recent" {
        if _, ok := recentThreads[compositeID]; !ok { continue }   // L298 — recency filtered AFTER limit
    }
```

When `SUMMARY_CHAT_CANDIDATE_LIMIT > 0` (config plumbed through `NewCandidateHandler` → `applyLimit`; only the default `-1` is safe), the SQL `LIMIT` returns N rows across the user's *entire* thread/direct set, and the loop then discards the non-recent ones — so `filter=recent` can return far fewer items than exist, or zero. The thread query also has **no `ORDER BY`**, so which N rows survive the limit is arbitrary. The directs path is partly mitigated by its existing `Order("ce.updated_at DESC")` (L345), which overlaps `recentDirects`, but threads are fully exposed.

The inconsistency with the groups path indicates this is an oversight rather than intent. Suggested fix: mirror the groups approach — apply the recency filter before `applyLimit` (e.g. push an `IN` clause on `channel_id`/composite key, or skip the SQL `LIMIT` and limit after the in-loop recency filter). Note all current tests use `NewCandidateHandler(imDB, -1)`, so the `queryLimit > 0` path is entirely uncovered; a regression test with a positive limit would lock this down.

## Non-blocking suggestions (P2)

- **Default-mode enrichment is silently incomplete (L168–191, L245, L309, L363).** The `recentGroups/Threads/Directs` maps that feed `last_active_at` are built from a top-50 `conversation_extra` window (`.Order("updated_at DESC").Limit(50)`, L175–180) on *every* authenticated request. In default mode (`filter=""`) a group/thread/direct outside that top-50 gets `last_active_at: ""` even though it has activity. Acceptable if `last_active_at` is documented as best-effort/recent-only; otherwise the 50 cap will surprise clients.

- **Two extra queries run on every authenticated request regardless of `filter` (L146–159 `group_setting`, L168–191 `conversation_extra`).** Both are gated only on `currentUIDStr != ""`, never on `filter`. For the common `filter=""` call this is added latency for enrichment that may not be needed. Consider gating the `conversation_extra` query on `filter == "recent" || <enrichment desired>`.

- **`filter=followed` re-derives the followed set in SQL despite already holding it in a map (L146–159 vs L214/L271).** `followedGroupNos` is already populated; the groups/threads queries then add an `INNER JOIN group_setting ... save=1`. Functionally fine, but the JOIN duplicates work already in memory — could filter against the map (or drop the redundant map build in this path).

- **`filter=recent` top-50 cap has no test for the truncation boundary (L179).** A hardcoded `Limit(50)` (magic number) bounds the entire recent result; worth a named const and a boundary test.

- **`goto groupsDone` / `goto threadsDone` (L227/L237, L282/L292)** hurt readability. The same early-skip is expressible with a guard (`if len(...) == 0 { /* skip query */ } else { ...Find... }`) without labels.

## Nits

- **`channel_type` switch defaults unknown types into `recentGroups` (L181–190).** `case 1`/`case 5` handle direct/thread; everything else (including real groups `=2` but also any other WuKongIM type) falls into the group bucket. In practice junk `channel_id`s simply fail to match the `group` table so impact is cosmetic, but an explicit `case 2:` with unknown types ignored is more defensive.
- **Missing `is_followed` on direct chats (L357–364).** Groups (L244) and threads (L308) set `is_followed`; directs omit it, so clients see an inconsistent shape across `chat_type`. Directs are never "followed", but emitting `is_followed: false` for uniformity would be cleaner.
- **Recency sort uses lexical string compare on `updated_at` (L370–381).** `ai > aj` on the raw string is correct only while the timestamp stays fixed-width and lexicographically monotonic (RFC3339 Z is fine); a format change would silently misorder. The empty-string-sorts-last handling is fine.
- **Dead `else if filter == "recent" {}` branch (L201–203)** carries only a comment — can be removed.
- **Test gaps:** `filter=followed` directs-exclusion is untested, and the directs `is_followed`/`>50` truncation paths are uncovered.

## Verdict

The feature is close, but the `filter=recent` limit-vs-recency ordering (P1) is a real correctness bug in the new code path and should be fixed (or the limit path explicitly tested) before merge. Requesting changes for that item; the remaining points are improvements at the author's discretion.

pkuWMH:
Round 3 fixes in 83a94e8:

**Jerry-Xin** is_followed on directs: Added is_followed:false to all direct candidates.

**yujiawei P1** LIMIT-before-filter: Threads skip applyLimit when filter=recent (composite ID match runs in Go loop, needs full set). Directs push recentDirects keys into SQL IN clause before LIMIT.

Cleanup: gofmt applied, removed dead else-if branch, explicit case 2 for groups in channel_type switch.

All 11 tests pass, go build and go vet clean.

Jerry-Xin (changes requested):
Summary: The PR is in scope for `octo-smart-summary`, adds the requested chat-candidate filtering behavior, and the full test suite passes locally.

💬 Non-blocking

🟡 Warning — [internal/api/handler/candidates.go](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-gdaoi_b2/repo/internal/api/handler/candidates.go:273): `filter=recent` skips `applyLimit` for threads and loads all matching threads before filtering by composite ID in Go. On large IM datasets this can make the candidate endpoint expensive. Consider pushing the composite match into SQL, for example with `CONCAT(t.group_no, '____', t.short_id) IN ?`, or otherwise bounding the query.

🟡 Warning — [internal/api/handler/candidates.go](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-gdaoi_b2/repo/internal/api/handler/candidates.go:175): the recent conversation seed is hard-limited to 50 before applying `chat_type`, keyword, space, membership, and direct-human filters. This can hide valid recent results when the top 50 raw `conversation_extra` rows are mostly another type or later filtered out. If `chat_type=thread` or `chat_type=group` is expected to mean “recent within that type,” the channel-type filter should happen before the limit.

✅ Highlights

🔵 Suggestion — The membership checks are preserved for groups and threads, and direct chats still go through the existing user/space/bot filtering path.

🔵 Suggestion — The added tests cover followed filtering, recent group/direct/thread behavior, parent-group leakage, ordering, and backward compatibility.

Verification: `go test ./internal/api/handler` and `go test ./...` both pass.

yujiawei (approved):
# Code Review — PR #67 (octo-smart-summary)

Independent review of the new `filter` parameter (`followed` / `recent` / empty) on `SearchChatCandidates`, at head `83a94e8`. The feature is well-structured and backward-compatible, all four previously-flagged blocking issues are resolved, the empty-filter path is behavior-preserving, all SQL is parameterized, and the suite (11 handler tests) plus `go build ./...`, `go vet`, and `gofmt -l` are all clean. No blocking issues remain. **Approving**; the items below are non-blocking quality follow-ups at the author's discretion.

## Verification of prior blocking issues (all ✅ fixed)

| Prior blocker | Status | Evidence |
|---|---|---|
| `filter=recent` global recency ordering | ✅ | `sort.SliceStable` over the combined `list` by `last_active_at` DESC, empty-last (`candidates.go:392-404`) |
| Thread recency leaking via parent group | ✅ | `recentThreads` keyed by composite `group_no____short_id` (channel_type=5), matched in-loop on the composite id (`candidates.go:166, 302-309`); covered by `FilterRecentThread` + `FilterRecentGroupNotLeakingThreads` |
| `LIMIT`-before-filter truncation | ✅ | Threads skip `applyLimit` under `filter=recent` (`candidates.go:295-299`); directs push the recent-peer set into an SQL `IN` clause before `applyLimit` (`candidates.go:356-366`); groups constrain `group_no IN` before limit (`candidates.go:223`) |
| Untyped recent map / cross-type collision | ✅ | Split into `recentGroups` / `recentThreads` / `recentDirects` by `channel_type` (`candidates.go:165-192`) |
| `is_followed` missing on directs | ✅ | Directs now emit `is_followed: false` (`candidates.go:385`) |
| gofmt | ✅ | `gofmt -l` clean |

## Non-blocking findings

### P2-1 — `filter=recent` groups path applies `LIMIT` with no `ORDER BY`
`candidates.go:223,236`

The groups query under `filter=recent` is constrained by `g.group_no IN (recentGroupNos)` but has **no `.Order(...)`**, then `h.applyLimit(q).Find(&groups)` applies the SQL `LIMIT`. With no ordering, MySQL returns an arbitrary N rows, so a user with more recent groups than the configured limit can have the *actually-most-recent* groups dropped before the global Go sort ever sees them — the final sort then orders an incomplete set. The directs path avoids this by ordering `ce.updated_at DESC` before its limit (`candidates.go:366`), and threads avoid it by skipping the limit entirely. Groups are the odd one out.

This is **not blocking** because the default `CandidateQueryLimit` is `-1` (`config/config.go:123`), making `applyLimit` a no-op — it only bites operators who set `SUMMARY_CHAT_CANDIDATE_LIMIT > 0`. Suggested fix: mirror the threads path (skip `applyLimit` for groups under `filter=recent`, since the `IN` clause already bounds the set), or add a recency `ORDER BY` before the limit.

### P2-2 — `queryLimit > 0` path is entirely uncovered by tests
`candidates_test.go` (all 11 handlers use `NewCandidateHandler(imDB, -1)`)

Every test runs the no-limit branch, so the precise code the PR set out to fix (the `LIMIT`-before-filter handling) has no regression guard. The fix is correct *by construction*, so this hides no live bug — but a future refactor that re-applies `applyLimit` to the recent-thread query, or drops the directs `IN` constraint, would silently reintroduce the original truncation with no failing test. A single test with a positive limit would lock this down (and would also surface P2-1 above).

### P2-3 — Unconditional enrichment queries on every authenticated request
`candidates.go:152-155` (`group_setting`), `candidates.go:175-180` (`conversation_extra`)

Both enrichment queries run for any logged-in caller regardless of `filter`, including the common `filter=""` path, purely to populate `is_followed` / `last_active_at`. Both are bounded/indexed (the `conversation_extra` one is `LIMIT 50`; `group_setting` is an indexed point-lookup on `uid` but has **no** `LIMIT`), so cost is modest, but it is strictly added DB work on the hot no-filter path. Consider gating the `conversation_extra` enrichment on `filter == "recent"` (or on whether the caller needs `last_active_at`), and confirming the saved-group count per user is naturally small.

## Nits

- **Unknown `filter` values silently behave like `""`** (`candidates.go:200-216`): a typo such as `filter=followd` falls through both branches and returns everything unfiltered, with no `400`. Correct-by-construction graceful degradation, but worth documenting or validating — and untested.
- **Lexical string compare in the recency sort** (`candidates.go:402`, `ai > aj`): correct only while `updated_at` is fixed-width/zero-padded (RFC3339 `Z` and MySQL `DATETIME` both are). The SQL `Order("updated_at DESC")` makes the same assumption, so they stay consistent — just confirm the production column type is `DATETIME`/`TIMESTAMP`, not an unpadded string or epoch int.
- **Top-50 recent cap spans all channel types combined** (`candidates.go:179`): a user with >50 active conversations could have one channel type crowd out another in the "recent" source set. Reasonable product cap; flagging only so it's an intentional, documented choice (also a magic number worth a named const).
- **Test confidence gaps**: `FilterFollowed` seeds no direct candidate, so the `includeDirects=false` exclusion is not actually exercised; and no test asserts `is_followed: false` on a direct entry. Both are trivially-correct paths — cheap to close.

## Highlights
- Composite thread-id matching (recent thread surfaces by its own id, recent parent group does not leak its threads) is a genuinely subtle correctness detail, handled correctly and well-tested.
- Backward compatibility for `filter=""` is preserved and explicitly covered.
- All user input is bound-parameterized; no injection surface introduced.

pkuWMH:
Closing: the followed/recent data already exists in the client-side conversation list (synced from IM layer). No backend API changes needed — the frontend ChatSelectorModal should filter candidates using the local conversation data directly, same as the main conversation list tabs do.

PR #69, opened 2026-06-06
fix(docker): add Asia/Shanghai timezone to api and worker images (#68)

### LINKED ISSUE #68: fix(docker): add timezone config to Dockerfile.api and Dockerfile.worker
## Bug Description

The summary worker container runs in UTC timezone by default. This causes `PreRetrievalNarrow` in `internal/pipeline/narrow.go` to pass incorrect "current time" to the LLM time-range extraction prompt, resulting in summaries returning **"在当前范围内未找到与主题相关的聊天记录"** even when messages clearly exist.

## Root Cause

`narrow.go:80` uses `time.Now()` to get "current time" for the LLM prompt. The prompt instructs the LLM to use `+08:00` timezone, but the container runs in UTC, so `time.Now().Format("15:04")` returns UTC time (e.g., `08:43`) instead of Beijing time (`16:43`).

The LLM then interprets this as `08:43+08:00` and sets `endTS` accordingly. Messages sent after `08:43 Beijing time` (which is `00:43 UTC`) are excluded from the query, returning 0 results.

### Example (from actual logs)

- User asks: "我最近发的最后一条消息是什么" at **16:43 Beijing time**
- Container `time.Now()` returns **08:43 UTC**
- LLM prompt says current time is `08:43`, timezone `+08:00`
- LLM narrows range to `[06-03 00:00 ~ 06-06 08:43] +08:00`
- `endTS` = `1780706580` (Beijing 08:43)
- Messages at 15:25 and 16:18 Beijing time have timestamps `1780730723` and `1780733881`
- Both timestamps > `endTS` → **0 messages fetched** → "未找到相关记录"

## Fix

Add timezone configuration to both `Dockerfile.api` and `Dockerfile.worker`, matching the pattern used by `octo-server` and `octo-speech`:

### Dockerfile.api

```diff
 FROM alpine:3.19
 RUN apk add --no-cache ca-certificates tzdata
+COPY --from=builder /usr/share/zoneinfo/Asia/Shanghai /etc/localtime
+RUN echo "Asia/Shanghai" > /etc/timezone
+ENV TZ=Asia/Shanghai
 COPY --from=builder /bin/summary-api /bin/summary-api
 EXPOSE 8080 8081
 ENTRYPOINT ["/bin/summary-api"]
```

### Dockerfile.worker

```diff
 FROM alpine:3.19
 RUN apk add --no-cache ca-certificates tzdata
+COPY --from=builder /usr/share/zoneinfo/Asia/Shanghai /etc/localtime
+RUN echo "Asia/Shanghai" > /etc/timezone
+ENV TZ=Asia/Shanghai
 COPY --from=builder /bin/summary-worker /bin/summary-worker
 EXPOSE 8082
 ENTRYPOINT ["/bin/summary-worker"]
```

### Reference

| Project | Timezone setup | Status |
|---------|---------------|--------|
| octo-server | `/etc/localtime` + `/etc/timezone` + `ENV TZ` | ✅ |
| octo-speech | `/etc/localtime` + `/etc/timezone` + `ENV TZ` | ✅ |
| octo-smart-summary (api) | only `tzdata` installed | ❌ |
| octo-smart-summary (worker) | only `tzdata` installed | ❌ |

## Verification

After fix, `time.Now()` in the container will return Beijing time, and the LLM time-range narrowing will produce correct `endTS` values that include recent messages.

boris-clark:
Closes #68.

## Problem

The api and worker containers ran in UTC because their runtime stage only installed `tzdata` without configuring the actual timezone. As a result `time.Now()` in `internal/pipeline/narrow.go` returned UTC wall-clock time, while the LLM time-range extraction prompt hardcodes `+08:00`. This shifted `endTS` ~8 hours early and filtered out same-day messages, producing "在当前范围内未找到与主题相关的聊天记录" even when messages existed.

## Fix

Configure the timezone to `Asia/Shanghai` in the runtime stage of both `Dockerfile.api` and `Dockerfile.worker`, matching the pattern already used by octo-server and octo-speech:

```dockerfile
FROM alpine:3.19
ENV TZ=Asia/Shanghai
RUN apk add --no-cache ca-certificates tzdata \
    && ln -snf /usr/share/zoneinfo/$TZ /etc/localtime \
    && echo $TZ > /etc/timezone
```

Note: instead of `COPY --from=builder /usr/share/zoneinfo/Asia/Shanghai /etc/localtime` (as drafted in the issue), the symlink is built from the runtime stage's own `tzdata`. The `golang:1.25-alpine` builder image does not ship `/usr/share/zoneinfo/Asia/Shanghai` by default, so copying from the builder risks a build failure.

## Scope

Dockerfile-only change. No Go code modified — all logic that requires UTC already calls `.UTC()` explicitly, and the remaining bare `time.Now()` uses are duration measurements (`time.Since`) unaffected by the local timezone.

## Verification

- Both images build successfully.
- `date` inside both containers returns `CST +0800`; `/etc/timezone` is `Asia/Shanghai`.
- Deployed to the local stack and confirmed via running containers (`TZ=Asia/Shanghai`, `+0800`); worker query timestamps now use Beijing time.

### DIFF
--- a/Dockerfile.api
+++ b/Dockerfile.api
@@ -7,7 +7,10 @@ COPY . .
 RUN CGO_ENABLED=0 go build -o /bin/summary-api ./cmd/summary-api
 
 FROM alpine:3.19
-RUN apk add --no-cache ca-certificates tzdata
+ENV TZ=Asia/Shanghai
+RUN apk add --no-cache ca-certificates tzdata \
+    && ln -snf /usr/share/zoneinfo/$TZ /etc/localtime \
+    && echo $TZ > /etc/timezone
 COPY --from=builder /bin/summary-api /bin/summary-api
 EXPOSE 8080 8081
 ENTRYPOINT ["/bin/summary-api"]
--- a/Dockerfile.worker
+++ b/Dockerfile.worker
@@ -7,7 +7,10 @@ COPY . .
 RUN CGO_ENABLED=0 go build -o /bin/summary-worker ./cmd/summary-worker
 
 FROM alpine:3.19
-RUN apk add --no-cache ca-certificates tzdata
+ENV TZ=Asia/Shanghai
+RUN apk add --no-cache ca-certificates tzdata \
+    && ln -snf /usr/share/zoneinfo/$TZ /etc/localtime \
+    && echo $TZ > /etc/timezone
 COPY --from=builder /bin/summary-worker /bin/summary-worker
 EXPOSE 8082
 ENTRYPOINT ["/bin/summary-worker"]


Jerry-Xin (approved):
This PR is in scope for `octo-smart-summary` and correctly fixes the API/worker runtime timezone mismatch behind the reported same-day filtering issue.

🔴 Blocking: None.

💬 Non-blocking: None.

✅ Highlights:
- `Dockerfile.api:10` and `Dockerfile.worker:10` set `TZ=Asia/Shanghai`, aligning container-local `time.Now()` behavior with the `+08:00` prompt assumption in `internal/pipeline/narrow.go`.
- `Dockerfile.api:11` and `Dockerfile.worker:11` install `tzdata` before linking `/etc/localtime`, so the target zoneinfo file exists in the runtime image.
- I built both images locally and verified each runtime reports `TZ=Asia/Shanghai`, `/etc/timezone` as `Asia/Shanghai`, `/etc/localtime` linked to `/usr/share/zoneinfo/Asia/Shanghai`, and `date` as `CST +0800`.

yujiawei (approved):
# Code Review — PR #69 (octo-smart-summary)

## Verdict: APPROVED

A correct, minimal, well-reasoned fix. The api and worker runtime images installed `tzdata` but never configured an actual zone, so the containers ran in UTC. Any bare `time.Now()` that feeds a wall-clock value into an LLM prompt (the time-range extractor, which hardcodes `+08:00`) was therefore 8 hours off, filtering out same-day messages. Pinning `Asia/Shanghai` in both runtime stages resolves this. No blocking issues found.

## Verification

- ✅ **Diff is symmetric and consistent** — `Dockerfile.api` and `Dockerfile.worker` receive the identical 3-line change (`ENV TZ`, plus `ln -snf .../localtime` and `echo $TZ > /etc/timezone` folded into the existing `apk add`). No layer-count regression, no extra packages.
- ✅ **Builder-vs-runtime tzdata reasoning is correct.** The PR deliberately symlinks from the runtime stage's own `tzdata` rather than `COPY --from=builder /usr/share/zoneinfo/...`. Confirmed sound: `golang:*-alpine` does not ship the zoneinfo database, so a copy-from-builder would risk a missing-path failure. Building the symlink after `apk add ... tzdata` in the same stage is the right pattern.
- ✅ **Bug root cause confirmed in code.** `internal/pipeline/narrow.go:80` uses a bare `time.Now()` whose `.Format("2006-01-02")` / `.Format("15:04")` / `.Weekday()` output is injected into the time-parser system prompt that mandates `+08:00` (`narrow.go:97`). Under UTC this produced an inconsistent reference date/time. This is precisely what the TZ fix corrects.
- ✅ **No UTC-dependent path is broken by the change.** All persistence and lease/deadline logic calls `time.Now().UTC()` explicitly (e.g. `internal/worker/processor.go:92,185,218`, `scheduler.go`, `api/handler/task.go`), so it is immune to the container local zone. Remaining bare `time.Now()` calls are either duration measurements via `time.Since` (`narrow.go:111`, `llm.go:247`, the `*Start` markers in `personal_processor.go`/`fetch.go`) or cache TTLs (`auth/cached_resolver.go`) — none are sensitive to wall-clock zone.
- ✅ **Build passes** — `go build ./...` succeeds at head SHA `4ce5b40`.

## Notes (non-blocking)

- **PR-description nit (no code impact):** the description states "the remaining bare `time.Now()` uses are duration measurements." That is slightly imprecise — `narrow.go:80`, `llm.go:401`, and `llm.go:446` are bare `time.Now()` used for *date/weekday formatting in LLM prompts*, not durations. They are not a problem; they are exactly the calls the fix is meant to correct (they now resolve to Beijing local time). Worth correcting the wording for the record, but nothing to change in the diff.
- **Latent consistency observation (P2, out of scope for this PR):** `internal/worker/processor.go:323` creates a `SummaryEvent` without setting `CreatedAt`, so GORM auto-fills it via its default `NowFunc` (local time). After this change that column is written in Beijing time, whereas the rest of the codebase standardizes on explicit `.UTC()` for DB writes. This is harmless here — `summary_event.created_at` is never read by value (the progress query in `task.go:984` orders by `MAX(id)`, not by time) — but if any future code starts comparing that column against a UTC value, the 8-hour skew would resurface. Not a reason to block; flagging so it's on record. Pinning `loc`/`parseTime` in the MySQL DSN would make timestamp handling zone-explicit end-to-end if the team wants to remove the ambiguity later.

## Summary

Scope is exactly two Dockerfiles, the pattern matches the stated convention used by sibling services, the build is green, and no UTC-dependent logic regresses. Approving.

PR #75, opened 2026-06-08
ci: pass TRIAGE_WEBHOOK_URL to octo-issue-notify

lml2468:
The issue notification caller only forwarded OCTO_BOT_TOKEN, so the triage webhook mode was silently skipped and the triage agent was never invoked. Forward the org-level TRIAGE_WEBHOOK_URL secret so triage runs on new issues.

### DIFF
--- a/.github/workflows/octo-issue-feed.yml
+++ b/.github/workflows/octo-issue-feed.yml
@@ -18,3 +18,4 @@ jobs:
       event_action: ${{ github.event.action }}
     secrets:
       OCTO_BOT_TOKEN: ${{ secrets.OCTO_BOT_TOKEN }}
+      TRIAGE_WEBHOOK_URL: ${{ secrets.TRIAGE_WEBHOOK_URL }}


Jerry-Xin (approved):
Single-line addition passes `TRIAGE_WEBHOOK_URL` secret to the reusable `octo-issue-notify` workflow. Correct placement under `secrets:` block, consistent with existing `OCTO_BOT_TOKEN` pattern. No logic change, no security concern (secret remains encrypted in GitHub). LGTM.

yujiawei (approved):
# Code Review — PR #75 (octo-smart-summary)

## Summary

This PR adds a single line to `.github/workflows/octo-issue-feed.yml`, forwarding the org-level `TRIAGE_WEBHOOK_URL` secret to the reusable `octo-issue-notify.yml@v1` workflow. The stated goal is to enable the (previously silently-skipped) triage-webhook path so new issues are forwarded to the triage agent.

The change is minimal, correct, and consistent with the design of the reusable workflow and its underlying transport action. **Verdict: APPROVED.**

## Verification (the cross-file contract holds)

Because the caller pins the reusable workflow at `@v1`, the secret must be declared at that exact ref — otherwise Actions fails with "secret X is not defined in the called workflow." I checked out `Mininglamp-OSS/.github` and verified every relevant tag:

- ✅ `octo-issue-notify.yml@v1` (and `v1.0.0`, `v1.1.0`, `v1.1.1`) all declare `TRIAGE_WEBHOOK_URL` under `secrets:` as `required: false`. The pinned `@v1` ref currently resolves to the same commit as `v1.1.1` (`392c7a9`), so the secret name matches and the call will not error.
- ✅ The reusable forwards the secret to the shared composite action `octo-notify@v1` as `webhook-url`. That action enables "webhook mode" only when `webhook-url` is non-empty; when the secret is unset it resolves to an empty string and the webhook POST is simply skipped (IM notification still runs). So the change degrades gracefully and is safe even before the secret is provisioned in a given repo.
- ✅ The secret is passed via the `secrets:` block (not hardcoded), and downstream it is consumed via `env:` and POSTed without any `${{ }}` shell/JSON interpolation. This is the correct handling for a capability secret in a public repo — no leakage into logs or shell.

## Security notes (PR was flagged security-sensitive)

- The webhook URL is treated as a capability secret throughout the chain (caller → reusable → composite action), passed by secret reference and never interpolated into a shell or string-built payload. No new secret-exposure surface is introduced.
- The forwarded payload is minimal (repo, issue number, title, url, author, action) and the title/author are sanitized for control characters in the reusable; the issue body is not egressed. No change here, but worth noting the existing posture is sound.

## For a human to confirm (operational, not a code blocker)

This PR's effectiveness depends entirely on `TRIAGE_WEBHOOK_URL` actually being available to this repository at runtime (org-level secret with the repo in scope, or a repo secret). If it is not provisioned, the line is a harmless no-op and triage still won't fire. Please confirm the secret is configured before relying on the triage path going live.

## Non-blocking observation

`octo-issue-feed.yml` still subscribes to `issues: [opened, reopened]`, while the reusable workflow centrally guards to `opened` only and skips everything else. This is pre-existing (not touched by this PR) and harmless — `reopened` events will invoke the reusable and immediately short-circuit. Optionally, the caller's trigger could be narrowed to `[opened]` to avoid spawning no-op runs, but that is outside this PR's scope.

## Conclusion

No P0/P1 issues. The cross-repo secret contract is satisfied at the pinned ref, secret handling is correct, and the change fails safe when the secret is absent. Approving.

PR #80, opened 2026-06-09
fix: support archived threads in smart summary when explicitly selected (#78)

### LINKED ISSUE #78: bug: smart summary backend hard-excludes archived threads (empty summary when selected)
## Problem

Archived threads are hard-excluded from smart summary at the backend in two independent layers, and the candidate response never tells the frontend whether a thread is archived. As a result, archived threads cannot be summarized, and even if a client supplies an archived thread id explicitly, it is silently dropped (producing an empty summary).

Related frontend issue: Mininglamp-OSS/octo-web#346

## Root cause (two layers)

1. Candidate listing — `internal/api/handler/candidates.go:170-173`: thread query hard-filters `WHERE t.status = 1`, so archived threads (`status = 2`) never reach the candidate list. The response object also has no `is_archived` / `status` field (`candidates.go:188-195`), so the frontend cannot distinguish active vs archived.

2. Fetch pipeline — `internal/pipeline/fetch.go:118-130` (`GetUserChannels`) filters `t.status = 1`; `ApplySourceConstraints` (`fetch.go:211-240`) intersects user-selected sources against `GetUserChannels` output. An explicitly-selected archived thread is absent from the allowed set and gets silently removed → empty summary. Ownership resolver (`internal/pipeline/resolve_channel.go:333,381`) also narrows by `status = 1`.

## Status model

Thread status: Active = 1, Archived = 2, Deleted = 3. Deleted (3) must remain excluded everywhere. Only Archived (2) becomes conditionally includable.

### 1. Candidate endpoint contract
- Accept `include_archived` query param (default `false`).
- When `false` → keep `t.status = 1`. When `true` → `t.status IN (1,2)` (keep `g.status = 1`, `message_count > 0`).
- Add `is_archived` to every candidate object (`true` iff thread `status = 2`; groups/directs always `false`).

### 2. Fetch pipeline
- Allow an explicitly-selected archived thread to survive `GetUserChannels` + `ApplySourceConstraints` (relax to `status IN (1,2)` only for explicitly-selected sources, so auto/background summaries keep `status = 1` and never leak archived threads).
- Relax ownership resolver thread queries (`resolve_channel.go:333,381`) for explicitly-selected ids; keep Deleted (3) excluded.

### 3. Default behavior
- Default `include_archived = false` (archived excluded, opt-in to include). Preserves current behavior, avoids performance regression, and is semantically correct (summary is read-only retrospection; forwarding always excludes archived).

## End-to-end contract (with octo-web#346)

| Direction | Name | Type | Default |
|---|---|---|---|
| Request → backend | `include_archived` | bool (query) | `false` |
| Response → frontend | `is_archived` | bool (per candidate) | `false` |

## Critical risk

Picker/fetch mismatch: if candidate listing is relaxed but the fetch pipeline is not, users select a now-visible archived thread and get a silent empty summary. Both layers must ship together; pipeline tests are the gate.

## Tests
- `candidates_test.go`: seed status 1/2/3; assert default → only status=1; `include_archived=true` → status 1 and 2, never 3; `is_archived` correct.
- Pipeline: with `specifiedSources` = one archived thread, assert it is retained and messages are fetched; with no explicit sources, assert archived stays out of auto-discovery; deleted thread never fetched; cover ownership resolver for archived selected thread.

jeff-wilson2010:
Closes #78. Frontend counterpart: Mininglamp-OSS/octo-web#346.

## Summary

Archived threads (status=2) were hard-excluded from smart summary in two independent backend layers, and the candidate response never told the frontend whether a thread was archived. This change lets an explicitly-selected archived thread be summarized, while keeping auto/background summaries to active threads only and always excluding deleted (status=3).

## Changes

- `internal/api/handler/candidates.go`: parse `include_archived` query param (default false). Thread candidate query uses `t.status = 1` by default and `t.status IN (1,2)` when `include_archived=true` (keeps `g.status = 1`, `message_count > 0`; never status=3). Added `is_archived` to all candidate objects (threads: true iff status=2; groups/directs: always false).
- `internal/pipeline/fetch.go`: relax thread discovery to `status IN (1,2)` ONLY for explicitly-selected sources, threaded through `GetUserChannels` / `IntersectParticipantChannels` so auto/background summaries (no explicit sources) keep `status = 1` and never include archived threads.
- `internal/pipeline/resolve_channel.go`: relax thread ownership queries (creator/member) to `status IN (1,2)` for explicitly-selected ids; deleted always excluded (defense-in-depth, documented in comments).

## End-to-end contract (with octo-web#346)

- Request: `include_archived` (bool query param, default false).
- Response: `is_archived` (bool per candidate, default false).

## Tests

- `candidates_test.go`: seeds status 1/2/3; asserts default returns only active; `include_archived=true` returns active + archived, never deleted; `is_archived` correct per type.
- `fetch_archive_test.go`: archived excluded from auto-discovery; explicitly-selected archived retained through the full pipeline (single-person AND multi-person participant intersect) and produces non-empty messages; selected-deleted never retained; ownership resolver covers archived selected thread for creator/member roles.

`go build ./...`, `go vet ./...`, `go test ./...` all pass.

## Coordination

Must ship together with octo-web#346 — relaxing the candidate listing without relaxing the fetch pipeline would let users select an archived thread and get an empty summary.

### DIFF
--- a/internal/api/handler/candidates.go
+++ b/internal/api/handler/candidates.go
@@ -120,6 +120,9 @@ func (h *CandidateHandler) SearchChatCandidates(c *gin.Context) {
 
 	keyword := c.Query("keyword")
 	chatType := c.Query("chat_type") // "group", "direct", or "" = all
+	// include_archived: when true, threads with status=2 (Archived) are also
+	// returned. Defaults to false. Deleted threads (status=3) are NEVER returned.
+	includeArchived := c.Query("include_archived") == "true" || c.Query("include_archived") == "1"
 
 	// Resolve current user from context (set by AuthMiddleware).
 	currentUID, _ := c.Get("user_id")
@@ -155,6 +158,7 @@ func (h *CandidateHandler) SearchChatCandidates(c *gin.Context) {
 				"chat_type":    "group",
 				"name":         g.Name,
 				"member_count": nil,
+				"is_archived":  false,
 			})
 		}
 	}
@@ -165,12 +169,21 @@ func (h *CandidateHandler) SearchChatCandidates(c *gin.Context) {
 			ShortID string `gorm:"column:short_id"`
 			Name    string `gorm:"column:name"`
 			GroupNo string `gorm:"column:group_no"`
+			Status  int    `gorm:"column:status"`
 		}
 		var threads []imThread
 		q := h.imDB.Table("thread t").
-			Select("DISTINCT t.short_id, t.name, t.group_no").
-			Joins("INNER JOIN `group` g ON g.group_no" + h.collate + " = t.group_no").
-			Where("t.status = 1 AND g.status = 1 AND t.message_count > 0")
+			Select("DISTINCT t.short_id, t.name, t.group_no, t.status").
+			Joins("INNER JOIN `group` g ON g.group_no" + h.collate + " = t.group_no")
+		// Default: only Active threads (status=1). When include_archived is set,
+		// also return Archived threads (status=2). Deleted (status=3) is never
+		// included. Parent group must be active (g.status=1) and the thread must
+		// have messages.
+		if includeArchived {
+			q = q.Where("t.status IN (1, 2) AND g.status = 1 AND t.message_count > 0")
+		} else {
+			q = q.Where("t.status = 1 AND g.status = 1 AND t.message_count > 0")
+		}
 		if currentUIDStr != "" {
 			// Use group_member instead of thread_member so that all threads in the
 			// user's groups are returned, not just threads the user has posted in.
@@ -191,6 +204,7 @@ func (h *CandidateHandler) SearchChatCandidates(c *gin.Context) {
 				"name":            t.Name,
 				"member_count":    nil,
 				"parent_group_no": t.GroupNo,
+				"is_archived":     t.Status == 2,
 			})
 		}
 	}
@@ -238,6 +252,7 @@ func (h *CandidateHandler) SearchChatCandidates(c *gin.Context) {
 				"name":         name,
 				"member_count": nil,
 				"is_bot":       d.Robot == 1,
+				"is_archived":  false,
 			})
 		}
 	}
--- a/internal/api/handler/candidates_test.go
+++ b/internal/api/handler/candidates_test.go
@@ -47,6 +47,19 @@ func doCandidateRequest(r *gin.Engine, userID, chatType, keyword string) *httpte
 	return w
 }
 
+// doCandidateRequestRaw issues a request with a verbatim query string, allowing
+// arbitrary params such as include_archived.
+func doCandidateRequestRaw(r *gin.Engine, userID, query string) *httptest.ResponseRecorder {
+	w := httptest.NewRecorder()
+	req := httptest.NewRequest("GET", "/api/v1/summary-chat-candidates?"+query, nil)
+	if userID != "" {
+		req.Header.Set("Token", userID)
+	}
+	req.Header.Set("X-Space-Id", "space1")
+	r.ServeHTTP(w, req)
+	return w
+}
+
 func TestSearchChatCandidates_ThreadUsesGroupMember(t *testing.T) {
 	imDB := setupCandidateImDB(t)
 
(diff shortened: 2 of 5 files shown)


Jerry-Xin (approved):
Summary: The PR is in scope for `octo-smart-summary` and correctly updates candidate discovery plus the fetch pipeline so explicitly selected archived threads can be summarized while deleted threads remain excluded.

🔴 Blocking: None.

✅ Highlights:
- `internal/api/handler/candidates.go:123` adds an explicit archived opt-in, and `internal/api/handler/candidates.go:182` keeps deleted threads excluded by enumerating only statuses `1` and `2`.
- `internal/api/handler/candidates.go:207` exposes `is_archived`, matching the frontend contract.
- `internal/pipeline/fetch.go:129` scopes archived discovery to selected thread IDs, avoiding archived leakage into auto/background summaries.
- `internal/pipeline/fetch.go:622` and `internal/pipeline/fetch.go:632` correctly thread the selected archived IDs through creator and participant discovery.
- `internal/pipeline/resolve_channel.go:339` and `internal/pipeline/resolve_channel.go:389` keep ownership filtering defensive by admitting active/archived but still excluding deleted threads.
- The new tests cover default exclusion, explicit inclusion, deleted exclusion, single-person fetch, participant intersection, and ownership behavior.

Verification:
- `go test ./internal/api/handler ./internal/pipeline` passed.
- `go test ./...` passed.
- `go vet ./...` passed.
- `go build ./...` passed.

lml2468 (approved):
## Verdict: APPROVED

Three-layer fix that closes #78 cleanly. The blast radius is correctly scoped — archived threads (status=2) only escape exclusion when the caller explicitly named them; auto/background paths and Deleted (status=3) are unchanged. Verified against head `981e4cf`.

### Architecture is sound

The "two backend layers excluded archived" framing in the PR description matches the codebase: `internal/api/handler/candidates.go` (listing) and `internal/pipeline/fetch.go` + `resolve_channel.go` (resolution). Patching the listing alone would have shipped a UX trap where the user picks an archived thread and gets an empty summary; relaxing only the pipeline would have left the picker unable to surface it. The PR fixes both **and** documents the coordination requirement with `octo-web#346` in the description.

### Scope guarantee is correctly enforced

The relaxation is gated on **caller-supplied selected ids**, not on a global feature flag:

- `GetUserChannels(ctx, uid, db, selectedThreadIDs...)` — when the variadic slice is empty (auto/background), discovery stays on `t.status = 1`; when ids are passed, it widens to `(t.status = 1 OR (t.status = 2 AND CONCAT(t.group_no, '____', t.short_id) IN ?))`. Deleted (3) is never admitted — verified by enumerating `IN (1,2)` rather than excluding `!= 3`, which is the right way to avoid future status-value bugs.
- `selectedThreadChannelIDs(specifiedSources)` extracts only frontend `source_type=2` (thread) — DM/group ids cannot accidentally widen the thread filter.
- Layer 1.5 (`IntersectParticipantChannels`) threads the same slice into each participant's discovery call. This is the **non-obvious** load-bearing piece: without it, the participant lookup would default to `status=1` only and drop the creator-side archived thread in the intersect, producing the exact "user selected an archived thread and got empty summary" bug the PR claims to prevent. Test `TestResolveAndFetch_MultiPerson_SelectedArchivedThread_Retained` is a direct regression guard for this, and the inverse guard `TestIntersectParticipantChannels_EmptySelection_NoArchived` proves the gating actually works in the unselected case.

### Defense-in-depth in `resolve_channel.go` is honest about its scope

The ownership branch relaxation to `status IN (1, 2)` is correctly annotated as defense-in-depth: `ResolveChannelScope` (Layer 1.7) is gated on `len(specifiedSources) == 0` in `fetch.go`, so archived threads never actually reach `filterByOwnership` in the live pipeline. The comment is explicit about that — future maintainers won't misread it as load-bearing. Critically, `status IN (1, 2)` (not `!= 3`) preserves the "Deleted never matches" invariant even if a future refactor changes the gating.

### Test coverage is genuinely end-to-end

`fetch_archive_test.go` (+330 lines) does not just hit `GetUserChannels` in isolation — it drives `ResolveAndFetchMessagesForPersonal` through to message bytes, with a sqlite-backed schema that includes the `utf8mb4_unicode_ci` collation the raw SQL joins on. The 9 archived/deleted assertions cover the matrix that matters:
- auto-discovery (empty selection) → archived excluded ✅
- explicit selection of archived → retained AND messages returned ✅
- explicit selection of *deleted* → never retained ✅
- multi-person intersect with archived → retained ✅
- multi-person intersect *without* selection scope → archived dropped (gating proof) ✅
- ownership filter (creator + member roles) → archived retained, deleted dropped ✅

`candidates_test.go` (+157 lines) covers the three include_archived states (default / =false / =true), the `is_archived` field per candidate type, and explicitly asserts Deleted is never returned under `include_archived=true`.

`go build ./... && go vet ./... && go test ./internal/api/handler/... ./internal/pipeline/... -count=1` all clean locally.

### Nits (non-blocking)

- `includeArchived := c.Query("include_archived") == "true" || c.Query("include_archived") == "1"` is case-sensitive — `"True"` / `"TRUE"` / `"yes"` silently fall back to false. Fine as a strict contract with octo-web#346, but worth a one-line comment that the values are `"true"`/`"1"` exclusively, so a future frontend doesn't get a silent mismatch.
- The `status IN (1, 2)` defense-in-depth comments in `resolve_channel.go` are accurate today but become live load-bearing if anyone later removes the `len(specifiedSources) == 0` gate on Layer 1.7. Consider adding a TODO/note pointing back to that gate so the invariant chain stays visible.

Neither blocks merge; both are documentation-only.

### Coordination ack

The PR description correctly flags the must-ship-together coupling with `Mininglamp-OSS/octo-web#346`. Merge order should be: web PR lands first (so users see the new picker) or both flipped behind the same release — backend-only deploy is safe (default `include_archived=false` means no behavior change for existing callers) but exposes new query params the FE isn't sending yet, which is harmless.

https://github.com/Mininglamp-OSS/octo-smart-summary/pull/80

mochashanyao (approved):
[Octo-Q · automated review]

---

# octo-smart-summary PR#80 Review Report (automated review)

**PR**: https://github.com/Mininglamp-OSS/octo-smart-summary/pull/80
**Title**: fix: support archived threads in smart summary when explicitly selected (#78)
**Head SHA**: `981e4cfe283ee967ea298c7b01db937a027ed791`
**Files**: 5 changed (+585 / −21)
**Build/Vet/Tests**: ✅ all pass

---

## 1. Verification Summary

| Check | Result | Evidence |
|-------|--------|----------|
| `go build ./...` | ✅ pass | exit 0 |
| `go vet ./...` | ✅ pass | exit 0 |
| `go test ./...` | ✅ pass | all packages green |
| Candidate API: `include_archived` gating | ✅ correct | `candidates.go:123,178-185` |
| Candidate API: `is_archived` on all types | ✅ correct | `candidates.go:161,207,255` |
| Pipeline Layer 1: scoped archived relaxation | ✅ correct | `fetch.go:131-143` |
| Pipeline Layer 1.5: selectedThreadIDs threaded | ✅ correct | `fetch.go:418,629` |
| Pipeline: deleted (status=3) never admitted | ✅ correct | all SQL conditions enumerate `(1,2)` only |
| Ownership filter: defense-in-depth | ✅ correct | `resolve_channel.go:338-345,389-395` |
| Test coverage | ✅ comprehensive | 15 new test functions across 2 files |

## 2. Findings

### No P0 or P1 issues found.

### P2 Observations

**P2-1: `filterByPersonIntersection` / `filterByPersonUnion` don't thread `selectedThreadIDs`**
- **Path**: `resolve_channel.go:487,519` — both call `GetUserChannels(ctx, uid, imDB)` without `selectedThreadIDs`.
- **Diff-scope**: pre-existing (these functions are not touched by this PR).
- **Impact**: None today — Layer 1.7 only runs when `len(specifiedSources) == 0` (`fetch.go:637`), which means `selectedThreadIDs` is always empty in that path. But if the Layer 1.7 guard is ever relaxed, archived threads would be silently dropped from person-based narrowing.
- **Recommendation**: Add a `// NOTE:` comment near the Layer 1.7 guard or in `filterByPersonIntersection`/`filterByPersonUnion` noting that `selectedThreadIDs` is intentionally not threaded because Layer 1.7 only runs in the no-explicit-sources path.

**P2-2: `include_archived` param parsing is strict**
- **Path**: `candidates.go:123` — `c.Query("include_archived") == "true" || c.Query("include_archived") == "1"`
- **Diff-scope**: new (introduced by this PR).
- **Impact**: Values like `"True"`, `"TRUE"`, `"yes"` silently default to false. This is standard Go/Gin behavior and the frontend contract is well-defined (bool → `"true"`), but worth noting for robustness.
- **Recommendation**: No action needed if the frontend contract is controlled. Consider `strings.EqualFold` if third-party callers are expected.

**P2-3: `DISTINCT` with added `status` column**
- **Path**: `candidates.go:172` — `Select("DISTINCT t.short_id, t.name, t.group_no, t.status")`
- **Diff-scope**: new (status column added by this PR).
- **Impact**: `DISTINCT` now operates on 4 columns instead of 3. Since each thread has exactly one status, this doesn't change dedup behavior. No issue.

## 3. Data-Flow Tracing

### Flow A: Candidate API (`include_archived` → SQL → response)
1. `c.Query("include_archived")` → `includeArchived` bool (`candidates.go:123`)
2. `includeArchived=true` → SQL `WHERE t.status IN (1, 2) AND g.status = 1 AND t.message_count > 0` (`candidates.go:181`)
3. `includeArchived=false` → SQL `WHERE t.status = 1 AND g.status = 1 AND t.message_count > 0` (`candidates.go:183`)
4. `imThread.Status` scanned from DB → `is_archived: t.Status == 2` in response (`candidates.go:207`)
5. Groups and directs always get `is_archived: false` (`candidates.go:161,255`)
✅ Data flows correctly from query param to SQL to response.

### Flow B: Pipeline (`specifiedSources` → `selectedThreadChannelIDs` → `GetUserChannels`)
1. `specifiedSources` (caller-provided `[]map[string]interface{}`) → `selectedThreadChannelIDs()` (`fetch.go:244-255`)
2. Filters by `mapFrontendSourceType(sourceType(s)) == 5` (frontend type 2 = thread) (`fetch.go:249`)
3. Extracts `source_id` string → `[]string` of channel IDs (`fetch.go:251-253`)
4. Passed to `GetUserChannels(ctx, creatorUID, imDB, selectedThreads...)` (`fetch.go:621`)
5. Inside `GetUserChannels`: when `len(selectedThreadIDs) > 0`, SQL condition becomes `(t.status = 1 OR (t.status = 2 AND CONCAT(t.group_no, '____', t.short_id) IN ?))` (`fetch.go:134-136`)
6. `threadArgs` properly binds `selectedThreadIDs` as GORM parameter (`fetch.go:137`)
✅ Data flows correctly. Archived threads only admitted when their exact channel ID is in the selected set.

### Flow C: Multi-person (Layer 1.5 intersect)
1. `selectedThreads` computed once at `fetch.go:620`
2. Passed to both `GetUserChannels` (Layer 1, `fetch.go:621`) and `IntersectParticipantChannels` (Layer 1.5, `fetch.go:629`)
3. `IntersectParticipantChannels` forwards `selectedThreadIDs...` to each participant's `GetUserChannels` call (`fetch.go:418`)
4. Intersection logic at `fetch.go:421-432` correctly uses channel ID sets
✅ Archived thread survives both creator and participant discovery when explicitly selected.

### Flow D: Ownership filter (defense-in-depth)
1. `filterByOwnership` only called from `executeRule` → `ResolveChannelScope` → Layer 1.7
2. Layer 1.7 guard: `len(specifiedSources) == 0` (`fetch.go:637`)
3. When `specifiedSources` is empty → `selectedThreadChannelIDs` returns `[]` → `GetUserChannels` uses `t.status = 1` → no archived threads reach Layer 1.7
4. Therefore `status IN (1, 2)` in ownership queries (`resolve_channel.go:343,393`) is unreachable today
✅ Correctly documented as defense-in-depth. No data-flow issue.

## 4. Additional Observations

- **Backward compatibility**: Both `GetUserChannels` and `IntersectParticipantChannels` use variadic `...string` for the new parameter, so existing callers (without the arg) get the original `status=1` behavior. Clean.
- **`sourceType` helper extraction** (`fetch.go:227-234`): Good refactor — eliminates duplicated type-assertion boilerplate. The `onlyDMSources` loop at `fetch.go:720` is now cleaner.
- **Test quality**: Tests use real SQLite in-memory DBs with MySQL collation registered, covering the full pipeline path including multi-person intersect. The test data is seeded through the same SQL patterns as production, not hand-fed structs.

## 5. Conclusion

This is a well-structured, well-tested change that correctly implements the stated goal: archived threads are admitted only when explicitly selected, auto/background summaries remain active-only, and deleted threads are always excluded. The data-flow tracing confirms that the `selectedThreadIDs` scope is properly threaded through all three pipeline layers (L1, L1.5, and the ownership filter). No P0 or P1 issues found.
[Octo-Q] verdict: APPROVE — no blocking issues; only P2/nit observations (pre-existing gap in filterByPerson*, strict param parsing, DISTINCT column addition — none affect correctness).

yujiawei (approved):
# Code Review — PR #80 (octo-smart-summary)

## Verdict: APPROVED

The change is correct, well-scoped, and well-tested. It cleanly achieves the stated goal: an **explicitly-selected** archived thread (`status=2`) can now be summarized, while auto/background summaries stay `status=1`-only and **deleted** threads (`status=3`) are never returned in any path. `go build`, `go vet`, and `go test ./...` all pass on head SHA `981e4cf`.

No P0/P1 issues found. The notes below are non-blocking suggestions (P2) that can be addressed in this PR or a follow-up.

## 1. Verification of the core invariants

I traced all three layers that gate thread status:

- ✅ **Candidate listing** (`internal/api/handler/candidates.go:182-186`): default `t.status = 1`; with `include_archived` → `t.status IN (1, 2)`. `status=3` is never admitted in either branch. `is_archived` is set correctly per type (threads: `t.Status == 2`; groups/directs: always `false`).
- ✅ **Discovery / intersect** (`internal/pipeline/fetch.go:268-290, 421`): `GetUserChannels` defaults to `t.status = 1`; the relaxation `(t.status = 1 OR (t.status = 2 AND CONCAT(...) IN ?))` only triggers when `selectedThreadIDs` is non-empty. `status=3` can never match because the condition enumerates allowed statuses rather than excluding `3`. The same scope is correctly threaded into each participant's lookup in `IntersectParticipantChannels` (`fetch.go:421`) — this is the actual fix for the multi-person empty-summary regression, and `TestResolveAndFetch_MultiPerson_SelectedArchivedThread_Retained` guards it.
- ✅ **Ownership filter** (`internal/pipeline/resolve_channel.go:341-346, 391-397`): relaxed to `status IN (1, 2)`; `status=3` excluded. The comment correctly notes this is defense-in-depth (this path only runs when no explicit sources are present, and Layer 1 has already excluded archived there).

## 2. Cross-file / end-to-end contract

- ✅ The identifier used to scope the relaxation (`selectedThreadChannelIDs`, `fetch.go:244-255`) matches `source_id` against `CONCAT(t.group_no, '____', t.short_id)`, which is exactly the `chat_id` format the candidate listing emits (`candidates.go:202`) and the `channel_id` `GetUserChannels` builds (`fetch.go`). The frontend→backend source-type mapping (frontend `2`=thread → backend `5`) is consistent (`fetch.go:214-225, 247`).
- ✅ `ApplySourceConstraints` (Layer 2) does not drop the selected archived thread once it survives discovery, and Layer 1.7 channel-scope narrowing only runs when `len(specifiedSources) == 0` (`fetch.go:640`), so it cannot interfere with an explicit selection.
- ✅ The `sourceType` helper extraction (`fetch.go:227-237`) correctly dedups the previously-inlined int/float64 handling and is reused in the Layer 4.5 DM-only check (`fetch.go:720`).
- ✅ As the PR notes, this must ship together with octo-web#346; relaxing candidate listing without the pipeline relaxation would yield empty summaries. The backend half here is internally consistent.

## 3. Security / visibility

- ✅ All new/changed SQL is parameterized (`IN ?`, bound `uid`/ids); no string interpolation of user input. The `include_archived` parse is a strict equality check, not reflected into SQL.
- ✅ No visibility bypass introduced: archived threads are gated by the same `thread_member`/`group_member` membership and `g.status = 1` parent-group checks as active threads. Relaxing status does not relax membership.

## 4. Non-blocking suggestions (P2)

These are quality/coverage notes, not merge blockers:

1. **Non-sargable `CONCAT(...) IN (?)` in ownership queries** (`resolve_channel.go:343, 392`). `CONCAT(group_no, '____', short_id)` in the `WHERE` prevents index use on `(group_no, short_id)`. Mitigated by the pre-filtered candidate set and early `status IN (1,2)` filter, so impact is low today — worth keeping in mind if these lists grow. (Pre-existing pattern, only extended here.)
2. **Untested `include_archived=1` numeric form** (`candidates.go:125`). The `== "1"` branch is supported but only `true`/`false` are exercised in tests — the branch could be silently dropped without a failing test.
3. **Misleading test name** (`candidates_test.go:330`, `TestSearchChatCandidates_GroupAndDirectHaveIsArchivedFalse`). It only exercises `chat_type=group`; the direct-message `is_archived=false` path (`candidates.go:255`) is never hit because the DM schema/tables aren't seeded. Consider renaming to reflect group-only coverage, or seed the DM path.
4. **No direct unit tests for `selectedThreadChannelIDs` / `sourceType` helpers**, including a mixed-source selection (thread + group + DM in one request) and the `float64` (JSON-decoded) branch of `sourceType`. The logic is exercised indirectly via integration tests on the pure-thread happy path; a small table-driven unit test would lock in the filtering behavior against future refactors.

Overall: solid, defensive, and the regression-guard tests (multi-person intersect, selected-deleted rejection, empty-selection-no-archived) are the right ones. Approving.

PR #77, opened 2026-06-08
feat(schedule): scheduled summary updates (clean re-integration of PR #62)

Ranwanglc:
## 概述

在从 `main` 新建的干净分支上**重新整合** PR #62 的「定时总结」能力，剔除其 15 轮 review 迭代过程中累积的冗余（7 个迁移文件、13 个 `pr62_roundN` 过程测试），并修复一处在 merge 过程中引入的回归。

本 PR 是 #62 的**整理替代版**，业务能力等价，但迁移与测试更干净。

## 业务功能（取 #62 最终 review 状态）

- 间隔式周期（天 / 周 / 月）与 legacy cron 并存；支持 `run_time` 运行时刻、`day_of_week` / `day_of_month` 锚点，以及持久化的 `anchor_dom` 防止月模式短月钳位后逐月漂移
- 任务↔定时**一对一活动绑定**；删除联动（级联软删，且尊重定时归属权限）；定时到点**原地覆盖执行**，含 CAS 抢占与防重叠
- 各环节计时日志 + 每次运行的 LLM 调用汇总报告（路径可配）
- 时区统一在代码层处理（`internal/timezone`，Asia/Shanghai）

## 迁移：7 → 2，全部 forward-only

- `20260608-01-schedule-interval-fields.sql`：`summary_schedule` 的 6 个周期相关列合并为一条 `ALTER`
- `20260608-02-task-schedule-binding.sql`：`live_schedule_id` 生成列 + 唯一绑定索引；`summary_participant(task_id,user_id)` 唯一索引

相较 #62 删除的内容及理由：
- 删除 `0604-03` 脏数据清理 / 回填脚本——脏数据是 #62 中间版本自身造成的，从 `main` 干净建支不会产生
- 删除 `0609-*` 中基于 `information_schema` 的「列是否已存在」幂等防御 DDL——干净库中目标列必然不存在，无需 `PREPARE/EXECUTE` 体操

**已在真实 MySQL 8.0 上验证**：完整迁移链正向 apply 全绿、down 回滚干净；生成列语义正确（两个活动任务绑同一 schedule 触发 1062；软删其一后另一可绑定）。

## 修复一处 #62 回归

`Regenerate` 接口在某次 merge `main` 时丢失了**请求体校验 + 改 topic 更新标题 + 终态原子转换防并发**。表现：畸形 JSON 不再返回 400、并发重新生成会重复干活。本 PR 已恢复（`main` 上原有的 4 个 `Regenerate` 测试重新变绿）。

[...]

PR #86, opened 2026-06-10
feat(schedule): schedule one-to-many summary_task

Ranwanglc:
## 背景
将 `schedule` 与 `summary_task` 由一对一改为一对多：每次定时到点 INSERT 一条全新 `summary_task`（本次窗口独立产出），不再重置老任务回 Pending。从根上消除 `uk_live_schedule_binding` 脏数据问题，对齐 Airflow DAG/DAGRun、K8s CronJob/Job 通用模型。

## 改动
- **migration**：删除 `live_schedule_id` 生成列与 `uk_live_schedule_binding` 唯一键，`schedule_id` 退化为普通外键；新增 `(space_id, schedule_id, deleted_at, id)` 联合索引覆盖「按 schedule_id 取最新」；保留 participant 唯一键 `(task_id, user_id)`（worker ON CONFLICT 依赖）。
- **scheduler**：`claimAndCreateScheduledTask` 改为 INSERT 新 task；重叠保护——上一条 Processing 未过期则跳过并照推 `next_run_at`，已过期则标 Failed 再新建（防卡死永久阻塞）；`last_run_at` 仅在成功新建时推进。
- **api/schedule**：拆除一对一绑定/双锁/409 守卫等遗留逻辑。
- **api/task**：删除定时类总结 = 软删该 schedule 名下全部 task + 停 schedule，仅创建者可删（否则 403）；手动类删除仍只软删自身。
- **ListSummaries**：定时按 schedule_id 折叠只露最新，手动不受影响。

## 验证
- `go build ./...` 通过，`go test ./internal/worker/...` 通过。
- 本地重建 summary-api / summary-worker 镜像并重启，健康检查 200。
- 数据库迁移已在本地实测应用（schema 为一对多目标态）。


mochashanyao (changes requested):
[Octo-Q · automated review]

---

# octo-smart-summary PR#86 Review Report
**PR:** feat(schedule): schedule one-to-many summary_task
**Head SHA:** e9faf0d73a3ad93ecdff32302c0774312126e3f0
**Routing:** complexity=security_sensitive (automated review)

---

## 1. Verification Summary

| Area | Status | Evidence |
|------|--------|----------|
| Scheduler 1->N INSERT logic | ✅ | `scheduler.go:130-155` — new task created with correct fields from `lockedSched` |
| Overlap guard (Processing skip) | ✅ | `scheduler.go:115-122` — checks latest task status, preserves `last_run_at` |
| `buildScheduledTaskChildren` | ✅ | `scheduled_replace_helpers.go:49-58` — creates source/participant/personal_result from schedule config |
| Multi-person guard (schedule config) | ⚠️ | `scheduler.go:237-260` — silently swallows JSON errors (fail-open) |
| `ListSummaries` folding | ✅ | `task.go:382-395` — ROW_NUMBER PARTITION BY COALESCE(schedule_id, id) correctly folds scheduled, keeps manual |
| `DeleteSummary` group delete | ✅ | `task.go:960-990` — creator-only group delete, non-creator gets unbind+pause |
| `DeleteSchedule` group soft-delete | ✅ | `schedule.go:1130-1160` — batch UPDATE, preserves schedule_id attribution |
| Migration completeness | ❌ | `20260610-01` does NOT drop `live_schedule_id` / `uk_live_schedule_binding` from deleted `20260608-02` |
| Test coverage | ✅ | `scheduler_lastrun_test.go` — covers first-run, overlap, multi-person disable |

## 2. Findings

### F1 — P1: Migration does not drop `uk_live_schedule_binding` unique key (BLOCKING)

**Diff-scope:** NEW (introduced by this PR's migration gap)

**What:** The deleted migration `20260608-02-task-schedule-binding.sql` added:
- `live_schedule_id` GENERATED ALWAYS AS stored column on `summary_task`
- `uk_live_schedule_binding` UNIQUE KEY on `live_schedule_id`

The new migration `20260610-01-schedule-one-to-many.sql` does NOT drop either. It only adds:
- `uk_summary_participant_task_user` on `summary_participant`
- `idx_space_schedule_deleted_id` on `summary_task`

**Impact:** If `20260608-02` was applied to any environment (staging/production), the unique key `uk_live_schedule_binding` still exists. The `live_schedule_id` generated column evaluates to `schedule_id` for every non-deleted task with a bound schedule. The UNIQUE constraint therefore **blocks the second task INSERT** under the same schedule — the core 1->N feature.

**Data-flow trace:**
1. `claimAndCreateScheduledTask` (`scheduler.go:130`) → `tx.Create(&newTask)` with `ScheduleID = &lockedSched.ID`
2. MySQL computes `live_schedule_id = schedule_id` (non-null, deleted_at IS NULL)
3. `uk_live_schedule_binding` fires: `Duplicate entry 'X' for key 'uk_live_schedule_binding'` → MySQL 1062
4. Transaction rolls back → `last_run_at` NOT advanced → schedule stuck, re-scanned every 60s, fails every time

**Severity (R1):** Makes the core 1->N feature non-functional in any environment where the old migration was applied. User-visible: scheduled summaries stop producing after the first run. **P1 blocking.**

**Fix:** Add to `20260610-01` UP section (idempotent form):
```sql
-- Defensive cleanup: drop generated column + unique key from 20260608-02 if present.
-- MySQL does not support IF EXISTS for ALTER TABLE DROP COLUMN/INDEX, so wrap in
-- a procedure or check information_schema first.
ALTER TABLE `summary_task` DROP INDEX `uk_live_schedule_binding`;
ALTER TABLE `summary_task` DROP COLUMN `live_schedule_id`;
```

### F2 — P2: `scheduleConfigMultiPerson` silently swallows JSON parse errors (fail-open)

**Diff-scope:** NEW (new function introduced by this PR)

**What:** `scheduler.go:243-245`:
```go
if err := json.Unmarshal(sched.ParticipantConfig, &participants); err != nil {
    return false, nil  // silently treats corrupt config as single-person
}
```

The API handler's equivalent check (`storedParticipantConfigSubsetOfCreator` at `schedule.go:155-162`) **fails closed** on parse errors:
```go
if err := json.Unmarshal(raw, &stored); err != nil {
    return false  // "Unparseable stored config is treated as unsafe (fail-closed)"
}
```

**Impact:** If `ParticipantConfig` becomes corrupt (e.g., manual DB edit, migration bug), the scheduler treats it as single-person and creates tasks with potentially wrong participants. The API handler would have refused to bind such a config. Divergent fail-safe directions.

**Severity (R1/R2):** Low probability (config is validated at write time), but the inconsistency is real. Not user-visible under normal operation. **P2.**

**Fix:** Change to `return false, fmt.Errorf("invalid participant_config JSON: %w", err)` or match the API handler's fail-closed direction.

### F3 — P2: `saveLatestResultAndCompleteTask` `isScheduled` pruning is dead code under 1->N

**Diff-scope:** PRE-EXISTING (not changed by this PR, but rendered dead by the 1->N model)

**What:** `scheduled_replace_helpers.go:273-298` — the `isScheduled` branch prunes prior `SummaryResult` rows for the same `task_id`. Under 1->N, each scheduled run has a unique `task_id` with exactly one result. The DELETE matches zero rows.

**Impact:** No bug — the SQL is harmless. But the code and comments ("Scheduled runs overwrite in place") are misleading under the new model.

**Severity:** P2 (maintainability).

### F4 — P2: `DeleteSchedule` does not set `status = -1` on group soft-delete (inconsistency)

**Diff-scope:** AMPLIFIED (pre-existing pattern, now affects batch of tasks instead of one)

**What:** `schedule.go:1150-1155` — `DeleteSchedule` sets only `deleted_at` on tasks:
```go
Update("deleted_at", &now)
```

But `DeleteSummary` group delete (`task.go:980-985`) sets both:
```go
Updates(map[string]interface{}{"status": -1, "deleted_at": now})
```

**Impact:** GORM soft-delete plugin filters `deleted_at IS NULL` on all queries, so deleted tasks are excluded regardless of status. However, any raw SQL or external tool that checks `status` instead of `deleted_at` would see inconsistent state. Low practical impact.

**Severity:** P2.

## 3. Suggestions

1. **Migration (F1 — must fix):** Add defensive DROP for `uk_live_schedule_binding` and `live_schedule_id` to `20260610-01`. Use `information_schema` checks or a stored procedure for idempotency since MySQL lacks `DROP COLUMN IF EXISTS`.

2. **Fail-safe consistency (F2):** Align `scheduleConfigMultiPerson` with the API handler's fail-closed pattern. Return an error on JSON parse failure so the scheduler does not silently create tasks from corrupt config.

3. **Dead code cleanup (F3):** Consider removing the `isScheduled` parameter from `saveLatestResultAndCompleteTask` or adding a comment that the pruning is a no-op under 1->N (kept for backward compatibility with any in-flight 1->1 tasks).

4. **Delete consistency (F4):** Align `DeleteSchedule` task soft-delete with `DeleteSummary` — set `status = -1` alongside `deleted_at` for consistency.

## 4. Additional Findings

- **Stale error vars:** `errTaskScopeScheduleBound` and `errLiveBindingDuplicate` are still declared (`schedule.go:72-76`) and handled in error switches, but the code paths that produced them have been removed. Dead code — harmless but should be cleaned up.
- **`scanStuckTasks` + overlap guard interaction:** When a stuck Processing task is reset to Pending, the overlap guard no longer blocks, so the scheduler creates a NEW task while the old one is retried. Two tasks run concurrently for the same schedule. No data corruption (independent task_ids), but duplicate LLM cost. Acceptable for this version.

## 5. Data-Flow Trace

| Consumed Data | Upstream Source | Verified? |
|---|---|---|
| `lockedSched.SpaceID/CreatorID/Title/SummaryMode` | FOR UPDATE re-read of `summary_schedule` | ✅ flows to `newTask` |
| `lockedSched.ID` (as `ScheduleID`) | FOR UPDATE re-read | ✅ set on `newTask.ScheduleID` |
| `latest.OriginChannelID/Type` | Latest task by `id DESC` under `schedule_id` | ✅ copied to `newTask` when `hasLatest` |
| `start/end` (time range) | `service.ComputeTimeRange(lockedSched.TimeRangeType, ...)` | ✅ set on `newTask` |
| `scheduleConfigMultiPerson` input | `lockedSched.ParticipantConfig` (FOR UPDATE re-read) | ✅ but JSON error silently swallowed |
| `buildScheduledTaskChildren` input | `lockedSched.SourceConfig` + `lockedSched.ParticipantConfig` | ✅ creates source/participant/personal_result |
| `ListSummaries` folding | `COALESCE(schedule_id, id)` in ROW_NUMBER | ✅ correct for both scheduled and manual |
| `DeleteSummary` group delete scope | `schedule_id = lockedSched.ID AND deleted_at IS NULL` | ✅ batch UPDATE, not per-row |

## 6. R5 Blind-Spot Checklist (security_sensitive)

**C1 — Dual-path parity:**
- **create ↔ delete:** CreateSchedule no longer enforces 1:1 binding (removed `boundCount` check). DeleteSchedule now does group soft-delete. The create path allows multiple tasks per schedule; the delete path cleans up all of them. **Symmetric. CLEAR.**
- **bind ↔ unbind:** `DeleteSummary` non-creator path unbinds (`schedule_id = nil`) + pauses schedule. `CreateSchedule` rebind path (`scope=task`) sets `schedule_id` on the task. No asymmetry found. **CLEAR.**

**C2 — Control-flow ordering / nested reuse:**
- `buildScheduledTaskChildren` is called once per new task creation, inside the claim transaction. No nested/double-call risk. **CLEAR.**
- `scheduleConfigMultiPerson` is called once per claim cycle. No reuse at different call-sites. **CLEAR.**

**C3 — Authorization boundary ≠ capability boundary:**
- Group delete in `DeleteSummary` is gated by `lockedSched.CreatorID == userID` (schedule creator). Non-creators can only delete their own task. **Authorization correctly scoped. CLEAR.**
- `loadBoundTaskForScheduleUpdate` checks `task.CreatorID != userID` → 403. Schedule update/toggle gated by creator. **CLEAR.**

---

**[Octo-Q] verdict: REQUEST_CHANGES**

**Reason:** P1 finding F1 — migration `20260610-01` does not drop `uk_live_schedule_binding` unique key from the deleted `20260608-02`. If the old migration was applied to any environment, the core 1->N INSERT will fail with MySQL 1062 on the second scheduled run, making the feature non-functional. Must add defensive DROP statements to the new migration.

Jerry-Xin (changes requested):
The PR is relevant to this repository, but several blocking issues prevent the one-to-many schedule model from working safely.

**🔴 Blocking**
- 🔴 Critical: Existing databases that already ran `20260608-02-task-schedule-binding.sql` will fail or keep the old one-to-one constraint. The new migration blindly adds `uk_summary_participant_task_user` again and never drops the existing `live_schedule_id` / `uk_live_schedule_binding`, so startup can fail on duplicate index creation, or the second live task for a schedule will still violate the old unique binding. Fix `migrations/sql/20260610-01-schedule-one-to-many.sql:3` and `migrations/sql/20260610-01-schedule-one-to-many.sql:7` to be a true forward migration from the previously released schema.

- 🔴 Critical: Non-creators can still mutate another user’s schedule through `DELETE /summaries/:id`. The PR description says scheduled deletion is creator-only and otherwise returns 403, but this branch allows a participant to unbind their task and set the schedule inactive. That is a permission bypass and a denial-of-service path against the schedule owner. See `internal/api/handler/task.go:961` through `internal/api/handler/task.go:1006`.

- 🔴 Critical: Expired `Processing` overlap handling is not implemented. `claimAndCreateScheduledTask` skips every `Processing` latest task without checking `processing_deadline`, while the PR contract says expired processing runs should be marked failed and a fresh task created. Combined with `scanStuckTasks`, expired runs can be reset back to `Pending`, after which the scheduler may create additional overlapping tasks instead of resolving the old run deterministically. See `internal/worker/scheduler.go:145` through `internal/worker/scheduler.go:152` and `internal/worker/scheduler.go:292` through `internal/worker/scheduler.go:314`.

- 🔴 Critical: `ListSummaries` can hide unrelated manual tasks. The fold key uses `COALESCE(schedule_id, id)`, which mixes two independent ID domains. A manual task with `id = 42` and scheduled tasks with `schedule_id = 42` end up in the same partition, so one visible summary disappears. Use a partition key that separates manual and scheduled rows, for example `PARTITION BY CASE WHEN schedule_id IS NULL THEN id ELSE schedule_id END, schedule_id IS NULL`, or equivalent. See `internal/api/handler/task.go:406` through `internal/api/handler/task.go:411`.

**💬 Non-blocking**
- 🟡 Warning: The raw list queries ignore `Scan` errors, so SQL/runtime errors silently return empty or partial responses. Check and return errors for both count and page queries. See `internal/api/handler/task.go:413` through `internal/api/handler/task.go:421`.

- 🟡 Warning: The new behavior needs tests for upgrade migration from the old binding schema, non-creator scheduled deletion returning 403, expired processing overlap recovery, and the `COALESCE(schedule_id, id)` collision case.

**✅ Highlights**
- The scheduler now creates fresh task rows for scheduled runs and rebuilds child rows from schedule config, which aligns with the intended run-history model.
- The list query correctly whitelists sortable columns before interpolating `ORDER BY`, avoiding an obvious SQL injection path.

yujiawei (changes requested):
# Code Review — PR #86 (octo-smart-summary)

Converting `schedule ↔ summary_task` from 1:1 to 1:N is the right model and the scheduler/handler rewrite is mostly clean and well-commented. `go build ./...` and `go test ./internal/worker/...` both pass on a **fresh** database. However, the migration is **not idempotent against any database that already applied the now-deleted `20260608-02`** (i.e. every environment deployed off `main`, where `20260608-02` shipped in #77). That is a hard deploy blocker, so this needs changes before merge.

## P0 — Migration is broken on already-deployed databases

This PR **deletes the file** `migrations/sql/20260608-02-task-schedule-binding.sql`, but the migration runner is **Up-only** — `migrate.Exec(db, dialect, source, migrate.Up)` at `internal/db/migrate.go:52`. Deleting a migration file does **not** run its `Down`; rubenv/sql-migrate simply stops tracking that ID. So on any DB where `20260608-02` already ran, all three objects it created still physically exist:
- generated column `summary_task.live_schedule_id`
- `UNIQUE KEY uk_live_schedule_binding`
- `UNIQUE KEY uk_summary_participant_task_user`

The new `migrations/sql/20260610-01-schedule-one-to-many.sql` neither drops the first two nor guards the third. Two distinct failures result:

### P0-a — Boot failure: duplicate key name (MySQL 1061)
`20260610-01` line 3-4 runs:
```sql
ALTER TABLE `summary_participant` ADD UNIQUE KEY `uk_summary_participant_task_user` (`task_id`, `user_id`);
```
That key was already created by `20260608-02`. `ADD UNIQUE KEY` cannot take `IF NOT EXISTS`, so on an existing DB this throws **MySQL 1061 "Duplicate key name"** → `migrate.Exec` returns an error → `RunMigrations` fails → **the service aborts on startup**. The feature never even runs.

### P0-b — Runtime: live-binding unique key blocks the 2nd INSERT (MySQL 1062)
Even past the migration, `uk_live_schedule_binding` (a UNIQUE over the `live_schedule_id` generated column, which equals `schedule_id` whenever `deleted_at IS NULL`) still enforces **at most one live task per schedule**. But the new scheduler INSERTs a brand-new live task each run without soft-deleting the predecessor:
```go
// internal/worker/scheduler.go:177-198
newTask = model.SummaryTask{ ... Status: model.StatusPending, ScheduleID: &lockedSched.ID, ... }
...
if err := tx.Create(&newTask).Error; err != nil { return err }
```
The prior task (read at `scheduler.go:132-143`) is only inspected for the overlap guard, never soft-deleted. For any terminal prior state (Completed/Failed) the code falls through to the INSERT, so two live rows share the same `schedule_id` → both compute `live_schedule_id = schedule_id` → **MySQL 1062** on the 2nd run. The error returns from the tx closure, rolling back the `next_run_at` advance (`scheduler.go:119-125`); the caller just logs and continues (`scheduler.go:47-49`), so `next_run_at <= now` stays true and the schedule re-scans forever, never producing a second summary.

**Fix:** make `20260610-01` idempotent against the deleted migration. Add (guarded against fresh DBs where these don't exist):
```sql
ALTER TABLE `summary_task` DROP INDEX `uk_live_schedule_binding`;
ALTER TABLE `summary_task` DROP COLUMN `live_schedule_id`;
-- only ADD uk_summary_participant_task_user if it does not already exist
```
Because MySQL DDL has no portable `DROP ... IF EXISTS` for indexes/columns pre-8.0, either split into a dedicated cleanup migration that you confirm runs against prod, or check `information_schema` first. This is the same deploy-safety class that prior schedule migrations in this repo had to handle explicitly.

## P1 — `DeleteSchedule` soft-delete is not honored by the worker (ghost runs, wasted LLM cost)

`DeleteSchedule` now soft-deletes bound tasks by setting **only** `deleted_at`, leaving `status` untouched:
```go
// internal/api/handler/schedule.go:1132-1134
return tx.Model(&model.SummaryTask{}).Where("id IN ?", taskIDs).Update("deleted_at", &now).Error
```
But `SummaryTask.DeletedAt` is a plain `*time.Time`, **not** `gorm.DeletedAt` (`internal/model/model.go:120`), so GORM applies no automatic soft-delete scope — every query must add `deleted_at IS NULL` by hand. The worker claim query does **not**:
```go
// internal/worker/processor.go:103-105 — no deleted_at filter
p.db.Where("status = ? AND retry_count < ? AND (processing_deadline IS NULL OR processing_deadline < ?)",
    model.StatusPending, p.cfg.WorkerMaxRetry, now)...
```
A scheduled task is born `StatusPending` (`scheduler.go:185`). If a schedule is deleted while a freshly-created run is still Pending, the worker still claims and fully processes it (LLM call, results written) — the deletion is silently ignored for that in-flight cycle, and the result is hidden from the list (which *does* filter `deleted_at IS NULL`), so the work is pure waste. `scanStuckTasks` (`scheduler.go:293-308`) similarly lacks the `deleted_at` filter and will resurrect a soft-deleted Processing task to Pending. Note `scanConfirmTimeouts` at `scheduler.go:270` *does* include `deleted_at IS NULL`, confirming the filter is expected.

This is a regression: before this PR, `DeleteSchedule` unbound (`schedule_id = nil`) rather than soft-deleting, so this state could not arise.

**Fix:** mirror the sibling `DeleteSummary` group-delete path (`task.go:982-988`), which sets **both** `status` and `deleted_at`, OR add `deleted_at IS NULL` to the worker queries at `processor.go:103` and `scheduler.go:294/307`, OR convert `DeletedAt` to `gorm.DeletedAt`. Relying on `status` alone is what `DeleteSchedule` currently fails to do.

## P1 — Overlap guard only covers `Processing`, not `Pending` → duplicate concurrent runs

With `uk_live_schedule_binding` intended to be gone (fresh DB), the only same-schedule dedup is the in-process guard:
```go
// internal/worker/scheduler.go:145
if hasLatest && latest.Status == model.StatusProcessing { ... return nil }
```
It skips only when the latest task is `Processing`. But new tasks are inserted as `Pending` and the scheduler does **not** trigger the worker for them — they wait for the poll loop (`processor.go:100`, 10/interval, oldest-first). For a short-interval schedule or a backed-up worker, the previous run can still be `Pending` when the next cycle fires → a 2nd Pending task is inserted for the same schedule → two full summaries for overlapping windows, doubled cost. `Regenerate` (`task.go:760-768`) and `scanStuckTasks` (`scheduler.go:297`) both deterministically push the latest task back to `Pending`, widening the window. The deleted unique key previously made this structurally impossible.

**Fix:** skip when the latest task is in **any** non-terminal state, e.g. `latest.Status NOT IN (Completed, Failed, Cancelled, -1)`, rather than only `Processing`. Optionally trigger the worker on insert to shrink the Pending window.

## Human-verify (security-sensitive) — MySQL version for `ROW_NUMBER()`

`ListSummaries` now folds with a window function:
```go
// internal/api/handler/task.go:411
innerSQL := "SELECT *, ROW_NUMBER() OVER (PARTITION BY COALESCE(schedule_id, id) ORDER BY id DESC) AS rn FROM summary_task WHERE " + whereSQL
```
`ROW_NUMBER()` requires **MySQL 8.0+**. The deleted migration relied on STORED generated columns (5.7+), so the prior floor was 5.7. There is no deployment manifest in this repo to confirm the target version — please verify production runs 8.0+, otherwise the summary list will error. (The raw SQL itself is safe: all WHERE fragments use bound `?` placeholders, `orderClause` is whitelisted to `{created_at,updated_at}`×`{asc,desc}` at `task.go:398-404`, and count/page args are correctly ordered.)

## P2 / nits (non-blocking)

- **Dead/misleading code:** on a fresh DB the `uk_live_schedule_binding` constraint never exists, so the `isMySQLDuplicateKey → errLiveBindingDuplicate → 409` handling in `CreateSchedule`/`UpdateSchedule` (`schedule.go:506-507, 897-898`) and the comment at `schedule.go:72` ("MySQL 1062 on uk_live_schedule_binding mapped to a clean 409") are dead and reference a constraint that no longer exists. Remove or update once the migration is fixed.
- **Magic number:** `DeleteSummary` group-delete writes `status: -1` (`task.go:984, 1011`), which is not a defined constant (`StatusPending=0…StatusCancelled=5`). Introduce a `StatusDeleted` constant or reuse `StatusCancelled` for clarity, and make both delete paths consistent.
- **History never reclaimed:** soft-deleted tasks leave subtable rows (`result/chunk/participant/personal_result`) intact with no purge. Intentional per the comments and not a correctness leak (reads are all gated by `deleted_at IS NULL` on the parent), but worth a retention follow-up.

## Verdict

The 1:N direction and most of the implementation are sound, but the migration will fail at boot on every already-deployed environment (P0), and two delete/overlap correctness gaps were introduced. Requesting changes — primarily the idempotent migration fix and the `DeleteSchedule` status/worker-filter gap.

Ranwanglc:
补充说明一下当前迁移处理的前提，避免我们对部署环境状态的理解不一致：目标部署环境此前从未成功执行过旧迁移 `20260608-02-task-schedule-binding.sql`，因此该迁移里创建的 `uk_live_schedule_binding` 唯一键以及 `live_schedule_id` 生成列，在目标环境中实际上都不存在。

基于这个前提，本 PR 当前的处理方式是符合目标环境实际情况的：移除这条从未成功生效的错误迁移 `20260608-02`；新的迁移 `20260610-01` 不包含任何 DROP 操作（因为目标环境里没有对应约束/列可删），只新增 participant 维度的唯一键，以及折叠查询所需的索引。

唯一需要额外说明的是本地环境差异：如果你本地此前已经成功执行过 `20260608-02`，已经建出了 `uk_live_schedule_binding` 唯一键和 `live_schedule_id` 生成列，那么本地在执行本 PR 迁移前，需要先手动将这两个对象 DROP 掉，再继续迁移；如果本地从未建过这两个对象，则无需做任何额外处理。

这里一并说明，主要是想把目标部署环境和本地历史环境区分清楚，避免我们基于不同前提讨论。

yujiawei:
Follow-up to my review above — one additional non-blocking item (verdict unchanged: changes requested, driven by the migration issue).

**P2 — `scheduleConfigMultiPerson` fails *open* on a JSON parse error, contradicting the API's fail-*closed* path.**

`internal/worker/scheduler.go:247-249`:
```go
if err := json.Unmarshal(sched.ParticipantConfig, &participants); err != nil {
    return false, nil  // corrupt config => treated as single-person => proceeds
}
```
The API equivalent `storedParticipantConfigSubsetOfCreator` (`internal/api/handler/schedule.go:145-150`) does the opposite — it returns `false` (unsafe → reject) on a parse error, and its comment even states *"This mirrors the worker, which would also fail to deserialize and skip the cycle."* That claim is inaccurate: the worker does **not** skip; it swallows the error and creates a task as if single-person. If `participant_config` ever became corrupt (manual DB edit, partial migration), the API would refuse to bind it but the scheduler would still run it. Low probability (config is validated at write time) but the divergent fail-safe direction is worth aligning — return the error (fail-closed) and fix the misleading comment.

Ranwanglc:
## 复核结论与修复说明（commit a0fa0b0）

针对此前提出的三个问题，逐条处理如下：

### 问题1：删除接口权限绕过 —— 已修复（方案 A）
原 `DeleteSummary` 在删除定时类 task 时，若调用者是参与者而非 schedule 创建者，会执行「解绑该 task（schedule_id=NULL）+ 将他人 schedule 置 is_active=0 暂停」，这相当于参与者可以静默暂停创建者的定时总结，属越权。

已改为方案 A：参与者（非 schedule 创建者）删除定时类总结**直接返回 403**（BizError code 40006，事务回滚），不再解绑该 task、不再暂停他人 schedule、不删除任何数据。整组删除（软删 schedule 下全部 task + 停 schedule）仍**仅限创建者**；手动类 task（schedule_id 为 NULL）删除逻辑保持不变。

### 问题2：过期任务未重置为失败 —— 本 PR 不改（非漏洞）
经核查为职责分离，不是漏洞，故本 PR 不动 `scheduler.go`：
- 调度器的重叠保护（scheduler.go ~145 行）在上一条 task 仍处于 Processing 时**跳过新建**，避免并发堆叠；
- 「过期 / 卡死的 Processing 任务」的重置或标记失败由独立的 `scanStuckTasks`（~287 行）负责：对 `processing_deadline IS NULL OR < now` 的任务，可重试的重置、超过 maxRetry 的标记 Failed，并兼容 `processing_deadline` 为 NULL 的历史老数据。

两者各司其职，重叠保护不负责回收过期任务，因此不存在「过期任务无人重置」的漏洞。

### 问题3：折叠键 COALESCE 混淆手动与定时任务 —— 已确认为真问题并修复
`ListSummaries` 原折叠键为 `ROW_NUMBER() OVER (PARTITION BY COALESCE(schedule_id, id) ...)`。`summary_schedule.id` 与 `summary_task.id` 是两套**独立自增主键**且值域重叠（实测当前库 schedule_id 落在 37–48，手动 task.id 含 38–43，已发生碰撞），COALESCE 会把数值相同的「手动 task id=N」与「定时 schedule_id=N」错误折叠进同一分组，导致列表丢行。

已改为带类型前缀的分组键，彻底隔离两个命名空间：
```sql
PARTITION BY (CASE WHEN schedule_id IS NULL THEN CONCAT('t', id) ELSE CONCAT('s', schedule_id) END)
```
ORDER BY id DESC、分页与 total 计数逻辑均保持不变（count 与分页复用同一 innerSQL，已同步生效）。

---
go build ./... 通过；go test ./internal/...（含 worker、api/handler）全部通过。

Jerry-Xin (changes requested):
This PR is in scope, but it has blocking migration and scheduler correctness issues that can break existing deployments and the one-to-many execution model.

**🔴 Blocking**
- 🔴 Critical: The migration is not compatible with databases that already applied the deleted `20260608-02-task-schedule-binding.sql`. [migrations/sql/20260610-01-schedule-one-to-many.sql](/Users/mlamp/.openclaw/tmp/pr-review-cd84uphi/repo/migrations/sql/20260610-01-schedule-one-to-many.sql:3) tries to add `uk_summary_participant_task_user` again, so those deployments will fail with a duplicate-key-name error. It also never drops the old generated `live_schedule_id` column / `uk_live_schedule_binding` unique key, so even if the duplicate index issue were bypassed, the database would still reject multiple live tasks for the same schedule. Add a forward migration that conditionally or explicitly removes the old live-binding constraint/column and handles the participant unique key safely.

- 🔴 Critical: The overlap guard does not implement the stated expired-processing behavior and can create duplicate runnable tasks for the same schedule window. [internal/worker/scheduler.go](/Users/mlamp/.openclaw/tmp/pr-review-cd84uphi/repo/internal/worker/scheduler.go:145) skips every latest `Processing` task without checking `processing_deadline`. Separately, `scanStuckTasks` resets expired processing tasks back to `Pending` while retries remain at [internal/worker/scheduler.go](/Users/mlamp/.openclaw/tmp/pr-review-cd84uphi/repo/internal/worker/scheduler.go:293). Once that happens, the next schedule claim sees latest status `Pending` and inserts a new scheduled task, while the old pending task is still eligible for workers. This violates the PR’s “expired Processing is marked Failed then new task is created” contract and can produce duplicate summaries for the same preserved `last_run_at` window.

**💬 Non-blocking**
- 🟡 Warning: [internal/api/handler/schedule.go](/Users/mlamp/.openclaw/tmp/pr-review-cd84uphi/repo/internal/api/handler/schedule.go:1122) loads all bound tasks into memory just to build an `IN` list before batch-updating them. Since the comment explicitly expects long-lived schedules with many tasks, update by `schedule_id` directly to avoid large result sets and huge `IN` clauses.

**✅ Highlights**
- The list folding correctly namespaces manual task IDs and schedule IDs before partitioning, avoiding `COALESCE(schedule_id, id)` collisions.
- The creator-only delete path for scheduled groups is a good permission hardening.

Sanity checks run: `go test ./internal/worker/...` and `go test ./internal/api/handler/...` both passed.

Ranwanglc:
已修复 reviewer 发现的 2 个 bug（commit f7a6fa8）：

**Bug1 — DeleteSchedule 软删 task 状态不对齐 + worker 未过滤软删**
- DeleteSchedule 整组软删 task 时现一并置 `status=-1`，与 DeleteSummary（task.go 整组删）的写法对齐，消除「同 PR 内一处带 status:-1、一处不带」的自相矛盾。
- worker 候选查询补 `AND deleted_at IS NULL`。因 `summary_task.deleted_at` 是 `*time.Time` 而非 `gorm.DeletedAt`，GORM 不会自动过滤软删行；此前若 DeleteSchedule 软删的 task 状态仍为 Pending，会被 worker 候选查询捡起执行。补上过滤从根上兜底。

**Bug2 — 重叠保护未覆盖过期复活任务**
- 重叠保护由「仅当最新一条 latest 为 Processing 才跳过新建」改为「该 schedule 下存在任何非终态(Pending/WaitingConfirm/Processing)且未软删 task 即跳过新建」（事务内 COUNT 判断）。
- 修复点：scanStuckTasks 会把过期 Processing 复活成 Pending，此时 latest 不是 Processing，旧判据失效 → 同 schedule 同窗口新建第二条可执行任务 → 重复执行竞态（无 DB 唯一键）。把所有非终态都视为占位即可堵住。
- 跳过时不推进 last_run_at，与原 transient overlap 语义一致，窗口由下个周期覆盖。任务达 WorkerMaxRetry 被 scanStuckTasks 标 Failed(终态)后自然解除跳过，不会永久阻塞。
- latest 查询保留（新建 task 仍需从其继承 origin channel）。

验证：`go build ./...` 通过；`go test ./internal/...` 全绿（含 internal/worker、internal/api/handler）。

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# octo-smart-summary PR#86 Review Report
**PR:** Mininglamp-OSS/octo-smart-summary #86
**Head SHA:** a0fa0b0d42337bd37e08061457a4a234611a1141
**Title:** feat(schedule): schedule one-to-many summary_task
**Files changed:** 7 (+344 / -200)

---

## 1. Verification Summary

| Area | Status | Evidence |
|------|--------|----------|
| 1->N scheduler INSERT model | ⚠️ Blocked by migration gap | `scheduler.go:195` tx.Create(&newTask); `20260610-01` migration missing DROP |
| Overlap guard (Processing skip) | ✅ Correct | `scheduler.go:155-160` hasLatest && latest.Status check |
| Multi-person guard | ✅ Correct (minor: silent JSON error swallow) | `scheduler.go:237-258` scheduleConfigMultiPerson |
| DeleteSummary 403 fix (commit 2) | ✅ Correct | `task.go:999-1003` non-creator → BizError 40006 |
| DeleteSchedule group soft-delete | ⚠️ status not set to -1 | `schedule.go:1133` only sets deleted_at |
| ListSummaries folding + namespace | ✅ Correct (commit 2 fixes collision) | `task.go:420-425` CONCAT('t'/'s', ...) |
| buildScheduledTaskChildren | ✅ Correct data flow | `scheduled_replace_helpers.go:38-145` |
| Migration completeness | ❌ Missing DROP for live_schedule_id + uk_live_schedule_binding | Old migration deleted, no DROP in new migration |
| Test coverage | ✅ Key paths covered | scheduler_lastrun_test.go: first-run, overlap, multi-person |

---

## 2. Findings

### F1 — P1: Migration gap — live_schedule_id generated column + unique key not dropped

**Diff-scope:** NEW — PR deletes old migration file `20260608-02-task-schedule-binding.sql` without providing a DROP migration for the artifacts it created.

**Problem:** The old migration (merged to main in PR #77, commit `04a360e`) created:
1. `live_schedule_id` BIGINT GENERATED ALWAYS AS column on `summary_task`
2. `uk_live_schedule_binding` UNIQUE KEY on `live_schedule_id`

The PR deletes the old migration file but the new migration `20260610-01-schedule-one-to-many.sql` does NOT include:
```sql
ALTER TABLE `summary_task`
    DROP INDEX `uk_live_schedule_binding`,
    DROP COLUMN `live_schedule_id`;
```

**Data-flow trace:** On any database where the old migration was already applied:
- `live_schedule_id` = `schedule_id` WHEN `deleted_at IS NULL AND schedule_id IS NOT NULL` (generated column)
- `uk_live_schedule_binding` enforces at most ONE live task per schedule_id
- Scheduler `claimAndCreateScheduledTask` (`scheduler.go:195`) INSERTs a new task with `ScheduleID: &lockedSched.ID` and `deleted_at IS NULL`
- The generated column computes `live_schedule_id = schedule_id`
- If any live task already exists under that schedule → MySQL 1062 duplicate key → tx rolls back → scheduler logs error → schedule never produces a second run

**Impact:** The core 1->N feature is completely broken on any deployed database. The scheduler would fail silently (logged error, schedule keeps retrying every 60s but always hitting the unique key).

**Fix:** Add to `20260610-01-schedule-one-to-many.sql` Up section:
```sql
ALTER TABLE `summary_task`
    DROP INDEX `uk_live_schedule_binding`,
    DROP COLUMN `live_schedule_id`;
```
And corresponding re-add in the Down section.

---

### F2 — P2: DeleteSchedule group soft-delete doesn't set status=-1

**Diff-scope:** NEW — PR changes DeleteSchedule from unbind (`schedule_id = nil`) to soft-delete (`deleted_at = &now`), but only sets `deleted_at`, not `status`.

**Problem:** `DeleteSchedule` (`schedule.go:1133`):
```go
return tx.Model(&model.SummaryTask{}).
    Where("id IN ?", taskIDs).
    Update("deleted_at", &now).Error
```

Compare with `DeleteSummary` (`task.go:990-993`):
```go
Updates(map[string]interface{}{
    "status":     -1,
    "deleted_at": now,
})
```

**Data-flow trace:** A task in `StatusProcessing` (status=2) that gets group-deleted via DeleteSchedule retains `status=2` with `deleted_at` set. `scanStuckTasks` (`scheduler.go:295`) queries:
```go
Where("status = ? AND (processing_deadline IS NULL OR processing_deadline < ?) AND retry_count < ?",
    model.StatusProcessing, now, maxRetry-1)
```
This query does NOT filter `deleted_at IS NULL`. The zombie task would be picked up and reset to Pending (incrementing retry), or marked Failed. The schedule is already soft-deleted so no new runs spawn, but the zombie task gets an unnecessary status transition.

**Impact:** Cosmetic — soft-deleted tasks get spurious status updates from the stuck-task scanner. Not user-visible (deleted tasks don't appear in ListSummaries which filters `deleted_at IS NULL`), but creates noise in logs and DB state.

**Fix:** In DeleteSchedule, change to:
```go
Updates(map[string]interface{}{
    "status":     -1,
    "deleted_at": &now,
})
```

---

### F3 — P2: scheduleConfigMultiPerson silently swallows JSON parse errors

**Diff-scope:** NEW — function introduced by this PR.

**Problem:** `scheduleConfigMultiPerson` (`scheduler.go:240`):
```go
if err := json.Unmarshal(sched.ParticipantConfig, &participants); err != nil {
    return false, nil  // silently treats malformed JSON as "not multi-person"
}
```

**Data-flow trace:** If `ParticipantConfig` contains invalid JSON:
1. `scheduleConfigMultiPerson` returns `false, nil` → multi-person guard bypassed
2. Code proceeds to create new task (`scheduler.go:195`)
3. `buildScheduledTaskChildren` → `buildScheduledTaskParticipants` (`scheduled_replace_helpers.go:91`) tries the same `json.Unmarshal` and returns `BizError(40000, "定时参与者配置无效")`
4. Transaction rolls back — no data corruption

**Impact:** No data corruption (tx rolls back at step 3). But the multi-person guard is silently bypassed for malformed JSON — the error surfaces as "invalid participant config" rather than the more accurate multi-person guard failure. In a scenario where `buildScheduledTaskParticipants` is later modified to be more lenient, this silent swallow could become a real gap.

**Fix:** Return the error instead of swallowing:
```go
if err := json.Unmarshal(sched.ParticipantConfig, &participants); err != nil {
    return false, fmt.Errorf("participant_config JSON: %w", err)
}
```

---

### F4 — P2: Dead error-mapping paths after 1->N conversion

**Diff-scope:** AMPLIFIED — PR removes the bound-count checks that produced `errTaskScopeScheduleBound`, and the `uk_live_schedule_binding` unique key that triggered `errLiveBindingDuplicate`, but leaves declarations and error-mapping cases.

**Evidence:**
- `schedule.go:69` — `errTaskScopeScheduleBound` declared but never returned (bound-count checks removed at lines ~448 and ~894)
- `schedule.go:73` — `errLiveBindingDuplicate` still returned at lines 507 and 898 via `isMySQLDuplicateKey(err)` check on `Update("schedule_id", ...)`. On a fresh DB (no `uk_live_schedule_binding`), this never triggers. On an existing DB (where old migration was applied), this masks the P1 migration gap (F1) behind a 409 "already bound" error.
- `schedule.go:523-526, 1016-1019` — error mapping cases for both variables

**Impact:** Dead code on fresh databases. On existing databases, the `errLiveBindingDuplicate` path could misleadingly return 409 "该定时已绑定其它总结" when the real issue is the leftover unique key from the old migration.

**Fix:** After fixing F1 (adding DROP migration), remove `errTaskScopeScheduleBound`, `errLiveBindingDuplicate`, and their mapping cases. Also remove the `isMySQLDuplicateKey` checks at lines 507 and 898.

---

### F5 — Nit: ListSummaries filter semantics — filters apply inside the fold

**Diff-scope:** NEW — new query structure.

The PR moves all business filters (status, trigger_type, keyword, etc.) INSIDE the window subquery. "Latest per schedule" is computed over the already-filtered set. This means a `status=completed` filter can surface an older completed task as the "representative" for a schedule group whose actual latest task is processing. This is arguably correct (showing the latest matching task per schedule), but is a semantic change from the old flat-list behavior. Not blocking.

---

## 3. Suggestions

1. **F1 fix (mandatory):** Add DROP statements for `live_schedule_id` column and `uk_live_schedule_binding` unique key to the new migration. Consider an `IF EXISTS` guard or a pre-check for environments where the old migration may or may not have been applied.

2. **F2 fix:** Align DeleteSchedule's group soft-delete with DeleteSummary by also setting `status = -1`.

3. **F3 fix:** Propagate JSON parse errors from `scheduleConfigMultiPerson` instead of silently swallowing.

4. **F4 cleanup:** Remove dead error variables and mapping cases after F1 is resolved.

5. **Consider:** Add `deleted_at IS NULL` filter to `scanStuckTasks` to prevent zombie task recovery on soft-deleted tasks (pre-existing gap, amplified by this PR).

---

## 4. Additional Findings

- **scanStuckTasks lacks deleted_at filter** (pre-existing): `scheduler.go:295` queries `status = StatusProcessing` without `deleted_at IS NULL`. Any soft-deleted processing task (from any delete path) could be picked up. This PR amplifies the impact by introducing group soft-delete that doesn't change status (F2).

- **errMultiPersonNotSupported still actively used** at lines 160, 165, 428, 1070, 1230 — these are in the API schedule create/update/toggle paths (not the scheduler worker). These are NOT dead code and correctly guard multi-person at the API level. No issue here.

---

## 5. Data-Flow Trace

|---|---|---|
| `newTask.ScheduleID` | `&lockedSched.ID` (scheduler.go:192) | ✅ Set from FOR UPDATE re-read |
| `newTask.OriginChannelID/Type` | `latest.OriginChannelID/Type` (scheduler.go:199-200) | ✅ Only when hasLatest=true |
| `newTask.CreatorID` | `lockedSched.CreatorID` (scheduler.go:188) | ✅ Schedule is source of truth |
| `buildScheduledTaskChildren(task.ID)` | `newTask.ID` after `tx.Create` (scheduler.go:195) | ✅ GORM auto-populates ID on INSERT |
| `scheduleConfigMultiPerson(lockedSched)` | `lockedSched.ParticipantConfig` (FOR UPDATE re-read) | ✅ Fresh from locked row |
| ListSummaries `schedule_id` output | `t.ScheduleID` from raw SQL Scan | ✅ GORM ignores extra `rn` column |
| DeleteSummary group delete scope | `WHERE schedule_id = ? AND deleted_at IS NULL` (task.go:990) | ✅ Catches all live tasks in group |
| `loadBoundTaskForScheduleUpdate` representative | `ORDER BY id DESC LIMIT 1` (schedule.go:280) | ✅ Latest task as representative |

---

## 6. Blind-Spot Checklist (R5 — security_sensitive)

**C1 — Dual-path parity:**
- CreateSchedule ↔ UpdateSchedule: Both remove bound-count checks symmetrically. ✅ Clear.
- DeleteSchedule ↔ DeleteSummary: Both do group soft-delete for creator. ⚠️ **Hit** — DeleteSchedule doesn't set status=-1 (F2). Parity gap between the two delete paths.
- subscribe ↔ unsubscribe (schedule binding): The PR removes the 1:1 binding enforcement. The create path (CreateSchedule) and rebind path (UpdateSchedule) both allow multiple tasks per schedule. Symmetric. ✅ Clear.

**C2 — Control-flow ordering / nested reuse:**
- `buildScheduledTaskChildren` is called once per new task creation in the scheduler. No nested/double-call risk. ✅ Clear.
- `scheduleConfigMultiPerson` is called once before task creation. No ordering issue with `buildScheduledTaskChildren`. ✅ Clear.
- ListSummaries raw SQL: `orderClause` is validated (whitelist: created_at/updated_at + asc/desc). No injection via sort params. ✅ Clear.

**C3 — Authorization boundary ≠ capability boundary:**
- DeleteSummary 403 fix (commit a0fa0b0): Previously a participant (non-creator) could unbind a task and pause (`is_active=0`) someone else's schedule — privilege escalation. Now correctly returns 403. ✅ Fixed.
- DeleteSchedule: Creator-only gate (`lockedSched.CreatorID != userID` → 40004). Group delete is scoped to the creator's schedule. ✅ Correct.
- `authorizeTaskAccess` (task-level) is broader than schedule-creator check. The DeleteSummary function correctly layers: task access first, then schedule-creator check for group operations. ✅ Correct.

---

[Octo-Q] verdict: REQUEST_CHANGES — P1 migration gap (F1) blocks the core 1->N feature on any database where the old migration was applied. The scheduler INSERT would fail with MySQL 1062 on every second run of any schedule. Must add DROP statements for `live_schedule_id` and `uk_live_schedule_binding` before landing.

yujiawei (changes requested):
# Code Review — PR #86 (octo-smart-summary)

This PR converts the schedule↔summary_task relationship from 1:1 to 1:N (each cron tick INSERTs a brand-new task instead of resetting the bound one). The latest commit (`a0fa0b0`) correctly resolves two earlier issues — the list-folding key collision and a privilege-escalation in `DeleteSummary`. Thanks for those. However, a deploy-blocking migration issue and two runtime correctness issues remain.

Build, `go vet`, and `go test ./internal/...` all pass — but the unit suite runs on SQLite `:memory:` with `AutoMigrate`, so it never exercises the real MySQL migration files or MySQL UNIQUE constraints. The blockers below are invisible to the current tests by construction.

---

## P0 — Deleting an already-applied migration file breaks boot on any non-fresh DB

This PR deletes `migrations/sql/20260608-02-task-schedule-binding.sql`. That migration is **already on `main`** (it shipped via PR #77 / commit `04a360e`), so any database that has run `main` has a `gorp_migrations` row with id `20260608-02-task-schedule-binding.sql`.

The migration runner uses the default `rubenv/sql-migrate` settings:

- `internal/db/migrate.go:52` → `migrate.Exec(db, dialect, source, migrate.Up)` (package-level `migSet`, `IgnoreUnknown = false`).

With `IgnoreUnknown = false`, `sql-migrate`'s `planMigrationCommon` checks every recorded migration against the embedded files and returns `newPlanError(..., "unknown migration in database")` when a recorded id has no matching file. That error propagates `Exec → RunMigrations → boot failure`, **before any SQL runs**. So on staging/prod/dev/CI databases that already applied `main`, the service will refuse to start.

A forward-only runner means **deleting a migration file is not the same as rolling it back**. The fix is to keep `20260608-02` in place and add a new forward migration `20260610-01` that explicitly `DROP`s what the 1:N model no longer wants — guarded for fresh DBs via `information_schema` — rather than deleting the original file.

Note this is independent of the team's "all target environments are fresh DBs" assumption: that is an out-of-band operational promise, not a code-level guarantee, and it does not hold for any environment that ran `main` (including CI that reuses a volume).

**Secondary consequence on existing DBs:** because `20260608-02` is not dropped, on an existing DB the `uk_live_schedule_binding` UNIQUE key (over the `live_schedule_id` generated column = at most one live task per schedule) *survives*. That directly contradicts the new 1:N scheduler, which INSERTs a second live task per schedule (`internal/worker/scheduler.go:197`) → MySQL 1062 → the INSERT's transaction rolls back (including the `next_run_at` advance), so the schedule re-scans every 60s and never produces a second summary. The new `20260610-01` also blindly re-adds `uk_summary_participant_task_user`, which already exists from `20260608-02` on an existing DB (1061 duplicate-key) — though the unknown-migration guard above aborts first.

## P1 — Soft-deleted (but still Pending) scheduled tasks are claimed and processed by the worker

`SummaryTask.DeletedAt` is a plain `*time.Time` (`internal/model/model.go:120`), **not** `gorm.DeletedAt`, so GORM applies no automatic `deleted_at IS NULL` scope. Every query must filter explicitly.

The worker claim query does not:

- `internal/worker/processor.go:103-105` selects candidates `WHERE status = ? AND retry_count < ? AND (processing_deadline IS NULL OR processing_deadline < ?)` with `StatusPending` — **no `deleted_at IS NULL`**.

`DeleteSchedule` now soft-deletes its bound tasks by setting **only `deleted_at`**, leaving `status` at Pending:

- `internal/api/handler/schedule.go:1131-1134` → `Update("deleted_at", &now)` (no status change).

So a task that was Pending when its schedule is deleted still matches the claim predicate (`status = Pending`) and is claimed + fully processed → wasted LLM cost and a "ghost" summary for a schedule the API reports as deleted. The pipeline never re-checks `deleted_at`.

This is a regression introduced by this PR: on `main`, `DeleteSchedule` *unbound* the task (`Update("schedule_id", nil)`, kept alive) rather than soft-deleting it. Note the `DeleteSummary` group/single paths (`task.go:986-991`, `task.go:1008`) set `status = -1` **and** `deleted_at`, so they are correctly excluded from a Pending claim — the gap is specific to `DeleteSchedule`. Other scan jobs (`scheduler.go:270` `scanConfirmTimeouts`) do filter `deleted_at IS NULL`, which makes the omission in `poll()` look like an oversight.

**Fix:** add `AND deleted_at IS NULL` to the worker claim query (and ideally convert the field to `gorm.DeletedAt`), or have `DeleteSchedule` also set a terminal `status` on the bound tasks the way `DeleteSummary` does.

## P1 — Overlap guard ignores the Pending state after the structural unique key was removed

The same-window dedup guard skips a new run only when the previous one is **Processing**:

- `internal/worker/scheduler.go:145` → `if hasLatest && latest.Status == model.StatusProcessing { ... skip }` (the comment at line 150 explicitly notes "no DB unique key").

New tasks are inserted as **Pending** (`scheduler.go:185`) and the scheduler does not trigger the worker directly (the worker is poll-based). Previously, `uk_live_schedule_binding` made >1 live task per schedule structurally impossible; this PR removes that key (on a fresh DB). So if the predecessor is still Pending (worker hasn't claimed it yet) when the schedule next becomes due, the guard does **not** skip → a second live Pending task is inserted → duplicate concurrent runs / duplicate summaries.

Reachability is bounded: the default poll interval is ~2s vs a ≥1-minute cron cadence and a 60s scan, so under normal load Pending→Processing happens well within the gap. The duplicate requires the worker to be down/backlogged for ~1 cron period. On an existing DB the surviving unique key still rejects the duplicate (with the side effects in P0). Still — dropping a hard DB invariant while leaving the app-level guard blind to Pending is a real weakness on fresh-DB deployments.

**Fix:** skip on *any* non-terminal status (Pending or Processing), not just Processing.

## P2 — Dead / misleading 1062→409 binding code

`errLiveBindingDuplicate`, `isMySQLDuplicateKey`, and the 1062→409 mapping remain in `internal/api/handler/schedule.go:72-73, 506-507, 526, 897-898, 1019`. They reference `uk_live_schedule_binding`, which on a fresh DB no longer exists (its creating migration is deleted and not recreated), so these branches can never fire there. On an existing DB the constraint persists and the code *can* fire — but as P0 notes, that constraint should not exist under the 1:N model at all. Once the migration is fixed correctly, please prune this stale handling (and the now-unused `errTaskScopeScheduleBound` returns).

---

## Confirmed-fixed (no action needed)

- **List folding key collision** — `task.go:416` now uses `PARTITION BY (CASE WHEN schedule_id IS NULL THEN CONCAT('t', id) ELSE CONCAT('s', schedule_id) END)`, correctly namespacing the two independent auto-increment id spaces. The raw SQL is injection-safe (bound `?`, whitelisted `orderClause`, correct count/page arg order). Minor nit: `CONCAT()` is MySQL-specific, so any future SQLite-backed test of this query path would diverge.
- **`DeleteSummary` privilege escalation** — the non-creator (participant) branch now returns `service.NewBizError(40006, ..., http.StatusForbidden)` and rolls back the transaction, instead of unbinding the task + pausing someone else's schedule. The `BizError` is matched via `errors.As` first in the outer handler (`task.go:1012-1017`) and rendered as 403, ahead of the NotFound/500 fallbacks. Authorization still runs before the transaction. Correctly fixed.

## Security-sensitive note for a human verifier

- Confirm the actual deploy/migration target. If **any** environment (prod, staging, CI with a reused volume) has run `main`, the P0 boot-failure and the surviving `uk_live_schedule_binding` constraint are live problems, not hypotheticals.
- `ListSummaries` raw SQL was re-checked for injection and is safe.

## Verdict

Requesting changes. The two earlier fixes are good, but the migration handling is a deploy blocker on any non-fresh DB, and the two P1 runtime issues should be addressed before merge.

lml2468 (approved):
## ✅ APPROVE — f7a6fa8（附迁移假设声明）

经多轮 review（e9faf0d → a0fa0b0 → f7a6fa8），所有阻塞项已解决或基于项目负责人对线上环境状态的确认撤销。`go build ./...` + 全套 `go test ./... -shuffle=on` 在 f7a6fa8 全过。

### 已修复并逐行核验
- **ListSummaries 折叠键碰撞**：`COALESCE(schedule_id, id)` 改为类型前缀 `CASE WHEN schedule_id IS NULL THEN CONCAT('t', id) ELSE CONCAT('s', schedule_id) END`，两个独立自增序列命名空间隔离，不再互相吞行。
- **删除接口提权**：非创建者参与者删定时 task 从「unbind + 暂停他人 schedule」收紧为直接 403 + 回滚（Option A），提权面消除。
- **软删 task 仍被 worker 执行**：双重修复 —— `DeleteSchedule` group soft-delete 补 `status = -1`（`schedule.go`），worker 候选查询补 `deleted_at IS NULL`（`processor.go`，根因防御）。
- **overlap 守卫漏过期复活任务**：从「只看 latest 是否 Processing」升级为「任何非终态 task（Pending/WaitingConfirm/Processing 且未软删）存在即 skip」（`scheduler.go`），堵住 scanStuckTasks 把 Processing 复活成 Pending 后产生第二条可执行 task 的竞态。
- **last_run_at**：仍只在成功新建 task 后推进，#77 R8 的修复未回退。

### ⚠️ 迁移假设（合并前必须满足，已与负责人确认）
本 PR 物理删除了 `20260608-02-task-schedule-binding.sql`（#77 已合并到 main），新迁移 `20260610-01` 不含对旧对象的清理。这**仅对从未应用过 `20260608-02` 的库安全**。负责人已确认线上环境未执行过该迁移，故当前最简迁移对线上正合适（对干净库加裸 `DROP` 反而会因对象不存在而阻塞）。

前提：
1. 任何应用过 `20260608-02` 的环境（如曾跑过 #77 集成测试的 staging/test）部署本 PR 前，需手动 `DROP INDEX uk_live_schedule_binding`、`DROP INDEX uk_summary_participant_task_user`、`DROP COLUMN live_schedule_id`，否则 `rubenv/sql-migrate` 会因 `unknown migration in database` 失败、且残留旧唯一键会挡住 1→N 的第二次 INSERT。
2. 建议在 PR 描述或迁移注释中注明此假设（沿用 #77 先例），让独立读迁移的人知道为何没有防御代码。

### 🟡 非阻塞（建议跟进，不挡合并）
- `scheduleConfigMultiPerson` 的 `json.Unmarshal` 失败时静默 `return false, nil`，会把坏配置当单人放行；建议至少 log 一行或保守 disable。
- `ListSummaries` 的 raw `h.db.Raw(...).Scan(...)` 未检查 `.Error`，出错会静默返回空列表；建议接错误返回 500。
- `schedule.go` 残留的 `uk_live_schedule_binding` 1062→409 错误映射（`errLiveBindingDuplicate` 等）在 1→N 下已是死代码，建议清理或更新注释。
- 缺回归测试：建议补「删 schedule 后 worker 不捡 Pending task」「Pending 占位时 claim skip 不新建」两个用例。

方向（1→N 对齐 Airflow DAGRun / K8s Job）正确，核心修复扎实。满足迁移假设后可合并。

Ranwanglc:
补充说明一下本次迁移脚本的适用前提，避免大家对执行路径有歧义。

本 PR 的线上目标环境，从未成功执行过已在本 PR 中删除的旧迁移 ，因此线上并不存在该旧迁移创建的对象，包括： 列、 唯一键，以及  唯一键。

基于这个事实，当前 PR 中的迁移脚本对线上环境是正合适的： 直接创建  及相关折叠索引即可，不会发生重复创建。如果在脚本里额外加入“兼容已应用旧迁移”的判断或 DROP 逻辑，反而会让线上在一个本来不存在这些对象的环境里多走一层分支，增加失败面，影响正常迁移执行。因此这里没有在 PR 中补这类兼容脚本。

需要单独说明的是本地环境的例外情况：如果你本地此前成功执行过 ，已经生成了上述旧对象，那么在此基础上再跑本 PR 的 ，确实会在  上遇到 。这种情况只需要在本地重新执行迁移前，先手动清理旧迁移残留对象即可：



清理后再执行本 PR 迁移即可正常通过；如果本地从未执行过 ，则无需做任何处理。

这里补充这些背景，主要是想把线上目标环境和本地已执行旧迁移环境区分开，方便大家对迁移脚本的适用范围形成一致理解。

Ranwanglc:
补充说明一下本次迁移脚本的适用前提，避免大家对执行路径有歧义。

本 PR 的线上目标环境，从未成功执行过已在本 PR 中删除的旧迁移 `20260608-02-task-schedule-binding.sql`，因此线上并不存在该旧迁移创建的对象，包括：`summary_task.live_schedule_id` 列、`uk_live_schedule_binding` 唯一键，以及 `uk_summary_participant_task_user` 唯一键。

基于这个事实，当前 PR 中的迁移脚本对线上环境是正合适的：`20260610-01` 直接创建 `uk_summary_participant_task_user` 及相关折叠索引即可，不会发生重复创建。如果在脚本里额外加入“兼容已应用旧迁移”的判断或 DROP 逻辑，反而会让线上在一个本来不存在这些对象的环境里多走一层分支，增加失败面，影响正常迁移执行。因此这里没有在 PR 中补这类兼容脚本。

需要单独说明的是本地环境的例外情况：如果你本地此前成功执行过 `20260608-02`，已经生成了上述旧对象，那么在此基础上再跑本 PR 的 `20260610-01`，确实会在 `uk_summary_participant_task_user` 上遇到 `Duplicate key name`。这种情况只需要在本地重新执行迁移前，先手动清理旧迁移残留对象即可：

```sql
DROP INDEX uk_live_schedule_binding ON summary_task;
DROP INDEX uk_summary_participant_task_user ON summary_participant;
ALTER TABLE summary_task DROP COLUMN live_schedule_id;
```

清理后再执行本 PR 迁移即可正常通过；如果本地从未执行过 `20260608-02`，则无需做任何处理。

这里补充这些背景，主要是想把线上目标环境和本地已执行旧迁移环境区分开，方便大家对迁移脚本的适用范围形成一致理解。

Jerry-Xin (approved):
Re-review at f7a6fa88bb49 — all previously-raised blocking items (B1–B3) have been resolved or explicitly accepted.

**Summary**

The one-to-many schedule→task refactor is solid. Prior blockers are addressed:
- ✅ B2 (ListSummaries key collision) — fixed via `CASE WHEN` namespace isolation in the window function
- ✅ B3 (DeleteSchedule soft-delete status) — now sets `status:-1` and worker queries filter `deleted_at IS NULL`
- ✅ Overlap guard upgraded to check ANY non-terminal predecessor task
- ✅ Permission bypass on delete restricted to 403 for non-creators
- ℹ️ B1 (migration compatibility) — accepted per author confirmation that production never applied the deleted `20260608-02` migration; no residual columns/indexes exist

**💬 Non-blocking suggestions (carried forward)**
- 🟡 `ListSummaries` raw `Scan` errors at `internal/api/handler/task.go:420,426` are silently swallowed — consider returning an API error rather than `200` with empty/partial data
- 🟡 The MySQL error-code 1062 mapping in `createSummaryTask` is currently unreachable given the new schema; consider removing dead code or adding a comment explaining future intent
- 🟡 Additional integration tests for the one-to-many happy path would strengthen confidence

**✅ Highlights**
- Scheduler task-creation flow cleanly supports one-to-many
- Non-creator delete path now correctly returns 403
- Existing tests pass (`go test ./internal/api/handler -run TestListSummaries`)

LGTM — approving.

lml2468 (approved):
## ✅ APPROVE — f7a6fa8 (independent re-review)

Independent review pinned to `f7a6fa8`. Cross-checked the four already-resolved items line by line and agree with the prior APPROVE. Adding one should-fix that earlier passes appear to have missed; not blocking.

### Verified fixes (line-by-line)

- **B3 soft-delete enforcement (double defense)** — `internal/api/handler/schedule.go:1132-1135` group soft-delete sets both `status=-1` and `deleted_at`; `internal/worker/processor.go:103` candidate query adds `deleted_at IS NULL`. ✓
- **Overlap guard upgrade** — `internal/worker/scheduler.go:145-160` now skips when any non-terminal task (Pending / WaitingConfirm / Processing) exists; Failed / Completed / Cancelled release the gate as expected. ✓
- **B2 collapse-key isolation** — `internal/api/handler/task.go:416` uses `t{id}` / `s{schedule_id}` namespaces, no cross-collision. ✓
- **Delete-endpoint authorization** — `internal/api/handler/schedule.go:1100`, `internal/api/handler/task.go:976,1004`: non-creator returns 403, no unbind/pause of other users' schedules. ✓
- **B1 migration** — non-blocking based on owner's confirmation of production state (no env has applied `20260608-02`). Acknowledged.

### New should-fix not covered in prior reviews (non-blocking)

🟡 `scanStuckPersonalTasks` does not filter soft-deleted tasks (`internal/worker/scheduler.go:336-385`).

- After `DeleteSchedule` soft-deletes the group task (status=-1 + deleted_at), the `participant` / `personal_result` sub-rows are intentionally retained (no soft-delete column, preserving history).
- The stuck scanner queries only by participant/personal_result, with no join to `summary_task.deleted_at` / `status`.
- Result: every 5/10 minutes the scanner flips `personal_result` back to Pending and re-triggers the personal worker. The worker enters its CAS, observes the task is deleted, and aborts — but it has already written participant/personal_result to Processing.
- Background churn loop on every deleted schedule that had stuck participants at delete time.

Suggested fix: LEFT JOIN `summary_task` in both scanner queries and filter `task.deleted_at IS NULL AND task.status >= 0`, or have the personal worker entry point pre-check the task is not soft-deleted before any state write.

### Other 🟡 (already noted in prior reviews — included for completeness)

- Raw `.Scan` calls don't check error (`internal/api/handler/task.go:420,426,1204`) — silent zero values on failure.
- Migration assumption not embedded in the SQL (`migrations/sql/20260610-01-schedule-one-to-many.sql:1`). Suggest one header comment per #77 precedent: "Assumes no environment applied `20260608-02`; environments that did must manually drop `live_schedule_id` / `uk_live_schedule_binding` / `uk_summary_participant_task_user` before applying."
- Missing regression tests for B3 cascade (assert group task `status=-1` after `DeleteSchedule`) and overlap revival path (current overlap test only seeds `StatusProcessing`).

### 🟢 Nit

- 1062 mapping comments are dead code post 1→N (`internal/api/handler/schedule.go:72,506,897`); safe to clean up later.

### Verification

- `go build ./...` clean on f7a6fa8.
- Full diff (8 files, +364 / -201) reviewed against base `main`.

Net: agree with prior APPROVE. None of the above blocks merge; the scanner soft-delete filter is the one I would prioritize as a follow-up.

yujiawei (approved):
Re-review at f7a6fa88 — withdrawing my earlier CHANGES_REQUESTED.

My prior P0 was the migration-compatibility concern: deleting the already-released `20260608-02-task-schedule-binding.sql` while the runner is Up-only means already-deployed DBs would keep `live_schedule_id` / `uk_live_schedule_binding` / `uk_summary_participant_task_user`, causing 1061 boot failure or 1062 on the second live task.

That blocker was premised on production having applied `20260608-02`. The owner has confirmed production never ran that migration, so no residual columns/indexes exist there, and adding an unconditional `DROP` would instead break clean DBs (dropping non-existent objects). On that confirmed production state the current minimal migration is correct — this matches the owner's clarification.

**Merge precondition (carried, not blocking):** any environment that *did* apply `20260608-02` (e.g. staging/test that ran #77 integration) must, before deploying this PR, manually `DROP INDEX uk_live_schedule_binding`, `DROP INDEX uk_summary_participant_task_user`, `DROP COLUMN live_schedule_id` — otherwise sql-migrate fails with "unknown migration in database" and the stale unique key blocks the 1→N second INSERT. Recommend noting this assumption in the migration comment (per #77 precedent).

The other previously-raised blockers are resolved on f7a6fa88 (list-folding key namespace isolation, non-creator delete → 403, soft-delete `status=-1` + worker `deleted_at IS NULL`, overlap guard upgraded to any non-terminal predecessor). Build / vet / test green.

LGTM — approving.

PR #88, opened 2026-06-10
fix: keep person narrow symmetric with channel narrow for explicit sources (#87)

### LINKED ISSUE #87: bug: explicitly-selected thread summary returns empty when topic wording triggers self-reference person narrow
## Summary

When a user explicitly selects a single thread (子区) as the summary source and phrases the topic with the word "群" (group) — e.g. `总结这个群的信息` — the summary returns "找不到对应的聊天数据" (no chat data). Changing the wording from "群" to "子区" (`总结这个子区的信息`) makes it work, even though the selected source and time range are identical.

This is **not** a model/LLM randomness issue: production runs on claude-sonnet-4-6 and the bug reproduces deterministically. Swapping models does not help.

## Reproduction

1. Create a personal summary task.
2. Explicitly select one thread (frontend `source_type=2`) as the source.
3. Set topic title to `总结这个群的信息`.
4. Run → result: "找不到对应的聊天数据" (0 messages).
5. Change topic to `总结这个子区的信息`, keep everything else identical → normal summary is produced.

## Root Cause

The personal pipeline has two orthogonal narrow-down dimensions:

- **channel narrow** (`ResolveChannelScope`, Layer 1.7, `internal/pipeline/fetch.go`) — guarded by `len(specifiedSources)==0`, so it is correctly **skipped** when the user explicitly selects a source.
- **person narrow** (`ResolveTopicTarget` at `internal/worker/personal_processor.go:328` + `FilterWithContext`) — runs **unconditionally**, with no such guard.

The chain:

1. The selected thread's messages are fetched correctly (the fetch layers are wording-independent).
2. `ResolveTopicTarget(task.Title)` runs an LLM function call. The word "群" in "总结这个群的信息" gets interpreted as a **self-reference** (`has_target=true, include_self=true`), so it returns the creator's UID.
3. `FilterWithContext` then keeps only the creator's messages. But the creator never posted in the selected thread (only a member) → `filter.go` hits `if len(targetIndices)==0 { return nil }` and returns **zero messages with no fallback**.
4. `personal_processor.go` sees 0 messages → returns the "no relevant content" message.

With "子区" wording, the LLM returns `has_target=false` → no person narrow → all messages kept → summary works.

The asymmetry is the core problem: when a source is explicitly selected, channel narrow is correctly disabled, but person narrow is not — and `FilterWithContext` is the only layer in the whole pipeline that returns empty with no fallback.

## Fix Direction

Make person narrow symmetric with channel narrow:

1. When a source is explicitly selected, a pure self-reference (`include_self` with no named person) should **not** become a filter target (mirror the `len(specifiedSources)==0` guard).
2. Named persons (real names in the topic) should still trigger filtering — so "选源 + 点名某人" keeps working.
3. Add an empty-result fallback in the caller so a named target with no messages in the selected source falls back to all messages instead of returning zero.

Affected files: `internal/pipeline/resolve_topic.go`, `internal/worker/personal_processor.go` (and the call-site signature in `internal/worker/processor.go`).

ploy-elison:
Closes #87.

## Problem

When a user explicitly selects a thread (子区) as the summary source and the topic wording triggers a self-reference (e.g. `总结这个群的信息`), the personal pipeline narrowed the result down to the creator and produced an empty summary ("no chat data"), even though the selected source had messages. Changing the wording to `总结这个子区的信息` worked. This reproduced deterministically regardless of the model.

## Root Cause

The personal pipeline has two orthogonal narrow-down dimensions:

- **channel narrow** (`ResolveChannelScope`) is guarded by `len(specifiedSources)==0`, so it is correctly skipped when a source is explicitly selected.
- **person narrow** (`ResolveTopicTarget` + `FilterWithContext`) ran unconditionally, with no such guard.

When a source was explicitly selected, a pure self-reference topic was still resolved to the creator UID. If the creator had no messages in the selected source, `FilterWithContext` returned empty with no fallback, surfacing as "no chat data".

## Fix

Make person narrow symmetric with channel narrow:

1. `ResolveTopicTarget` takes a new `hasExplicitSource` flag. With explicitly-selected sources, a pure self-reference (no concrete person named) no longer narrows to the creator. A concretely named person still narrows as before, so "select source + name a person" keeps working.
2. Empty-result fallback in the caller: when a named target had no messages in the explicitly-selected source(s), fall back to all fetched messages instead of returning empty. The fallback is scoped to `hasExplicitSource` so non-explicit-source self queries keep their previous behavior.

## Tests

- New unit tests covering: explicit-source self-reference skips narrow; non-explicit-source self-reference still returns creator (unchanged); explicit-source named person still narrows; explicit-source named person + include_self still keeps creator.
- `go build ./...` and `go test ./...` pass.

## Verification

End-to-end verified locally with a real worker running the real pipeline against a controlled dataset (a thread where the creator is a member but never posted): the bug scenario (explicit thread source + topic containing "群") now produces a normal summary instead of "no chat data"; the "子区"-wording control and the non-explicit-source self-query path behave as expected.


yujiawei (approved):
# Code Review — PR #88 (octo-smart-summary)

## Verdict: APPROVED

This is a well-scoped, correct fix for the asymmetry described in #87. The root cause is accurately diagnosed and the change is minimal, tested, and consistent with the existing channel-narrow behavior. No P0/P1 issues. One P2 test-fidelity nit below.

## Summary of the change

The personal pipeline has two orthogonal narrowing dimensions:
- **channel narrow** (`ResolveChannelScope`) — already guarded by `len(specifiedSources)==0` (`internal/pipeline/fetch.go:640`), so it is skipped when the user explicitly picks a source.
- **person narrow** (`ResolveTopicTarget` + `FilterWithContext`) — previously ran unconditionally.

The asymmetry meant a topic the LLM misreads as a self-reference (e.g. `总结这个群的信息`) narrowed to the creator UID; if the creator never posted in the explicitly-selected source, `FilterWithContext` returned zero messages with no fallback → "no chat data".

The PR fixes this in two places:
1. `ResolveTopicTarget` gains a `hasExplicitSource` flag; with explicit sources, a *pure* self-reference (`include_self` + empty/all-invalid UIDs) returns `nil` instead of `[creator]`, mirroring channel narrow.
2. A caller-side empty-result fallback in `executePersonalPipeline`: when a resolved target produced zero filtered messages **and** `hasExplicitSource`, it falls back to all fetched messages rather than returning empty.

## Verification

- ✅ **Root cause matches** — confirmed the channel-narrow guard at `internal/pipeline/fetch.go:640` (`len(specifiedSources) == 0`); the new person-narrow skip is a faithful mirror.
- ✅ **Named-person path preserved** — `TestResolveTopicTarget_ExplicitSource_NamedPerson_Narrows` and `..._IncludeSelf_KeepsCreator` confirm concrete names still narrow, and `include_self` still appends the creator. "select source + name a person" keeps working.
- ✅ **Backward compatibility** — non-explicit-source self-reference still returns `[creator]` (`TestResolveTopicTarget_NoExplicitSource_SelfReference_ReturnsCreator`); the new caller fallback is gated on `hasExplicitSource`, so the non-explicit self-query path still surfaces "no relevant content" as before.
- ✅ **Both new branches covered** — the `validUIDs==0 && IncludeSelf` split (creator fallback vs. skip) is exercised by `..._AllUIDsInvalid_IncludeSelf_SkipsNarrow`.
- ✅ **Signature change is safe** — `ResolveTopicTarget` has a single production caller (`internal/worker/personal_processor.go:331`), updated correctly; all 12 existing test call sites updated to pass `false`.
- ✅ **Build & tests** — `go build ./...`, `go vet ./internal/...`, and `go test ./internal/pipeline/... ./internal/worker/...` all pass; the 7 new tests pass individually.

## Findings

### P2 — Test helper does not mirror the real caller's `hasExplicitSource` gate (test fidelity)

`applyTargetFilterWithFallback` in `internal/pipeline/filter_test.go` is documented as mirroring "the caller-side logic in executePersonalPipeline", but it omits the `hasExplicitSource` condition that gates the real fallback:

```go
// helper (test):
filtered := FilterWithContext(messages, targetUIDs, contextWindow)
if len(filtered) == 0 {
    return messages          // unconditional fallback
}
```

```go
// real caller (personal_processor.go):
if len(userMessages) == 0 && hasExplicitSource {  // gated
    userMessages = messages
}
```

Because the helper is a re-implementation rather than a call into the production code, `TestApplyTargetFilterWithFallback_*` validate the *concept* but not the actual guarded behavior — they would still pass even if the production `hasExplicitSource` gate regressed. This is a maintainability/drift risk, not a current correctness bug (the gated branch is independently covered by the `ResolveTopicTarget` tests). Consider either renaming the helper/comment to drop the "mirrors the caller" claim, or adding the `hasExplicitSource` parameter so the helper matches the real branch it stands in for.

### Nit — `targetMsgCount` is 0 on the fallback path

When the explicit-source fallback assigns `userMessages = messages`, those messages bypass `FilterWithContext` and therefore never get `IsTargetUser` set, so `targetMsgCount` (computed at `personal_processor.go:371-377`) stays 0 for this run. This is semantically reasonable — in the fallback there is effectively no specific target — and only affects the count passed to summary generation, not correctness. Flagging for awareness; no change required.

## Conclusion

The fix is correct, minimal, and matches the documented root cause. The two new code paths are covered by tests and the named-person/backward-compat behavior is preserved. The single substantive note (P2) concerns test fidelity, not runtime behavior, and does not block merge.

yujiawei (dismissed):
# Code Review — PR #88 (octo-smart-summary)

## Verdict: APPROVED

This is a well-scoped, correct fix for the asymmetry described in #87. The root cause is accurately diagnosed and the change is minimal, tested, and consistent with the existing channel-narrow behavior. No P0/P1 issues. One P2 test-fidelity nit below.

## Summary of the change

The personal pipeline has two orthogonal narrowing dimensions:
- **channel narrow** (`ResolveChannelScope`) — already guarded by `len(specifiedSources)==0` (`internal/pipeline/fetch.go:640`), so it is skipped when the user explicitly picks a source.
- **person narrow** (`ResolveTopicTarget` + `FilterWithContext`) — previously ran unconditionally.

The asymmetry meant a topic the LLM misreads as a self-reference (e.g. `总结这个群的信息`) narrowed to the creator UID; if the creator never posted in the explicitly-selected source, `FilterWithContext` returned zero messages with no fallback → "no chat data".

The PR fixes this in two places:
1. `ResolveTopicTarget` gains a `hasExplicitSource` flag; with explicit sources, a *pure* self-reference (`include_self` + empty/all-invalid UIDs) returns `nil` instead of `[creator]`, mirroring channel narrow.
2. A caller-side empty-result fallback in `executePersonalPipeline`: when a resolved target produced zero filtered messages **and** `hasExplicitSource`, it falls back to all fetched messages rather than returning empty.

## Verification

- ✅ **Root cause matches** — confirmed the channel-narrow guard at `internal/pipeline/fetch.go:640` (`len(specifiedSources) == 0`); the new person-narrow skip is a faithful mirror.
- ✅ **Named-person path preserved** — `TestResolveTopicTarget_ExplicitSource_NamedPerson_Narrows` and `..._IncludeSelf_KeepsCreator` confirm concrete names still narrow, and `include_self` still appends the creator. "select source + name a person" keeps working.
- ✅ **Backward compatibility** — non-explicit-source self-reference still returns `[creator]` (`TestResolveTopicTarget_NoExplicitSource_SelfReference_ReturnsCreator`); the new caller fallback is gated on `hasExplicitSource`, so the non-explicit self-query path still surfaces "no relevant content" as before.
- ✅ **Both new branches covered** — the `validUIDs==0 && IncludeSelf` split (creator fallback vs. skip) is exercised by `..._AllUIDsInvalid_IncludeSelf_SkipsNarrow`.
- ✅ **Signature change is safe** — `ResolveTopicTarget` has a single production caller (`internal/worker/personal_processor.go:331`), updated correctly; all 12 existing test call sites updated to pass `false`.
- ✅ **Build & tests** — `go build ./...`, `go vet ./internal/...`, and `go test ./internal/pipeline/... ./internal/worker/...` all pass; the 7 new tests pass individually.

## Findings

### P2 — Test helper does not mirror the real caller's `hasExplicitSource` gate (test fidelity)

`applyTargetFilterWithFallback` in `internal/pipeline/filter_test.go` is documented as mirroring "the caller-side logic in executePersonalPipeline", but it omits the `hasExplicitSource` condition that gates the real fallback:

```go
// helper (test):
filtered := FilterWithContext(messages, targetUIDs, contextWindow)
if len(filtered) == 0 {
    return messages          // unconditional fallback
}
```

```go
// real caller (personal_processor.go):
if len(userMessages) == 0 && hasExplicitSource {  // gated
    userMessages = messages
}
```

Because the helper is a re-implementation rather than a call into the production code, `TestApplyTargetFilterWithFallback_*` validate the *concept* but not the actual guarded behavior — they would still pass even if the production `hasExplicitSource` gate regressed. This is a maintainability/drift risk, not a current correctness bug (the gated branch is independently covered by the `ResolveTopicTarget` tests). Consider either renaming the helper/comment to drop the "mirrors the caller" claim, or adding the `hasExplicitSource` parameter so the helper matches the real branch it stands in for.

### Nit — `targetMsgCount` is 0 on the fallback path

When the explicit-source fallback assigns `userMessages = messages`, those messages bypass `FilterWithContext` and therefore never get `IsTargetUser` set, so `targetMsgCount` (computed at `personal_processor.go:371-377`) stays 0 for this run. This is semantically reasonable — in the fallback there is effectively no specific target — and only affects the count passed to summary generation, not correctness. Flagging for awareness; no change required.

## Conclusion

The fix is correct, minimal, and matches the documented root cause. The two new code paths are covered by tests and the named-person/backward-compat behavior is preserved. The single substantive note (P2) concerns test fidelity, not runtime behavior, and does not block merge.

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# octo-smart-summary PR#88 Review Report
**PR:** Mininglamp-OSS/octo-smart-summary#88
**Head SHA:** `451741e08e24dc008cf28639b7d8ba8df9c36ff4`
**Title:** fix: keep person narrow symmetric with channel narrow for explicit sources (#87)

---

## 1. Verification Summary

| Item | Status | Evidence |
|------|--------|----------|
| Build (`go build ./...`) | ✅ PASS | Clean build, no errors |
| Tests (`go test ./...`) | ✅ PASS | All packages pass, including 5 new tests |
| Root-cause analysis | ✅ Correct | Person narrow ran unconditionally while channel narrow was guarded by `len(specifiedSources)==0` — verified at `fetch.go:645` vs old `personal_processor.go:327` |
| Symmetry fix | ✅ Correct | `hasExplicitSource := len(sources) > 0` at `personal_processor.go:331` uses the same data source as `specifiedSources` at line 270 |
| Backward compat (non-explicit-source paths) | ✅ Preserved | All existing tests pass with `hasExplicitSource=false`; self-reference still returns `[creatorUID]` |

## 2. Findings

### F1 — `IsTargetUser` not set on fallback messages → `targetMsgCount=0` in reduce prompt & DB [P2]

**Severity:** P2 (misleading metadata, not a broken path)
**Diff-scope:** **new** — introduced by this PR's fallback code path.

**Data-flow trace:**

1. When the empty-result fallback fires (`personal_processor.go:349-352`), `userMessages = messages` (the raw fetch result from `ResolveAndFetchMessagesForPersonal`).
2. `IsTargetUser` is **only** set inside `FilterWithContext` (`filter.go:40`). The raw `messages` from the fetch layer have `IsTargetUser = false` (zero value).
3. The downstream loop at `personal_processor.go:371-377` counts `targetMsgCount` by checking `IsTargetUser`:
   ```go
   if userMessages[i].IsTargetUser { targetMsgCount++ }
   ```
   Since no message has `IsTargetUser=true`, `targetMsgCount = 0`.
4. `targetMsgCount` flows to two consumers:
   - **Reduce prompt** (`llm.go:518`): `"消息总量：%d 条"` → tells the LLM "0 messages" when there are actually N messages. In multi-chunk scenarios this could confuse the reduce phase.
   - **DB record** (`personal_processor.go:561` → `completedPersonalResultUpdates`): `msg_count = 0` stored in `SummaryResult`.

**Trigger condition:** Explicit source + named person (valid UID) + person has zero messages in the selected source + enough messages to produce multiple chunks. Narrow edge case, but reachable in production.

**Suggested fix:** After the fallback assignment, set `IsTargetUser` on the messages based on `targetUIDs`:
```go
if len(userMessages) == 0 && hasExplicitSource {
    userMessages = messages
    // Re-tag target messages so downstream counts are correct
    targetSet := make(map[string]bool, len(targetUIDs))
    for _, uid := range targetUIDs { targetSet[uid] = true }
    for i := range userMessages {
        userMessages[i].IsTargetUser = targetSet[userMessages[i].SenderUID]
    }
}
```

### F2 — Test helper `applyTargetFilterWithFallback` duplicates production logic [P2/nit]

**Severity:** P2/nit (maintainability)
**Diff-scope:** **new** — test helper added by this PR.

`applyTargetFilterWithFallback` in `filter_test.go:138-148` mirrors the fallback logic in `personal_processor.go:338-352`. The tests validate the helper, not the production code path. If the production logic is refactored independently, the tests could remain green while the production behavior drifts.

**Suggested fix:** Extract the fallback into a package-level function in `pipeline/` (e.g., `FilterWithFallback`) and test that directly, or add an integration-level test that exercises `executePersonalPipeline` with a mock LLM.

## 3. Suggestions

1. **F1 fix** (above) — re-tag `IsTargetUser` after fallback to keep `targetMsgCount` accurate.
2. **F2 fix** (above) — extract fallback to a testable function, or accept the duplication risk given the logic is 4 lines.

## 4. Additional Observations

- **Single production caller confirmed:** `ResolveTopicTarget` is only called from `executePersonalPipeline` (`personal_processor.go:331`). The group pipeline (`processor.go:466`) does not use it. No cross-pipeline impact.
- **Channel-scope / person-scope symmetry verified:** Channel scope guard at `fetch.go:645` (`len(specifiedSources)==0`) and person scope guard at `personal_processor.go:331` (`hasExplicitSource := len(sources) > 0`) derive from the same DB query result. Symmetry is correct.
- **Signature change is breaking but safe:** Adding `hasExplicitSource bool` to `ResolveTopicTarget` is a breaking API change, but with only one production caller, this is safe.

## 5. Data-Flow Trace

|---|---|---|
| `hasExplicitSource` | `len(sources) > 0` where `sources` ← DB `Find(&sources, task_id)` (`personal_processor.go:264`) | ✅ Same data as `specifiedSources` passed to fetch layer |
| `targetUIDs` (self-ref + explicit source) | `ResolveTopicTarget` returns `nil` (`resolve_topic.go:152-154`) | ✅ → `len(targetUIDs)==0` → else branch → `userMessages = messages` |
| `targetUIDs` (named person + explicit source) | `ResolveTopicTarget` returns `[validUID]` (`resolve_topic.go:192-194`) | ✅ → FilterWithContext narrows; fallback if empty |
| `targetUIDs` (named person + include_self + explicit source) | `ResolveTopicTarget` returns `[validUID, creatorUID]` (`resolve_topic.go:186-191`) | ✅ Creator appended after valid UIDs |
| `targetUIDs` (all-invalid UIDs + include_self + explicit source) | `ResolveTopicTarget` returns `nil` (`resolve_topic.go:178-180`) | ✅ No creator fallback when explicit source |
| `userMessages` after fallback | Raw `messages` from fetch layer | ⚠️ `IsTargetUser` not set → `targetMsgCount=0` (F1) |

## 6. R5 Blind-Spot Checklist

- **C1 — Dual-path parity:** ✅ Clear. The PR explicitly addresses the asymmetry between channel narrow (guarded) and person narrow (unguarded). After the fix, both are guarded by the same condition. Add/remove, subscribe/unsubscribe patterns: N/A for this PR.
- **C2 — Control-flow ordering / nesting reuse:** ✅ Clear. `ResolveTopicTarget` is called exactly once in the pipeline. No nested or repeated calls. The fallback logic is a simple linear check.
- **C3 — Authorization boundary ≠ capability boundary:** ✅ N/A. This PR does not touch authorization, permissions, or tool exposure.

---

**[Octo-Q] verdict: APPROVE**

The fix is correct, well-scoped, and achieves its stated goal. The root cause analysis is accurate and the symmetry argument is verified. Two P2 findings (F1: `IsTargetUser` not re-tagged after fallback → misleading `targetMsgCount`; F2: test helper duplicates production logic) are noted but do not block landing. F1 is worth fixing in a follow-up or before merge if the author has bandwidth — it only affects a narrow edge case (explicit source + named person absent from source + multi-chunk reduce).

Jerry-Xin (changes requested):
One-line summary: The PR is in-scope for this repository, but the fix regresses legitimate explicit-source self queries.

🔴 Blocking

- 🔴 Critical: Explicit-source self queries are now always widened to all messages. In [internal/pipeline/resolve_topic.go](/Users/mlamp/.openclaw/tmp/pr-review-wyvh0a1y/repo/internal/pipeline/resolve_topic.go:150), any `include_self=true` result with no concrete UID returns `nil` when `hasExplicitSource` is true. Because [internal/worker/personal_processor.go](/Users/mlamp/.openclaw/tmp/pr-review-wyvh0a1y/repo/internal/worker/personal_processor.go:330) sets that flag for every selected source, a legitimate request like “summarize what I said in this selected thread” will skip person narrowing and summarize the whole selected source instead. This fixes the false-positive “this group” case by disabling the valid “me in this source” case. The resolver needs a way to distinguish actual first-person intent from the LLM’s false self-reference classification, or the worker needs a narrower guard than `len(sources) > 0`.

💬 Non-blocking

- 🟡 Warning: The fallback tests in [internal/pipeline/filter_test.go](/Users/mlamp/.openclaw/tmp/pr-review-wyvh0a1y/repo/internal/pipeline/filter_test.go:137) test a local helper that only mirrors worker logic, not the actual `executePersonalPipeline` path. That gives some unit coverage for the intended behavior, but it will not catch integration drift in the worker. A focused worker-level test for explicit-source target filtering would be stronger.

✅ Highlights

- ✅ The PR correctly identifies that channel narrowing already skips explicit sources and that person narrowing was asymmetric.
- ✅ `go test ./...` passes locally.
- ✅ The added `ResolveTopicTarget` tests cover the new flag behavior and preserve the non-explicit-source self-reference path.

ploy-elison:
Updated the approach based on review feedback.

The previous commit used a `hasExplicitSource` blanket guard that skipped person narrow for any self-reference when a source was explicitly selected. As pointed out, that regressed legitimate first-person queries (e.g. "summarize what I said in this selected thread"), because that case is indistinguishable from the misclassified "this group" case once it reaches the worker.

Root cause is upstream: the model was classifying source-range references ("this group" / "here") as `include_self=true`. The fix now addresses it at the resolver layer:

- Tightened the `include_self` definition (schema + prompt): scope words ("这个群/这里/本群/这个子区"...) are source-range references and never trigger `include_self`; only an explicit first-person "我" as a participant does. Added positive/negative examples.
- Reverted the `hasExplicitSource` parameter. A pure self-reference resolves to the creator again, so a genuine "what did I say here" query is narrowed correctly.
- When a true first-person query finds no creator messages in the selected source, return a clear notice ("你在所选范围内没有发言记录。") instead of silently widening to the whole source. Named-other empty hits still fall back to the whole source.
- Normalized `targetMsgCount` on untrimmed/fallback paths to the actual message count.
- Included both empty-placeholder notices in the scheduled-task skip check so a transient empty window never overwrites a previous valid result.

`go build ./...`, `go vet ./...` and `go test ./...` all pass.

mochashanyao (approved):
[Octo-Q · automated review]

---

# octo-smart-summary PR#88 Review Report
**PR:** Mininglamp-OSS/octo-smart-summary#88
**Head SHA:** `bd5148265c4394f3ac2cff762ebddca5e8cf70f9`
**Files:** `resolve_topic.go` (+21/-4), `personal_processor.go` (+56/-4), `personal_processor_test.go` (+144, new)

---

## 1. Verification Summary

| Item | Status | Evidence |
|------|--------|----------|
| Build (`go build ./...`) | ✅ PASS | Clean exit, no warnings |
| Unit tests (`go test ./...`) | ✅ PASS | All 8 new tests + full suite green |
| Prompt tightening logic | ✅ Correct | Tool schema + system prompt + examples all consistent |
| `decidePersonalMessages` decision table | ✅ Correct | All 4 branches verified against data flow |
| `isEmptyPlaceholder` + scheduled skip | ✅ Correct | Both placeholders covered, backward-compatible |
| `targetMsgCount` F1 normalization | ✅ Correct | Only fires on fallback/no-target paths; true narrow unaffected |
| Data flow: fetch → resolve → filter → decide → count | ✅ Traced end-to-end | See §5 below |

---

## 2. Findings

### F1 — LLM-dependent classification is the sole guard against scope-word misclassification (P2, pre-existing risk, mitigated)

**Diff-scope:** The PR replaces a code-level `hasExplicitSource` blanket skip (commit 1) with prompt-layer tightening (commit 2). The fix now relies on the LLM correctly classifying `include_self=false` for scope-word topics like "总结这个群的信息".

**Risk:** If the LLM still misclassifies `include_self=true` for a scope-word topic, the resolver returns `targetUIDs=[creator]`. If the creator has no messages in the selected source, the user now sees "你在所选范围内没有发言记录" instead of the previous silent empty summary. This is a **better** failure mode (clear message vs. confusing "no chat data"), but still incorrect.

**Mitigation quality:** The tightened prompt is well-structured with:
- Explicit "iron rules" (判定铁律) with numbered priority ordering
- 12 concrete examples covering both positive and negative cases
- Clear separation of scope words vs. first-person pronouns
- The `include_self` tool description mirrors the system prompt rules

**Verdict:** P2 — acceptable risk. The prompt engineering is thorough, and the residual misclassification produces a clearer error message than before. Not a blocker.

### F2 — `isEmptyPlaceholder` uses exact string matching (P2/nit, new)

**Diff-scope:** New function `isEmptyPlaceholder` at `personal_processor.go:49-52`.

```go
func isEmptyPlaceholder(content string) bool {
    trimmed := strings.TrimSpace(content)
    return trimmed == noRelevantContentMessage || trimmed == noSelfMessagesMessage
}
```

If a future placeholder message is added but not registered here, scheduled tasks could overwrite previous valid results with the new placeholder. Low risk since both constants are package-level and co-located, but worth a comment in the code pointing to this function when adding new placeholders.

**Verdict:** P2/nit — maintainability note, not a blocker.

---

## 3. Suggestions

1. **Consider a code-level safety net for scope-word misclassification (optional, not blocking):** A lightweight post-resolver check — if `include_self=true` was returned but the topic text contains no literal "我" character — could override the LLM's `include_self` to `false`. This would make the fix deterministic rather than LLM-dependent. Something like:

   ```go
   if parsed.IncludeSelf && !strings.Contains(topic, "我") {
       parsed.IncludeSelf = false
   }
   ```

   This is a belt-and-suspenders option; the prompt alone may be sufficient in practice.

2. **Add a code comment near `isEmptyPlaceholder`** listing the contract: "any new empty-result placeholder message must be registered here to prevent scheduled-task overwrite."

---

## 4. Additional Observations

- **Commit 2 is a clean improvement over commit 1.** The first commit's `hasExplicitSource` blanket skip would have prevented genuine "what did I say in this thread" queries from narrowing correctly. The refactor to resolve at the prompt layer preserves correct behavior for true first-person queries while fixing the scope-word bug.
- **The `noSelfMessagesMessage` early-return path is well-integrated.** It correctly flows through the caller's `shouldSkipScheduledPlaceholderResult` check, ensuring scheduled tasks preserve previous valid results while manual tasks show the user a clear notice.
- **Test coverage is comprehensive:** 5 `decidePersonalMessages` tests covering all 4 branches + edge case, 2 `targetMsgCount` normalization tests, 1 scheduled-placeholder test with 4 sub-cases.

---

## 5. Data Flow Trace

### Full pipeline chain (post-PR):

```
ResolveAndFetchMessagesForPersonal (fetch.go:604)
  → Layer 1.7: ResolveChannelScope [gated by len(specifiedSources)==0] ← unchanged
  → returns []Message (all IsTargetUser=false)

ResolveTopicTarget (resolve_topic.go:57)
  → LLM resolves has_target, uids, include_self
  → Tightened prompt: scope words ≠ include_self
  → Returns targetUIDs or nil

if len(targetUIDs) > 0:
  FilterWithContext (filter.go:4)
    → Sets IsTargetUser=true on target messages (filter.go:40)
    → Returns nil if no target messages found
  decidePersonalMessages (personal_processor.go:37)
    → filtered non-empty → use filtered (IsTargetUser populated)
    → filtered empty + self-only → nil + "你在所选范围内没有发言记录"
    → filtered empty + other targets → fallback to all (IsTargetUser all false)

targetMsgCount loop (personal_processor.go:402-415)
  → Counts IsTargetUser=true messages
  → F1: if count==0 → normalize to len(userMessages)
  → True narrow paths have ≥1 IsTargetUser → unaffected
```

### Key data-flow invariants verified:

1. **`IsTargetUser` is only set in `FilterWithContext` (filter.go:40).** Messages on fallback/no-target paths never pass through this function, so `IsTargetUser` stays `false`. The F1 normalization correctly compensates.
2. **`decidePersonalMessages` pure-self check (`len==1 && [0]==creator`) is safe.** Multi-target queries like "我和老王" always have `len(targetUIDs) > 1`, so they never trigger the self-empty notice.
3. **Early return with `noSelfMessagesMessage` flows correctly through the caller.** `msgCount=0`, `isEmptyPlaceholder` returns `true`, scheduled tasks skip content overwrite, manual tasks show the notice.
4. **`ResolveChannelScope` guard (`len(specifiedSources)==0`) is unchanged** and orthogonal to this PR's person-narrow fix, confirming the symmetry described in the PR body.

---

## 6. Blind-Spot Checklist (R5)

- **C1 — Dual-path parity:** N/A. This PR fixes an asymmetry between channel narrow (already guarded) and person narrow (now guarded at the prompt layer). The fix correctly makes person narrow symmetric. No create/delete or subscribe/unsubscribe pairs touched.
- **C2 — Control-flow ordering / nesting reuse:** Clear. `decidePersonalMessages` is called at exactly one site. `isEmptyPlaceholder` is called at exactly one site (`shouldSkipScheduledPlaceholderResult`). No nesting or double-application risk.
- **C3 — Authorization boundary ≠ capability boundary:** N/A. No auth/permission/tool-exposure changes in this PR.

---

## [Octo-Q] verdict: APPROVE

No P0/P1 findings. The fix is well-designed, well-tested, and correctly addresses the root cause. Two P2 observations (LLM-dependent classification risk, exact-match placeholder detection) are noted but not blocking. The optional code-level safety net suggestion in §3 could further harden the fix but is not required for landing.

Jerry-Xin (approved):
## Re-review: R1 blocker resolved ✅

The R1 blocker (explicit-source self-query regression) has been properly fixed via a two-pronged approach:

1. **Hardened `include_self` prompt** — The tool description and system prompt now have strict rules preventing false self-reference classification when the topic only contains scope words (这个群/这个子区/这里). This was the root cause of the original misclassification.

2. **`decidePersonalMessages` helper** — When `ResolveTopicTarget` correctly identifies a self-reference (returns `[creator]`) and the creator has no messages in the filtered source, the pipeline now returns `noSelfMessagesMessage` instead of falling back to all messages. This eliminates the "query about my messages returns everyone's messages" regression.

The fix is symmetric with the channel narrow path: explicit sources constrain *which channels to fetch from* (Layer 2), while person narrowing constrains *whose messages to summarize*. When both combine and the target has no messages, the user gets a clear answer rather than an incorrect fallback.

**✅ Highlights**
- `decidePersonalMessages` cleanly separates the four cases (no target / filtered non-empty / self-empty / named-other-empty) with appropriate behavior for each
- Comprehensive unit tests cover all branches including the edge case of self+other targets
- `shouldSkipScheduledPlaceholderResult` correctly extended to the new placeholder
- `targetMsgCount` normalization (F1 fix) addresses the R1 warning about stats being 0 on fallback paths
- The prompt hardening with explicit examples is a solid defense-in-depth measure

**💬 Non-blocking**
- 🟡 Consider adding an integration test that exercises the full `executePersonalPipeline` path with a mocked explicit-source + self-query scenario, to prevent future regressions at the wiring level (not blocking since the unit tests validate the helper logic thoroughly)

yujiawei (approved):
# Code Review — PR #88 (octo-smart-summary)

## Verdict: APPROVED

The change builds (`go build ./...`), the full suite passes (`go test ./...`), and it correctly resolves the happy path of #87: with an explicitly-selected thread source and a range-word topic like `总结这个群的信息`, the pipeline no longer narrows to the creator and produces a normal summary. No P0/P1 defects found. One **major design/robustness concern** is documented below as a strong, non-blocking recommendation for a follow-up.

Note for reviewers: this PR contains two commits, and they take *different* approaches. The first (`fix: keep person narrow symmetric...`) added an `hasExplicitSource` code guard. The second (`refactor: resolve self-reference at the prompt layer instead of a blanket guard`) **removed** that guard and shifted the disambiguation entirely into the LLM prompt. The final tree is the prompt-layer approach — the PR description still describes the first commit's `hasExplicitSource` design, so the description is stale relative to what actually ships.

---

## What ships (final state)

1. `ResolveTopicTarget` no longer takes `hasExplicitSource`. The signature is back to the original. Disambiguation of "range word vs first-person" is now done by a substantially expanded system prompt + few-shot examples that explicitly classify `这个群 / 这里 / 子区 / 频道` as *source scope* (never `include_self`), and require a literal `我` for `include_self=true`.
2. `decidePersonalMessages(targetUIDs, creatorUID, all, filtered)` centralizes the post-filter decision:
   - no target → all messages;
   - filtered non-empty → filtered (normal narrow);
   - filtered empty + target is exactly the creator → early-return `noSelfMessagesMessage`;
   - filtered empty + other/named target → fall back to all messages.
3. `noSelfMessagesMessage` is added and treated as an empty-placeholder alongside `noRelevantContentMessage`, so scheduled tasks won't overwrite a prior valid result with it.
4. `targetMsgCount` is normalized to `len(userMessages)` on untrimmed/fallback paths (where no message carries `IsTargetUser`), so Reduce and persistence see a real count instead of 0.

## Strengths

- The decision logic is extracted into a small, pure, well-tested function (`decidePersonalMessages`) with clear unit coverage for all four branches.
- The `targetMsgCount` normalization (F1) is a genuine correctness fix for the newly-introduced fallback paths — previously those paths would have persisted a count of 0.
- Treating both placeholder strings uniformly in `shouldSkipScheduledPlaceholderResult` is correct and tested, and prevents a transient empty window from clobbering a good scheduled result.
- `go build` and `go test ./...` both pass locally.

---

## Findings

### Major (design / robustness) — non-blocking, recommend a follow-up safety net

Issue #87 stressed that the bug is **deterministic and not LLM randomness** ("production runs on claude-sonnet-4-6 and the bug reproduces deterministically"). The first commit fixed it deterministically with a code guard. The final commit removes that guard and makes the headline fix **entirely dependent on the LLM obeying the prompt**.

Concretely, for the exact reported scenario (explicit thread source + `总结这个群的信息` + creator never posted in that thread):

- If the model classifies `这个群` as `include_self=false` → `has_target=false` → `targetUIDs=nil` → all messages → correct summary. ✅
- If the model regresses and still classifies `这个群` as `include_self=true` → `targetUIDs=[creator]` → `FilterWithContext` returns empty → `decidePersonalMessages` hits the *pure-self* branch and returns `noSelfMessagesMessage` ("你在所选范围内没有发言记录。"). ❌ — i.e. the original "empty result for what should be a full-source summary" symptom recurs, just with different wording.

The asymmetry worth calling out: the **named-other-person** empty case falls back to the whole source, but the **pure-self** empty case does **not** — and the original bug was a pure-self misclassification. So the one scenario #87 was filed about is the one scenario with no code-level backstop; its correctness rides 100% on prompt adherence.

This is a defensible tradeoff (the blanket `hasExplicitSource` guard genuinely broke legitimate "explicit source + 我说了什么" queries, and intent classification *is* the right layer to disambiguate). The expanded prompt with targeted few-shot is reasonable mitigation. But given #87 explicitly framed determinism as a requirement, I'd recommend a defensive net in a follow-up: when the source is explicitly selected and a *pure-self* target resolves to zero messages, prefer falling back to the whole selected source over `noSelfMessagesMessage`. The user explicitly picked a source; an empty self-result there is more likely a misclassification than a true "you said nothing" — and a full-source summary is the safer failure mode than an empty one. (This would require threading an `hasExplicitSource`-style signal back into `decidePersonalMessages`, which the refactor deliberately dropped.)

### P2 — description/behavior parity for the non-explicit-source self path

The PR description says non-explicit-source self queries "keep their previous behavior." The *control flow* is preserved (no fallback, empty placeholder), but the **message text changes**: a non-explicit-source self query where the creator never spoke now returns `noSelfMessagesMessage` ("你在所选范围内没有发言记录。") instead of the previous `noRelevantContentMessage` ("在当前范围内未找到与主题相关的聊天记录。"). Both are placeholders so scheduled-skip handling is unaffected, but the user-visible string did change for this path. Worth a one-line note in the description, or aligning the wording if parity was intended. Also, for a non-explicit-source query the phrase "所选范围" (selected range) is slightly inaccurate since the user didn't select a source.

### P2 — lost regression coverage around the actual bug

The refactor deleted the `applyTargetFilterWithFallback` tests in `internal/pipeline/filter_test.go` and trimmed `resolve_topic_test.go`. The new `personal_processor_test.go` covers `decidePersonalMessages` branches well, but there's no test that pins the headline behavior — i.e. "a range-word topic must not narrow to the creator." Since correctness now depends on prompt classification, consider adding a test that stubs `ResolveTopicTarget`'s tool-call function to return `include_self=false / has_target=false` for `总结这个群` (asserting the prompt's *intended* contract), plus a test documenting the pure-self misclassification fallback gap above. These guard the exact regression #87 describes.

### Nit

- `decidePersonalMessages` treats a target list of exactly `[creator]` as "pure self." If the LLM ever returns the creator's UID as a *named* person (real name in topic, not via `include_self`), that user would also get `noSelfMessagesMessage` rather than a fallback. Edge case, acceptable, but the doc comment's "true first-person query" framing slightly overstates what the code can distinguish.

---

## Summary

Correct, tested, and it fixes the reported happy path; no blocking defects. The one thing I'd flag for the team's awareness is that the final approach trades the first commit's deterministic guard for prompt-layer classification, which reintroduces LLM dependence for precisely the scenario #87 said was deterministic — and the pure-self/explicit-source empty case has no code-level fallback. Approving on the basis that the code is correct and the prompt mitigation is reasonable, with a recommendation to add a defensive fallback + regression test in a follow-up.

Issue #94, opened 2026-06-12
Backfill historical thread summaries with correct source_name (follow-up of #93)


boris-clark:
## Background

This is a follow-up to #93. The fix in #93 corrects the thread-source resolution logic so that **newly created** summaries store the correct thread `source_name` (real thread name instead of the parent group name).

However, `source_name` is resolved once at summary-creation time and persisted into `summary_source`. The list API only reads it back. Therefore the #93 fix does **not** retroactively correct historical thread summaries that were already saved with the wrong (parent-group) name.

## Problem

Existing rows in `summary_source` where `source_type = 2` (thread) may still hold an incorrect `source_name` (parent group name, or an indistinguishable placeholder). These historical summaries continue to render ambiguously in the list UI even after #93 lands.

## Expected behavior

Provide a one-time, safe backfill that recomputes `source_name` for historical thread summaries using the corrected resolution logic, so the existing summary list also shows the correct thread names.

## Requirements

- One-time backfill that iterates `summary_source` rows with `source_type = 2` and recomputes `source_name` with the corrected resolver.
- Must support a **dry-run** mode that prints the intended diffs without writing.
- Must run in **batches** and be **idempotent / re-runnable** (safe to interrupt and resume).
- Must run as a maintenance command / guarded startup task, **not** mixed into the #93 code fix.
- Must not touch group (`source_type = 1`) or DM (`source_type = 3`) rows.
- Recommended to run during off-peak hours.

## Acceptance criteria

- [ ] Dry-run mode lists every row that would change, with old and new `source_name`.
- [ ] Actual run updates only `source_type = 2` rows; group and DM rows untouched.
- [ ] Backfill is idempotent: re-running produces no further changes once complete.
- [ ] Batched execution with bounded query size.
- [ ] Group and DM summaries verified unaffected after the run.

## Related

- Follow-up of #93 (thread source_name resolution fix for new summaries).

Issue #93, opened 2026-06-12
Summary list API does not distinguish thread vs group source (source_name shows parent group name, missing source_type field)


Jerry-Xin:
## Problem

The summary list API does not distinguish between **thread (sub-topic)** and **group** sources. When a summary is generated for a specific thread, the `source_name` field still shows the **parent group name**, so users cannot tell the real source of the summary.

API: `GET https://im.deepminer.com.cn/summary/api/v1/summaries?page=1&page_size=20`

## Expected behavior

- `source_name` should reflect the **actual thread name** when the summary source is a thread, instead of falling back to the parent group name.
- Add an extra field that explicitly identifies the **source type** (group / DM / thread), so the frontend can render differentiated UI for group chats, direct messages, and threads.

## Why it matters

Without distinguishing source type and showing the correct thread name, the summary list is ambiguous — multiple summaries from different threads of the same group all look identical. Adding a type flag + correct source name unblocks the frontend to present group / DM / thread summaries differently.

## Acceptance criteria

- [ ] Summary list API returns the correct `source_name` for thread sources (real thread name, not parent group name).
- [ ] A new field (e.g. `source_type`) distinguishes group / DM / thread.
- [ ] Backward compatible with existing group and DM summaries.

Issue #89, opened 2026-06-11
Scheduled summary group name degrades to fallback placeholder (source_name dropped + nil IM DB in resolver)


Ranwanglc:
## Summary

After a scheduled summary is triggered, the generated source record `summary_source.source_name` degrades into a fallback placeholder name such as `来源-<id前8位>(群聊)` ("Source-<first 8 chars of id>(Group)"), instead of the chat's real name (e.g. `<real group name>(群聊)`). Manually created summary tasks are not affected — only scheduled tasks exhibit this problem.

> Note: This is **not** a database charset/encoding issue. Verified with `HEX(source_name)` that the value stored in the DB is genuine text (`来源-...`); it is a placeholder name written by the fallback logic.

## Reproduction

1. Create a scheduled summary (`summary_schedule`) with a group chat as the source.
2. Wait for the scheduled task to trigger and generate `summary_source` records.
3. Inspect `summary_source.source_name`: some records show `来源-<first 8 chars of source_id>(群聊)` instead of the real group name.

Example data (same group `<group_no>`, real name `<real group name>`):

| source_name | Note |
| --- | --- |
| `<real group name>(群聊)` | Correct |
| `来源-<first 8 chars>(群聊)` | Degraded to fallback placeholder |

## Root Cause

Two independent defects combine to cause this (both are required):

### Defect A — Backend entry point drops `source_name` (the root cause)

The `sourceReq` struct in `internal/api/handler/task.go` only has `SourceType` and `SourceID` — it is **missing the `SourceName` field**:

```go
// internal/api/handler/task.go:121-124
type sourceReq struct {
    SourceType int    `json:"source_type"`
    SourceID   string `json:"source_id"`
    // SourceName is missing
}
```

When creating/updating a schedule (`internal/api/handler/schedule.go`), `req.Sources []sourceReq` deserializes the frontend payload. Because the struct has no such field, **the `source_name` sent by the frontend is silently dropped during deserialization**. The subsequent `json.Marshal(req.Sources)` stored into `summary_schedule.source_config` therefore never contains `source_name`.

```go
// internal/api/handler/schedule.go:351-353
if len(req.Sources) > 0 {
    b, _ := json.Marshal(req.Sources) // the marshaled sourceReq has no source_name
    sourceConfig = b
}
```

### Defect B — Scheduled worker cannot resolve the name, and the fallback resolver is disabled

[...]

PR #95, opened 2026-06-12
fix(summary): correct thread source_name resolution & enforce backend-authoritative naming (#93)

Ranwanglc:
## 背景 / Issue

修复 #93：智能总结中子区（`source_type=2`）的「信息来源名称」`source_name` 显示为父群名、同一父群下多个子区无法区分。同时统一即时总结与定时总结两条链路的命名口径。

历史数据回填作为独立 issue **#94** 处理，本 PR 仅交付代码修复。

### 问题一：子区 source_name 解析（`internal/service/summary.go`）
- `case 2`（thread）改为用 **联合键 `(group_no, short_id)`** 的单条 JOIN 查询，联 `group` 表取父群名，跨表比较加 `COLLATE utf8mb4_unicode_ci`（与 `internal/pipeline/fetch.go`、`internal/api/handler/candidates.go` 既有写法一致）。替换原先 `WHERE short_id = ?` 的单键查询（可能命中其它群的同名 `short_id`）。
- 不再用 `_ =` 吞掉查询错误，改为 `log.Printf` 打日志，便于线上定位。
- 新增 `threadFallbackName()`：未命名 / 未命中的子区占位名**永远包含 `short_id` 片段**，保证同一父群下多个未命名子区互不撞名（#93 根因正是旧逻辑退化成纯父群名）。
- 子区 `source_id` 形态前置校验，非法值走确定性兜底。

### 问题二：即时 / 定时链路命名口径统一（`internal/worker/scheduled_replace_helpers.go`）
- `buildScheduledTaskSources` / `syncScheduledTaskSources` 改为**始终**调用 `ResolveSourceNameWithType` 重新解析，不再信任前端传入的 `source_name`（定时表单可能把脏值如 `groupNo____shortId` 当 name 传上来）。与即时总结链路口径对齐。

### 测试（`internal/worker/scheduled_replace_test.go`）
- 更新 `TestBuildScheduledTaskSources`：断言落库的 `source_name` 是**后端重解析值**、且**不等于**配置里的脏值。

## 验证

- `go test ./...` 全绿。
- 线上真实链路验证：同一父群 `qwen3-omni-speech-te` 下子区 `子区问题` 的定时总结，配置 `source_name` 为空，生成的 `summary_source.source_name` = `qwen3-omni-speech-te-子区问题(子区)`（HEX 还原确认正确 UTF-8），证明子区名正确带入 + 生成层重解析 + COLLATE 联合键 JOIN 均生效。

## 不涉及

- 无 DB schema 变更：`source_name` 仍 `varchar(200)`；`source_type` 枚举沿用 `1/2/3`。
- 历史数据回填 → #94 单独 PR。
- 前端 `octo-web` 不再发送 `source_name` → 跨仓、非阻塞，单独排期。

### DIFF
--- a/internal/service/summary.go
+++ b/internal/service/summary.go
@@ -2,6 +2,7 @@ package service
 
 import (
 	"fmt"
+	"log"
 	"math/rand"
 	"strings"
 
@@ -58,29 +59,50 @@ func ResolveSourceNameWithType(sourceID string, sourceType int, imDB *gorm.DB) s
 			return name + "(群聊)"
 		}
 	case 2: // thread
-		// Thread source_id format: {group_no}____{short_id}
-		// Note: For thread, ____ separates group_no and short_id; both are used
+		// 子区身份是 (group_no, short_id) 联合键；source_id 形如 {group_no}____{short_id}。
+		// 解析严格依赖 source_type==2，不靠 "____" 的出现与否猜测是否为子区
+		// （群 source_id 也可能含 "____{space_id}" 后缀，见 case 1）。
 		parts := strings.SplitN(sourceID, "____", 2)
-		if len(parts) == 2 {
-			groupNo := parts[0]
-			shortID := parts[1]
-			
-			var groupName string
-			_ = imDB.Table("group").Where("group_no = ?", groupNo).Pluck("name", &groupName).Error
-			
-			var threadName string
-			_ = imDB.Table("thread").Where("short_id = ?", shortID).Pluck("name", &threadName).Error
-			
-			if groupName != "" && threadName != "" {
+		if len(parts) != 2 || parts[0] == "" || parts[1] == "" {
+			// 非法子区 source_id：走确定性兜底，绝不退化成父群名。
+			log.Printf("[service] ResolveSourceNameWithType: malformed thread source_id=%q", sourceID)
+			return fallbackSourceName(sourceID, sourceType)
+		}
+		groupNo := parts[0]
+		shortID := parts[1]
+
+		// 一条 JOIN 同时取子区名与父群名；group_no 跨表比较必须加 COLLATE，
+		// 与 internal/pipeline/fetch.go:141、internal/api/handler/candidates.go:177/190 保持一致。
+		var row struct {
+			ThreadName string `gorm:"column:thread_name"`
+			GroupName  string `gorm:"column:group_name"`
+		}
+		err := imDB.Raw(
+			"SELECT t.name AS thread_name, g.name AS group_name "+
+				"FROM thread t "+
+				"LEFT JOIN `group` g ON g.group_no COLLATE utf8mb4_unicode_ci = t.group_no "+
+				"WHERE t.group_no = ? AND t.short_id = ? "+
+				"LIMIT 1",
+			groupNo, shortID,
+		).Scan(&row).Error
+		if err != nil {
+			// 不再吞错：打日志便于线上定位；返回确定性可区分兜底（含 short_id）。
+			log.Printf("[service] ResolveSourceNameWithType: thread query failed source_id=%q: %v", sourceID, err)
+			return threadFallbackName("", shortID)
+		}
+
+		threadName := strings.TrimSpace(row.ThreadName)
+		groupName := strings.TrimSpace(row.GroupName)
+
+		// 兜底顺序：子区名优先；其次“父群名-占位”；最后纯占位。
+		// 关键：未命名子区永远带 short_id 片段，保证同群多子区互不相同。
+		if threadName != "" {
+			if groupName != "" {
 				return groupName + "-" + threadName + "(子区)"
 			}
-			if threadName != "" {
-				return threadName + "(子区)"
-			}
-			if groupName != "" {
-				return groupName + "(子区)"
-			}
+			return threadName + "(子区)"
 		}
+		return threadFallbackName(groupName, shortID)
 	case 3: // DM (private chat)
 		var name string
 		err := imDB.Table("user").Where("uid = ?", sourceID).Pluck("name", &name).Error
@@ -106,6 +128,20 @@ func fallbackSourceName(sourceID string, sourceType int) string {
 		return "来源-" + sourceID[:8] + suffix
 	}
 	return "来源-" + sourceID + suffix
+}
+
+// threadFallbackName 为“未命名 / 未命中”的子区生成确定性、可区分的占位名。
+// 永远包含 short_id 的前若干位，确保同一父群下的不同子区不会撞名
+// （这正是 #93 的根因：旧逻辑退化成纯父群名导致同群子区全部同名）。
+func threadFallbackName(groupName, shortID string) string {
+	short := shortID
+	if len(short) > 6 {
+		short = short[:6]
+	}
+	if groupName != "" {
+		return groupName + "-子区" + short + "(子区)"
+	}
+	return "子区-" + short + "(子区)"
 }
 
 // ResolveSourceName returns source name from IM DB, fallback to placeholder.
--- a/internal/worker/scheduled_replace_helpers.go
+++ b/internal/worker/scheduled_replace_helpers.go
@@ -63,10 +63,12 @@ func buildScheduledTaskSources(tx *gorm.DB, imDB *gorm.DB, taskID int64, raw mod
 		if src.SourceID == "" {
 			return fmt.Errorf("scheduled source_id is required")
 		}
-		sourceName := src.SourceName
-		if sourceName == "" {
-			sourceName = service.ResolveSourceNameWithType(src.SourceID, src.SourceType, imDB)
-		}
+		// Always resolve the canonical source name from the IM DB; never trust a
+		// client-supplied source_name (the schedule-management UI can submit a raw
+		// group_no/thread id as the "name", e.g. "groupNo____shortId"). Resolving
+		// unconditionally keeps the stored name consistent with the instant-summary
+		// path, which already drops source_name and lets the backend look it up.
+		sourceName := service.ResolveSourceNameWithType(src.SourceID, src.SourceType, imDB)
 		if err := tx.Create(&model.SummarySource{
 			TaskID:     taskID,
 			SourceType: src.SourceType,
@@ -156,10 +158,9 @@ func syncScheduledTaskSources(tx *gorm.DB, imDB *gorm.DB, taskID int64, raw mode
 		if src.SourceID == "" {
 			return fmt.Errorf("scheduled source_id is required")
 		}
-		sourceName := src.SourceName
-		if sourceName == "" {
-			sourceName = service.ResolveSourceNameWithType(src.SourceID, src.SourceType, imDB)
-		}
+		// Always resolve the canonical source name from the IM DB; never trust a
+		// client-supplied source_name (see buildScheduledTaskSources).
+		sourceName := service.ResolveSourceNameWithType(src.SourceID, src.SourceType, imDB)
 		if err := tx.Create(&model.SummarySource{
 			TaskID:     taskID,
 			SourceType: src.SourceType,
--- a/internal/worker/scheduled_replace_test.go
+++ b/internal/worker/scheduled_replace_test.go
@@ -167,21 +167,24 @@ func TestSyncScheduledTaskParticipants_AlwaysIncludesCreatorAndDedups(t *testing
 	}
 }
 
-// buildScheduledTaskSources must persist the source_name carried in the
-// schedule config verbatim (the fix for the dropped-name bug). It must NOT
-// degrade a config-provided real name into the fallback placeholder.
-func TestBuildScheduledTaskSources_UsesConfigSourceName(t *testing.T) {
+// buildScheduledTaskSources must ALWAYS re-resolve the source_name from the IM
+// DB and never trust the source_name carried in the schedule config (issue #93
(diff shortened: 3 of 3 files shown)


Jerry-Xin (approved):
Summary: The PR is relevant to this repository and correctly moves thread source-name resolution to a backend-authoritative `(group_no, short_id)` lookup while aligning scheduled summaries with the instant-summary path.

💬 Non-blocking

- 🟡 Warning: `internal/service/summary.go:136` truncates fallback thread IDs to 6 bytes. This makes fallback names shorter, but it does not strictly guarantee uniqueness for multiple unnamed or missing threads under the same parent group if two `short_id` values share the same prefix. Consider using the full `shortID`, or a longer bounded suffix, since `summary_source.source_name` has room for it.

- 🔵 Suggestion: `internal/worker/scheduled_replace_test.go:175` verifies that client-provided names are ignored, but it only covers the `imDB == nil` fallback path. A focused resolver test with two rows sharing the same `short_id` under different groups would lock in the main regression fix around the composite key lookup.

✅ Highlights

- The PR passes the relevance gate: it fixes behavior in the project’s summary source naming and scheduled summary persistence paths.
- `internal/service/summary.go:80` uses parameterized SQL, so the new JOIN does not introduce SQL injection risk.
- `internal/worker/scheduled_replace_helpers.go:71` and `:163` now consistently make the backend the source of truth for persisted `source_name`, matching the instant-summary creation path.
- Verification: `go test ./...` passes.

mochashanyao (approved):
[Octo-Q · automated review]

---

# octo-smart-summary PR#95 Review Report
**PR:** Mininglamp-OSS/octo-smart-summary #95
**Head SHA:** `d609417fa6afac34f16ce3b2695cc51904ab689c`
**Title:** fix(summary): correct thread source_name resolution & enforce backend-authoritative naming (#93)

---

## 1. Verification Summary

| Item | Status | Evidence |
|------|--------|----------|
| Thread source_name resolution via composite key `(group_no, short_id)` | ✅ | `internal/service/summary.go:65-100` — single JOIN query replaces two separate queries |
| COLLATE pattern consistency | ✅ | Matches `internal/pipeline/fetch.go:141` and `internal/api/handler/candidates.go:177/190` |
| Error no longer swallowed | ✅ | `summary.go:91-93` — `log.Printf` + deterministic fallback |
| Scheduled path always re-resolves source_name | ✅ | `scheduled_replace_helpers.go:71` and `:163` — both `build` and `sync` paths updated symmetrically |
| Test updated to assert re-resolution | ✅ | `scheduled_replace_test.go:170-204` — asserts dirty config value is NOT stored |
| Malformed source_id guard | ✅ | `summary.go:68-72` — `len(parts) != 2 || parts[0] == "" || parts[1] == ""` → deterministic fallback |

## 2. Findings

### F1 — No test coverage for thread path (case 2) of `ResolveSourceNameWithType` [P2]

**Diff-scope:** new (the thread path was substantially rewritten; `threadFallbackName` is brand new)

The rewritten `case 2` in `ResolveSourceNameWithType` (lines 63-100) and the new `threadFallbackName` function (lines 134-145) have **zero direct test coverage**. No `summary_test.go` exists in `internal/service/`. The only test exercising `ResolveSourceNameWithType` uses `sourceType=1` (group) with `imDB=nil` (fallback path only).

Key untested paths:
- The JOIN query returning both thread_name and group_name (happy path)
- The JOIN query returning no rows (thread not found → `threadFallbackName` with groupName)
- The JOIN query erroring (→ `threadFallbackName` with empty groupName)
- Malformed source_id (→ `fallbackSourceName`)
- `threadFallbackName` with short_id ≤ 6 chars vs > 6 chars

**Recommendation:** Add a `summary_test.go` with SQLite-backed tests for case 2, covering at minimum: happy path (both names found), not-found fallback, and malformed source_id. Note: the hard-coded `COLLATE utf8mb4_unicode_ci` will fail on SQLite — consider either gating it (see `candidates.go`'s `h.collate` pattern) or testing with a MySQL integration fixture.

### F2 — `threadFallbackName` byte-slices short_id without UTF-8 awareness [P2]

**Diff-scope:** new (new function)

`summary.go:139`: `short = short[:6]` operates on bytes, not runes. If `short_id` ever contains multi-byte UTF-8 characters (the column is `TEXT` in a `utf8mb4` database), this could produce invalid UTF-8 by slicing mid-rune.

**Mitigating factors:** All observed `short_id` values in test fixtures are pure ASCII (`thA`, `th001`, etc.). The pre-existing `fallbackSourceName` has the same pattern (`sourceID[:8]`), so this follows established convention.

**Recommendation:** Low priority. If short_id is guaranteed ASCII by the upstream IM system, document that invariant. Otherwise, use `[]rune(shortID)` slicing.

### F3 — Hard-coded COLLATE reduces testability vs `candidates.go` pattern [P2]

**Diff-scope:** new (new SQL in this PR)

`summary.go:83` hard-codes `COLLATE utf8mb4_unicode_ci` in the raw SQL string. `candidates.go` uses a configurable `h.collate` field (set to `" COLLATE utf8mb4_unicode_ci"` in production, `""` in tests) for SQLite compatibility. The hard-coded approach means any future unit test for `ResolveSourceNameWithType` case 2 with SQLite will fail on this SQL.

**Recommendation:** Consider extracting the COLLATE clause to a package-level variable or function parameter, mirroring the `candidates.go` approach, to enable SQLite-based testing.

## 3. Suggestions

1. **Add thread-path tests** (addresses F1): Even a minimal test with SQLite (after addressing F3) covering the happy path and the malformed-source_id guard would lock in the fix for #93.
2. **Consider COLLATE abstraction** (addresses F3): A simple `var threadCollate = " COLLATE utf8mb4_unicode_ci"` at package level, overridable in `_test.go` init, would unblock SQLite testing without the full handler-injection pattern.

## 4. Additional Observations

- **Behavioral change in thread lookup (correct):** Old code queried `WHERE short_id = ?` alone, which could match a thread from a different group with the same short_id. New code uses `(group_no, short_id)` composite — strictly more correct. This is the core fix for #93.
- **`SplitN` vs `Split` divergence (acceptable):** `summary.go` uses `SplitN(sourceID, "____", 2)` while `resolve_channel.go:298` uses unbounded `Split`. The comment at `summary.go:62-64` explicitly documents this design choice and the rationale (group source_ids can also contain `____`).
- **Scheduled path symmetry (correct):** Both `buildScheduledTaskSources` and `syncScheduledTaskSources` are updated identically — no asymmetric drift.

## 5. Data-Flow Trace

|---|---|---|
| `sourceID` (thread) | Schedule config JSON → `scheduleSourceConfig.SourceID` → `ResolveSourceNameWithType` param | ✅ Parsed by `SplitN("____", 2)` into `(groupNo, shortID)` |
| `groupNo` from split | `sourceID` left of `____` | ✅ Used in `WHERE t.group_no = ?` (parameterized, safe) |
| `shortID` from split | `sourceID` right of `____` | ✅ Used in `WHERE t.short_id = ?` (parameterized, safe) |
| `row.ThreadName` from JOIN | `thread.name` via `LEFT JOIN group` | ✅ TrimSpace → empty check → fallback chain |
| `row.GroupName` from JOIN | `group.name` via `LEFT JOIN` | ✅ TrimSpace → used in fallback naming |
| `sourceName` in scheduled path | `ResolveSourceNameWithType(src.SourceID, src.SourceType, imDB)` — always called, config value ignored | ✅ Symmetric in both `build` and `sync` functions |
| `imDB` nil guard | `ResolveSourceNameWithType` line 44: `if imDB == nil { return fallbackSourceName(...) }` | ✅ Prevents nil-pointer on DB calls; test relies on this |

## 6. R5 Blind-Spot Checklist

This PR is **not security_sensitive** (bug fix for name resolution, no auth/permission/jail changes). Checklist run for completeness:

- **C1 — Dual-path parity:** N/A. No add/remove or subscribe/unsubscribe symmetry involved. Both `build` and `sync` scheduled-source functions are updated identically — parity confirmed.
- **C2 — Control-flow ordering / nested reuse:** Clear. `ResolveSourceNameWithType` is called from 3 sites (2 scheduled + 1 instant). No nested/double invocation risk. No security controls to bypass.
- **C3 — Authorization boundary ≠ capability boundary:** N/A. No new endpoints, tools, or capabilities exposed. The `source_name` from client config is now ignored (defense in depth — correct direction).

---

**[Octo-Q] verdict: APPROVE**

All findings are P2 (test coverage gap, UTF-8 byte-slicing latent risk, COLLATE testability). No P0/P1 issues identified. The core fix is correct: composite-key thread lookup eliminates the #93 root cause (same-short_id cross-group collision), error swallowing is replaced with logging + deterministic fallback, and the scheduled path is symmetrically hardened against dirty client-supplied names. The test update correctly asserts the new "never trust config" contract.

yujiawei (approved):
# Code Review — PR #95 (octo-smart-summary)

**Scope reviewed:** head `d609417` against merge-base `cdecb3b` (one commit; 3 files, +82/-37). Build, vet, and full `go test ./...` pass locally.

## Verdict: APPROVED

This is a focused, well-reasoned fix for #93. Both changes are correct, and the thread-resolution rewrite is consistent with the existing cross-table query conventions in the repo. No blocking (P0/P1) issues found. A few non-blocking suggestions below.

---

## 1. Verification of the fix

### Problem 1 — thread (`source_type=2`) name resolution (`internal/service/summary.go`)
✅ **Correct.** The old logic queried `thread WHERE short_id = ?` as a single-key lookup, which can match a same-named `short_id` belonging to a *different* parent group. The new code uses the composite key `(group_no, short_id)` in one JOIN, with `COLLATE utf8mb4_unicode_ci` on the cross-table `group_no` comparison. This matches the established convention elsewhere:
- `internal/pipeline/fetch.go:140` — `INNER JOIN \`group\` g ON g.group_no COLLATE utf8mb4_unicode_ci = t.group_no`
- `internal/api/handler/candidates.go:177,190` — same collate clause via `h.collate`

✅ **Root cause of #93 addressed.** `threadFallbackName()` always embeds a `short_id` fragment, so multiple unnamed sub-threads under the same parent group no longer collapse to an identical name. The old fallback chain degraded to the bare parent-group name, which was the reported bug.

✅ **Type-driven parsing.** Keying the branch strictly on `source_type==2` (rather than inferring from the presence of `____`) is the right call — `case 1` (group) source IDs can also carry a `____{space_id}` suffix, so the separator alone is ambiguous.

✅ **Error handling improved.** Replacing the silenced `_ =` with `log.Printf` on query failure aids production diagnosis, and malformed/empty parts now route to a deterministic fallback instead of degrading to the parent-group name.

### Problem 2 — backend-authoritative naming for scheduled tasks (`internal/worker/scheduled_replace_helpers.go`)
✅ **Correct and consistent.** `buildScheduledTaskSources` / `syncScheduledTaskSources` now always call `ResolveSourceNameWithType` instead of trusting the client-supplied `source_name`. I verified this aligns the scheduled path with the instant-summary path, which already ignores the client value and re-resolves at `internal/api/handler/task.go:254`.

⚠️ **Note — this intentionally reverses the #89/#91 behavior** (which was "preserve the config-supplied real name"). The reversal is justified: the IM DB is the authoritative source, the schedule-management UI can submit dirty values (`groupNo____shortId` as a "name"), and consistency with the instant path is the explicit goal. The `imDB` is correctly threaded through in production (`scheduler.go:219 → buildScheduledTaskChildren → buildScheduledTaskSources`, wired from `cmd/summary-worker/main.go:60,86`), so this does not reintroduce the `nil`-imDB → placeholder regression that #89/#91 fixed.

### Tests (`internal/worker/scheduled_replace_test.go`)
✅ `TestBuildScheduledTaskSources_IgnoresConfigSourceName` correctly asserts the stored name is the re-resolved value and **not** the dirty config value. The expected fallback `来源-group-ab(群聊)` matches `fallbackSourceName` (first 8 chars of `group-abcdef123456` + `(群聊)`) under `imDB=nil`.

---

## 2. Non-blocking suggestions (P2 / nit)

- **P2 — no unit test for the core `summary.go` thread-resolution path.** The most important change (the `(group_no, short_id)` JOIN and `threadFallbackName` fallback ordering) has no direct test. The worker test only exercises the `imDB=nil` fallback. Consider a test with a stubbed `thread`/`group` table (SQLite, dropping the COLLATE clause as `candidates_test.go` already does via `h.collate=""`) covering: named thread + named group, named thread + missing group, and unnamed thread (fallback contains `short_id`).
- **nit — `threadFallbackName` truncates `short_id` to 6 chars.** Two distinct sub-threads whose `short_id` share the same first 6 characters would still collide in the fallback name. Low likelihood, and the result is still deterministic, but worth a comment or using the full `short_id` if length permits.
- **nit — `scheduleSourceConfig.SourceName` is now effectively unused** for persistence (still unmarshalled, never stored). Harmless and fine to keep for JSON compatibility; just flagging it as dead-ish.

---

## 3. CI status (informational — not blocking this review)

Build / Lint / Test / Vet all pass. Three failing checks are environmental/policy, unrelated to this diff:
- `secret-scan (gitleaks)` — flags `internal/service/llm_test.go:137` (`NewLLMClient("http://localhost", "key", ...)`), a **pre-existing** test fixture in a file this PR does not touch (present unchanged on the merge-base). False positive; consider a `.gitleaks.toml` allowlist or a `gitleaks:allow` annotation in a separate change.
- `dependency-review` — "Dependency review is not supported on this repository" (Dependency graph disabled in repo settings); no `go.mod`/`go.sum` changes in this PR anyway.
- `check-sprint` — `PROJECT_TOKEN is invalid or expired`; project-board automation, not code.

None of these reflect a defect introduced by PR #95.

PR #99, opened 2026-06-17
feat: expose creator_id in task detail + list DTOs (OCT-35 part 1)

lml2468:
## Summary
- Adds `creator_id` to both the `GetSummary` detail DTO and the `ListSummaries` list DTO (`internal/api/handler/task.go`). The model already persisted `creator_id` (`internal/model/model.go:104`); it was simply never serialized.
- Additive, non-breaking contract change. Existing fields and consumers are unaffected.

## Why
Split from backend review **OCT-32** of octo-web [#291](https://github.com/Mininglamp-OSS/octo-web/pull/291) (summary-notify). The client-only tip had no way to attribute the real author and fell back to the *observing* user's login. Exposing `creator_id` lets the client stamp the correct author and gate notify behavior on it.

This is **part 1** of OCT-35. The remaining work (server-side notify-policy gate + idempotent emit on the `COMPLETED` transition) is tracked separately because it needs a new IM outbound path and cross-repo coordination on sender identity + `contentType=21` payload.

## Test plan
- [x] `go build ./...`
- [x] `go vet ./internal/api/handler/`
- [x] `go test ./internal/api/handler/` (new `creator_id_test.go`: detail + list DTO presence, asserting a non-creator participant observer still sees the real `creator_id`)
- [ ] CI green

Co-Authored-By: Paperclip <noreply@paperclip.ing>

PR #100, opened 2026-06-17
feat(notify): server-side notify-policy gate + idempotent summary-notify emit (OCT-43)

lml2468:
## Summary

Server-side emit for the summary-notify tip — OCT-35 part 2/3, following part 1 (`creator_id` in DTOs, #99). Resolves all three defects from octo-web #291:

- **wrong author** → tip is sent by a stable system App-Bot (`from_uid`/`from_name`), not an arbitrary member.
- **channel-wide leak of a private summary** → tip goes only to the access-gated target set (creator + explicit participants) as directed DMs; never broadcast to `origin_channel_id`.
- **N-member duplicate fan-out** → exactly-once per task via a `notified_at` CAS anchor.

## What changed

- `internal/worker/notify.go` (new): `Notifier` + `botNotifier`.
  - `notifyTargetUserIDs` = creator + participants, deduped — mirrors `handler.canAccessTask` (channel membership alone grants no tip).
  - `NotifyCompleted` does a single-writer CAS `notified_at IS NULL -> now`; only the row-affecting winner sends → exactly-once even across restarts / concurrent winners.
  - `botAPISender` POSTs to octo-server `POST /v1/bot/sendMessage` (`Authorization: Bearer app_*`) with a `contentType=21` payload.
  - No-op until configured (`config.NotifyConfigured()`), so an unconfigured deploy never burns the `notified_at` anchor.
- Emit wired into the CAS-win branch of both completion paths: `personal_processor.go` (single-person) and `meta_processor.go` (multi-person, idempotent across dirty re-runs).
- `summary_task.notified_at` column + migration `20260618-01-add-summary-task-notified-at.sql`.
- Config: `SUMMARY_NOTIFY_ENABLED`, `SUMMARY_NOTIFY_BOT_API_URL`, `SUMMARY_NOTIFY_BOT_TOKEN`, `SUMMARY_NOTIFY_BOT_UID`, `SUMMARY_NOTIFY_BOT_NAME`.

## Tests

- `TestNotifyPolicyGate` — creator-only / creator+participants / creator-deduped; asserts a non-participant is never targeted and every emit is a person DM with `type=21`.
- `TestNotifyIdempotencyConcurrent` — 16 goroutines → single fan-out; race-clean.
- `TestNotifyAlreadyNotifiedNoop` — already-set `notified_at` → no re-emit.
- Full `internal/worker` package + `go vet` green.

## ⚠️ Draft — blocked on before-merge coordination

1. **Sender identity (CTO/infra):** provision the system App-Bot (`robotID` + `app_*` token) + token distribution — decision #1.
2. **`contentType=21` payload contract (Frontend):** confirm exact required keys vs. what octo-web #291 registered. `buildPayload` keys are proposed (`type`, `from_uid`, `from_name`, `task_id`, `task_no`, `title`) and flagged in code.
3. **Security review (SecurityEngineer):** new authenticated outbound sender posting into user channels with access-gating.

Refs OCT-43.

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -84,6 +84,23 @@ type Config struct {
 
 	// Tool call per-attempt timeout (seconds)
 	ToolCallTimeout int
+
+	// Summary-notify outbound (OCT-43). When a task completes the worker emits
+	// a directed contentType=21 tip to the creator + explicit participants via
+	// octo-server's bot API. Disabled unless NotifyEnabled is true AND both URL
+	// and token are set (see NotifyConfigured).
+	NotifyEnabled   bool   // SUMMARY_NOTIFY_ENABLED master switch
+	NotifyBotAPIURL string // SUMMARY_NOTIFY_BOT_API_URL full endpoint, e.g. http://octo-server:8080/v1/bot/sendMessage
+	NotifyBotToken  string // SUMMARY_NOTIFY_BOT_TOKEN system App-Bot token (app_ prefix) — provisioning is OCT-43 decision #1
+	NotifyBotUID    string // SUMMARY_NOTIFY_BOT_UID stable from_uid surfaced in the payload (server stamps the authoritative from_uid from the token)
+	NotifyBotName   string // SUMMARY_NOTIFY_BOT_NAME stable from_name display string
+}
+
+// NotifyConfigured reports whether the summary-notify emit path is fully
+// configured and enabled. The worker builds a no-op notifier otherwise, so an
+// unconfigured deploy never burns the notified_at idempotency anchor.
+func (c *Config) NotifyConfigured() bool {
+	return c.NotifyEnabled && c.NotifyBotAPIURL != "" && c.NotifyBotToken != ""
 }
 
 func Load() *Config {
@@ -133,6 +150,12 @@ func Load() *Config {
 		ChannelScopeEnabled: envBool("CHANNEL_SCOPE_ENABLED", true),
 
 		ToolCallTimeout: envInt("TOOL_CALL_TIMEOUT", 30),
+
+		NotifyEnabled:   envBool("SUMMARY_NOTIFY_ENABLED", false),
+		NotifyBotAPIURL: envStr("SUMMARY_NOTIFY_BOT_API_URL", ""),
+		NotifyBotToken:  envStr("SUMMARY_NOTIFY_BOT_TOKEN", ""),
+		NotifyBotUID:    envStr("SUMMARY_NOTIFY_BOT_UID", ""),
+		NotifyBotName:   envStr("SUMMARY_NOTIFY_BOT_NAME", "智能总结"),
 	}
 }
 
--- a/internal/model/model.go
+++ b/internal/model/model.go
@@ -115,9 +115,14 @@ type SummaryTask struct {
 	OriginChannelType  int        `gorm:"column:origin_channel_type;type:tinyint;not null;default:0" json:"origin_channel_type"`
 	ProcessingDeadline *time.Time `gorm:"column:processing_deadline" json:"processing_deadline"`
 	ConfirmDeadline    *time.Time `gorm:"column:confirm_deadline" json:"confirm_deadline"`
-	CreatedAt          time.Time  `gorm:"column:created_at;not null" json:"created_at"`
-	UpdatedAt          time.Time  `gorm:"column:updated_at;not null" json:"updated_at"`
-	DeletedAt          *time.Time `gorm:"column:deleted_at;index" json:"deleted_at,omitempty"`
+	// NotifiedAt is the crash-safe idempotency anchor for the summary-notify
+	// tip. A single-writer CAS (notified_at IS NULL → now) guarantees
+	// exactly-once emit per task even across worker restarts. NULL = not yet
+	// notified.
+	NotifiedAt *time.Time `gorm:"column:notified_at" json:"notified_at,omitempty"`
+	CreatedAt  time.Time  `gorm:"column:created_at;not null" json:"created_at"`
+	UpdatedAt  time.Time  `gorm:"column:updated_at;not null" json:"updated_at"`
+	DeletedAt  *time.Time `gorm:"column:deleted_at;index" json:"deleted_at,omitempty"`
 }
 
 func (SummaryTask) TableName() string { return "summary_task" }
--- a/internal/worker/meta_processor.go
+++ b/internal/worker/meta_processor.go
@@ -217,6 +217,10 @@ func (m *MetaProcessor) processMetaSummary(ctx context.Context, taskID int64) {
 			EventType: "META_SUMMARY_UPDATED",
 		})
 
+		// Emit the access-gated summary-notify tip (OCT-43). Idempotent via the
+		// notified_at CAS, so dirty re-runs of this loop never double-emit.
+		m.proc.emitSummaryNotify(taskID)
+
 		log.Printf("[meta-worker] task %d meta-summary version %d created (%d participants)",
 			taskID, result.Version, len(submitted))
 
--- a/internal/worker/notify.go
+++ b/internal/worker/notify.go
@@ -1,0 +1,203 @@
+package worker
+
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"io"
+	"log"
+	"net/http"
+	"time"
+
+	"github.com/Mininglamp-OSS/octo-smart-summary/internal/config"
+	"github.com/Mininglamp-OSS/octo-smart-summary/internal/model"
+	"gorm.io/gorm"
+)
+
+// contentTypeSummaryNotify is the IM message content type the octo-web client
+// (#291) registers to render the summary-notify tip. OCT-43 decision #2
+// (Frontend) owns the authoritative payload key contract for this type.
+const contentTypeSummaryNotify = 21
+
+// channelTypePerson is octo-server's DM (person) channel type
+// (octo-lib common.ChannelTypePerson). The tip is delivered as a directed DM
+// to each authorized user, never broadcast to the origin channel.
+const channelTypePerson = 1
+
+// Notifier emits the summary-completed tip for a finished task. Implementations
+// must be idempotent: NotifyCompleted may be invoked more than once for the
+// same task (concurrent CAS winners, restarts) and must emit at most once.
+type Notifier interface {
+	NotifyCompleted(taskID int64)
+}
+
+// noopNotifier is installed when summary-notify is unconfigured. It does
+// nothing — crucially it does NOT touch notified_at, so enabling the feature
+// later can still notify previously-completed tasks.
+type noopNotifier struct{}
+
+func (noopNotifier) NotifyCompleted(int64) {}
+
+// emitSummaryNotify fires the summary-completed tip for a task that this
+// goroutine just transitioned to Completed. Nil-safe so Processors built via
+// struct literal (tests) don't panic.
+func (p *Processor) emitSummaryNotify(taskID int64) {
+	if p.notifier == nil {
+		return
+	}
+	p.notifier.NotifyCompleted(taskID)
+}
+
+// messageSender is the outbound transport to octo-server's bot API. Extracted
+// so tests can assert target set + payload without a live HTTP server.
+type messageSender interface {
+	Send(channelID string, channelType int, payload map[string]interface{}) error
+}
+
+// botNotifier emits a directed contentType=21 tip to the access-gated target
+// set (creator + explicit participants) of a completed task.
+type botNotifier struct {
+	db       *gorm.DB
+	sender   messageSender
+	fromUID  string
+	fromName string
+}
+
+// NewNotifier builds the summary-notify emitter. Returns a no-op when the
+// feature is not fully configured (see config.NotifyConfigured).
+func NewNotifier(db *gorm.DB, cfg *config.Config) Notifier {
+	if !cfg.NotifyConfigured() {
+		log.Printf("[notify] summary-notify disabled (not configured); emit is a no-op")
+		return noopNotifier{}
+	}
+	return &botNotifier{
+		db: db,
+		sender: &botAPISender{
+			url:    cfg.NotifyBotAPIURL,
+			token:  cfg.NotifyBotToken,
+			client: &http.Client{Timeout: 5 * time.Second},
+		},
+		fromUID:  cfg.NotifyBotUID,
+		fromName: cfg.NotifyBotName,
+	}
+}
+
+// NotifyCompleted emits the tip exactly once per task. The notified_at CAS
+// (notified_at IS NULL → now) is the idempotency anchor: only the single
+// goroutine whose CAS affects one row proceeds to send, so concurrent callers
+// and post-restart retries are deduplicated even across processes.
+func (n *botNotifier) NotifyCompleted(taskID int64) {
+	now := time.Now().UTC()
+	cas := n.db.Model(&model.SummaryTask{}).
+		Where("id = ? AND notified_at IS NULL", taskID).
+		Update("notified_at", now)
+	if cas.Error != nil {
+		log.Printf("[notify] task %d notified_at CAS error: %v", taskID, cas.Error)
+		return
+	}
+	if cas.RowsAffected != 1 {
+		// Already notified (lost the CAS race or replayed after restart).
+		return
+	}
+
+	var task model.SummaryTask
+	if err := n.db.Where("id = ?", taskID).First(&task).Error; err != nil {
+		log.Printf("[notify] task %d load failed after CAS: %v", taskID, err)
+		return
+	}
+
+	targets := notifyTargetUserIDs(n.db, &task)
+	payload := n.buildPayload(&task)
+	sent := 0
+	for _, uid := range targets {
+		if err := n.sender.Send(uid, channelTypePerson, payload); err != nil {
+			// Best-effort per recipient: one failed DM must not block the rest.
+			log.Printf("[notify] task %d send to %s failed: %v", taskID, uid, err)
+			continue
+		}
+		sent++
+	}
+	log.Printf("[notify] task %d emitted summary-notify: %d/%d targets", taskID, sent, len(targets))
+}
+
+// notifyTargetUserIDs returns the access-gated notify target set for a task:
+// the creator plus every explicit participant, deduplicated and creator-first.
+// This mirrors handler.canAccessTask — origin-channel membership alone does NOT
+// receive a tip, which is what prevents the channel-wide leak (octo-web #291).
+func notifyTargetUserIDs(db *gorm.DB, task *model.SummaryTask) []string {
+	seen := make(map[string]struct{})
+	var targets []string
+	add := func(uid string) {
+		if uid == "" {
+			return
+		}
+		if _, ok := seen[uid]; ok {
+			return
+		}
+		seen[uid] = struct{}{}
+		targets = append(targets, uid)
+	}
+
+	add(task.CreatorID)
+
+	var participants []model.SummaryParticipant
+	db.Where("task_id = ?", task.ID).Find(&participants)
+	for _, p := range participants {
+		add(p.UserID)
+	}
+	return targets
+}
+
+// buildPayload constructs the contentType=21 tip payload.
+//
+// OCT-43 decision #2 (Frontend): these keys are the proposed contract; the
+// authoritative shape is whatever octo-web #291 registered client-side. Confirm
+// the exact required keys (link / task ref form) before merge. from_uid /
+// from_name are surfaced here for the renderer even though octo-server stamps
+// the authoritative from_uid from the bot token.
+func (n *botNotifier) buildPayload(task *model.SummaryTask) map[string]interface{} {
+	return map[string]interface{}{
+		"type":      contentTypeSummaryNotify,
+		"from_uid":  n.fromUID,
+		"from_name": n.fromName,
+		"task_id":   task.ID,
+		"task_no":   task.TaskNo,
+		"title":     task.Title,
+	}
+}
+
+// botAPISender posts to octo-server's POST /v1/bot/sendMessage as a system
+// App-Bot. octo-server stamps from_uid = the authenticated bot.
+type botAPISender struct {
+	url    string
+	token  string
+	client *http.Client
+}
+
+func (s *botAPISender) Send(channelID string, channelType int, payload map[string]interface{}) error {
+	body, err := json.Marshal(map[string]interface{}{
+		"channel_id":   channelID,
+		"channel_type": channelType,
+		"payload":      payload,
+	})
+	if err != nil {
+		return err
+	}
+	req, err := http.NewRequest(http.MethodPost, s.url, bytes.NewReader(body))
+	if err != nil {
+		return err
+	}
+	req.Header.Set("Content-Type", "application/json")
+	req.Header.Set("Authorization", "Bearer "+s.token)
+
+	resp, err := s.client.Do(req)
+	if err != nil {
+		return err
+	}
+	defer resp.Body.Close()
+	if resp.StatusCode < 200 || resp.StatusCode >= 300 {
+		b, _ := io.ReadAll(io.LimitReader(resp.Body, 512))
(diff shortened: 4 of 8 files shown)


lml2468:
## Frontend contract confirmation — contentType=21 (summary-notify)

Confirming against the actual renderer registered client-side in octo-web [#291](https://github.com/Mininglamp-OSS/octo-web/pull/291) (`SummaryNotifyContent`, `packages/dmworkbase/src/Messages/SummaryNotify/index.tsx`). This is the authoritative web parser for `type=21`.

### Keys the web client actually parses

`decodeJSON` reads **only** two keys:

```js
decodeJSON(content) {
  this.fromUID  = content["from_uid"]
  this.fromName = content["from_name"]
}
```

And `encodeJSON` emits exactly:

```json
{ "type": 21, "from_uid": "...", "from_name": "..." }
```

So the minimal payload the web renderer needs is:

```json
{
  "type": 21,
  "from_uid": "<stable system/actor uid>",
  "from_name": "<display-name fallback>"
}
```

`type` is the WuKongIM content-type discriminator (must be `21`).

### Answers to your questions

1. **Are `task_id`, `task_no`, `title` the keys the client parses?** — No. The web renderer **ignores** them. Extra keys are harmless (JSON decode won't error), but nothing is rendered from them. Safe to keep for forward-compat / mobile, but **not required by web** and **not currently consumed**.
2. **Is any required key missing?** — No required key is missing for web. `from_uid` + `from_name` is sufficient.
3. **Exact form of the task reference / deep-link the client expects** — **There is none today.** The web cell renders a single grey system tip and nothing else:
   - `<name> summarized the chat` / `<name>总结了群聊内容`
   - No task link, no deep-link, no summary-version rendering. If we want web to deep-link into the summary from this tip, that is a **follow-up octo-web change** (#291 does not implement it) — flag it and I'll spec/implement it. Don't block OCT-43 on it.

### `from_uid` / `from_name` semantics (please match)

Name resolution in the renderer:
- if `from_uid === <recipient's own login uid>` → renders **"You summarized the chat"**;
- else it looks up the **person-channel** info for `from_uid` and uses `orgData.displayName`;
- if no channel info is found, it falls back to **`from_name`**.

Implications for the stable system-bot identity:
- Prefer a `from_uid` that resolves to a real person/bot channel so the display name is consistent; otherwise it will fall back to whatever string you put in `from_name`.
- Since #100 sends a **directed DM to each authorized user**, if a recipient is also the summary creator and you set `from_uid` to their uid, that user sees "You summarized the chat". If you instead use a fixed system-bot `from_uid`, everyone sees the bot/creator name via `from_name`. Pick intentionally.

### One coordination flag (non-blocking)

#291 is a **client-initiated** tip sent to **group/thread sources** (it skips DM sources). #100 is a **server-initiated directed DM** to each authorized user. Both render through the same `type=21` cell, so rendering is fine — but if both ship, users could be **double-notified** (group system tip + directed DM). Worth deciding which path is authoritative before merge. Happy to disable/guard the client-side group tip in #291 if the server DM becomes the source of truth.

— Frontend

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# octo-smart-summary PR#100 Review Report
**PR:** https://github.com/Mininglamp-OSS/octo-smart-summary/pull/100
**Head SHA:** `1960350573ca7bcbed497cdd24155a97cd0a87c5`
**Title:** feat(notify): server-side notify-policy gate + idempotent summary-notify emit (OCT-43)
**Routing:** complexity=security_sensitive (automated review)

---

## 1. Verification Summary

| Item | Status | Evidence |
|------|--------|----------|
| Access-gated target set mirrors `canAccessTask` | ✅ | `notifyTargetUserIDs` (notify.go:127-149) = creator + participants, matches `canAccessTask` (task.go:41-55) |
| CAS idempotency anchor (`notified_at IS NULL → now`) | ✅ | notify.go:97-106; `TestNotifyIdempotencyConcurrent` 16 goroutines → exactly 1 fan-out, race-clean |
| No-op when unconfigured (preserves anchor) | ✅ | `NotifyConfigured()` (config.go:97-99) gates `NewNotifier` → returns `noopNotifier` |
| Emit wired into CAS-win branches only | ✅ | personal_processor.go:223 (after `saveLatestResultAndCompleteTask` / `completeTaskWithoutNewResult`); meta_processor.go:220 (after `saveLatestResultAndCompleteTask`) |
| DM-only delivery (no channel broadcast) | ✅ | `channelTypePerson = 1` used in all `Send` calls (notify.go:113); asserted in `TestNotifyPolicyGate` |
| Bot token not leaked in logs | ✅ | Token stored in `botAPISender.token`, used only in `Authorization: Bearer` header (notify.go:190); no struct-level logging |
| Migration matches model | ✅ | `20260618-01-add-summary-task-notified-at.sql` adds `notified_at DATETIME NULL`; model.go `NotifiedAt *time.Time` |
| `go vet` clean | ✅ | `go vet ./internal/worker/` → exit 0 |
| Tests pass (race detector) | ✅ | `go test -v -race -run TestNotify ./internal/worker/` → all 3 tests PASS |

## 2. Findings

### F1 — `notifyTargetUserIDs` ignores DB error on participants query [P2]

**File:** `internal/worker/notify.go:146`
**Code:**
```go
db.Where("task_id = ?", task.ID).Find(&participants)
for _, p := range participants {
    add(p.UserID)
}
```

**Issue:** The `Find` call does not check the returned error. If the query fails (transient DB error, connection timeout), `participants` remains nil and only the creator receives the notification. All other participants are silently dropped.

**Diff-scope (R2):**
1. **New** — this function is introduced by this PR.
2. The PR's diff creates this entire file; there is no pre-existing equivalent.
3. N/A (new code).

**Severity (R1):** Does this make a working path unavailable or produce user-visible wrong data? Partially — some users who should be notified are silently skipped. However: (a) the notification is a best-effort tip (the summary is still accessible via the UI), (b) the send loop itself already treats per-recipient failures as best-effort (notify.go:112-115), establishing a pattern where partial delivery is acceptable. The gap is the **silent** nature — no log line records the participant-query failure.

**Severity: P2** — should-fix before merge (add error check + log), but not blocking.

**Suggested fix:**
```go
if err := db.Where("task_id = ?", task.ID).Find(&participants).Error; err != nil {
    log.Printf("[notify] task %d participant query failed: %v", task.ID, err)
}
```

### F2 — Scheduled empty-window path still emits notify tip [Note]

**File:** `internal/worker/personal_processor.go:223`

When `isScheduledEmptyWindow` is true (scheduled trigger found no new messages), the task completes via `completeTaskWithoutNewResult` and `emitSummaryNotify` still fires. The user receives a "总结完成" notification but no new summary content was generated (the previous result is retained).

**Diff-scope:** New behavior introduced by this PR (emit call is new).

**Severity:** Not blocking — the task did complete processing, and the notification accurately reflects that. The UX question (notify vs. suppress for no-new-content scheduled runs) is a product decision, not a defect. Flag for product discussion.

## 3. Suggestions

1. **F1 fix** — add error check on the participants `Find` call in `notifyTargetUserIDs` (one-liner, see above).
2. **Consider** logging the target count before the send loop for easier debugging: `log.Printf("[notify] task %d targets: %v", taskID, targets)` — currently only the post-send summary line exists.
3. **Product decision** — confirm whether scheduled empty-window runs should suppress the notify tip (F2).

## 4. Additional Observations

- **No retry for partial send failures:** Once the `notified_at` CAS succeeds, the anchor is consumed. If 2 of 3 recipients fail to receive the DM, there is no retry mechanism. This is an acceptable trade-off (avoiding duplicate notifications is more important than guaranteed delivery for a tip), but worth documenting as a known limitation.
- **Test concurrency model:** `TestNotifyIdempotencyConcurrent` uses SQLite with `SetMaxOpenConns(1)`, which serializes all DB operations. This validates the CAS logic but doesn't exercise true DB-level concurrency. Production MySQL behavior is correctly handled by the `WHERE notified_at IS NULL` atomic UPDATE, so this is a test-infrastructure limitation, not a gap.
- **Payload contract (OCT-43 decision #2):** The `buildPayload` keys (`type`, `from_uid`, `from_name`, `task_id`, `task_no`, `title`) are flagged as proposed. Frontend confirmation is a pre-merge coordination item, not a code review concern.

## 5. Data-Flow Tracing

| Consumed Data | Upstream Source | Traced To | Verified |
|---|---|---|---|
| `task.CreatorID` | `SummaryTask.creator_id` column | DB load post-CAS (notify.go:108-110) | ✅ Flows from task creation → DB → reload |
| `task.ID` | `SummaryTask.id` (auto-increment PK) | CAS WHERE clause + post-CAS load | ✅ |
| `task.TaskNo`, `task.Title` | Same DB load | `buildPayload` (notify.go:163-170) | ✅ |
| `participants[].UserID` | `SummaryParticipant.user_id` column | `notifyTargetUserIDs` query (notify.go:146) | ⚠️ Error not checked (F1) |
| `cas.RowsAffected` | GORM `Update` result | Gate for send loop (notify.go:104) | ✅ Atomic CAS: only 1 winner |
| `cfg.NotifyBotToken` | Env `SUMMARY_NOTIFY_BOT_TOKEN` | `botAPISender.token` → `Authorization: Bearer` header | ✅ Not logged |
| `cfg.NotifyBotUID` / `NotifyBotName` | Env vars | `buildPayload` `from_uid`/`from_name` | ✅ Server stamps authoritative from_uid |
| `notifier` field on `Processor` | `NewNotifier(db, cfg)` in `NewProcessor` (processor.go:50) | `emitSummaryNotify` nil-safe dispatch (notify.go:48-52) | ✅ No-op when unconfigured |

## 6. R5 Blindspot Checklist (security_sensitive)

**C1 — Dual-path parity:** ✅ Clear
- Notify has no add/remove or subscribe/unsubscribe symmetry requirement. The emit is a one-shot per task.
- Both completion paths (personal single-person + meta multi-person) wire the emit identically after their respective CAS-win completions.

**C2 — Control-flow ordering / nested reuse:** ✅ Clear
- `emitSummaryNotify` is called from exactly 2 sites, both post-completion-CAS. No nested/double-call risk.
- The meta_processor's dirty-rerun loop calls emit inside the loop, but the `notified_at` CAS correctly deduplicates: only the first iteration's CAS succeeds.
- `botAPISender.Send` is a leaf function with no re-entrancy.

**C3 — Authorization boundary ≠ capability boundary:** ✅ Clear
- The bot API endpoint (`POST /v1/bot/sendMessage`) is authenticated via `Bearer app_*` token. Only the worker process holds this token.
- No new user-facing endpoint is exposed. The notification target set is access-gated (creator + participants only), mirroring `canAccessTask`.
- The bot sends DMs (`channelTypePerson=1`), never broadcasts to the origin channel — preventing the channel-wide leak that OCT-43 addresses.

**C4 — Authorization lifecycle / container-member state cascade:** ✅ N/A
- This PR does not modify any authorization or access-control logic. It adds a notification emit path that *consumes* the existing access-control set.
- The target set is derived from `SummaryTask.CreatorID` + `SummaryParticipant` rows. If a user is removed from the task (participant row deleted), they are correctly excluded from notifications.
- Space/org-level container status is not relevant here: the notification is a best-effort tip, not a data-access gate.

## 7. Cross-Round Blocker Recheck (R6)

---

**[Octo-Q] verdict: APPROVE**

No P0/P1 findings. One P2 should-fix (F1: add error check on participant query) and one product note (F2: scheduled empty-window notify behavior). The access-gating, CAS idempotency, and security posture are sound. The P2 is a one-liner fix that should be addressed before merge but does not block approval.

lml2468:
## Security review — changes requested before merge (OCT-48)

Reviewed `internal/worker/notify.go` + wiring against the four asks in the tracking issue. The idempotency, sender-auth handling, and payload shape are sound; one access-control gap must be fixed before merge, plus two hardening items to schedule.

### Must-fix before merge

**1. Notify path doesn't honor soft-delete — diverges from `canAccessTask`.**
The PR states `notifyTargetUserIDs` mirrors `handler.canAccessTask`, but the real gate is two-part: `authorizeTaskAccess` first loads the task with `WHERE id = ? AND deleted_at IS NULL` (soft-deleted → 404), *then* applies creator-or-participant. The emit replicates only the second half. Because `SummaryTask.DeletedAt` is a plain `*time.Time` (not `gorm.DeletedAt`), GORM does **no** automatic soft-delete scoping — every query must add the clause by hand (the handler does; the notify CAS/load do not). Net effect: a task soft-deleted before/during completion still DMs its title + task ref to creator+participants, even though the API treats it as nonexistent. Failure of complete mediation / fail-closed.

Fix — add the predicate to the CAS so a deleted task neither wins the CAS nor burns the anchor:
```go
cas := n.db.Model(&model.SummaryTask{}).
    Where("id = ? AND notified_at IS NULL AND deleted_at IS NULL", taskID).
    Update("notified_at", now)
```
and, defense in depth, bail after the load if `task.DeletedAt != nil`. Ship with a `TestNotifyDeletedTaskNoop` regression (0 sends, `notified_at` stays NULL).

### Schedule (hardening)

**2. Privileged bearer token over cleartext transport.** The config example normalizes `http://octo-server:8080/...`. The `app_*` system-bot token (can DM any user as the bot) would cross the network in cleartext. Require `https://` (or validate the scheme and refuse non-loopback `http://`), or document an explicit trusted-mesh/mTLS requirement. Token scope is OCT-43 decision #1 — flag for least-privilege provisioning (contentType=21 person-DM send only, server-side rate limit).

**3. Synchronous, uncapped fan-out can stall the worker.** `emitSummaryNotify` runs inline in the completion path and `NotifyCompleted` sends sequentially, 5s timeout each, with no cap on `len(targets)`. A large participant set or a slow/hung octo-server blocks summary processing for up to N×5s. Cap the target count and/or emit with an overall deadline (or off the processing goroutine).

### Confirmed OK
- **Idempotency:** single-writer `notified_at` CAS with `RowsAffected==1` gate is correct; 16-goroutine test is solid. CAS-before-send is fail-closed (no amplification/replay) — losing a notification on a transient octo-server outage is the accepted tradeoff.
- **Token handling:** never logged; request headers not logged; error path echoes only the bounded (512B) response body, not the token.
- **Payload:** no secrets — `task_id`/`task_no`/`title` + bot identity only, to access-gated recipients. `from_uid` is client-surfaced but server stamps the authoritative value from the token (please confirm octo-server ignores the client-supplied `from_uid`).
- **SSRF/injection:** endpoint is operator env config (not request-derived); gorm is parameterized; payload is `json.Marshal`-encoded. No injection vector.

Tracking: internal OCT-48 (review) / OCT-59 (remediation). Happy to re-review once #1 lands.

yujiawei (changes requested):
# Code Review — PR #100 (octo-smart-summary)

**Scope:** independent review of the server-side summary-notify emit path. Reviewed at head SHA `1960350573ca7bcbed497cdd24155a97cd0a87c5` against merge-base `main`. This PR is a **Draft** and was routed as security-sensitive (new authenticated outbound sender posting into user channels), so this review doubles as the security check.

**Verdict: Changes requested** — not because the design is wrong (it is sound), but because the change is not mergeable as-is: the `contentType=21` payload contract is still unconfirmed, and a few robustness items on the new outbound path are worth addressing before this is enabled in production. None of the findings is a P0.

---

## 1. Verification summary

| Area | Result | Evidence |
|---|---|---|
| Access-gating (no channel-wide leak) | ✅ Correct | `notifyTargetUserIDs` (`internal/worker/notify.go:127-149`) = creator + explicit participants, deduped — set-enumeration of the same predicate as `canAccessTask` (`internal/api/handler/task.go:45-54`). The only `Send` call site (`notify.go:113`) uses `channelTypePerson` (=1) with a per-user `uid`; `OriginChannelID` never reaches a send. No broadcast path exists. |
| Idempotency / no double-emit | ✅ Correct | `notified_at` CAS (`notify.go:91-101`) is a correct single-writer anchor under InnoDB: PK + conditional `notified_at IS NULL`, gated on `RowsAffected == 1`. Holds across goroutines, processes, restarts. |
| Build / vet / tests | ✅ Green | `go build ./...`, `go vet ./internal/worker/`, and `go test ./internal/worker/ -run TestNotify` all pass locally. |
| Migration | ✅ Consistent | `20260618-01-add-summary-task-notified-at.sql` matches existing Up/Down `+migrate` style; nullable column, reversible. |
| Payload contract | ⚠️ Unconfirmed | `buildPayload` keys (`notify.go:151-167`) are self-described as "proposed"; not validated against octo-web #291. |

The security-sensitive concern that motivated routing — a private summary leaking channel-wide — is **not present**. The tip is delivered only as directed DMs to the access-gated set, and I tried three refutation vectors (alternate send paths, `OriginChannelID` reaching a target, channel-membership-derived participants); all are dead ends. Good design, and the test (`notify_test.go:152-159`) explicitly asserts a non-participant outsider is never targeted.

## 2. Findings

### P1 — `contentType=21` payload contract is unconfirmed (merge gate)
`internal/worker/notify.go:151-167`. The doc comment states the keys (`type`, `from_uid`, `from_name`, `task_id`, `task_no`, `title`) are *proposed*, and the PR body lists "confirm exact required keys vs. what octo-web #291 registered" as a before-merge blocker. The test only asserts `payload["type"] == 21` (`notify_test.go:160-161`); the other five keys are never checked against any registered schema. If the client expects a different shape (e.g. a deep link / task-ref form), the tip will fail to render once the feature is enabled. The no-op-until-configured guard (`config.NotifyConfigured`) prevents breakage on an unconfigured deploy, but does not remove the risk once it ships. **Confirm the contract with octo-web #291 before this is enabled.**

### P2 — Idempotency anchor is committed before the send, so a failed fan-out loses the tip permanently
`internal/worker/notify.go:91-121`. The CAS sets `notified_at = now` and commits *before* the task load (`:104`), the participant query (`:144`, whose error is swallowed), and the send loop. If the post-CAS load fails, or the participant query fails, or all sends fail (octo-server down during the completion window), `notified_at` is already set and there is no reset/retry path anywhere in the repo — the tip is lost with at-most-once semantics. Impact is bounded: this is a best-effort UI nudge, and the summary itself is persisted by `saveLatestResultAndCompleteTask` *before* the emit, so the user can always open the task. Worth hardening: either reset `notified_at` to NULL when `sent == 0` / on post-CAS error, or only commit the anchor after at least one successful send.

### P2 — Synchronous, sequential sends on the worker path can stall the pool
`internal/worker/notify.go:112-119` + `personal_processor.go:225` / `meta_processor.go:222`. `emitSummaryNotify` runs synchronously inside the pooled worker goroutine, and sends to targets sequentially with a 5s-timeout client. A slow/unreachable octo-server blocks the goroutine for up to `5s × len(targets)` per completion while it still holds a worker-pool slot (`WORKER_MAX_CONCURRENT_TASKS`, default 20). Realistic target sets are small, so this is throughput degradation under a degraded dependency, not data loss. Suggest running the emit in its own goroutine, parallelizing per-target sends, or adding an overall context deadline.

### P2 — Response body not drained on success (connection reuse)
`internal/worker/notify.go` `botAPISender.Send`. On a 2xx the body is closed without being read to EOF, which prevents keep-alive connection reuse and can churn connections / exhaust ephemeral ports under load. Add `io.Copy(io.Discard, resp.Body)` before close on the success path. *(Surfaced by the secondary review leg.)*

### P2 — Participant query error is ignored
`internal/worker/notify.go:144` — `db.Where("task_id = ?", task.ID).Find(&participants)` discards the returned error. Combined with the P2 above (anchor already committed), a transient DB error here silently notifies only the creator and permanently drops every participant. Check and log the error; consider treating it as a retryable failure.

### P2 — Docs overclaim "exactly-once / crash-safe"; concurrency test doesn't exercise contention
`internal/model/model.go:118-122` and `notify.go:85-88` say "exactly-once"; the real guarantee is **at-most-once** (anchor commits before best-effort send — see P2 above). Separately, `TestNotifyIdempotencyConcurrent` pins `sqlDB.SetMaxOpenConns(1)` (`notify_test.go:26`), which serializes all 16 goroutines through one connection — so it validates the predicate logic but does not actually exercise concurrent row contention or the InnoDB locking path. Recommend softening the comment to "at-most-once" and noting the test's fidelity limit (or exercising it against MySQL).

### nit — Soft-delete not filtered in the CAS/load
`internal/worker/notify.go:92,104`. `SummaryTask.DeletedAt` is a plain `*time.Time` (not `gorm.DeletedAt`), so GORM applies no automatic `deleted_at IS NULL` clause, and neither the CAS nor the load adds one (every other query in the repo spells it out by hand). A task soft-deleted in the microsecond window between completion and notify could still be notified. Impact is negligible (a stray DM to the owners who just deleted their own task; no leak), but adding `AND deleted_at IS NULL` to both statements would match the rest of the codebase.

## 3. Recommendations
1. **Before enabling in prod (P1):** confirm the `contentType=21` payload keys against octo-web #291 and the actual `POST /v1/bot/sendMessage` contract, then add a test that asserts the full key set, not just `type`.
2. Harden the emit path: commit `notified_at` only after a successful send (or reset on total failure), check the participant-query error, drain the response body, and move the fan-out off the worker goroutine.
3. Soften the "exactly-once/crash-safe" wording to "at-most-once" and document the concurrency-test fidelity limit.

## 4. Additional notes
- **Security gate (PR before-merge item #3): cleared.** The outbound sender is access-gated correctly and never broadcasts; the only outstanding security-relevant dependency is provisioning the system bot token (item #1), which is operational, not a code defect. One thing for a human to confirm operationally: the bot's `app_*` token must be a system identity scoped only to post these tips, since this worker can now post into any creator/participant DM.
- The other two listed Draft blockers (bot provisioning, security review) are coordination items; the code correctly no-ops until `NotifyConfigured()` is true, so merging disabled is safe if desired, but the P1 payload contract should be resolved before the feature is switched on.

yujiawei:
One additional non-blocking note for the revision pass:

**Scheduled empty-window completions still emit the tip.** In `internal/worker/personal_processor.go`, `emitSummaryNotify(taskID)` sits after the `if/else` (line ~225), so the `isScheduledEmptyWindow` branch (which completes via `completeTaskWithoutNewResult` and intentionally retains the previous result, no new content generated) also fires a "总结完成" tip. The task did complete processing, so the tip isn't strictly wrong, but recipients get a "summary completed" nudge with no new summary behind it. Whether to suppress the tip for no-new-content scheduled runs is a product decision rather than a defect — flagging it so it's a conscious choice before the feature is enabled.

Jerry-Xin (changes requested):
The PR is in scope for `octo-smart-summary`, but the notify idempotency design currently drops notifications permanently on transient delivery failures.

**🔴 Blocking**
- 🔴 Critical: [internal/worker/notify.go](/Users/mlamp/.openclaw/tmp/pr-review-tknelq1o/repo/internal/worker/notify.go:96) sets `notified_at` before loading targets and before any outbound send succeeds. Any failure after that point, including task reload failure at lines 108-111, participant query issues at lines 153-154, or bot API/send failures at lines 122-127, permanently suppresses retries because later calls see `notified_at IS NOT NULL` and return. A configured but temporarily unhealthy bot API can therefore produce `0/N` sends while marking the task notified. This violates the PR’s “exactly-once emit” claim and makes the new server-side notification unreliable. Use an outbox/attempt table or mark delivery only after successful completion with retryable state; for per-recipient fan-out, track recipient-level delivery if partial failure matters.

**💬 Non-blocking**
- 🟡 Warning: [internal/worker/notify.go](/Users/mlamp/.openclaw/tmp/pr-review-tknelq1o/repo/internal/worker/notify.go:97) does not require `status = Completed` in the CAS. Current call sites invoke it after completion, but adding the status predicate would make the notifier fail closed if a future caller or retry path invokes it too early.
- 🟡 Warning: [internal/worker/notify.go](/Users/mlamp/.openclaw/tmp/pr-review-tknelq1o/repo/internal/worker/notify.go:153) ignores the participant query error. That can silently send only to the creator while burning the global anchor. This should return/log an error and leave retry state intact.
- 🔵 Suggestion: [internal/config/config.go](/Users/mlamp/.openclaw/tmp/pr-review-tknelq1o/repo/internal/config/config.go:102) considers UID/name optional. If the frontend contract requires stable sender fields, include `NotifyBotUID` and `NotifyBotName` in `NotifyConfigured()` or validate them at startup.

**✅ Highlights**
- The target policy matches the API access model: creator plus explicit participants, not origin-channel broadcast.
- The tests cover deduping, concurrent CAS behavior, already-notified no-op, and soft-deleted tasks.
- `go test ./...`, `go test ./internal/worker`, and `go vet ./internal/worker` pass locally.

PR #115, opened 2026-06-27
ci(octo-pr-review-feed): opt into webhook-only triage

lml2468:
Adds enable_im_notify: false + wires PR_TRIAGE_WEBHOOK_URL secret. Backed by Mininglamp-OSS/.github#81 (which added the input + secret to the reusable). PR fan-out triage agent now files a tracked review issue; the IM ping becomes redundant.

### DIFF
--- a/.github/workflows/octo-pr-review-feed.yml
+++ b/.github/workflows/octo-pr-review-feed.yml
@@ -17,6 +17,8 @@ jobs:
       pr_url: ${{ github.event.pull_request.html_url }}
       pr_author: ${{ github.event.pull_request.user.login }}
       event_action: ${{ github.event.action }}
+      enable_im_notify: false
       project_group_id: '8d68238743664ba1a49eec75b4e296bd'
     secrets:
       OCTO_BOT_TOKEN: ${{ secrets.OCTO_BOT_TOKEN }}
+      PR_TRIAGE_WEBHOOK_URL: ${{ secrets.PR_TRIAGE_WEBHOOK_URL }}


yujiawei (approved):
# Code Review — PR #115 (octo-smart-summary)

## 1. Spec compliance

**Spec: ✅**

The PR does exactly what its title and description state — opt this repo into webhook-only PR triage — and nothing more:

- **Missing:** none. Both intended lines are present in `.github/workflows/octo-pr-review-feed.yml`:
  - line 20: `enable_im_notify: false`
  - line 24: `PR_TRIAGE_WEBHOOK_URL: ${{ secrets.PR_TRIAGE_WEBHOOK_URL }}`
- **Extra:** none. The diff is +2 / -0, single file.
- **Deviation:** none. Both the `enable_im_notify` input and the `PR_TRIAGE_WEBHOOK_URL` secret are defined in the called reusable workflow `Mininglamp-OSS/.github/.github/workflows/octo-pr-review-feed.yml@v1` (input `enable_im_notify` is `boolean`, default `true`; secret `PR_TRIAGE_WEBHOOK_URL` is optional). So this caller passes only declared parameters — no "unexpected input/secret" failure.

## 2. Code quality

**Quality: Approved**

End-to-end behavior verified against the reusable workflow and the `octo-notify@v1` composite action:

- With `enable_im_notify: false`, the reusable computes `im-message: ${{ inputs.enable_im_notify && steps.build.outputs.im_message || '' }}` → empty string → `octo-notify` skips IM mode. Correct.
- The `PR_TRIAGE_WEBHOOK_URL` secret is forwarded to `octo-notify`'s `webhook-url`; when non-empty, webhook mode POSTs the sanitized JSON payload. Correct.
- Security hygiene is sound for a security-sensitive change: the webhook URL is consumed as a capability **secret** (not hardcoded), passed via the `secrets:` block, and the composite action consumes all inputs via `env:` rather than `${{ }}` shell interpolation. `permissions: {}` keeps least privilege. `pull_request_target` is pre-existing and safe here — no PR code is checked out or executed; only PR metadata is read and sanitized.

### Items for a human to verify (security-sensitive PR)

- **P2 — Secret must be configured in repo settings.** This change makes the workflow depend on `PR_TRIAGE_WEBHOOK_URL`. If that secret is **not** set while `enable_im_notify: false`, `octo-notify` receives neither an `im-message` nor a `webhook-url` and exits with code 2 (`octo-notify invoked with neither im-message nor webhook-url`) — i.e. every future `ready_for_review` / `review_requested` event would fail the workflow with no notification sent. The diff itself is correct given the documented precondition (org `.github#81`); just confirm the repo (or org) actually has `PR_TRIAGE_WEBHOOK_URL` populated before relying on it. Not blocking, since the absence would be a config gap, not a defect in this diff.

### Nit

- **P2 (nit) — Missing trailing newline.** The file now ends without a final newline (`\ No newline at end of file`). Harmless for YAML, but adding one keeps POSIX-friendly tooling/diffs clean.

## 3. Overall verdict

**APPROVED** — Spec ✅ and Quality Approved. No P0/P1 issues. The single follow-up (confirm `PR_TRIAGE_WEBHOOK_URL` is configured) is an operational precondition to verify, not a code change required in this PR.

PR #117, opened 2026-06-27
fix: reverse DM channel ID normalization to match WuKongIM storage

### LINKED ISSUE #72: Bug: DM channel ID normalization order reversed — NormalizeDMChannelID uses max@min but WuKongIM stores min@max

lml2468:
## Summary

Fix DM smart summary returning empty results by correcting the `NormalizeDMChannelID` ordering to match how WuKongIM stores DM channel IDs in the message table.

### Root cause

`NormalizeDMChannelID` normalized DM channel IDs to `max(uid1,uid2)@min(uid1,uid2)`, but WuKongIM stores them as `min(uid1,uid2)@max(uid1,uid2)`. This mismatch caused `WHERE channel_id = ?` to never match any rows, producing the `0 rows fetched` symptom and empty summaries for all DM conversations.

### Fix

- **`internal/pipeline/fetch.go`**: Flip the comparison in `NormalizeDMChannelID` from `a < b` → `a > b` so the result is `min@max`, matching WuKongIM's storage order. Updated the doc comment accordingly.
- **`internal/pipeline/fetch_test.go`**: Reversed expected values in all four DM test cases to match the corrected ordering. Added a regression test using the real channel IDs from the bug report.

Fixes #72

## Test plan

- [x] `go test ./internal/pipeline/ -run TestNormalizeDMChannelID -v` — all 7 subtests pass
- [x] `go test ./...` — full suite passes, no regressions
- [ ] Verify on staging with a real WuKongIM DM conversation to confirm summaries are populated


Jerry-Xin (dismissed):
The PR is relevant to this repository and fixes smart-summary DM fetch behavior in the pipeline by aligning DM channel ID normalization with the storage lookup path.

💬 Non-blocking

🔵 Suggestion — Consider adding one fetch-level regression test, not just a pure normalization test. `internal/pipeline/fetch.go:343-371` normalizes the channel ID, computes the shard table, and then queries `WHERE channel_id = ?`; a test that inserts a DM message under `min@max` and calls `FetchMessagesFromChannel` with a peer-only or reversed ID would guard the complete failure mode from #72. The current tests in `internal/pipeline/fetch_test.go:5-73` cover the helper well, so this is not blocking.

✅ Highlights

The normalization change in `internal/pipeline/fetch.go:207` is small and consistently updates both discovery/source filtering and fetch paths because callers route through `NormalizeDMChannelID`.

The regression case in `internal/pipeline/fetch_test.go:42-47` uses the real reported ID shape and directly verifies the corrected `min@max` ordering.

Verification run:
`go test ./internal/pipeline/ -run TestNormalizeDMChannelID -v` passed.
`go test ./...` passed.

OctoBoooot (commented):
## Review: fix: reverse DM channel ID normalization to match WuKongIM storage (#117)

**Verdict:** Approve with comments

⚠ This PR is marked **draft** — treat this as a provisional pre-review; the verdict isn't a merge gate until you mark it ready. CI is red (`dependency-review` failing — pre-existing repo-settings issue, not introduced here), which also keeps the verdict floor below APPROVE per the CI-precondition rule.

The fix itself is correct, minimal, well-tested, and the root-cause analysis in the PR body matches the code. Closes #72 plausibly (the issue title states the exact `max@min` vs `min@max` mismatch this fixes).

### What was verified
- `internal/pipeline/fetch.go:207` — comparator flip `a < b → a > b` now yields `min(a,b)@max(a,b)`. Walked every branch:
  - peer-only (`channelID=peer`, `b=selfUID`): swaps to min@max ✓
  - `peer@self` already-`@` form: parses both sides, swaps to min@max ✓
  - all 5 test cases recompute correctly under the new comparator (e.g. `5904fca8`/`2c56cb` → `2c56cb@5904fca8`; `b@a` → `a@b`) ✓
- §7.1 inverse-case check (comparator swap is the classic "fixes one direction, breaks the other" shape): the inverse would break if anything *wrote* or *expected* the old `max@min`. It doesn't — all three call sites (`:108`, `:279`, `:343`) use `NormalizeDMChannelID` purely at query time to build `WHERE channel_id = ?` lookups against the WuKongIM message table. WuKongIM is the source of truth storing `min@max`, normalization isn't persisted anywhere, so flipping reads to match storage is the correct and only direction. No symmetric bug introduced.
- The `non-DM unchanged` (channelType != 1) early return is untouched — group/thread channel IDs pass through verbatim, correct.

### Question
- The test plan's third box (`Verify on staging with a real WuKongIM DM conversation`) is unchecked. Since the original bug was precisely a storage-format mismatch that unit tests didn't catch (the old tests asserted the *wrong* expected values and still passed), the staging check against real WuKongIM data is the load-bearing verification here — the unit tests now encode the corrected assumption but can't prove it matches production storage. Worth completing that box before marking ready, given the unit tests were previously green on the wrong order.

### Praise
- The added regression test using the real channel IDs from the bug report (`9a45a51c...@1d60945f...` / self `1d60945f...` → `1d60945f...@9a45a51c...`) is exactly right — it pins the fix to the actual failing production input, not just synthetic short strings, so a future refactor can't silently re-break it. This is the test that would have caught the original bug.
- Doc comment updated in lockstep with the behavior change (`max@min` → `min@max` with the WuKongIM rationale) — no stale comment left contradicting the code, which is the most common bit-rot on a one-line logic flip.
- Tight root-cause writeup: names the exact symptom (`0 rows fetched` / empty summaries), the storage-order mismatch, and the one-line fix. Easy to verify against the code.

mochashanyao (approved):
[Octo-Q · automated review]

---

# octo-smart-summary PR#117 Review Report

**PR:** https://github.com/Mininglamp-OSS/octo-smart-summary/pull/117
**Head SHA:** a5147fd0775ccbe91b9a6b0e60cb7d6345ec3504
**Title:** fix: reverse DM channel ID normalization to match WuKongIM storage order

## 1. Verification Summary

- ✅ **Build passes** (`go build ./...`) — clean
- ✅ **All unit tests pass** (`go test ./...`) — all packages green
- ✅ **Targeted test `TestNormalizeDMChannelID`** — 7/7 subtests pass
- ✅ **Normalization consistency** — All 3 call sites (`GetUserChannels`, `ApplySourceConstraints`, `FetchMessagesFromChannel`) use the same `NormalizeDMChannelID` function; change is applied uniformly
- ✅ **No external storage impact** — All persistence layers (`summary_source.source_id`, `summary_task.origin_channel_id`, `summary_schedule.source_config`) store raw peer UIDs, not normalized `a@b` format

## 2. Findings

### P2 — Test case names semantically swapped (nit)
**Diff-scope:** new (introduced by this PR's test update)

After reversing the sort order, two test case names no longer describe their behavior accurately:

- `"already has @ reorder"` (input `a@b`, want `a@b`) — no reorder happens now; the input is already in `min@max` order. Should be named `"already correct order"`.
- `"already correct order"` (input `b@a`, want `a@b`) — now actually reorders to `min@max`. Should be named `"already has @ reorder"`.

**Impact:** Cosmetic only. Tests are functionally correct, just the names are misleading for future readers.

**Suggestion:** Swap the two test names.

## 3. Root Cause Analysis (4-Phase)

### Phase 1: Symptom
DM summaries return zero messages. Users requesting a summary of a DM conversation get empty results.

### Phase 2: Root Cause
`NormalizeDMChannelID` produced `max@min` format (e.g., `"9a45a51c...@1d60945f..."`), but WuKongIM stores messages in the message table with channel_id in `min@max` format (e.g., `"1d60945f...@9a45a51c..."`). This mismatch caused:
1. **SQL query miss:** `FetchMessagesFromChannel` queries `WHERE channel_id = ?` with the wrong format → 0 rows
2. **Shard mismatch:** `MessageTable()` uses `CRC32(channelID) % tableCount` — `"a@b"` and `"b@a"` hash to different values, potentially routing to the wrong shard table

### Phase 3: Fix Verification
The fix reverses the sort comparison from `if a < b` (producing `max@min`) to `if a > b` (producing `min@max`). This is a minimal, targeted change that directly addresses the root cause.

### Phase 4: Blast Radius
- **All 3 production call sites** use the same function — change propagates uniformly
- **No persistence-layer impact** — stored source IDs are raw peer UIDs; normalization only happens at runtime
- **Shard routing changes** — DM channel queries will now hit the correct shard table (the one WuKongIM actually writes to)
- **No cross-service impact** — candidates API returns raw peer UIDs to frontend; frontend sends them back as `source_id`; normalization happens inside the pipeline

## 4. Data Flow Traceback

| Consumer | Data | Upstream Source | Verified Flow |
|----------|------|-----------------|---------------|
| `GetUserChannels` (fetch.go:108) | `d.ChannelID` from `conversation_extra` | SQL query (line 97) → raw peer UID or `a@b` from WuKongIM | ✅ Normalized before use in `ChannelInfo.ChannelID` |
| `ApplySourceConstraints` (fetch.go:279) | `source_id` from frontend `specifiedSources` | Client request → raw peer UID | ✅ Normalized via `NormalizeDMChannelID(id, selfUID, backendChType)` before map lookup |
| `FetchMessagesFromChannel` (fetch.go:343) | `channelID` parameter | Comes from `ChannelInfo.ChannelID` (already normalized) or direct call | ✅ Re-normalized for safety before SQL query |
| `MessageTable` (shard.go:14) | `channelID` string | Receives normalized `min@max` from caller | ✅ CRC32 hash now matches WuKongIM's shard assignment |

**Critical path verified:** `conversation_extra.channel_id` (raw peer UID) → `NormalizeDMChannelID(peer, self, 1)` → `min@max` → `MessageTable(min@max)` → correct shard → SQL `WHERE channel_id = 'min@max'` → matches WuKongIM storage → messages returned.

## 5. Suggestions

1. **(P2)** Swap test names `"already has @ reorder"` and `"already correct order"` to match the new sort semantics.

## 6. Blindspot Checklist (R5)

- **C1 — Dual-path parity:** N/A. No symmetric add/remove or create/delete paths are touched by this change. The normalization function is a pure utility with no side effects.
- **C2 — Control-flow ordering / nested reuse:** Clear. `NormalizeDMChannelID` is called at 3 sites, all with consistent inputs. No nested/double normalization occurs (each call site normalizes independently, and the function is idempotent).
- **C3 — Authorization boundary:** N/A. No permission/jail/tool exposure changes.
- **C4 — Authorization lifecycle / container-member cascade:** N/A. No auth changes.
- **C5 — Build ≠ runtime path:** Clear. This is a Go backend service. The build output directly matches runtime behavior. `go build` and `go test` both pass. No browser extension / CLI / packaging involved.
- **C6 — Governance / policy document:** N/A. No documentation or policy changes.

## 7. Cross-Round Blocker Re-check (R6)

---

**[Octo-Q] verdict: APPROVE** — The fix is minimal, correct, and well-tested. All 3 call sites are consistently updated through the shared function. No P0/P1 issues found. Only a cosmetic P2 nit on test naming.

yujiawei (changes requested):
# Code Review — PR #117 (octo-smart-summary)

**Head SHA reviewed:** `a5147fd0775ccbe91b9a6b0e60cb7d6345ec3504`

## Verdict: Changes Requested

The diff is internally clean (single point of normalization, all three call sites symmetric, tests green, build green). However, the **core premise of the fix is incorrect**, so the change does not actually fix the bug for the general case and regresses conversations that previously worked.

## Root-cause analysis is wrong: WuKongIM does not order DM channel IDs lexicographically

The PR assumes WuKongIM stores DM channel IDs as **lexicographic** `min(uid)@max(uid)`, and flips the comparison from `a < b` to `a > b` to produce that ordering.

WuKongIM actually orders the two UIDs by their **CRC32 hash**, not by string comparison:

- Legacy `GetFakeChannelIDWith` (the function the IM server uses to build the persisted channel ID):

  ```go
  func GetFakeChannelIDWith(fromUID, toUID string) string {
      fromUIDHash := wkutil.HashCrc32(fromUID)
      toUIDHash := wkutil.HashCrc32(toUID)
      if fromUIDHash > toUIDHash {
          return fromUID + "@" + toUID
      }
      // ... tie-break on hash collision ...
      return toUID + "@" + fromUID
  }
  ```

- Current `EncodePersonChannel` (`internal/runtime/channelid/person.go`) is identical in spirit:

  ```go
  leftHash := crc32.ChecksumIEEE([]byte(leftUID))
  rightHash := crc32.ChecksumIEEE([]byte(rightUID))
  if leftHash > rightHash { return leftUID + "@" + rightUID }
  ...
  return rightUID + "@" + leftUID
  ```

The persisted order is the UID with the **higher CRC32 hash first** — independent of lexicographic order.

## Why the regression test passes anyway (false confidence)

The regression case uses the real report IDs:
- `selfUID  = 1d60945f0d4e488fa8f237e270ad34af`
- `peerUID  = 9a45a51cd76c4e668e12cf10c2b797dd`

For this specific pair:
- `crc32(1d60…) = 4029552095`, `crc32(9a45…) = 3572916565` → higher-hash-first → `1d60…@9a45…`
- lexicographic `min@max` → `1d60…@9a45…`

Both orderings coincidentally produce the same string, so this single data point cannot distinguish lexicographic ordering from CRC32 ordering. The added test therefore does not validate the root cause — it only confirms the code reproduces one coincidental value.

## The fix is a coin flip and regresses ~half of DM channels

Comparing each lexicographic strategy against WuKongIM's true CRC32 ordering over 200,000 random 32-hex-char UID pairs:

| Strategy | Matches WuKongIM CRC32 storage order |
|---|---|
| OLD code (`a < b`, lex max@min) | **49.8%** |
| NEW code (`a > b`, lex min@max) | **50.2%** |

So the flip does not fix the bug — it merely swaps *which ~50% of DM conversations* resolve to the wrong `channel_id`. Conversations whose CRC32 order happens to equal lexicographic `max@min` worked under the old code and will now break; the symptom (`0 rows fetched`, empty summary) just moves to a different set of users. Because the normalized ID feeds **both** the `WHERE channel_id = ?` filter **and** the CRC32 shard table selection (`MessageTable`, `internal/pipeline/shard.go:18`), a wrong order points the query at the wrong shard *and* the wrong row.

Note the repository already replicates WuKongIM's CRC32 IEEE convention for shard selection (`shard.go`, `shard_test.go:53` "matches Python's binascii.crc32"), so the hashing primitive is understood here — it just is not applied to the DM-ID ordering.

## Required change

Replace the lexicographic comparison in `NormalizeDMChannelID` (`internal/pipeline/fetch.go:207`) with WuKongIM's CRC32-based ordering, mirroring `GetFakeChannelIDWith` / `EncodePersonChannel` (higher CRC32 hash first, with the documented tie-break for hash collisions). Something equivalent to:

```go
ah := crc32.ChecksumIEEE([]byte(a))
bh := crc32.ChecksumIEEE([]byte(b))
if ah < bh || (ah == bh && a < b) {
    a, b = b, a
}
return a + "@" + b
```

(Confirm the exact tie-break direction against the deployed WuKongIM version before finalizing.)

Then update the doc comment and tests to describe **CRC32-hash ordering**, not `min@max`, and add at least one regression case whose CRC32 order differs from its lexicographic order (e.g. a pair where `crc32(a) > crc32(b)` but `a < b`) so the test can actually distinguish the two strategies — the current real-report case cannot.

## Suggestions

- Add a test asserting `NormalizeDMChannelID(channelID, selfUID, 1)` agrees with a local CRC32 reference for several pairs, including at least one where lexicographic and CRC32 orders disagree.
- Before merge, verify on staging against a real WuKongIM DM conversation whose UIDs have CRC32 order opposite to lexicographic order (the test-plan item that is still unchecked). The current verification cannot catch this class of bug.

## Additional finding (non-blocking)

`NormalizeDMChannelID` with an empty `selfUID` and a no-`@` input returns `"@" + channelID` (any non-empty string sorts after `""`). This is a degenerate, non-production path and is reached identically by all call sites, so it is not introduced by this change — but a guard for empty UIDs would harden it.

Jerry-Xin (changes requested):
Retracting my earlier APPROVE. After verifying WuKongIM's actual DM channel ID composition against upstream source, **yujiawei is correct: this fix sorts by the wrong key.**

**🔴 Blocking**

🔴 Critical — `NormalizeDMChannelID` orders the two UIDs by **lexical** string comparison (`if a > b { a, b = b, a }`), but WuKongIM composes a personal channel ID by comparing **`crc32.ChecksumIEEE(uid)`**, not lexical order. Upstream canonical impl (WuKongIM `PersonChannelID` / `encodeBenchPersonChannel`):
```
leftHash  := crc32.ChecksumIEEE([]byte(leftUID))
rightHash := crc32.ChecksumIEEE([]byte(rightUID))
if leftHash > rightHash { return leftUID + "@" + rightUID }
if leftHash == rightHash && leftUID > rightUID { return leftUID + "@" + rightUID }
return rightUID + "@" + leftUID
```
WuKongIM's design doc states it explicitly: "两端 uid 的前后顺序按 `crc32(uid)` 比较决定，而不是按字典序" (lexical UID compare is only the tie-breaker on a crc32 collision). CRC32 order and lexical order are uncorrelated, so both the old `max@min` lexical sort AND this PR's new `min@max` lexical sort produce the wrong `@`-ordering for roughly half of DM UID pairs → those DMs still resolve to a non-existent channel ID and fetch zero messages. This does not fix #72; it just flips which ~50% of DMs break.

🔴 Critical — the tests are non-discriminating. `internal/pipeline/fetch_test.go` asserts lexical `min@max` expectations, which cannot tell correct (crc32-ordered) output apart from incorrect (lexically-ordered) output. They pass only because the assertions were written to match the lexical implementation, not WuKongIM's real storage order. A valid test must compute the expected `@`-order via `crc32.ChecksumIEEE` for the chosen UIDs (or assert against a real stored channel ID), and should include a UID pair whose crc32 order is the **opposite** of its lexical order to actually catch the bug.

**Fix**
- Replace the lexical comparator in `NormalizeDMChannelID` with the crc32 rule above (hash compare, lexical tie-break on collision), matching `GetFakeChannelIDWith` / `PersonChannelID`. Note the repo already imports `hash/crc32` in `internal/pipeline/shard.go`.
- Rebuild the tests around crc32-derived expectations + an opposite-order pair.

**Correction note**
My earlier APPROVE was wrong: I confirmed the comparator was applied consistently across callers but did not verify that the *ordering key itself* (lexical vs crc32) matches WuKongIM's storage rule — the exact thing that determines correctness here. Credit to yujiawei for catching the wrong sort key.

Jerry-Xin:
Thanks for the contribution and the detailed write-up on the root cause — much appreciated.

We're going to close this one. The DM summary bug (#72) has already been fixed on `main` via #124, which aligns `NormalizeDMChannelID` with WuKongIM's actual `GetFakeChannelIDWith` rule (CRC32-based ordering). Since the storage order is determined by CRC32 rather than lexicographic comparison, flipping to `min@max` doesn't correlate with the real ordering and would still resolve roughly half of the UID pairs to the wrong direction.

Closing as superseded by #124. Thanks again for digging into this!

PR #118, opened 2026-06-27
fix: prevent channel scope over-narrowing on unmatched topic keywords (#79)

### LINKED ISSUE #79: bug: topic-based channel scope over-narrows to an irrelevant empty channel, producing an empty summary
## Problem

When a summary topic names a specific keyword that matches NO candidate channel, the LLM channel-scope resolver (`resolve_channel_scope`) still forces `has_constraint=true` and narrows the candidate set down to a single irrelevant channel. That channel has no messages, so the pipeline returns an empty summary with "no chat records matching the topic in the current scope", instead of summarizing across all candidate channels.

## Reproduction

Topic: `总结 octo 的项目进展` (summarize octo project progress). None of the 26 candidate channels is named/related to "octo" (closest is `hello-v2`).

Worker log:

```
Layer 1 (channel discovery) ... (26 channels)
Layer 1.5 (participant intersect) ... (26 channels)
[pipeline] CallWithTools: tool=resolve_channel_scope input={topic:"总结 octo 的项目进展", members:13, channels:26} result={has_constraint:true, rules:1}
[pipeline] ResolveChannelScope: narrowed 26 -> 1 channels (1 rules), reason=... 没有任何频道名称包含"octo"关键词 ... 然而用户明确提到了"octo的项目进展" ... 所以 has_constraint=true，只是无法匹配到具体频道
Layer 1.7 (channel scope) ... (1 channels)
Layer 2 (source constraints) ... (1 -> 1 candidates)
FetchMessagesFromChannel 2a08e6ef...: 0 rows fetched
Layer 4: fetched 0 messages from 1 channels
Total pipeline took 12096ms (0 messages final)
```

The model's own `reasoning` admits no channel matches "octo", yet it force-fit the keyword onto the lexically-nearest channel `hello-v2`.

## Root cause

Channel matching is done entirely by the LLM (semantic/fuzzy match of topic keywords against each channel's `名称`/name), not by Go code. The candidate list is serialized in `internal/pipeline/resolve_channel.go` (`callResolveChannelScope`, ~line 631) as `- ID: <id>, 名称: <name> (<type>)` and sent together with the topic via `buildChannelScopeUserMsg`. The rules live in `resolveChannelScopeSystemPrompt` (~lines 98-151).

Two layers, LLM primary:

1. Prompt gap (primary): the prompt defines only two outcomes — a match is found (`has_constraint=true`), or the topic is generic (`has_constraint=false`). The case "topic names a specific keyword but NO candidate channel matches it" is undefined, so the model improvises and force-fits an irrelevant channel. The existing "不得编造" guard only forbids inventing IDs, not selecting a real-but-irrelevant one.
2. Pipeline fallback too weak (secondary): `ResolveChannelScope` (`resolve_channel.go:187-191`) only falls back to all candidates when narrowing yields ZERO channels. A single force-fit (valid but wrong) channel id slips past this guard, so over-narrowing to a content-empty channel becomes a silent empty summary.

The user-facing message `noRelevantContentMessage` (`personal_processor.go:21`, returned at ~:300-302) is itself correct for genuine no-content; it is only triggered here by wrong upstream narrowing.

## Affected code

- `internal/pipeline/resolve_channel.go`: tool def (~:38-96), system prompt (~:98-151), `ResolveChannelScope` narrowing + fallback (~:153-196), `filterByChannelIDs` (~:259-283), `callResolveChannelScope` / `buildChannelScopeUserMsg` (~:631+).
- `internal/pipeline/fetch.go`: `FetchMessagesFromChannel` (0 rows).
- `internal/api/handler/personal_processor.go`: `noRelevantContentMessage` (~:21, :300-302).

## Proposed fix (both layers recommended)

1. Prompt fix: add a rule + example — if a named keyword matches no candidate channel, do NOT force a match; return `has_constraint=false` (or empty `channel_ids`) so the summary runs across all candidate channels.
2. Deterministic guard: when a rule's only effective constraint is `channel_ids`, validate the selected ids in Go and drop non-matching ones; combined with the existing `len==0` fallback this returns all candidates instead of one wrong channel.

Optional/reserve: after fetch, if a narrowed scope yields 0 messages, re-fetch across all candidates (more robust but more invasive).

## Notes

Unrelated to the archived-threads work (#78 / octo-web#346). This is in Layer 1.7 (LLM topic scope resolution); Layer 1 correctly discovered all 26 channels.

lml2468:
## Root cause

When a summary topic names a keyword (e.g. "octo") that matches NO candidate channel, the LLM channel-scope resolver has no defined "no-match" outcome. It improvises by force-fitting the keyword onto the lexically-nearest channel (e.g. "hello-v2"), returning `has_constraint=true` with a valid-but-irrelevant channel ID. The Go-side guard only catches 0-channel results, so the single wrong channel slips through → 0 messages fetched → empty summary.

## Fix

**Two-layer approach:**

1. **Prompt fix** (`resolveChannelScopeSystemPrompt`):
   - Added rule: "如果主题提到了特定关键词...但候选频道列表和成员列表中没有任何一项与该关键词匹配，则 has_constraint 为 false"
   - Added anti-force-fit instruction under `channel_ids`: "不得强行匹配：如果主题中的关键词与任何候选频道名称都不相关，不要选取'最接近'的频道 ID"
   - Added regression example: `"octo的项目进展"（但候选频道中没有名称包含"octo"的频道）→ has_constraint=false`

2. **Go guard** (`ResolveChannelScope`):
   - New `topicOverlapsChannel(topic, channelName)` helper: extracts multi-char ASCII tokens from topic, checks if any appears in channel name
   - After narrowing: if result has 1 channel from ≥3 candidates and zero keyword overlap → force-fit detected → fallback to all candidates

## Verification

- `TestResolveChannelScope_ForceFitGuard` — reproduces the bug scenario (topic "octo" → channel "hello-v2"). Confirms fallback to all candidates.
- `TestResolveChannelScope_ForceFitGuard_LegitimateNarrow` — confirms guard doesn't trigger when keyword legitimately matches (topic "octo" → channel "octo-dev").
- `TestTopicOverlapsChannel` — unit test for the overlap helper with 8 cases.
- Full `go test ./...` passes — no regressions.

Fixes #79

### DIFF
--- a/internal/pipeline/resolve_channel.go
+++ b/internal/pipeline/resolve_channel.go
@@ -113,6 +113,7 @@ const resolveChannelScopeSystemPrompt = `你是一个频道范围解析器。根
   - 精确匹配：主题明确指定频道名，选取对应 ID
   - 模糊匹配：主题含关键词（如"dev相关"），选取所有名称中包含该关键词的频道
   - 只能从候选频道列表中选取 ID，不得编造
+  - 不得强行匹配：如果主题中的关键词与任何候选频道名称都不相关，不要选取"最接近"的频道 ID；应改为设置 has_constraint=false
 - channel_type：数组形式，仅当主题明确限定频道类型时设置
   - "私聊"、"DM" → ["dm"]
   - "群"、"群组"、"群聊" → ["group"]
@@ -127,6 +128,7 @@ const resolveChannelScopeSystemPrompt = `你是一个频道范围解析器。根
 - 当主题包含多个独立约束条件需要并集时，拆分为多条 rules
   - 例："总结下我的全部私聊，以及我和Alice所在群聊" → 两条规则
 - 如果主题不包含任何频道范围约束（如"项目进度"、"最近在聊什么"），has_constraint 为 false，rules 为空数组
+- 如果主题提到了特定关键词（频道名、人名、项目名等），但候选频道列表和成员列表中没有任何一项与该关键词匹配，则 has_constraint 为 false，rules 为空数组。不得选取"最接近"的频道或成员强行匹配。
 
 示例：
 - "我和Alice聊了什么"
@@ -141,6 +143,8 @@ const resolveChannelScopeSystemPrompt = `你是一个频道范围解析器。根
   → has_constraint=true, rules=[{channel_ids:[所有含dev的频道ID], channel_type:["group"]}]
 - "最近大家在讨论什么"
   → has_constraint=false, rules=[]
+- "octo的项目进展"（但候选频道中没有名称包含"octo"的频道）
+  → has_constraint=false, rules=[]（不要强行选取不相关的频道）
 - "我建的群在聊什么"
   → has_constraint=true, rules=[{ownership:["creator"]}]
 - "我参与但不管理的群"
@@ -187,6 +191,16 @@ func ResolveChannelScope(
 	if len(result) == 0 {
 		log.Printf("[pipeline] ResolveChannelScope: all rules resulted in 0 channels, fallback to all %d candidates",
 			len(candidates))
+		return candidates
+	}
+
+	// Guard against LLM force-fitting: when narrowing drastically reduces the
+	// candidate set (≥3 candidates → 1 result), verify that at least one
+	// selected channel name shares keyword overlap with the topic. If none do,
+	// the LLM likely picked a valid-but-irrelevant channel — fall back to all.
+	if len(candidates) >= 3 && len(result) == 1 && !topicOverlapsChannel(topic, result[0].ChannelName) {
+		log.Printf("[pipeline] ResolveChannelScope: narrowed to %q but topic %q has no keyword overlap, fallback to all %d candidates (force-fit guard)",
+			result[0].ChannelName, topic, len(candidates))
 		return candidates
 	}
 
@@ -280,6 +294,42 @@ func filterByChannelIDs(current []ChannelInfo, selectedIDs []string, allCandidat
 		}
 	}
 	return result
+}
+
+// topicOverlapsChannel checks whether any multi-character token in the topic
+// appears in the channel name. Used as a deterministic guard against LLM
+// force-fitting an irrelevant channel when the topic keyword matches nothing.
+func topicOverlapsChannel(topic, channelName string) bool {
+	// Extract multi-character ASCII runs (≥2 chars) as meaningful tokens.
+	// Single CJK characters are too ambiguous for reliable overlap detection.
+	var tokens []string
+	start := -1
+	for i := 0; i <= len(topic); i++ {
+		if i < len(topic) && ((topic[i] >= 'a' && topic[i] <= 'z') ||
+			(topic[i] >= 'A' && topic[i] <= 'Z') ||
+			(topic[i] >= '0' && topic[i] <= '9')) {
+			if start < 0 {
+				start = i
+			}
+		} else {
+			if start >= 0 {
+				if i-start >= 2 {
+					tokens = append(tokens, strings.ToLower(topic[start:i]))
+				}
+				start = -1
+			}
+		}
+	}
+	if len(tokens) == 0 {
+		return true // no extractable tokens → can't judge, assume match
+	}
+	lowerName := strings.ToLower(channelName)
+	for _, t := range tokens {
+		if strings.Contains(lowerName, t) {
+			return true
+		}
+	}
+	return false
 }
 
 func filterByOwnership(ctx context.Context, candidates []ChannelInfo, creatorUID string, ownership []string, imDB *gorm.DB) []ChannelInfo {
--- a/internal/pipeline/resolve_channel_test.go
+++ b/internal/pipeline/resolve_channel_test.go
@@ -328,6 +328,94 @@ func TestResolveChannelScope_GlobalFallback(t *testing.T) {
 	}
 }
 
+// TestResolveChannelScope_ForceFitGuard reproduces issue #79: the LLM returns
+// has_constraint=true with a single channel_id that is a valid candidate but
+// has no keyword overlap with the topic (e.g. topic "octo" → channel "hello-v2").
+// The deterministic guard should detect this and fall back to all candidates.
+func TestResolveChannelScope_ForceFitGuard(t *testing.T) {
+	candidates := []ChannelInfo{
+		{ChannelID: "ch1", ChannelType: 2, ChannelName: "hello-v2"},
+		{ChannelID: "ch2", ChannelType: 2, ChannelName: "random-chat"},
+		{ChannelID: "ch3", ChannelType: 2, ChannelName: "water-cooler"},
+		{ChannelID: "ch4", ChannelType: 2, ChannelName: "announcements"},
+		{ChannelID: "ch5", ChannelType: 2, ChannelName: "general"},
+	}
+	memberMap := map[string]string{"uid_a": "Alice"}
+
+	// Simulate LLM force-fitting: topic mentions "octo" but no channel matches.
+	// LLM picks "hello-v2" as the "nearest" channel.
+	mockFn := mockChannelScopeToolCallFn(ChannelScopeResult{
+		HasConstraint: true,
+		Rules:         []ChannelScopeRule{{ChannelIDs: []string{"ch1"}}},
+		Reasoning:     "no channel named octo, picking hello-v2 as nearest",
+	})
+
+	result := ResolveChannelScope(context.Background(), "octo的项目进展", candidates, "creator", memberMap, nil, mockFn)
+	gotIDs := extractIDs(result)
+	wantIDs := extractIDs(candidates) // should fall back to all candidates
+	if !sliceEqual(gotIDs, wantIDs) {
+		t.Errorf("force-fit guard: got %v, want %v (should fallback to all candidates)", gotIDs, wantIDs)
+	}
+}
+
+// TestResolveChannelScope_ForceFitGuard_LegitimateNarrow ensures the guard
+// does NOT trigger when the topic keyword legitimately matches a channel name.
+func TestResolveChannelScope_ForceFitGuard_LegitimateNarrow(t *testing.T) {
+	candidates := []ChannelInfo{
+		{ChannelID: "ch1", ChannelType: 2, ChannelName: "octo-dev"},
+		{ChannelID: "ch2", ChannelType: 2, ChannelName: "random-chat"},
+		{ChannelID: "ch3", ChannelType: 2, ChannelName: "water-cooler"},
+		{ChannelID: "ch4", ChannelType: 2, ChannelName: "announcements"},
+	}
+	memberMap := map[string]string{"uid_a": "Alice"}
+
+	// Topic mentions "octo" and "octo-dev" is a real match.
+	mockFn := mockChannelScopeToolCallFn(ChannelScopeResult{
+		HasConstraint: true,
+		Rules:         []ChannelScopeRule{{ChannelIDs: []string{"ch1"}}},
+		Reasoning:     "octo matches octo-dev",
+	})
+
+	result := ResolveChannelScope(context.Background(), "octo的项目进展", candidates, "creator", memberMap, nil, mockFn)
+	gotIDs := extractIDs(result)
+	wantIDs := []string{"ch1"} // should NOT fall back; legitimate match
+	if !sliceEqual(gotIDs, wantIDs) {
+		t.Errorf("legitimate narrow: got %v, want %v", gotIDs, wantIDs)
+	}
+}
+
+// TestTopicOverlapsChannel tests the keyword overlap helper directly.
+func TestTopicOverlapsChannel(t *testing.T) {
+	tests := []struct {
+		topic       string
+		channelName string
+		want        bool
+	}{
+		// Force-fit: no overlap
+		{"octo的项目进展", "hello-v2", false},
+		{"octo项目进展", "random-chat", false},
+		// Legitimate match: ASCII keyword present in name
+		{"octo项目进展", "octo-dev", true},
+		{"octo项目进展", "my-octo-group", true},
+		// Multiple tokens, at least one matches
+		{"octo项目进展", "octo讨论组", true}, // "octo" matches
+		// "octo" not in name, only CJK tokens remain → no overlap
+		{"octo项目进展", "进展讨论组", false},
+		// All-Chinese topic: no extractable tokens → assume match (conservative)
+		{"项目进展总结", "random", true},
+		// Single-char ASCII tokens ignored, CJK tokens ignored → no extractable → conservative match
+		{"a的b项目", "hello", true},
+	}
+	for _, tt := range tests {
+		t.Run(tt.topic+"→"+tt.channelName, func(t *testing.T) {
+			got := topicOverlapsChannel(tt.topic, tt.channelName)
+			if got != tt.want {
+				t.Errorf("topicOverlapsChannel(%q, %q) = %v, want %v", tt.topic, tt.channelName, got, tt.want)
+			}
+		})
+	}
+}
+
 func TestResolveChannelScope_LLMError_Fallback(t *testing.T) {
 	candidates := []ChannelInfo{
 		{ChannelID: "g1", ChannelType: 2},


internal/pipeline/resolve_channel.go
```diff
@@ -190,6 +194,16 @@ func ResolveChannelScope(
 		return candidates
 	}
 
+	// Guard against LLM force-fitting: when narrowing drastically reduces the
+	// candidate set (≥3 candidates → 1 result), verify that at least one
+	// selected channel name shares keyword overlap with the topic. If none do,
```
OctoBoooot, line 199: **[major]** The force-fit guard fires on the final result regardless of which rule dimension narrowed it, but `topicOverlapsChannel` only makes sense for channel-*name* matching. Person/type/ownership rules narrow with no relation to the channel name, so the guard spuriously reverts them.

Failing input: 5 group candidates (`water-cooler`, `random-chat`, …), Alice is a member of `water-cooler`. Topic `"我和Alice聊了什么"`. LLM correctly returns `rules=[{persons:[AliceUID]}]` → narrows to `water-cooler`. Guard: candidates=5≥3, result=1, `topicOverlapsChannel("我和Alice聊了什么", "water-cooler")` extracts ASCII token `"alice"`, not in `"water-cooler"` → false → **guard fires → reverts to all 5 channels.** The legitimate person-narrow is destroyed; user asked about Alice and gets every channel summarized.

Same false-positive for `"总结我的DM"` (token `"dm"`) reverting a correct `channel_type:["dm"]` narrow.

Fix: gate the guard on rule shape — apply it only when the surviving narrowing came from `ChannelIDs` rules and none used `Persons`/`Ownership`/`ChannelType`. `parsed.Rules` is in scope; check rule dimensions before invoking `topicOverlapsChannel`.

OctoBoooot (changes requested):
## Review: fix: prevent channel scope over-narrowing on unmatched topic keywords (#118)

**Verdict:** Request changes

⚠ This PR is marked **draft** — provisional pre-review, not a merge gate until marked ready. CI is also red (`dependency-review` — pre-existing repo-settings issue, not this PR), which independently keeps the floor below APPROVE.

The two-layer approach (prompt rule + deterministic Go guard) is the right instinct for #79, and the prompt changes are sound. But the Go-side force-fit guard over-reaches: it reverts legitimate **non-name-based** narrowing, re-introducing the inverse of the bug it fixes.

### Major
- `internal/pipeline/resolve_channel.go:198-203` — the force-fit guard fires on `len(candidates) >= 3 && len(result) == 1 && !topicOverlapsChannel(topic, result[0].ChannelName)`, but it's applied to the final result regardless of **which rule dimension** did the narrowing. `topicOverlapsChannel` only makes sense for channel-*name* matching (`ChannelIDs` rules). Person/type/ownership rules narrow with no relation to the channel name, so the guard spuriously reverts them.
  - **Failing input:** 5 group candidates with arbitrary names (`water-cooler`, `random-chat`, …), Alice is a member of `water-cooler`. Topic `"我和Alice聊了什么"`. LLM correctly returns `rules=[{persons:[AliceUID]}]` → `filterByPersonUIDs` narrows to the 1 channel Alice shares (`water-cooler`). Now the guard: candidates=5 ≥3, result=1, and `topicOverlapsChannel("我和Alice聊了什么", "water-cooler")` extracts ASCII token `"alice"`, which is NOT in `"water-cooler"` → returns false → **guard fires → falls back to all 5 channels.** The user asked "what did Alice and I discuss" and gets a summary of every channel. The legitimate person-narrow is destroyed.
  - Same false-positive for type narrowing when the topic carries an ASCII token: `"总结我的DM"` → token `"dm"`, channel named `"random-chat"` → no overlap → reverts a correct `channel_type:["dm"]` narrow.
  - Fix: scope the guard to name-based narrowing only — apply it solely when the surviving rules used `ChannelIDs` (name/keyword matching) and none used `Persons`/`Ownership`/`ChannelType`. The `parsed.Rules` are in scope; gate on rule shape before invoking `topicOverlapsChannel`.

### Minor
- `internal/pipeline/resolve_channel.go:199` — the `≥3 candidates → exactly 1 result` shape is narrow: a force-fit to 1-from-2 candidates, or a force-fit to 2 wrong channels from 5, both slip through unguarded, so #79 still reproduces on those shapes. Acceptable to scope this PR to the 1-from-3+ case, but worth a comment noting the guard is intentionally narrow (or widening `len(result) == 1` to a ratio test).
- `resolve_channel_test.go` — the three new tests only exercise `ChannelIDs`-rule narrowing (force-fit + legitimate name match). The person/type/ownership false-positive class above is untested; a `TestResolveChannelScope_PersonNarrowNotReverted` case (topic with a romanized name, narrow-by-person to a differently-named channel, assert NOT reverted) would have caught the Major.

### Praise
- `topicOverlapsChannel`'s conservative default — "no extractable ASCII tokens → assume match" (`return true`) — is the right failure-safe direction for the all-CJK topic case: it avoids spuriously reverting when the helper simply can't judge. The `TestTopicOverlapsChannel` table covers that branch explicitly (`"项目进展总结"` → true). The instinct to fail open is correct; the gap is only that the guard is invoked on paths where channel-name overlap isn't the right signal at all.
- The prompt-layer fix is well-formed: the explicit "不得强行匹配" anti-force-fit rule + the `has_constraint=false` no-match instruction + the concrete `"octo的项目进展"` regression example give the LLM a defined no-match outcome, which is the actual root cause. If the prompt change alone reliably stops the force-fit, the Go guard is belt-and-suspenders — which is exactly why its false-positive cost should be minimized (scope it tightly per the Major).

mochashanyao (approved):
[Octo-Q · automated review]

---

# octo-smart-summary PR#118 Review Report
**PR:** https://github.com/Mininglamp-OSS/octo-smart-summary/pull/118
**Head SHA:** `20c631b05f48332882ce6669f3554314c7a716ad`
**Title:** fix: prevent channel scope over-narrowing on unmatched topic keywords (#79)
**Changed files:** 2 (`resolve_channel.go` +50, `resolve_channel_test.go` +88)

---

## 1. Verification Summary

- ✅ **Prompt anti-force-fit rules** — Clear, non-contradictory additions to `resolveChannelScopeSystemPrompt`. New rule at L116 ("不得强行匹配") + expanded no-match guidance at L131 + regression example at L144-145. Consistent with existing prompt structure.
- ✅ **Go guard (`topicOverlapsChannel` + caller check)** — Correctly placed after `len(result)==0` fallback (L192) and before the success log (L207). Guard fires only when `len(candidates) >= 3 && len(result) == 1 && !topicOverlapsChannel(...)` → falls back to all candidates.
- ✅ **`topicOverlapsChannel` helper** — Extracts ≥2-char ASCII runs, lowercases, substring-matches against channel name. Conservative default (no tokens → return true). Implementation is clean and well-bounded.
- ✅ **Test coverage** — 3 new tests: `ForceFitGuard` (bug reproduction), `LegitimateNarrow` (no false positive), `TestTopicOverlapsChannel` (8 cases). All pass.
- ✅ **Full `go test ./...`** — All packages pass, no regressions.

## 2. Findings

### F1 — P2/nit: Guard threshold `≥3` misses 2→1 narrowing

- **Location:** `resolve_channel.go:198` — `if len(candidates) >= 3 && len(result) == 1`
- **Diff-scope:** new (introduced by this PR)
- **Description:** If user has exactly 2 candidate channels and LLM force-fits to 1, the guard doesn't fire. The bug scenario in #79 specifically involves ≥3 candidates so this is not a gap in the reported fix, but the heuristic has a known blind spot at `len(candidates) == 2`.
- **Severity:** P2 — known limitation, not the reported bug scenario, safe fallback (worst case: slightly over-narrowed scope, not empty summary).
- **Suggestion:** Consider lowering to `>= 2` or document the intentional gap. Not blocking.

### F2 — P2/nit: Single-char ASCII keywords silently skipped

- **Location:** `resolve_channel.go:310` — `if i-start >= 2`
- **Diff-scope:** new
- **Description:** Topics referencing single-char project names (e.g., "X的项目") won't extract "X" as a token, so `topicOverlapsChannel` returns `true` (conservative). This means the guard won't fire even if LLM force-fits. Acceptable because the conservative direction is safe (shows more channels, not fewer), but worth noting.
- **Severity:** P2/nit — conservative direction is safe; single-char ASCII project names are rare in practice.

### F3 — P2/nit: No test for multi-rule union producing single force-fit result

- **Location:** `resolve_channel_test.go`
- **Diff-scope:** new
- **Description:** Existing tests only exercise single-rule scenarios. A case where two rules union to a single force-fit channel (e.g., rule1 selects ch1 via channel_ids, rule2 selects ch1 via person filter, union = {ch1}) would also trigger the guard. The guard handles this correctly, but no test demonstrates it.
- **Severity:** P2/nit — the guard code path is identical regardless of how `result` was constructed.

## 3. Suggestions

- Consider adding a comment near `len(candidates) >= 3` explaining why 3 (e.g., "2→1 is too common for legitimate narrow topics like 'my DM with Alice' when only 2 channels exist").
- The prompt additions are well-crafted. The regression example at L144-145 is particularly effective as a few-shot anchor for the LLM.

## 4. Additional Findings

None beyond F1–F3. The fix is well-scoped, the two-layer approach (prompt + deterministic guard) is sound defense-in-depth, and the conservative fallback direction (always safe — shows more channels, never fewer) is correct.

## 5. Data-Flow Tracing

| Consumed Data | Upstream Source | Traced Path | Verified |
|---|---|---|---|
| `topic` (string) | Caller in `fetch.go:646` → `ResolveAndFetchMessagesForPersonal` parameter | Passed unchanged to `ResolveChannelScope` → `callResolveChannelScope` (prompt) and `topicOverlapsChannel` | ✅ Flows correctly |
| `candidates[].ChannelName` | Set by Layer 1 (`GetUserChannels`) | `fetch.go:646` → `ResolveChannelScope` → `topicOverlapsChannel(topic, result[0].ChannelName)` | ✅ Populated by upstream DB query |
| `result` (filtered channels) | Built by `executeRule()` union loop (L175-188) iterating `parsed.Rules` | Each rule filters `candidates` by channel_ids/type/ownership/persons → union into `result` | ✅ Correct; guard checks `len(result)` after full union |
| `parsed.HasConstraint` / `parsed.Rules` | LLM structured output via `callResolveChannelScope` → JSON unmarshal | `toolCallFn` returns JSON → `json.Unmarshal` into `ChannelScopeResult` | ✅ Standard path; mock tests validate the flow |
| `topicOverlapsChannel` return value | Pure function of `topic` + `channelName` | ASCII token extraction → lowercase → `strings.Contains` | ✅ No external state; deterministic |

## 6. Blind-Spot Checklist (R5)

- **C1 (双路径 parity):** N/A — this PR only touches the channel scope narrowing path. No symmetric add/remove or create/delete paths involved.
- **C2 (control-flow ordering / 嵌套复用):** CLEAR — Guard is placed after the `len(result)==0` fallback and before the success return. No re-entrancy or nested invocation of `topicOverlapsChannel`. The function is a leaf helper with no side effects.
- **C3 (授权边界 ≠ 能力边界):** N/A — no authorization, permissions, or tool exposure changes.
- **C4 (授权生命周期 / 容器-成员状态级联):** N/A — no auth changes.
- **C5 (build/note ≠ 运行期路径):** N/A — pure Go backend change, no build artifacts, extensions, or relative URL concerns.
- **C6 (治理/策略文档自洽性):** N/A — no governance or policy documents changed.

## 7. Cross-Round Blocker Review (R6)

---

**[Octo-Q] verdict: APPROVE**

No P0 or P1 findings. The fix correctly addresses issue #79 with a sound two-layer approach (prompt improvement + deterministic Go guard). All three P2 findings are nits about known heuristic limitations, all in the safe-fallback direction. Tests are thorough and full suite passes clean.

yujiawei (changes requested):
# Code Review — PR #118 (octo-smart-summary)

## Summary

The fix targets a real bug (#79): when a topic names a keyword matching no candidate channel, the LLM force-fits the keyword onto a lexically-nearest channel, returning one valid-but-irrelevant channel → 0 messages → empty summary. The two-layer approach (prompt anti-force-fit rule + a deterministic Go guard) is the right shape, and the prompt changes and the `#79` reproduction are well done.

However, the deterministic guard as implemented introduces a **functional regression** in the existing channel-scope narrowing feature. It should be fixed before merge.

## Spec compliance

**Spec: ❌**

- Missing: none — both layers from the issue are present.
- Extra: none material.
- Deviation: Issue #79's proposed guard was explicitly scoped — *"when a rule's only effective constraint is `channel_ids`, validate the selected ids in Go and drop non-matching ones."* The implementation instead applies a **blanket** channel-name overlap check (`resolve_channel.go:201`) to **every** narrowing outcome, regardless of which dimension produced it (`channel_ids`, `channel_type`, `ownership`, or `persons`). This unscoped application is what causes the regression below.

## Code quality

### P1 — Force-fit guard is dimension-blind and reverts legitimate narrowing (`internal/pipeline/resolve_channel.go:201`, `topicOverlapsChannel` ~`:298-331`)

`ResolveChannelScope` narrows across four dimensions (`executeRule`: `channel_type`, `channel_ids`, `ownership`, `persons`). The guard validates the result **only** by `result[0].ChannelName` ASCII-token overlap. So a *legitimate* narrow-to-1 driven by a non-name dimension is wrongly reverted to all candidates whenever the topic carries an ASCII token that the channel name lacks.

Two concretely reproduced cases (both `FAIL` against this head):

1. **Person-scoped DM narrow.** DM channel names are always built as `私聊-<uid>` (`fetch.go:62`) — they never contain the person's display name. Topic `"和Alice聊了什么"` legitimately resolves (via the member map) to `channel_ids:[dm1]`, but `dm1`'s name `私聊-u_8a3f21c9` has no `alice` token → guard reverts to **all** channels. The user asked about their DM with Alice and instead gets a summary spanning every channel.

   ```
   [pipeline] ResolveChannelScope: narrowed to "私聊-u_8a3f21c9" but topic "和Alice聊了什么"
   has no keyword overlap, fallback to all 4 candidates (force-fit guard)
   ```

2. **CJK channel-name narrow + unrelated ASCII token.** Topic `"研发组Q3进展"` legitimately narrows to group `研发组`, but the extracted ASCII token is `q3`, absent from the CJK channel name → guard reverts to all. Any topic mixing a Chinese channel/person name with an incidental ASCII token (`Q3`, `PR`, `DM`, `OKR`…) hits this.

This is a regression: person-/type-/ownership-scoped narrowing is an existing, unit-tested feature (`TestResolveChannelScope_ByPersonUnion_IncludeSelf`, `TestResolveChannelScope_ByOwnership`), and this guard silently broadens it.

Scope note: candidates are creator-scoped (`fetch.go:623`, `GetUserChannels(ctx, creatorUID, …)`), so the fallback over-includes only within the *user's own* channels — there is **no cross-user privacy leak**. The impact is a wrong/over-broad summary, not data exposure.

**Suggested fix:** restrict the guard to the case the issue actually described — fire it only when the *only* effective constraint that produced the narrow was `channel_ids` (i.e. a name/keyword match). When the narrow came from `persons`/`channel_type`/`ownership`, the channel name is irrelevant and the guard must not run. Carrying the matched dimension out of `executeRule` (or validating against the constraint that fired) would close this cleanly.

### P2 — Guard only triggers on exactly one result (`resolve_channel.go:201`, `len(result) == 1`)

If the LLM force-fits **two** irrelevant channels instead of one, the guard is bypassed and the original #79 failure mode partially survives. Consider `len(result) <= K` or per-channel overlap filtering.

### P2 — `topicOverlapsChannel` fails open for pure-CJK topics (`resolve_channel.go` ~`:320`)

For an all-Chinese topic (e.g. `"财务部报表"`) no ASCII tokens are extracted, so the helper returns `true` (assume match) and the guard provides **zero** protection — which is the dominant input shape for this product. The original #79 repro (`总结 octo 的项目进展`) happens to contain `octo`, but Chinese-only force-fits won't be caught. This makes the fix incomplete for the common case rather than wrong; noting it so the limitation is explicit.

## Tests

Tests are well-constructed and pass (`go test ./internal/pipeline/...` green), but they only cover the channel-name happy path. They do not exercise narrowing via `persons`/`channel_type`/`ownership`, which is exactly where the P1 regression lives. Add cases for a person-driven narrow-to-1 DM and a CJK-name narrow with an incidental ASCII token in the topic.

## Verdict

The bug being fixed is real and the direction is sound, but the unscoped guard regresses an existing narrowing feature (P1). Recommend scoping the guard to channel-id/keyword-driven narrows before merge.

Jerry-Xin (changes requested):
This PR is in scope for `octo-smart-summary`, but the new guard can undo valid person-based narrowing and broaden summaries to unrelated channels.

**🔴 Blocking**

🔴 Critical: `internal/pipeline/resolve_channel.go:198` applies the channel-name overlap guard to **every** single-channel result, regardless of whether the narrowing came from `channel_ids`, `persons`, `include_self`, `channel_type`, or ownership. A valid request such as "What did Alice discuss?" can narrow through the person filter to one group named e.g. `project-updates`; since `alice` is not in the channel name, `topicOverlapsChannel` returns `false` and the resolver falls back to **all** candidates. This is a correctness **and privacy-sensitive** regression — summaries can then include channels outside the requested person scope. DMs make this worse: DM channels are named `私聊-<peerUID>` (see `internal/pipeline/fetch.go`), so a topic containing a display name will never overlap a DM channel name and will always trip the fallback. The guard should only run for single-channel results produced by channel-name / channel-id matching, or it should verify overlap against the actual constraint source (including member/display names), rather than the channel name alone.

**💬 Non-blocking**

🟡 The guard only runs when `len(candidates) >= 3`, so the original force-fit behavior can still occur with two candidate channels. May be an acceptable heuristic, but it leaves a known class of cases unfixed.

🟡 Please add regression coverage for valid person-based narrowing where the topic contains an English display name and the selected channel name does not contain that name (DM channels especially, since they are UID-named, not display-name-named).

**✅ Highlights**

- The prompt change (`resolveChannelScopeSystemPrompt` no-match / anti-force-fit rules + regression example) is directly aligned with #79.
- The new guard is correctly wired into the real resolve path (`ResolveChannelScope` is called from `internal/pipeline/fetch.go`), not dead code.
- Added tests cover the reported channel-name force-fit scenario and the legitimate-narrow counter-case.

Verification: `go build ./...` and `go test ./internal/pipeline/...` both pass locally (exit 0) on head `20c631b05f48332882ce6669f3554314c7a716ad`.

_Note: PR is a draft; the failing `dependency-review` CI check is the same pre-existing dependency issue also present on #117 and is not introduced by this PR's code — not treated as a blocking item against this change._

PR #145, opened 2026-07-09
feat(schedule): validate source access on Create/Update schedule (#143)

Ranwanglc:
## 概要
`UpdateSchedule` 直接把 `req.Sources` 写入 `source_config`（`internal/api/handler/schedule.go`），并未校验用户对新来源是否有访问权限；`CreateSchedule` 同样漏洞。攻击者/越权者可以把定时任务的来源改成自己无权访问的群/子区/私聊，等下次触发时越权拉取消息。

本 PR：
- 新增 `pipeline.ValidateUserAccessibleSources`，复用现有权威的 `pipeline.GetUserChannels` 拉全集再 diff。`imDB==nil` 直接放行（测试路径）。source_type（1群/2子区/3私聊）↔ pipeline ChannelType（2/5/1）映射封在 helper 内部；DM 归一化沿用 `NormalizeDMChannelID` 消除 peer/peer@self/self@peer 三形态。
- `ScheduleHandler` 加 `imDB` 字段 + 新构造 `NewScheduleHandlerWithIMDB`；router 切到新构造。旧 `NewScheduleHandler` / `NewScheduleHandlerWithFlag` 保留兼容既有测试。
- `UpdateSchedule` + `CreateSchedule` 在写库前调用 helper；不通过则 HTTP 403 / `Code: 40017 / Message: "存在无权访问或不存在的来源"`，`data.missing_sources` 逐项返回 `{source_type, source_id, source_name}`，前端可精确定位。message 保持通用文案避免文案注入面。
- 顺手清理死代码 `syncScheduledTaskConfig` / `syncScheduledTaskSources`（`internal/worker/scheduled_replace_helpers.go`）：`grep -RIn` 双确认无 caller、无测试引用；1→N 模型下每次触发都通过 `buildScheduledTaskChildren` 全量重建，原地增量路径不可达。

## 验收
- 编辑定时任务把来源改成用户**有权访问**的群 → 200 保存成功，`source_config` 更新。
- 改成用户**无权访问**的群 → 403 / 40017，`source_config` 不变。
- 创建定时任务给**有权/无权**来源 → 同样规则。

## 测试
- `internal/pipeline/access_test.go`：群/子区/DM 命中 + 未命中；`imDB==nil` 与空 sources 快速通道。
- `internal/api/handler/schedule_source_access_test.go`：Create × 有权/无权、Update × 有权/无权，共四条端到端 case，覆盖「拒绝路径下 `source_config` 不被改动」不变式。
[...]

PR #147, opened 2026-07-10
fix(schedule): 定时任务编辑消息来源加可访问性/权限校验 (#143 后端)

### LINKED ISSUE #143: 定时总结：支持在已创建的定时任务中修改消息来源
## 需求

智能总结的定时任务创建后，用户应能在后续编辑该定时任务时**修改其消息来源**（source：群/频道），修改后每次定时触发按新的消息来源执行。

## 现状

- API 层 `updateScheduleReq` 已含 `sources` 字段（source_type/source_id/source_name），结构上已支持传入新来源。
- 后端更新链路 `syncScheduledTaskSources` 会把 schedule 的 source 同步到定时生成的 task。
- 定时触发链路：`scanPendingSchedules` → `claimAndCreateScheduledTask`，使用 schedule 中存储的 source 配置。

## 待确认 / 待补

1. 前端编辑入口是否开放"修改消息来源"能力。
2. 更新 schedule 后，已排期的下一次执行是否正确应用新 source。
3. 修改来源是否需要权限/绑定校验（避免改成用户无权访问的群）。

## 验收标准

编辑已有定时任务 → 修改消息来源 → 保存 → 下一次定时触发使用新来源生成总结。

Ranwanglc:
## 背景 / Issue
Fixes #143 后端部分：定时任务编辑「消息来源」缺少可访问性/权限校验。

`CreateSchedule` / `UpdateSchedule` 此前把用户传入的 `sources` 直接写入 `source_config`，**未校验该用户是否有权访问这些群/子区/DM**。用户可把定时任务来源改成自己无权访问的渠道，后续定时触发即越权拉取那些渠道消息。

## 改动
对每个来源做 **fail-closed** 可访问性校验（`validateSourcesAccessible`），任一来源无权则整体 403（全或无），`source_config` 不落库：

- **群 (type=1)**：`group_member` 成员 + `group.status=1`（活跃群）。
- **子区 (type=2)**：`thread_member` 成员 + `thread.status IN (1,2)`（排除已删=3）+ 父群 `status=1`。
- **DM (type=3)**：`conversation_extra(uid,channel_id,channel_type=1)` 归属校验（与 `GetUserChannels` DM 发现路径同源），并拒绝空 `source_id`。
- 未知 type → 拒绝。
- `ScheduleHandler` 注入 `imDB` + 可注入 `collate`（生产默认 MySQL COLLATE，测试置空以兼容 sqlite）。
- Create/Update 两路复用同一校验函数。

## 测试
新增 10 个用例（合并进 `schedule_behavior_test.go`）：群有权/无权、子区有权/无权/已删、混合来源全或无、DM 空 id/非本人/本人、imDB=nil 跳过。`go build ./...` 与 `go test ./internal/api/handler/... -run "Schedule|SourceAccess"` 全绿。

## 部署前提 / 必配环境变量
无新增环境变量、**无 DB schema 变更 / 无新 migration**。仅要求 API 容器的 IM 数据库连接（`imDB`）已正常注入（现有 `SetupPublic` 已具备），线上默认满足。

## 说明
- 纯 API 侧校验逻辑，worker 未改动。
- 错误信息统一为「无权将来源修改为无权访问的群/子区」，不泄漏具体资源是否存在。

### DIFF
--- a/internal/api/handler/schedule.go
+++ b/internal/api/handler/schedule.go
@@ -6,6 +6,7 @@ import (
 	"log"
 	"net/http"
 	"strconv"
+	"strings"
 	"time"
 	"unicode/utf8"
 
@@ -22,21 +23,174 @@ import (
 // ScheduleHandler handles schedule endpoints.
 type ScheduleHandler struct {
 	db *gorm.DB
+	// imDB is the IM database used to verify that a user actually has access to
+	// the group/thread sources they attach to a schedule (#143). When nil the
+	// source-access check is skipped (fail-open) with a warning; production
+	// wiring in router.SetupPublic always injects it, so fail-open is limited to
+	// tests/tools that construct the handler without an IM DB.
+	imDB *gorm.DB
 	// featureTeamSchedule, when true, bypasses the multi-person rejection guards
 	// (FEATURE_TEAM_SCHEDULE). Default false keeps the existing 40015 behavior.
 	featureTeamSchedule bool
+	// collate is the SQL COLLATE clause used for cross-table group_no comparisons
+	// in the IM DB (MySQL collation mismatch between `thread`/`group` and their
+	// member tables). Production wiring uses scheduleSourceCollate; tests against
+	// sqlite set it to "" so the thread/group-status path can actually run (sqlite
+	// rejects the MySQL-only COLLATE name). Mirrors CandidateHandler.collate.
+	collate string
 }
 
 // NewScheduleHandler creates a new ScheduleHandler.
+//
+// Deprecated for production wiring: it leaves imDB nil, which disables the
+// source-access check. Kept for tests/tools; production must use
+// NewScheduleHandlerWithIMDB / NewScheduleHandlerWithFlag so imDB is injected.
 func NewScheduleHandler(db *gorm.DB) *ScheduleHandler {
-	return &ScheduleHandler{db: db}
+	return &ScheduleHandler{db: db, collate: scheduleSourceCollate}
 }
 
 // NewScheduleHandlerWithFlag creates a ScheduleHandler with the team-schedule
 // feature flag explicitly set. When featureTeamSchedule is true the multi-person
 // rejection guards are bypassed so multi-participant schedules can be created.
+//
+// Deprecated for production wiring: prefer NewScheduleHandlerWithIMDB which also
+// injects imDB for the source-access check. Retained for existing test callers.
 func NewScheduleHandlerWithFlag(db *gorm.DB, featureTeamSchedule bool) *ScheduleHandler {
-	return &ScheduleHandler{db: db, featureTeamSchedule: featureTeamSchedule}
+	return &ScheduleHandler{db: db, featureTeamSchedule: featureTeamSchedule, collate: scheduleSourceCollate}
+}
+
+// NewScheduleHandlerWithIMDB creates a ScheduleHandler with both the IM DB (for
+// the #143 source-access check) and the team-schedule feature flag. This is the
+// constructor production wiring must use so schedule sources are validated
+// against real group/thread membership.
+func NewScheduleHandlerWithIMDB(db *gorm.DB, imDB *gorm.DB, featureTeamSchedule bool) *ScheduleHandler {
+	return &ScheduleHandler{db: db, imDB: imDB, featureTeamSchedule: featureTeamSchedule, collate: scheduleSourceCollate}
+}
+
+// scheduleSourceCollate is the COLLATE clause required for cross-table group_no
+// comparisons in the IM DB (MySQL collation mismatch between `thread`/`group`
+// and `group_member`). Mirrors CandidateHandler.collate and the raw SQL in
+// pipeline/resolve_channel.go and service/summary.go. Production wiring passes
+// this via ScheduleHandler.collate; sqlite tests pass "" (see F3).
+const scheduleSourceCollate = " COLLATE utf8mb4_unicode_ci"
+
+// validateSourcesAccessible verifies that userID may access every source in the
+// list, using IM DB membership. Fail-closed: any source the user cannot be
+// shown to access returns an error. When imDB is nil the check is skipped
+// (fail-open) with a warning — production always injects imDB.
+//
+// collate is the cross-table COLLATE clause (scheduleSourceCollate in production,
+// "" under sqlite tests) applied to every group_no join, mirroring
+// pipeline/fetch.go (GetUserChannels) and candidates.go so the same membership
+// semantics are enforced here.
+//
+//   - source_type=1 (group): user must be a non-deleted member of a live group
+//     (g.status=1). source_id may carry a "____{space_id}" suffix; only the
+//     group_no prefix is used (see ResolveSourceNameWithType case 1).
+//   - source_type=2 (thread/子区): source_id is "{group_no}____{short_id}"; the
+//     user must be a member of a non-deleted thread (t.status IN (1,2), i.e.
+//     Active|Archived, NOT Deleted=3) whose parent group is live (g.status=1).
+//     group_no is compared across tables with COLLATE. A deleted thread with a
+//     residual thread_member row is rejected (F2).
+//   - source_type=3 (DM): fail-closed. An empty source_id is rejected (F1); a
+//     non-empty peer must be a conversation the user actually holds — verified
+//     against conversation_extra (uid=current, channel_id=peer, channel_type=1),
+//     the same DM-ownership source of truth as GetUserChannels. Unknown types
+//     are rejected fail-closed.
+func validateSourcesAccessible(imDB *gorm.DB, collate string, userID string, sources []sourceReq) error {
+	if imDB == nil {
+		log.Printf("[schedule] validateSourcesAccessible: imDB not injected, skipping source-access check (fail-open) for user=%q sources=%d", userID, len(sources))
+		return nil
+	}
+	if userID == "" {
+		return errors.New("missing user id for source access check")
+	}
+	for _, s := range sources {
+		switch s.SourceType {
+		case 1: // group
+			groupNo := s.SourceID
+			if idx := strings.Index(groupNo, "____"); idx > 0 {
+				groupNo = groupNo[:idx]
+			}
+			if groupNo == "" {
+				return errors.New("invalid group source id")
+			}
+			var cnt int64
+			// Membership + live-group guard (mirrors GetUserChannels: g.status=1 and
+			// gm.is_deleted=0). group_no is compared across `group`/group_member with
+			// COLLATE, matching candidates.go's group-thread joins.
+			if err := imDB.Raw(
+				"SELECT 1 FROM group_member gm "+
+					"INNER JOIN `group` g ON g.group_no"+collate+" = gm.group_no "+
+					"WHERE gm.uid = ? AND gm.group_no = ? AND gm.is_deleted = 0 AND g.status = 1 "+
+					"LIMIT 1",
+				userID, groupNo,
+			).Scan(&cnt).Error; err != nil {
+				log.Printf("[schedule] validateSourcesAccessible: group query failed user=%q group=%q: %v", userID, groupNo, err)
+				return err
+			}
+			if cnt == 0 {
+				return errors.New("user not a member of group source")
+			}
+		case 2: // thread / 子区
+			parts := strings.SplitN(s.SourceID, "____", 2)
+			if len(parts) != 2 || parts[0] == "" || parts[1] == "" {
+				return errors.New("invalid thread source id")
+			}
+			groupNo := parts[0]
+			shortID := parts[1]
+			var cnt int64
+			// Membership of a NON-DELETED thread whose parent group is live (F2):
+			//   - t.status IN (1,2): Active|Archived; Deleted=3 is excluded so a
+			//     residual thread_member row on a deleted thread no longer passes.
+			//   - g.status=1: parent group must still be live.
+			// group_no is compared across `group`/thread with COLLATE, matching
+			// GetUserChannels' thread join. The literal group_no bind on t.group_no
+			// disambiguates the (group_no, short_id) composite key.
+			if err := imDB.Raw(
+				"SELECT 1 FROM thread t "+
+					"INNER JOIN thread_member tm ON tm.thread_id = t.id "+
+					"INNER JOIN `group` g ON g.group_no"+collate+" = t.group_no "+
+					"WHERE tm.uid = ? AND t.group_no"+collate+" = ? AND t.short_id = ? "+
+					"AND t.status IN (1, 2) AND g.status = 1 "+
+					"LIMIT 1",
+				userID, groupNo, shortID,
+			).Scan(&cnt).Error; err != nil {
+				log.Printf("[schedule] validateSourcesAccessible: thread query failed user=%q group=%q short=%q: %v", userID, groupNo, shortID, err)
+				return err
+			}
+			if cnt == 0 {
+				return errors.New("user not a member of thread source")
+			}
+		case 3: // DM / 私聊
+			// F1 fail-closed: never accept an empty DM peer id (empty id would also
+			// break later source materialization). source_id is the peer uid, matching
+			// ResolveSourceNameWithType case 3 and conversation_extra.channel_id.
+			peer := s.SourceID
+			if peer == "" {
+				return errors.New("invalid dm source id")
+			}
+			var cnt int64
+			// Deep ownership check: the current user must actually hold this P2P
+			// conversation. conversation_extra(uid=current, channel_id=peer,
+			// channel_type=1) is the same DM source of truth GetUserChannels reads
+			// (internal/pipeline/fetch.go). No arbitrary/other users' DM peer can be
+			// attached.
+			if err := imDB.Raw(
+				"SELECT 1 FROM conversation_extra WHERE uid = ? AND channel_id = ? AND channel_type = 1 LIMIT 1",
+				userID, peer,
+			).Scan(&cnt).Error; err != nil {
+				log.Printf("[schedule] validateSourcesAccessible: dm query failed user=%q peer=%q: %v", userID, peer, err)
+				return err
+			}
+			if cnt == 0 {
+				return errors.New("user does not own dm source conversation")
+			}
+		default:
+			return errors.New("unknown source type")
+		}
+	}
+	return nil
 }
 
 type createScheduleReq struct {
@@ -536,6 +690,11 @@ func (h *ScheduleHandler) CreateSchedule(c *gin.Context) {
 
 	var sourceConfig model.JSON
 	if len(req.Sources) > 0 {
+		if err := validateSourcesAccessible(h.imDB, h.collate, userID, req.Sources); err != nil {
+			log.Printf("[schedule] CreateSchedule: source access denied user=%q: %v", userID, err)
+			c.JSON(http.StatusForbidden, apiResponse{Code: 40004, Message: "无权将来源修改为无权访问的群/子区"})
+			return
+		}
 		b, _ := json.Marshal(req.Sources)
 		sourceConfig = b
 	}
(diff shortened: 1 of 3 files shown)


internal/api/handler/schedule.go
```diff
@@ -22,21 +23,174 @@ import (
...
+//     Active|Archived, NOT Deleted=3) whose parent group is live (g.status=1).
+//     group_no is compared across tables with COLLATE. A deleted thread with a
+//     residual thread_member row is rejected (F2).
+//   - source_type=3 (DM): fail-closed. An empty source_id is rejected (F1); a
+//     non-empty peer must be a conversation the user actually holds — verified
+//     against conversation_extra (uid=current, channel_id=peer, channel_type=1),
+//     the same DM-ownership source of truth as GetUserChannels. Unknown types
+//     are rejected fail-closed.
+func validateSourcesAccessible(imDB *gorm.DB, collate string, userID string, sources []sourceReq) error {
+	if imDB == nil {
```
OctoBoooot, line 101: **[blocker]** `validateSourcesAccessible` fails OPEN on `imDB == nil` (returns nil = access granted), and nil is prod-reachable: `cmd/summary-api/main.go:61-64` sets `imDB = nil` when the IM DB connect fails (logged "non-fatal"), and `router.SetupPublic` passes it straight into `NewScheduleHandlerWithIMDB`. During an IM DB outage the access check is skipped and any user can set any group/thread/DM source — the exact IDOR this PR closes — and the unauthorized `source_config` persists past the outage. The doc comment ("production always injects it") is contradicted by main.go. This is the same fail-open yujiawei flagged as a blocker on #145, which was resolved there by failing CLOSED (return error → 500); the rewrite lost it.

Failing input: IM DB down at startup → imDB=nil; `PUT /summary-schedules/{own_id}` with `sources:[{source_type:1, source_id:"<group not a member of>"}]` → returns nil → 200, unauthorized source persisted.

Fix: fail closed on nil imDB (return an error; caller surfaces 500), matching #145. If a no-IM-DB test mode is needed, gate it behind an explicit flag, not the nil default; flip `TestUpdateSchedule_SourceAccess_NilIMDBSkipsCheck` to assert the closed behavior.

internal/api/handler/schedule.go
```diff
@@ -22,21 +23,174 @@ import (
...
+				return errors.New("invalid dm source id")
+			}
+			var cnt int64
+			// Deep ownership check: the current user must actually hold this P2P
+			// conversation. conversation_extra(uid=current, channel_id=peer,
+			// channel_type=1) is the same DM source of truth GetUserChannels reads
+			// (internal/pipeline/fetch.go). No arbitrary/other users' DM peer can be
+			// attached.
+			if err := imDB.Raw(
+				"SELECT 1 FROM conversation_extra WHERE uid = ? AND channel_id = ? AND channel_type = 1 LIMIT 1",
```
OctoBoooot, line 180: **[praise]** DM ownership check is the correct fix for the issue that dogged #145: `peer := s.SourceID` (raw peer uid) matched against `conversation_extra WHERE channel_id = peer AND channel_type = 1` — the raw-form comparison the picker storage contract (candidates.go:212, "channel_id … is the peer's uid") actually requires, with the test seeding raw `peerA` on both sides. The normalized-`a@b`-vs-raw mismatch that false-403'd legit DMs on #145 is gone, empty id is rejected fail-closed, and the F2 deleted-thread (status=3) + all-or-nothing contracts each have a targeted test.

OctoBoooot (changes requested):
## Review: fix(schedule): validate source access on Create/Update schedule (#147)

**Verdict:** Request changes

This is the reworked #143 backend (replacing #145's pipeline/access.go approach with an in-handler `validateSourcesAccessible`). The DM root cause that held #145 is **fixed correctly here** — but the fail-open-on-nil-imDB security bypass that #145 explicitly resolved to fail-closed has been **reintroduced** in this new implementation. That's the one blocker.

Note: this PR is a **draft** — functional CI (Build/Test/Vet/Lint/race) is `skipped` (draft-gated, ci.yml:74-147), so the code has not actually been compiled/tested by CI. Governance (check-sprint, dependency-review, secret-scan, osv-scan) is green. `mergeStateStatus` is BLOCKED (review required + draft).

Risk tier: high (API surface). `needs-human-review` applies.

### Blocker
- `internal/api/handler/schedule.go:~78` — `validateSourcesAccessible` **fails open** when `imDB == nil` (`return nil` = access granted), and nil is prod-reachable: `cmd/summary-api/main.go:61-64` sets `imDB = nil` when the IM DB connect fails (logged "non-fatal"), and `router.SetupPublic` passes that straight into `NewScheduleHandlerWithIMDB(db, imDB, …)`. So during an IM DB outage the access check is skipped and **any user can set any group/thread/DM source** — the exact IDOR this PR exists to close — and the unauthorized `source_config` persists, so the exposure outlives the outage (the scheduled task later pulls those sources once IM DB recovers). The code comment ("production wiring always injects it, so fail-open is limited to tests/tools") states a premise main.go contradicts. This is the same fail-open yujiawei flagged as a blocker on #145, which was resolved there by failing **closed** (return an error → HTTP 500); the rewrite lost that fix. _(confidence: high)_
  - **Failing input**: IM DB down at API startup → `imDB=nil`. Attacker calls `PUT /api/v1/summary-schedules/{own_id}` with `sources:[{source_type:1, source_id:"<group they're not in>"}]` → `validateSourcesAccessible` returns nil → 200, source_config persists the unauthorized group.
  - **Smallest fix**: fail closed on `imDB == nil` (return an error; caller surfaces 500), matching #145's resolved behavior. If a genuine no-IM-DB test/tool mode is needed, gate it behind an explicit opt-in flag, not the nil default. `TestUpdateSchedule_SourceAccess_NilIMDBSkipsCheck` then flips to asserting the closed behavior.

### Resolved vs #145 ✓ (verified)
- **DM check now matches the storage contract** — `peer := s.SourceID` (raw peer uid) is matched against `conversation_extra WHERE channel_id = peer AND channel_type = 1`. This is the raw-form comparison the picker contract (candidates.go:212, "channel_id … is the peer's uid") requires, and the test seeds the raw `"peerA"` on both sides (`schedule_behavior_test.go`). The normalized-`a@b`-vs-raw mismatch that false-403'd legit DMs on #145 is gone. Empty DM id rejected (F1).
- **Group / thread checks** are sound: membership + live-group (`g.status=1`, `gm.is_deleted=0`) for groups; thread membership with `t.status IN (1,2)` (excludes deleted=3, F2) + live parent group; cross-table `group_no` COLLATE matches candidates.go. All-or-nothing (any denied source → whole 403, source_config not persisted) is enforced at both Create and Update.

### Praise
- The DM ownership check is exactly the fix the prior PR needed — raw peer uid against the real `conversation_extra` store, with the test seeding the production format rather than the code's assumed one. The F2 deleted-thread-with-residual-membership rejection and the all-or-nothing contract are the right edge cases, and each has a targeted test. The one gap is the nil-imDB direction: a fail-open on a security gate should fail closed.

Jerry-Xin (changes requested):
Summary: The PR is in scope and the membership checks are generally aligned with the schedule source authorization issue, but the new security control still fails open when the API has no IM DB connection.

**🔴 Blocking**

🔴 Critical — Source authorization is skipped when `imDB` is nil, reopening the exact permission bypass this PR is meant to close. [internal/api/handler/schedule.go:100](/Users/mlamp/.openclaw/tmp/pr-review-stftge_z/repo/internal/api/handler/schedule.go:100) returns `nil` on missing `imDB`, and [cmd/summary-api/main.go:60](/Users/mlamp/.openclaw/tmp/pr-review-stftge_z/repo/cmd/summary-api/main.go:60) treats IM DB connection failure as non-fatal before [internal/api/router/router.go:41](/Users/mlamp/.openclaw/tmp/pr-review-stftge_z/repo/internal/api/router/router.go:41) injects that nil into the schedule handler. In that state, authenticated users can create/update schedules with arbitrary `sources`, and the test at [internal/api/handler/schedule_behavior_test.go:1417](/Users/mlamp/.openclaw/tmp/pr-review-stftge_z/repo/internal/api/handler/schedule_behavior_test.go:1417) explicitly locks in this fail-open behavior. For a permission check, this should fail closed: either make API startup fail if IM DB is unavailable, or make `validateSourcesAccessible` return an error when `imDB == nil` for create/update source writes.

**💬 Non-blocking**

🟡 Warning — The new denial matrix is update-heavy. The create path is only exercised through an accessible setup helper at [internal/api/handler/schedule_behavior_test.go:1352](/Users/mlamp/.openclaw/tmp/pr-review-stftge_z/repo/internal/api/handler/schedule_behavior_test.go:1352), but there is no explicit create-denied test for inaccessible group/thread/DM sources. Since create and update share the validator this is not a blocker, but a direct create-denial regression test would make the security contract clearer.

🔵 Suggestion — The public error message still names only groups/subsections even when DM validation fails at [internal/api/handler/schedule.go:179](/Users/mlamp/.openclaw/tmp/pr-review-stftge_z/repo/internal/api/handler/schedule.go:179). Consider making the localized message cover “message sources” generically without revealing resource existence.

**✅ Highlights**

The group, thread, unknown-type, empty-DM, and all-or-nothing validation shape is good, and the update path avoids persisting rejected `source_config`.

Verification:
- `go test ./internal/api/handler/... -run "Schedule|SourceAccess"` passed.
- `go test ./internal/pipeline/...` passed.
- `go build ./...` could not complete locally because the linker could not find `libtokenizers`; this appears environment-related.

yujiawei (approved):
# Code Review — PR #147 (octo-smart-summary)

**Scope reviewed:** head SHA `bca7434957e0d49e8ee8601b926e54aaa0bd588e` vs `main` base. 3 files, +490/-3. Security-sensitive (auth) — reviewed with an adversarial pass and cross-file verification against the runtime data path.

## Verdict: Approve ✅

No correctness/security blockers. The fix is well-targeted, fail-closed, and its membership semantics line up with the runtime fetch path. The findings below are all non-blocking (test-coverage + observability hardening).

---

## 1. Spec compliance (#143)

The stated goal — reject schedule sources the creator cannot access on both create and update — is fully met, with no over-build:

- **All three source types gated on both paths.** `validateSourcesAccessible` handles group (`case 1`), thread (`case 2`), DM (`case 3`), with a fail-closed `default → "unknown source type"`. Both `CreateSchedule` (schedule.go:692–698) and `UpdateSchedule` (schedule.go:1298–1304) call it with identical args before persisting `source_config`.
- **Fail-closed + all-or-nothing.** Any inaccessible source returns 403 (code 40004) and `source_config` is not written. Verified: the only two `source_config` writers are these two handlers; no clone/import/copy path bypasses the check. The manual-summary path in `task.go` is a separate pre-existing feature out of #143 scope.
- **No over-build / no schema change.** Diff adds only two functions plus the constructor swap in `router.SetupPublic` (`imDB` was already a `SetupPublic` param). No migration, no new endpoint/field, no new env var. `sourceReq` is reused from `task.go:202`, not redefined.
- **Membership semantics mirror the runtime source of truth** (`internal/pipeline/fetch.go` `GetUserChannels`): group = `group_member` + `g.status=1` + `is_deleted=0`; thread = `thread_member` + `t.status IN (1,2)` + live parent group; DM = `conversation_extra(uid, channel_id=peer, channel_type=1)`, where `channel_id` is the bare peer uid (consistent with `candidates.go:214`). The `_type=2` thread source_id `{group_no}____{short_id}` and the group `____{space_id}` suffix stripping both match `ResolveSourceNameWithType`.

**Spec: ✅** — no missing / over-built / deviating behavior.

## 2. Code quality

**Quality: Approved.**

Notably solid: raw SQL uses bound parameters throughout (no injection surface), the `____` split guards against empty group_no/short_id, DM empty-id is explicitly rejected, and there is a defense-in-depth backstop — at trigger time `ApplySourceConstraints` (fetch.go:277) re-intersects stored sources with live `GetUserChannels`, so even a source that somehow bypassed validation would pull zero messages. This bounds the blast radius of any residual gap in this check.

### Non-blocking findings (P2 / nit)

**P2 — Production `COLLATE` clause is not exercised by tests.** The sqlite test harness blanks `sh.collate = ""` (schedule_behavior_test.go:1344), so the production string `t.group_no COLLATE utf8mb4_unicode_ci = ?` never runs against MySQL in CI. The two collated *joins* are byte-identical to already-in-prod queries (fetch.go:159, candidates.go:177/190), so they are effectively prod-validated — but the `COLLATE ... = ?` applied to a **bound parameter** in the WHERE clause (schedule.go thread branch) is a construction with no prod precedent and is only ever run with an empty collate. Worth a MySQL-backed test or a lint that the collate string parses.

**P2 — `imDB == nil` is fail-open.** When `imDB` is nil the check is skipped entirely (`return nil`). This is documented and only reachable via the two `Deprecated`-annotated constructors; the sole prod caller uses `NewScheduleHandlerWithIMDB`. Consider a startup assertion (or removing the deprecated constructors) so a future wiring regression can't silently disable the gate. Backstopped at runtime by `ApplySourceConstraints`.

**P2 — DB/infra error is conflated with authz denial.** Every query error path does `return err`, and both call sites map any non-nil return to `403 / 40004 "无权..."`. This is correctly fail-closed (an IM-DB outage denies rather than allows), but an outage will surface to operators as "permission denied" rather than 500. Structured logs already capture the underlying error, so this is observability polish, not a bug.

**P2 — Group rejection branches under-tested.** The group query carries `AND gm.is_deleted = 0 AND g.status = 1`, but the only group-denial test (`DeniedGroupRejected`) denies via a *missing* member row. A regression dropping either guard would still pass all current tests. The thread path does test its analogous `status=3` guard (`DeletedThreadRejected`); the group path deserves the same (member of a `status!=1` group, and an `is_deleted=1` row).

**nit — Redundant/inconsistent COLLATE placement.** The group join adds `COLLATE` on the `group ↔ group_member` join, which all three runtime paths (fetch.go:90, candidates.go:145, resolve_channel.go:371) omit; and the thread query collates the `t.group_no = ?` bind, which no runtime path does and which its own group query does not. Harmless (values are system IDs; runtime Go-exact re-filter is the real gate; direction skews toward under-denial, not blocking legit users), but mirror the runtime queries to avoid confusion — the inline comment claiming it "matches candidates.go's group-thread joins" is inaccurate for a group↔group_member join.

**nit — `CreateSchedule` denial path & archived-thread (`status=2`) accept path have no direct test.** The shared validator is exhaustively covered via the six `UpdateSchedule` tests, and `status IN (1,2)` is intentional and consistent with the run-time archived-thread relaxation (thread sources are always passed as `selectedThreadIDs`), so nothing is broken — but a `CreateSchedule` 403 test and a `status=2` accept test would pin the symmetric wiring.

## 3. Blind spots for a human to confirm

- The `COLLATE ... = ?` bound-parameter WHERE clause runs correctly against **real MySQL** with the production collation (only exercised with empty collate in CI).
- All prod handler wiring keeps `imDB` non-nil (no path constructs the schedule handler via the deprecated nil-imDB constructors).
- `conversation_extra.channel_id` for P2P DMs stores the **bare peer uid** in your production data (the check and `candidates.go` both assume this; `GetUserChannels` additionally `@`-splits, which is a display-normalization concern, not a membership one).

None of these are blockers — the runtime `ApplySourceConstraints` re-filter is a second independent membership gate on the actual message-fetch path.

yujiawei (changes requested):
# Code Review — PR #147 (octo-smart-summary)

**Scope:** head SHA `bca7434957e0d49e8ee8601b926e54aaa0bd588e` vs `main`. 3 files, +490/-3. Security-sensitive (auth/IDOR).

> **Correction to my earlier Approve:** I initially approved after verifying the membership-check logic in isolation. On a cross-file re-check of the handler's wiring (`cmd/summary-api/main.go`), the fail-open path is **production-reachable**, which two other reviewers correctly flagged as blocking. I'm superseding my approval with the verdict below. Apologies for the churn.

## Verdict: Request changes ❌

The membership-validation logic is correct and well-tested, but the gate **fails open** in a production-reachable state, which reopens the exact IDOR this PR closes.

---

## 1. Spec compliance (#143)

The intended behavior is implemented for the normal path: all three source types (group/thread/DM) are validated on both `CreateSchedule` and `UpdateSchedule`, fail-closed per-source, all-or-nothing, `source_config` not persisted on denial, `default → unknown type` rejected. No schema/migration/endpoint over-build. Membership semantics mirror the runtime source of truth in `internal/pipeline/fetch.go` (`GetUserChannels`): group = `group_member` + `g.status=1` + `is_deleted=0`; thread = `thread_member` + `t.status IN (1,2)` + live parent; DM = `conversation_extra(uid, channel_id=peer, channel_type=1)` with the bare peer uid (matches `candidates.go:214`, and the DM `a@b`-vs-raw mismatch from the prior attempt is gone).

**Spec: ❌** — deviation: the security control does not hold when `imDB` is nil, and that state is reachable in production (detail below). The check must be fail-closed to satisfy #143's intent.

## 2. Code quality

**Quality: Changes-Requested.**

### 🔴 P0 — `validateSourcesAccessible` fails open when `imDB == nil`, and nil is production-reachable

`schedule.go` (the `if imDB == nil { ... return nil }` guard) grants access when the IM DB handle is nil. The inline comment asserts *"production wiring in router.SetupPublic always injects it, so fail-open is limited to tests/tools"* — but `cmd/summary-api/main.go:61-64` treats an IM DB connect failure as **non-fatal**:

```go
imDB, err := db.New(cfg.IMMySQLDSN)
if err != nil {
    log.Printf("[main] connect IM DB (non-fatal): %v", err)
    imDB = nil
}
```

and then `main.go:98` passes that `imDB` straight into `router.SetupPublic → NewScheduleHandlerWithIMDB(db, imDB, …)`. So if the IM DB is unavailable at API startup, the schedule handler runs with `imDB == nil` and the access check is silently skipped.

**Concrete exploit (IM DB down at API startup):** an authenticated user calls `PUT /api/v1/summary-schedules/{their_own_id}` with `sources:[{source_type:1, source_id:"<a group they are not in>"}]` → `validateSourcesAccessible` returns nil → 200, and the unauthorized `source_config` is **persisted**. The bad config outlives the outage.

Blast radius is partially bounded — the worker re-intersects stored sources against live membership at trigger time (`ApplySourceConstraints` in `fetch.go:277`, and the worker's own `imDB` is fatal-on-failure, so it won't summarize with a nil handle) — so the *message-fetch* leak is contained. But that is defense-in-depth on a different process, not a substitute for the gate: the IDOR write still succeeds, unauthorized config is stored, and a security control that is supposed to deny is instead granting. This is also a **regression**: the prior approach (#145) was explicitly resolved to fail *closed*; this rewrite lost that.

**Fix:** fail closed on `imDB == nil` — return an error so the caller surfaces 500 (or make API startup fatal when the IM DB is unavailable, matching the worker). If a genuine no-IM-DB test/tool mode is needed, gate it behind an explicit opt-in flag rather than the nil default. `TestUpdateSchedule_SourceAccess_NilIMDBSkipsCheck` should then flip to asserting the closed behavior.

### 🟡 P2 — non-blocking hardening

- **Production `COLLATE` clause is untested.** The sqlite harness blanks `sh.collate = ""` (test:1344), so the prod string `t.group_no COLLATE utf8mb4_unicode_ci = ?` never runs against MySQL. The collated *joins* mirror in-prod queries (fetch.go:159, candidates.go:177/190), but the `COLLATE … = ?` applied to a **bound parameter** is novel to this PR. A MySQL-backed test would close it.
- **Group rejection branches under-tested.** The only group-denial test denies via a *missing* member row, so a regression dropping `AND gm.is_deleted = 0` or `AND g.status = 1` would still pass. The thread path tests its analogous `status=3` guard; the group path should too.
- **No direct `CreateSchedule`-denied test.** The validator is shared, so it's covered transitively, but a create-denial regression test would pin the symmetric wiring.
- **DB/infra error conflated with authz denial (nit).** Query errors `return err`, mapped by both call sites to `403/40004 "无权…"`. Correctly fail-closed, but an IM-DB outage reads as "permission denied" to operators. Logs already capture the underlying error.

### 🔵 nits

- **Inconsistent COLLATE placement.** The group join collates `group ↔ group_member` (runtime paths at fetch.go:90 / candidates.go:145 / resolve_channel.go:371 do not), and the thread query collates the `t.group_no = ?` bind (no runtime path does). Harmless (system-generated IDs, runtime Go-exact re-filter is the real gate) but mirror the runtime queries; the "matches candidates.go's group-thread joins" comment is inaccurate for a group↔group_member join.
- **Error message wording** (schedule.go:179): the 403 message names only groups/subsections even on DM-validation failure. Consider a generic "message source" wording (without revealing resource existence).

## 3. Blind spots for a human to confirm

- The nil-`imDB` fail-open is the load-bearing issue — confirm the fix fails closed and that no other handler relies on the nil-passthrough for a legitimate degraded mode.
- **CI note:** this PR is a draft, so functional CI (Build/Test/Vet/Lint/race) is `skipped` (draft-gated). The code has not been compiled/tested by CI — `go build ./...` + the schedule tests do pass locally.

PR #153, opened 2026-07-14
feat(notify): deliver summary notifications as octo/v1 cards (#150 worker side)

### LINKED ISSUE #150: 总结完成通知支持点击跳转到总结详情页
总结完成后的通知目前不能点击直接跳转到总结详情页。

希望：在同一个页面中，点击某一条「总结完成」的通知，能直接跳转到该总结对应的详情页。

Ranwanglc:
## Summary
Switches smart-summary terminal-state notifications (completed/failed) from a plain-text DM to the octo-server **structured card** path introduced by octo-server #579/#580 (`POST /v1/internal/notify` `card` field). smart-summary sends raw fields only; octo-server owns card layout/labels and the `/s/{task_no}?sp={space_id}` deep link.

Closes #150 (worker side). Pairs with octo-web `/s/:taskNo` route (separate PR) and the smart-summary `task_no` detail lookup (separate PR).

## Why
The old plain text appended an unreachable `/summary/detail?taskId=` link. Moving to the server card gives a real "查看详情" button whose deep link (`/s/{task_no}?sp=`) opens the detail page once the web route lands.

## Info parity (owner requirement: card must carry no less than the text)
- The server card has no dedicated space-name field, so the **space name is folded into the card title**: `空间「X」·<title>` / `空间「X」总结` / `<title>` / `总结`.
- completed: title(+space), 时间范围, 参与成员, 消息数量, 生成时间, + 查看详情 button.
- failed: title(+space), 失败原因 (sanitized).
- `members`/`msg_count` omitted when `<=0`; `reason` is failed-only (`omitempty`).

## Behavior preserved
- `payload` and `card` are mutually exclusive — a card send omits `payload` entirely.
- claim/markSent/markFailed/Sweep dedup state machine unchanged; the sweep **redeliver** path also sends the card (no stale text).
- Failure reason still goes through the error sanitizer.

## Deployment prerequisites — required env vars (unchanged from #135, listed for completeness)
| Var | Default | Effect if unset |
|---|---|---|
| `SUMMARY_NOTIFY_ENABLED` | `false` | Master switch. Off ⇒ notifier=nil, terminal callback returns early, **no notification at all** (silent; dev only warns). |
| `SUMMARY_NOTIFY_TOKEN` | — | `X-Internal-Token` for `/v1/internal/notify`; must equal octo-server `NOTIFY_INTERNAL_TOKEN`. Prod fatal / dev warn if missing while enabled. |
| `OCTO_API_URL` | — | internal-notify delivery target. Empty ⇒ notifications disabled (warning). |
| `APP_ENV` | `dev` | `prod` ⇒ fatal on missing token; `dev` ⇒ warn. |

Card deep link origin is owned by **octo-server** (`external.webLoginURL`, must be https) — a non-https value degrades the card to plain text server-side. `SUMMARY_WEB_BASE_URL` on the worker is now unused by the link (retained for config compat).

## Testing
`go build ./...` clean; `go test ./internal/notify/... ./internal/worker/... -count=1` green. Tests assert card shape, title folding, wire omission of `reason`/`members`/`msg_count`, payload absence on card sends, and the sanitized failure path. Independently re-verified + adversarially reviewed (2 review rounds, clean).

### DIFF
--- a/internal/notify/client.go
+++ b/internal/notify/client.go
@@ -37,13 +37,24 @@ type Deliverer interface {
 }
 
 // SendMessageRequest is the state-machine-internal message shape. deliver()
-// fills ChannelID with the recipient uid, ChannelType=WireChannelDM, and Payload
-// with the IM-recognized shape {type:1, content:text, space_id?}. The
-// InternalNotifyDeliverer forwards Payload verbatim into the NotifyReq.
+// fills ChannelID with the recipient uid, ChannelType=WireChannelDM, and Card
+// with the octo-server notification card fields.
 type SendMessageRequest struct {
 	ChannelID   string         `json:"channel_id"`
 	ChannelType int            `json:"channel_type"`
-	Payload     map[string]any `json:"payload"`
+	Payload     map[string]any `json:"payload,omitempty"`
+	Card        *notifyCard    `json:"card,omitempty"`
+}
+
+type notifyCard struct {
+	TaskNo      string `json:"task_no"`
+	Kind        string `json:"kind"`
+	Title       string `json:"title"`
+	TimeRange   string `json:"time_range,omitempty"`
+	Members     int    `json:"members,omitempty"`
+	MsgCount    int    `json:"msg_count,omitempty"`
+	GeneratedAt string `json:"generated_at,omitempty"`
+	Reason      string `json:"reason,omitempty"`
 }
 
 // oboReservedKeys are OBO markers that must never appear in a payload.
@@ -73,7 +84,8 @@ type notifyReq struct {
 	Event    string         `json:"event"`
 	Targets  []string       `json:"targets"`
 	ActorUID string         `json:"actor_uid"`
-	Payload  map[string]any `json:"payload"`
+	Payload  map[string]any `json:"payload,omitempty"`
+	Card     *notifyCard    `json:"card,omitempty"`
 }
 
 const (
@@ -91,7 +103,7 @@ const (
 type InternalNotifyDeliverer struct {
 	baseURL string
 	token   string
-	// webBaseURL builds the result link folded into payload.result_url. Empty => omitted.
+	// Kept for constructor compatibility; card deep links are built server-side.
 	webBaseURL string
 	client     *http.Client
 }
@@ -120,21 +132,17 @@ type notifyResp struct {
 	Filtered  map[string]string `json:"filtered"`
 }
 
-// SendMessage posts one recipient's message to /v1/internal/notify. Payload is
-// forwarded verbatim: octo-server passes it straight to the IM message builder,
-// which requires the {type:1, content} shape — reshaping it here would render an
-// empty message under a 2xx and silently lose the notification.
+// SendMessage posts one recipient's message to /v1/internal/notify.
 func (d *InternalNotifyDeliverer) SendMessage(ctx context.Context, spaceID string, msg SendMessageRequest) error {
 	if payloadHasOBOReserved(msg.Payload) {
 		return fmt.Errorf("notify payload contains forbidden OBO reserved field")
 	}
-	// Copy so appending result_url never mutates the caller's payload.
-	payload := make(map[string]any, len(msg.Payload)+1)
-	for k, v := range msg.Payload {
-		payload[k] = v
-	}
-	if d.webBaseURL != "" {
-		payload["result_url"] = d.webBaseURL
+	var payload map[string]any
+	if msg.Card == nil && len(msg.Payload) > 0 {
+		payload = make(map[string]any, len(msg.Payload))
+		for k, v := range msg.Payload {
+			payload[k] = v
+		}
 	}
 
 	// NotifyReq.space_id is required (binding); fall back to payload.space_id.
@@ -152,6 +160,7 @@ func (d *InternalNotifyDeliverer) SendMessage(ctx context.Context, spaceID strin
 		Targets:  []string{msg.ChannelID}, // single recipient — keep per-uid granularity
 		ActorUID: "",
 		Payload:  payload,
+		Card:     msg.Card,
 	}
 	return d.post(ctx, notifyEndpoint, req, msg.ChannelID)
 }
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -126,8 +126,7 @@ type Config struct {
 	// summary_notification.last_error. Prod + NotifyEnabled but empty => startup
 	// fatal (fail-fast on misconfig); dev only warns.
 	NotifyInternalToken string
-	// SummaryWebBaseURL is the base for a result link in the notification text.
-	// Empty => no link appended.
+	// SummaryWebBaseURL is retained for compatibility; the server card path builds deep links.
 	SummaryWebBaseURL string
 	// AppEnv is the deployment environment ("prod" | "dev"). Drives the
 	// prod-fatal / dev-warn handling of a missing NotifyInternalToken.
--- a/internal/notify/notify.go
+++ b/internal/notify/notify.go
@@ -173,8 +173,8 @@ func (n *Notifier) deliverToRecipient(task model.SummaryTask, kind, errMsg strin
 	}
 
 	// 2) Build + deliver.
-	text := n.buildText(task, kind, errMsg)
-	deliverErr := n.deliver(task.ID, target, text)
+	card := n.buildCard(task, kind, errMsg)
+	deliverErr := n.deliver(task.ID, target, card)
 
 	// 3) Persist outcome.
 	if deliverErr == nil {
@@ -247,7 +247,7 @@ func (n *Notifier) claimRetry(taskID int64, kind, uid string) (bool, error) {
 	return res.RowsAffected == 1, nil
 }
 
-func (n *Notifier) deliver(taskID int64, target deliveryTarget, text string) error {
+func (n *Notifier) deliver(taskID int64, target deliveryTarget, card notifyCard) error {
 	ctx, cancel := context.WithTimeout(context.Background(), 15*time.Second)
 	defer cancel()
 
@@ -270,24 +270,10 @@ func (n *Notifier) deliver(taskID int64, target deliveryTarget, text string) err
 			return fmt.Errorf("ensureFriend: %w", err)
 		}
 	}
-	// octo-server identifies a plain-text bot message by its ContentType enum
-	// (type=1=Text) carrying the body in "content"; a bare {"text":...} payload is
-	// not recognized and renders as an empty/unopenable message. Send the
-	// server-recognized shape instead.
-	//
-	// summary is registered as a system bot, so its messages must carry space_id:
-	// octo-server's filterPersonMessagesBySpace drops a system-bot message whose
-	// payload has an empty space_id (it would otherwise be unopenable in the UI).
-	// Include space_id only when known, to hit the "exact match keep" branch and
-	// avoid regressing the empty-SpaceID case.
-	payload := map[string]any{"type": 1, "content": text}
-	if target.SpaceID != "" {
-		payload["space_id"] = target.SpaceID
-	}
 	msg := SendMessageRequest{
 		ChannelID:   target.ChannelID,
 		ChannelType: target.ChannelType,
-		Payload:     payload,
+		Card:        &card,
 	}
 	if err := n.deliverer.SendMessage(ctx, target.SpaceID, msg); err != nil {
 		return fmt.Errorf("sendMessage: %w", err)
@@ -298,14 +284,10 @@ func (n *Notifier) deliver(taskID int64, target deliveryTarget, text string) err
 // recipientIsActiveMember reports whether uid is an active member (status=1) of
 // an active Space (status=1), mirroring octo-server's space.CheckMembership.
 //
-// Why this matters: for a system-bot DM, octo-server only adopts the X-Space-ID
-// header (and thus injects the authoritative payload.space_id) when the
-// RECIPIENT is an active member of that Space. If they are not, the server
-// STRIPS space_id yet still returns 200 — the message is then dropped by
-// filterPersonMessagesBySpace and the user silently never sees it, while the
-// worker would otherwise record it as "sent". Pre-checking here turns that
-// silent loss into an explicit delivery failure (retried/visible), instead of a
-// false success.
+// Why this matters: octo-server only binds a system-bot DM to a Space when the
+// recipient is an active member of that Space. If they are not, the server may
+// still return 200 while the user never sees the message. Pre-checking here
+// turns that silent loss into an explicit delivery failure.
 //
 // Conservative failure mode: on a nil imDB or a query error we return true
 // (do NOT block delivery) — the server-side CheckMembership still guards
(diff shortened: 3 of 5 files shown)


Jerry-Xin (approved):
Clean, well-tested worker-side migration from plain-text notify payloads to structured octo/v1 `card` fields; delivery semantics, error sanitization, and degraded-path behavior are all preserved.

## 🔴 Blocking
None.

## 💬 Non-blocking
- `internal/notify/notify.go:406` — `buildText` now has **no production callers** (delivery + redeliver both call `buildCard`; only `notify_test.go` still exercises it). It's effectively legacy/dead code kept alive by tests. Consider dropping it (and its helpers if they become unused) in a follow-up to avoid the two render paths drifting, or keep it deliberately as the fallback and add a one-line comment saying so.
- `internal/worker/processor.go:130` — comment still says sanitization happens "inside notify.buildText"; the single render point for the card path is now `sanitizedFailureReason`. Stale comment worth updating so future readers trace the right function.
- `internal/notify/client.go` `SendMessage` — the `spaceID == "" → payload["space_id"]` fallback is now inert for card sends (payload is nil). That's fine because `deliver()` passes `target.SpaceID` directly, but the fallback branch is dead for the card path; a short note or removal would reduce confusion.
- `internal/config/config.go` — `SummaryWebBaseURL` is now "retained for compatibility" but still wired through the constructor in `cmd/summary-worker/main.go`. Fine for this PR; flag for eventual cleanup once the server card path is the only consumer.

## ✅ Highlights
- Failure-reason sanitization is correctly preserved: `buildCard` routes through `sanitizedFailureReason` (same `errorSanitizer` + hardcoded safe fallback), and the regression tests assert no raw markers (`dial tcp`, DSN creds, IPs) leak into `card.Reason` on both the synchronous and sweep/redeliver paths.
- Wire-shape discipline: `Payload` and `Card` are mutually exclusive with `omitempty`; tests assert card sends omit `payload` entirely and completed cards omit `reason`/non-positive `members`/`msg_count` on the wire.
- `foldedTitle` has full nil/empty coverage (space+title / space-only / title-only / default "总结"), with degraded-path tests for nil imDB and missing space row that confirm delivery is never blocked.
- Card fields are a typed struct (no arbitrary map), so JSON escaping is native and layout/deep-link ownership moves cleanly server-side — no worker-side injection surface.
- Test suite thoroughly rewritten to match the card shape rather than deleted.

internal/notify/client.go
```diff
@@ -120,21 +132,17 @@ type notifyResp struct {
...
 	}
-	// Copy so appending result_url never mutates the caller's payload.
-	payload := make(map[string]any, len(msg.Payload)+1)
-	for k, v := range msg.Payload {
-		payload[k] = v
-	}
-	if d.webBaseURL != "" {
-		payload["result_url"] = d.webBaseURL
+	var payload map[string]any
+	if msg.Card == nil && len(msg.Payload) > 0 {
```
OctoBoooot, line 141: **[praise]** Clean mutual exclusion: the payload map is only built when `Card == nil`, and both fields are `omitempty` on the wire, so a card send never emits a stale `payload` — and the worker/notify tests assert the wire body omits it. Combined with sending raw typed card fields (octo-server owns layout), this removes the client-side string-interpolation-into-markup risk the old text path carried.

OctoBoooot (approved):
## Review: feat(notify): deliver summary notifications as octo/v1 cards (#153)

**Verdict:** Approve

Clean, well-tested refactor from a plain-text DM to a structured `card` field, with genuine information parity and no injection surface. Functional CI is fully green (Build/Vet/Lint/Test/Test-race-cgo/osv/secret-scan). The only red check is `check-sprint` (governance — maintainer clears out-of-band); `mergeStateStatus` stays BLOCKED until it passes. Approving on code.

Risk tier: normal.

## Parity checked
The PR claims "info parity" and "behavior preserved" — verified base→head:
- **Fields**: the plain text carried title(+space), 时间范围, 参与成员, 消息数量, 生成时间, and a detail link. The card carries `Title` (folded space via the same `resolveSpaceName`), `TimeRange`, `Members`, `MsgCount`, `GeneratedAt` (same `resultMeta`/`participantCount` sources), plus `TaskNo` for the server-built deep link. Failed: text had title + sanitized 失败原因; card has `Title` + `Reason` via the **same** sanitizer (refactored into the shared `sanitizedFailureReason`, used by both `buildCard` and the legacy `buildText`). No field lost.
- **The one removed capability** — the worker-built `/summary/detail?taskId=` link — was the *unreachable/broken* one this change intentionally replaces with octo-server's `/s/{task_no}` button. Documented, and an improvement, not a silent regression.
- **space_id**: previously folded into `payload.space_id` for octo-server's system-bot filter; now passed as the top-level `NotifyReq.space_id` (required binding) via `SendMessage(ctx, target.SpaceID, …)`. Same authoritative value, cleaner path. The `recipientIsActiveMember` silent-drop pre-check is unchanged.
- **State machine**: claim/markSent/markFailed/Sweep and the `redeliver` path are untouched except for building a card instead of text — the worker test confirms sweep/redeliver also sends a card with the sanitized reason (no raw-error leak).

### Praise
- `payload`/`card` are cleanly mutually exclusive — both `omitempty` on the wire structs and enforced in `SendMessage` (payload map only built when `Card == nil`), so a card send never emits a stale `payload`, and the test asserts the wire body omits it. Sending raw typed fields and letting octo-server own layout removes the client-side string-interpolation-into-markup risk entirely (JSON marshaling handles escaping). Folding the failure sanitizer into one shared `sanitizedFailureReason` used by both the card and legacy-text paths means the scrub can't drift between them. Tests assert exact card shape, title-folding across all four branches (space+title / space / title / neither), omitempty for non-positive counts + failed-only reason, and payload absence — reverting the feature would fail them.

### Not a merge signal
`mergeStateStatus` BLOCKED on `check-sprint` (governance). Approved on code; the sprint gate must still clear.

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #153 (octo-smart-summary)

## Summary

This PR migrates summary notification delivery from plain-text messages (`{type:1, content:text}` payload) to structured octo/v1 notification cards (`notifyCard`). The `buildCard` function replaces `buildText` as the production render path, carrying structured fields (task number, title, time range, member count, message count, generated-at timestamp, failure reason) that the server renders as a card UI with deep links. The legacy `buildText` is retained but no longer called from production code. The failure-reason sanitization is extracted into `sanitizedFailureReason`, shared between the new `buildCard` and the legacy `buildText`. Tests are updated to assert against the card struct rather than raw payload text. The sweep/redeliver path is correctly migrated to `buildCard`.

## Verification

Static analysis only at head `a90329b2298f2861dedb24f914eaa9a558a344fc`; build and tests not executed in this environment.

- ✅ **Data flow tracing** — `buildCard` consumes `task.TaskNo`, `kind` (from `kindForStatus`), `foldedTitle` (from `resolveSpaceName` + `task.Title`), `formatTimeRange` (from `task.TimeRange{Start,End}`), `resultMeta` (from `summary_result` table), `participantCount` (from `summary_participant`), and `sanitizedFailureReason` (from `errMsg` via `errorSanitizer`). All sources traced to DB queries or function parameters; all degrade gracefully on nil DB / query error / zero values.
- ✅ **Deliver path** — `deliver()` builds `SendMessageRequest{Card: &card}` with nil `Payload`. `InternalNotifyDeliverer.SendMessage` correctly passes `Card` to `notifyReq.Card` and omits `Payload` (omitempty). `space_id` is carried in `notifyReq.SpaceID` (from the function arg `target.SpaceID`), not in `Payload` — correct for the card contract.
- ✅ **Sweep/redeliver path** — `redeliver()` at `internal/notify/notify.go:692` uses `buildCard` and follows the same `buildCard → deliver → mark` flow. The R3 regression test (`TestSweep_RedeliverSanitizesRawError`) asserts `card.Reason` on both synchronous and sweep paths.
- ✅ **AUTH/VISIBILITY unchanged** — No changes to `X-Internal-Token` auth, `EnsureFriend` precondition, `recipientIsActiveMember` pre-check, or space membership filtering. All guard logic is unchanged; only the message payload format changed.
- ✅ **Dedup/retry unchanged** — `claim`, `claimRetry`, `claimStalePending`, `markSent`, `markFailed`, and the sweep logic are untouched. The card migration does not affect the state machine.
- ✅ **OBO guard safe on nil payload** — `payloadHasOBOReserved(nil)` iterates zero times and returns false. Safe.
- ✅ **`omitempty` on card zero-values** — `Members int`, `MsgCount int`, `Reason string` all have `omitempty`; zero/empty values are correctly omitted from JSON wire output. Tests assert this explicitly.

## Findings

### P2 — Stale `buildText` references in comments (`internal/notify/notify.go:459`)

The `WithErrorSanitizer` doc comment at `:70` and `:77` still says "buildText runs every failure reason" and "buildText falls back to a hardcoded safe string". The production path now uses `buildCard` → `sanitizedFailureReason`. The same stale references appear in `internal/worker/processor.go:130,137` ("Sanitization happens at a SINGLE render point inside notify.buildText") and `internal/worker/personal_processor.go:988` ("sanitizer in Notifier.buildText"). Update all four locations to reference `sanitizedFailureReason` (the actual shared render point).

### P2 — `buildText` is dead code with no documented retention reason (`internal/notify/notify.go:406`)

`buildText` is retained as "legacy" but has no production caller — both `deliverToRecipient` (`:176`) and `redeliver` (`:692`) call `buildCard`. If kept for rollback safety or a phased migration, a comment explaining the retention reason would prevent future maintainers from assuming it's still live. Otherwise, removing it eliminates a parallel render path that could drift from `buildCard`.

## Human-verify

1. **Server-side card contract** — This PR is the "worker side" of a paired change (title references "#150 worker side"). The octo-server `/v1/internal/notify` endpoint must accept `notifyReq.Card` and render it as a card UI. If the server still only processes `payload.{type,content}` and ignores the `card` field, notifications will arrive empty or be dropped. Confirm the server-side PR is deployed first or simultaneously. This is **not** a merge blocker for this worker PR — flagging for deployment-ordering awareness.

- `sanitizedFailureReason` correctly preserves the single-render-point sanitizer contract (PR#113 R3): nil sanitizer → hardcoded safe string; empty errMsg → empty string (omitted from card).
- `foldedTitle` handles all four combinations of space/title presence, matching the prior `buildText` logic.
- `notifyCard` struct fields with `omitempty` correctly suppress zero values on the wire — tested in `TestBuildCard_CompletedOmitsNonPositiveCountsOnWire`.
- `SendMessage` correctly handles `msg.Card != nil && msg.Payload == nil`: payload copy is skipped, `payloadHasOBOReserved` is safe on nil, `space_id` fallback from payload is harmless (nil map access returns zero value).
- All tests updated from `Payload["content"]` assertions to `sentCard(t, msg)` assertions, including the sweep sanitization regression test.
- `webBaseURL` is retained in `InternalNotifyDeliverer` for constructor compatibility but no longer used — clean deprecation.

## Verdict: APPROVED

Clean migration from plain-text to structured card notifications. No correctness, security, or functional defects. The data flow is sound, sanitization is preserved, tests cover the new contract, and the sweep/redeliver path is correctly migrated. Two P2 documentation/dead-code items are non-blocking. The server-side card contract compatibility is a deployment-ordering concern, not a code defect.

yujiawei (approved):
# Code Review — PR #153 (octo-smart-summary)

Switches terminal-state summary notifications (completed/failed) from a plain-text DM payload to the structured `card` field on `POST /v1/internal/notify`. Reviewed at head SHA `a90329b` against merge-base `87395d9`. Verified against the octo-server counterpart (`modules/notify` card ingress) since this is a cross-service contract change.

**Method**: build + full test run, `go vet`, field-by-field contract diff against octo-server, and adversarial re-verification across security / correctness / contract dimensions.

## 1. Spec compliance

**Spec: ✅**

- **Missing**: none. Completed cards carry title(+space), time range, members, msg count, generated-at + the server-owned detail button; failed cards carry title(+space) + sanitized reason. `payload` and `card` are mutually exclusive on the wire (`client.go:141`). The claim/markSent/markFailed/Sweep/redeliver dedup state machine is untouched (pure `text → card` signature swap at `notify.go:176`, `notify.go:692`).
- **Extra scope**: none. No new flags, endpoints, or fields beyond the card contract.
- **Deviation**: none. Field names and semantics match the octo-server contract exactly (see §2).

## 2. Cross-service contract (load-bearing — verified against octo-server)

The 8 JSON tags on the worker's `notifyCard` (`internal/notify/client.go:49-58`) match octo-server's `SummaryCardFields` (`modules/notify/model.go:24-33`) **1:1**: `task_no / kind / title / time_range / members / msg_count / generated_at / reason`. A tag mismatch here would silently drop fields, so this is the highest-risk surface — it is clean.

Server validation (`modules/notify/card.go:33-41`) requires non-empty `task_no` + `title`, `kind ∈ {completed, failed}`, and `1 ≤ targets ≤ 200`. The worker cannot violate any of these:
- `task_no` — sourced from `summary_task.task_no`, `not null` unique column (`model.go:116`).
- `title` — `foldedTitle` (`notify.go:379-392`) always returns a non-empty string (falls back to `"总结"`).
- `kind` — only ever `completed`/`failed` via `kindForStatus` (`notify.go:189-198`).
- `targets` — always exactly one uid (`client.go:160`), preserving per-recipient dedup/retry granularity.

The empty-delivery risk (server accepts card but delivers nothing) does **not** apply: octo-server's card path is fully wired (`api.go:201` routes `Card != nil` → `deliverCardNotification`) and covered by tests. On the card path the server still returns `delivered/filtered`, and the worker's `post()` (`client.go:204-212`) correctly treats "recipient not in `delivered`" as a retryable failure rather than a false success.

## 3. Code quality

**Quality: Approved**

Security (this PR is `security_sensitive` — verified with extra care):
- **Failure-reason sanitization is preserved on both paths.** `buildCard` routes `reason` through `sanitizedFailureReason` (`notify.go:374`), which uses the injected `errorSanitizer` and falls back to a hardcoded safe string when nil (`notify.go:394-403`). The sweep/redeliver path reloads the raw `ErrorMessage` and rebuilds via `buildCard` (`notify.go:689-692`), so the PR#113-R3 regression (raw DSN/IP/stack via the redeliver path) does not reappear. Raw `errMsg` never reaches `card.Reason`.
- **Token handling unchanged & safe.** `X-Internal-Token` is header-only, never echoed, and never folded into the card, payload, or `last_error` (`markFailed` runs `sanitize()` + `truncateForLastError`). The `notifyCard` struct has no token field.
- **No info-leak to non-members.** Server verifies membership before transport and reports non-members in `filtered`; `failed` notifications go only to the creator (`notify.go:113-116`). Card title/excerpt/facts are markdown-escaped and the deep-link is URL-escaped + https-enforced server-side.

Correctness: the state machine, space-binding, and `omitempty` behavior were re-verified. Space binding now flows top-level (`target.SpaceID → SendMessage(spaceID) → notifyReq.SpaceID`, which is `binding:"required"` server-side) instead of via `payload.space_id`; the old payload fallback is superseded, not regressed. `members`/`msg_count` are gated on `> 0` and `reason` is failed-only, matching the server's documented omit semantics; tests assert the wire omission explicitly.

Tests, `go build ./...`, and `go vet ./...` all pass clean at head SHA.

### P2 — non-blocking

- **Dead code from the removed `result_url` feature.** `Notifier.buildText` (`notify.go:405-466`) now has zero non-test call sites; `InternalNotifyDeliverer.webBaseURL` (`client.go:107,117`) is written but never read; `Config.SummaryWebBaseURL` (`config.go:130,210` → `main.go:110`) feeds only that dead field. These are one root cause (incomplete feature removal) and one cleanup PR clears all three. CI is unaffected (Go does not flag unused methods/fields, and the repo's golangci config runs only `govet`), so this is hygiene, not a build break. Additionally, the `WithErrorSanitizer` doc comment (`notify.go:70`) still says "buildText runs every failure reason through this sanitizer" — the live render point is now `buildCard`; update the comment when removing `buildText`.
- **Deploy-ordering dependency.** This worker sends card-only requests (no `payload`). An octo-server that predates the card ingress rejects those with `400 payload不能为空` (`api.go:176-178`), which the worker treats as a delivery failure → retried/swept → attempts exhausted. Failure mode is graceful (no crash, no silent loss), but octo-server #579/#580 must be deployed **before or with** this worker or terminal notifications will be dropped until it is. Worth calling out in the rollout runbook.
- **Test gap for the empty-title invariant.** The `foldedTitle` `default: "总结"` branch (`notify.go:390`) is what guarantees the worker never emits a title the server would reject. The server-side rejection is tested, but no worker-side test pins "worker always emits a non-empty title." Low risk given the fallback, but a one-line assertion would lock the cross-service invariant.

## 4. Verdict

**APPROVE.** No P0/P1 defects. Spec ✅ and Quality Approved. The P2 items (dead code, deploy ordering, one test gap) are cleanup/operational and should not block merge — recommend a follow-up hygiene PR to delete the dead `result_url`/`webBaseURL` chain and refresh the `WithErrorSanitizer` doc, and a note in the rollout runbook on the octo-server ordering dependency.

## 5. For a human to double-check (security-sensitive)

- Confirm octo-server with the card ingress is deployed **before/with** this worker (rollout ordering).
- Confirm `external.webLoginURL` on octo-server is https in the target environment; a non-https value silently degrades every card to plain text (server-side fallback — no data loss, but no button).

PR #154, opened 2026-07-14
feat(api): resolve GET /summaries/:id by task_no as well as numeric id

Ranwanglc:
## Summary
`GET /summaries/:id` now resolves EITHER the numeric id (unchanged) OR the varchar `task_no`, so the notification card deep link `/s/{task_no}?sp={space_id}` can fetch the detail. Numeric ids behave exactly as before.

Supports #150. Pairs with the notify-card PR (worker) and the octo-web `/s/:taskNo` route PR.

## Change
- `GetSummary`: when the `:id` param is non-numeric, resolve by `task_no` (`WHERE task_no = ? AND deleted_at IS NULL`), then run the **same** `authorizeTaskAccess` on the resolved numeric id.
- Cross-space / unknown `task_no` returns the **identical** `404` / biz code `40008` as the numeric path — the by-task_no path does not bypass authorization or leak task existence.
- Only `GetSummary` (detail read) changed; other handlers untouched.

## No schema change
`task_no` already exists as `varchar(32)` with `uniqueIndex uk_task_no`. **No migration, no schema edit.** Generated `task_no` values are prefixed (non-numeric), so a numeric param is unambiguously an id.

## Deployment prerequisites
None. No new env vars, no migration.

## Testing
`go build ./...` clean; `go vet` clean; `go test ./internal/api/handler/... -count=1` green. Tests: numeric id regression, valid task_no success, unknown task_no 404/40008, wrong-space task_no denial (compared to numeric wrong-space denial). Independently re-verified + adversarially reviewed (APPROVE).

### DIFF
--- a/internal/api/handler/task.go
+++ b/internal/api/handler/task.go
@@ -122,6 +122,21 @@ func (h *TaskHandler) pickDisplayResult(taskID int64) (model.SummaryResult, bool
 	return result, true
 }
 
+func (h *TaskHandler) resolveSummaryTaskParam(c *gin.Context) (int64, bool) {
+	param := c.Param("id")
+	taskID, err := strconv.ParseInt(param, 10, 64)
+	if err == nil {
+		return taskID, true
+	}
+
+	var task model.SummaryTask
+	if err := h.db.Where("task_no = ? AND deleted_at IS NULL", param).First(&task).Error; err != nil {
+		c.JSON(http.StatusNotFound, apiResponse{Code: 40008, Message: "任务不存在"})
+		return 0, false
+	}
+	return task.ID, true
+}
+
 // callerPlainCitationsVisible decides whether the caller may see a display
 // SummaryResult's plain (non-team) citations.
 //
@@ -626,9 +641,8 @@ func (h *TaskHandler) ListSummaries(c *gin.Context) {
 
 // GetSummary handles GET /api/v1/summaries/:id
 func (h *TaskHandler) GetSummary(c *gin.Context) {
-	taskID, err := strconv.ParseInt(c.Param("id"), 10, 64)
-	if err != nil {
-		c.JSON(http.StatusBadRequest, apiResponse{Code: 40000, Message: "invalid task id"})
+	taskID, resolved := h.resolveSummaryTaskParam(c)
+	if !resolved {
 		return
 	}
 
--- a/internal/api/handler/auth_test.go
+++ b/internal/api/handler/auth_test.go
@@ -74,14 +74,34 @@ func setupRouter(h *TaskHandler) *gin.Engine {
 }
 
 func doRequest(r *gin.Engine, method, path, userID string) *httptest.ResponseRecorder {
+	return doRequestWithSpace(r, method, path, userID, "space1")
+}
+
+func doRequestWithSpace(r *gin.Engine, method, path, userID, spaceID string) *httptest.ResponseRecorder {
 	w := httptest.NewRecorder()
 	req := httptest.NewRequest(method, path, nil)
 	if userID != "" {
 		req.Header.Set("Token", userID)
 	}
-	req.Header.Set("X-Space-Id", "space1")
+	if spaceID != "" {
+		req.Header.Set("X-Space-Id", spaceID)
+	}
 	r.ServeHTTP(w, req)
 	return w
+}
+
+func summaryTaskIdentity(t *testing.T, w *httptest.ResponseRecorder) (int64, string) {
+	t.Helper()
+	var resp struct {
+		Data struct {
+			TaskID int64  `json:"task_id"`
+			TaskNo string `json:"task_no"`
+		} `json:"data"`
+	}
+	if err := json.Unmarshal(w.Body.Bytes(), &resp); err != nil {
+		t.Fatalf("unmarshal GetSummary body: %v; body=%s", err, w.Body.String())
+	}
+	return resp.Data.TaskID, resp.Data.TaskNo
 }
 
 func TestAuthorizeTaskAccess_NonMember(t *testing.T) {
@@ -118,6 +138,70 @@ func TestAuthorizeTaskAccess_Creator(t *testing.T) {
 	if w.Code != http.StatusOK {
 		t.Errorf("expected 200 for creator, got %d: %s", w.Code, w.Body.String())
 	}
+}
+
+func TestGetSummary_NumericIDStillResolves(t *testing.T) {
+	db, imDB := setupTestDBs(t)
+	taskID := seedTask(t, db, imDB)
+	h := NewTaskHandler(db, imDB, "")
+	r := setupRouter(h)
+
+	w := doRequest(r, "GET", fmt.Sprintf("/api/v1/summaries/%d", taskID), "creator1")
+	if w.Code != http.StatusOK {
+		t.Fatalf("expected 200 for numeric id, got %d: %s", w.Code, w.Body.String())
+	}
+	gotID, gotNo := summaryTaskIdentity(t, w)
+	if gotID != taskID || gotNo != "TST-001" {
+		t.Fatalf("expected task_id=%d task_no=TST-001, got task_id=%d task_no=%s", taskID, gotID, gotNo)
+	}
+}
+
+func TestGetSummary_TaskNoResolvesSameTask(t *testing.T) {
+	db, imDB := setupTestDBs(t)
+	taskID := seedTask(t, db, imDB)
+	h := NewTaskHandler(db, imDB, "")
+	r := setupRouter(h)
+
+	w := doRequest(r, "GET", "/api/v1/summaries/TST-001", "creator1")
+	if w.Code != http.StatusOK {
+		t.Fatalf("expected 200 for task_no, got %d: %s", w.Code, w.Body.String())
+	}
+	gotID, gotNo := summaryTaskIdentity(t, w)
+	if gotID != taskID || gotNo != "TST-001" {
+		t.Fatalf("expected task_id=%d task_no=TST-001, got task_id=%d task_no=%s", taskID, gotID, gotNo)
+	}
+}
+
+func TestGetSummary_UnknownTaskNoReturnsNotFound(t *testing.T) {
+	db, imDB := setupTestDBs(t)
+	seedTask(t, db, imDB)
+	h := NewTaskHandler(db, imDB, "")
+	r := setupRouter(h)
+
+	w := doRequest(r, "GET", "/api/v1/summaries/NO-SUCH-TASK", "creator1")
+	if w.Code != http.StatusNotFound {
+		t.Fatalf("expected 404 for unknown task_no, got %d: %s", w.Code, w.Body.String())
+	}
+	assertBizCode(t, w, 40008)
+}
+
+func TestGetSummary_TaskNoWrongSpaceMatchesNumericDenial(t *testing.T) {
+	db, imDB := setupTestDBs(t)
+	taskID := seedTask(t, db, imDB)
+	h := NewTaskHandler(db, imDB, "")
+	r := setupRouter(h)
+
+	wNumeric := doRequestWithSpace(r, "GET", fmt.Sprintf("/api/v1/summaries/%d", taskID), "participant1", "other_space")
+	if wNumeric.Code != http.StatusNotFound {
+		t.Fatalf("expected numeric wrong-space 404, got %d: %s", wNumeric.Code, wNumeric.Body.String())
+	}
+	assertBizCode(t, wNumeric, 40008)
+
+	wTaskNo := doRequestWithSpace(r, "GET", "/api/v1/summaries/TST-001", "participant1", "other_space")
+	if wTaskNo.Code != wNumeric.Code {
+		t.Fatalf("expected task_no wrong-space status %d, got %d: %s", wNumeric.Code, wTaskNo.Code, wTaskNo.Body.String())
+	}
+	assertBizCode(t, wTaskNo, 40008)
 }
 
 func TestAuthorizeTaskAccess_Participant(t *testing.T) {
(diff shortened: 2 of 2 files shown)


Jerry-Xin (approved):
GET /summaries/:id now resolves by task_no as well as numeric id, and the change correctly preserves the existing access-control gate — approving.

## ✅ Highlights
- **Authz preserved on both paths.** `resolveSummaryTaskParam` (`internal/api/handler/task.go:125`) is resolution-only: numeric param → parsed id; non-numeric → task_no lookup returning `task.ID`. The resolved id then flows through the unchanged `authorizeTaskAccess` gate in `GetSummary` (`internal/api/handler/task.go:649`), which enforces `id = ? AND space_id = ? AND deleted_at IS NULL` + `canAccessTask`. No new resolution path bypasses the access check — a classic IDOR spot handled correctly.
- **No collision/ambiguity risk.** `SummaryTask.TaskNo` carries a global `uniqueIndex:uk_task_no` (`internal/model/model.go:116`), so a task_no maps to exactly one row. There is no way to fetch another record by supplying a task_no that numerically matches a different id — the numeric branch only triggers when the param parses as int64, and task_no lookup is a separate branch.
- **Cross-space sealed.** The task_no lookup itself has no space filter, but the resolved id is re-loaded under `space_id = ?` in `authorizeTaskAccess`, so a task_no belonging to another space resolves then returns 40008/404 — identical to numeric wrong-space denial. Verified by `TestGetSummary_TaskNoWrongSpaceMatchesNumericDenial`.
- **SQL safe.** `Where("task_no = ? AND deleted_at IS NULL", param)` is parameterized.
- **404 consistency / no info leak.** Unknown task_no and wrong-space both return the same `{Code:40008, "任务不存在"}`; no distinction between "exists but not yours" and "not found".
- Solid test coverage: numeric-still-resolves, task_no-resolves-same-task, unknown-task_no-404, and the wrong-space parity test.

## 💬 Non-blocking notes
- **Behavior change:** a non-numeric `:id` that used to return `400 "invalid task id"` (Code 40000) now returns `404` (Code 40008) when it doesn't match a task_no. This is the intended semantics of the feature and is fine; just flagging for any client that relied on the old 400.
- This resolver only wires GetSummary. If other `:id` endpoints (GetResult/Regenerate/DeleteSummary/CancelSummary) should also accept task_no, consider reusing `resolveSummaryTaskParam` there for consistency — not required for this PR.

CI: all required checks green (Vet/Test/Lint/Build/Test race-cgo). Only `check-sprint` (governance) is failing, which is not a code blocker.

internal/api/handler/task.go
```diff
@@ -122,6 +122,21 @@ func (h *TaskHandler) pickDisplayResult(taskID int64) (model.SummaryResult, bool
...
 
+func (h *TaskHandler) resolveSummaryTaskParam(c *gin.Context) (int64, bool) {
+	param := c.Param("id")
+	taskID, err := strconv.ParseInt(param, 10, 64)
+	if err == nil {
+		return taskID, true
+	}
+
+	var task model.SummaryTask
+	if err := h.db.Where("task_no = ? AND deleted_at IS NULL", param).First(&task).Error; err != nil {
```
OctoBoooot, line 133: **[praise]** Right shape for a dual-key lookup: this resolver only maps the param to a numeric taskID (parameterized `task_no = ?`, no injection) and does no authorization itself, so `GetSummary` runs the identical `authorizeTaskAccess` space+membership gate regardless of which branch resolved it. A task_no from another space resolves to its id and is then denied by the caller's-space re-scope — and unknown-task_no vs cross-space both return the same 40008, so there's no existence oracle. `TestGetSummary_TaskNoWrongSpaceMatchesNumericDenial` pins exactly that (task_no denial == numeric denial).

OctoBoooot (approved):
## Review: feat(api): resolve GET /summaries/:id by task_no as well as numeric id (#154)

**Verdict:** Approve

The dual-key resolution is done safely: it only *resolves* the `:id` param to a numeric taskID, then the identical `authorizeTaskAccess` space+membership gate runs regardless of branch — so the new task_no path cannot bypass access control. Functional CI is green (Build/Vet/Lint/Test/Test-race-cgo/osv/secret-scan); only `check-sprint` (governance) is red, cleared out-of-band. `mergeStateStatus` stays BLOCKED until it passes.

Risk tier: high (authz path). `needs-human-review` applies.

### Security — the IDOR surface, verified clean
This is the exact risk of a dual-key lookup, and it's handled correctly:
- `resolveSummaryTaskParam` (task.go:125) parses the param as a numeric id, or on parse failure looks it up by `task_no = ? AND deleted_at IS NULL` (parameterized — no injection) and returns the row's numeric `ID`. It performs **no** authorization itself.
- `GetSummary` then calls the **same** `authorizeTaskAccess(c, taskID)` on the resolved id as before, which enforces: auth required → fail-closed empty-`X-Space-Id` gate → load scoped `WHERE id=? AND space_id=? AND deleted_at IS NULL` → `canAccessTask` (creator/participant). So a task_no belonging to another space resolves to its numeric id and is then denied by the caller's-space re-scope, identical to the numeric path.
- **No existence oracle**: an unknown task_no (resolver) and a cross-space/other task (authorize) both return the same `40008 "任务不存在"`, so the task_no path leaks no more than the numeric path ever did.

### Parity checked
Body claims "Numeric ids behave exactly as before":
- The numeric branch is `strconv.ParseInt(param, 10, 64)` → same taskID as the old `GetSummary` first line; on success it returns immediately without touching the DB, so numeric lookups are byte-for-byte the prior behavior. The old `40000 "invalid task id"` on a non-numeric param is replaced by the task_no lookup (intended new capability) with a `40008` fallthrough — the only behavior change, and it's the feature. Test `TestGetSummary_NumericIDStillResolves` confirms the numeric path.

### Praise
- The tests target precisely the dangerous case: `TestGetSummary_TaskNoWrongSpaceMatchesNumericDenial` asserts a task_no from another space returns the **same** status + `40008` as the numeric cross-space denial — proving the new resolution path inherits space isolation rather than sidestepping it — alongside numeric-resolves, task_no-resolves-same-task, and unknown-task_no→404. That's the right invariant to pin for a dual-key lookup, and pinning it as "task_no denial == numeric denial" is exactly how you keep a future refactor from opening an IDOR here.

### Not a merge signal
`mergeStateStatus` BLOCKED on `check-sprint` (governance) + `needs-human-review` (authz path). Approved on code; the sprint gate + a human still clear before merge.

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #154 (octo-smart-summary)

## Summary

This PR adds `task_no` resolution to `GET /api/v1/summaries/:id`. The `:id` path parameter now accepts either a numeric task ID (existing behavior) or a `task_no` string (e.g. `ST20260604abcd1234`). A new helper `resolveSummaryTaskParam` tries integer parse first, falling back to a DB lookup by `task_no`. The change is clean, well-tested, and preserves the existing cross-space authorization contract through `authorizeTaskAccess`.

## Verification

- ✅ **Cross-space authorization preserved** — `resolveSummaryTaskParam` returns a task ID that `authorizeTaskAccess` (`internal/api/handler/task.go:91`) then validates with `space_id` scoping. The wrong-space test (`TestGetSummary_TaskNoWrongSpaceMatchesNumericDenial`) confirms identical 404/40008 denial for both numeric and task_no paths.
- ✅ **No information leakage** — the task_no query at `internal/api/handler/task.go:133` is not space-scoped, but the downstream `authorizeTaskAccess` returns the same 404 error for "task not found" and "task in wrong space", preventing enumeration of cross-space task existence.
- ✅ **Soft-delete handled** — explicit `deleted_at IS NULL` at `:133` is correct and necessary because `SummaryTask.DeletedAt` is `*time.Time` (manual soft-delete, not GORM's auto `DeletedAt`).
- ✅ **No SQL injection** — parameterized GORM query at `:133`.
- ✅ **No numeric/task_no collision** — `GenerateTaskNo()` (`internal/service/summary.go:31`) always produces `ST<date><8 alphanumeric>` strings, which can never parse as pure integers. The two resolution paths are disjoint.
- ✅ **Test coverage** — 4 new tests: numeric happy path, task_no happy path, unknown task_no (404), and cross-space parity between numeric and task_no denial.
- ✅ **INPUT-SHAPE** — `strconv.ParseInt` with base 10, 64-bit handles overflow correctly; Gin URL-decodes the path param before the handler sees it; no injection vector through the fallback query.
- ✅ **PERSISTENCE/CACHE** — read-only endpoint; no state mutation or cache write is affected.

## Findings

No P0/P1 issues. One P2 and one nit below.

### P2 — `resolveSummaryTaskParam` not shared with sibling endpoints (`internal/api/handler/task.go:126`)

Four other handlers on the same `/summaries/:id/*` route family still parse `:id` with `strconv.ParseInt` only: `GetResult` (`:860`), `Regenerate` (`:921`), `DeleteSummary` (`:1495`), `CancelSummary` (`:1617`). Clients that adopt `task_no` for detail lookups will find it unsupported on result, regenerate, delete, and cancel routes — receiving `400/40000` ("invalid task id") instead of a resolved lookup. This is not a security issue (authorization is correctly enforced downstream in all paths), but is an API consistency gap.

If the scope is intentionally GET-only for now, consider a brief code comment on `resolveSummaryTaskParam` stating the planned rollout. Otherwise, the helper is already extracted and could be reused by each sibling handler with a one-line replacement.

### Nit — Error code change for non-numeric `:id` values (`internal/api/handler/task.go:644`)

The old code returned `400 Bad Request` with biz-code `40000` ("invalid task id") for any non-numeric `:id`. The new code falls through to the `task_no` lookup and returns `404 Not Found` with biz-code `40008` ("任务不存在") when no match exists. This is arguably more RESTful, but is a behavioral change in the error contract. Any client that distinguished `400/40000` from other errors will silently see different behavior for the same input.

## Verdict: APPROVED

Clean, well-scoped feature addition with solid test coverage. The cross-space authorization contract is preserved end-to-end. The P2 endpoint-parity gap and the error-code nit are non-blocking and can be addressed in a follow-up.

yujiawei (approved):
# Code Review — PR #154 (octo-smart-summary)

## Summary

This PR lets `GET /api/v1/summaries/:id` resolve the path param as **either** a numeric task id (unchanged) **or** a varchar `task_no`, so the notification deep link `/s/{task_no}?sp={space_id}` can fetch the detail. The change is small, well-scoped, and correct. I independently re-verified it against the head SHA with a security focus (cross-space isolation, enumeration/oracle risk, contract change) — no blocking issues.

**Verdict: APPROVED.** The findings below are all non-blocking (P2 / nit) — forward-looking consistency and test-coverage hygiene.

## What I verified

- **No authorization bypass.** `resolveSummaryTaskParam` (`internal/api/handler/task.go:125`) only maps `task_no → id`; the resolved id then flows through the unchanged `authorizeTaskAccess` (`task.go:80`), which re-queries `WHERE id = ? AND space_id = ? AND deleted_at IS NULL`, keeps the fail-closed empty-`X-Space-Id` gate, and runs `canAccessTask`. The by-`task_no` path adds no new trust.
- **Cross-space isolation preserved.** The `task_no` lookup itself is space-agnostic (`WHERE task_no = ? AND deleted_at IS NULL`), but it returns only the id — the space check is enforced downstream. A `task_no` from another space returns the identical `404 / 40008` as a numeric cross-space miss, so it is no existence oracle. `TestGetSummary_TaskNoWrongSpaceMatchesNumericDenial` (`auth_test.go:188`) asserts exactly this, using a real member (`participant1`) who *would* be authorized in the correct space — that isolates the space gate from the membership gate and makes the test non-tautological.
- **Numeric disambiguation is unambiguous.** `task_no` is server-generated only (`service.GenerateTaskNo`, `summary.go:31`), emitting `ST + YYYYMMDD + 8 rand` — always non-numeric and never user-set. So a numeric param is unambiguously an id; `strconv.ParseInt` is a safe discriminator. No SQL injection risk (gorm placeholder). No schema change (the `uk_task_no` unique index already exists).
- **Scope is correct.** Only `GetSummary` was rewired; all mutating/sibling handlers remain numeric-only, matching the deep-link's read-only need.
- **Build/tests green locally.** `go build ./...` clean, `go vet` clean, `go test ./internal/api/handler/...` passes including the four new tests.

## Findings (non-blocking)

### P2 — Only `GetSummary` is `task_no`-aware; siblings on `/summaries/:id` still 400 on a non-numeric param
`GetResult` (`task.go:860`), `Regenerate` (`task.go:921`), `DeleteSummary` (`task.go:1495`), `CancelSummary` (`task.go:1617`), `personal.go:128`, and `edit.go:35` all still do a bare `strconv.ParseInt` and return `400 / 40000` for a non-numeric param. This is **not a security bug and not a regression** — those paths never accepted `task_no`, a 400 is fail-safe, and the deep-link page can be served entirely from `GetSummary` (which already embeds sources, participants, and the latest result). Flagging only as a forward-looking footgun: if the deep link later calls any sibling by `task_no`, it will silently 400. Consider documenting that only `GetSummary` accepts `task_no`, or routing the other read paths through the same resolver.

### nit — Non-numeric `:id` error changed from `400 / 40000` to `404 / 40008` for `GetSummary`
Previously a non-numeric id returned `400 {code:40000 "invalid task id"}`; it now falls through to the `task_no` lookup and returns `404 {code:40008 "任务不存在"}` when unknown. Non-breaking (the only pre-PR non-numeric caller was malformed input; no test or caller asserts `40000` for this endpoint) and arguably more secure, since a bad `task_no` now looks identical to a missing task rather than acting as an oracle. Noted for completeness.

### nit — Test-fixture fidelity & coverage hygiene
- `seedTask` uses `TaskNo: "TST-001"` (`auth_test.go:49`), which does not match the production `ST + date + 8rand` format. It still exercises the intended branch (it's non-numeric, so `ParseInt` fails), so the tests are logically valid — but aligning the fixture to a realistic `ST...` literal would guard against any future code that wrongly special-cases the prefix/hyphen.
- No dedicated test seeds a **soft-deleted** task and asserts its `task_no` returns `404`. The outcome is already guaranteed by `authorizeTaskAccess`'s own `deleted_at IS NULL` filter (the `deleted_at` clause in the new branch is redundant defense-in-depth), so this is coverage hygiene, not a gap.
- A numeric-but-nonexistent id (`/summaries/999999`) takes a different branch (resolver returns early, `404` comes from `authorizeTaskAccess`). That branch is already exercised by the wrong-space numeric test, so an explicit assertion would only make the enumeration-parity contract more self-documenting.

## Security note (per security-sensitive classification)

The one item worth a human's explicit sign-off: the `task_no` lookup deliberately runs **without** a `space_id` filter and relies on the downstream `authorizeTaskAccess` re-query for isolation. This is correct as written and covered by tests, but it means the isolation guarantee is load-bearing on that second query — any future refactor that consumes `resolveSummaryTaskParam`'s output without going through `authorizeTaskAccess` would leak cross-space. Worth keeping in mind for future changes to this resolver.

PR #161, opened 2026-07-17
fix(agent): first-turn citation [0] regression from #158 (4-reviewer P1 + 3 P2)

wanyiwang06:
## Summary

Post-merge follow-up to #158. Four reviewers (Jerry-Xin / yujiawei / lml2468 / mochashanyao) independently byte-verified a P1 regression at the squash-merged head of #158 (`7e888b1`): first-turn agent summaries emit `[0]` citation markers because `getSessionMessagePool` reads from `agent_message` tool rows that haven't been persisted yet during the run. `worker.BuildCitations` (`idx >= 1 && idx <= maxIdx`) then discards every marker at save time, so the summary body references non-existent citation rows on the dominant single-turn path. This PR closes the P1 plus the 3 concurrent P2s Octo-Q raised in the same round.

## Linked Spec

Follow-up to #158 (agent framework + chat-summary API); reviewer trace posted in the #158 comment thread and reproduced per-commit here. No separate brief — the P1 root cause and fix are captured in the `c1fb458` commit body.

## How verified

Docker `golang:1.25-bookworm` + `libtokenizers.a`, mirroring the CI `Test (race, cgo)` job:

```
CGO_ENABLED=1 CGO_LDFLAGS='-L/usr/local/lib' go vet -buildvcs=false ./...        → clean
CGO_ENABLED=1 CGO_LDFLAGS='-L/usr/local/lib' go build -buildvcs=false ./...      → clean
CGO_ENABLED=1 CGO_LDFLAGS='-L/usr/local/lib' go test -buildvcs=false -count=1 \
  ./internal/agent/... ./internal/api/handler/... ./internal/model/...
  → ok  internal/agent       0.075s   (includes 4 new regression tests, all PASS)
  → ok  internal/api/handler 11.270s
  → ok  internal/model       0.002s
```

Four new regression tests in `internal/agent/tool_summarize_chunk_first_turn_cgo_test.go` — all four fail against the pre-fix code (`7e888b1`):
- `TestGetSessionMessagePool_FirstTurn_NoAgentMessageRowsYet` — the exact 4-reviewer scenario: seeds evidence only with ZERO `agent_message` tool rows, asserts pool size N and CitationIndex is dense 1..N.
- `TestGetSessionMessagePool_CacheHotPath` — cache preferred over evidence JSON on the hot path.
- `TestGetSessionMessagePool_CacheMissEvidenceFallback` — long-session cold cache falls back to evidence JSON, symmetric with `buildCitationsForSession`.
[...]

PR #158, opened 2026-07-16
feat(agent): agent framework + chat-summary API with citations, SSE, …

wanyiwang06:
…refine, and reference-based context

Introduce a configurable agent framework and wire it into the summary service, exposing an agent-driven chat/summary/refine API that streams progress over SSE, persists citations and snapshots, and supports reference-based refinement using prior summaries as context.

What's included (see PR body for the B1-B9 breakdown):

- Agent framework: configurable prompts + tools, per-user cache isolation, fine-grained authz, multi-turn memory (DB-backed sliding window)
- POST /summaries/agent, GET /agent/chat/history, agent SSE streaming
- Citations for agent-generated summaries (SUM-20)
- SUM-21 SSE scenarios, SUM-24 optional origin_channel, SUM-25 summary_refine profile
- SUM-36 snapshot_json on PersonalResult
- SUM-43 refine endpoint with shared runner builder
- Reference-based chat: replace form-based refine with reference-driven agent chat; strip conversational preamble on save; borrow citations
- 24h background cleanup for stale agent_message rows

DB migrations: agent_message table, snapshot_json column, referenced_task_ids.

Verification:
- `CGO_ENABLED=1 go build -buildvcs=false ./...` in golang:1.25-bookworm → clean build
- Local end-to-end: exercised in the deployed OCTO stack against the companion octo-web frontend PR

## Checklist

- [ ] Code compiles with no new warnings
- [ ] Linter and formatter pass
- [ ] Tests added/updated for the changes
- [ ] All tests pass locally
- [ ] Rebased onto latest `main`
- [ ] Commit messages follow [Conventional Commits](https://www.conventionalcommits.org/)
- [ ] Self-reviewed the full diff from a reviewer's perspective

## Breaking Changes

N/A

PR #168, opened 2026-07-22
fix(agent): env-configurable step timeout + whitelisted detail on 500 responses

wanyiwang06:
## Summary

Two related fixes to make agent chat resilient and observable under
realistic refine load — both traced back from a reproducible production
500 in `im.deepminer.com.cn` on `claude-sonnet-4-6` at ~83s:

```
[agent] chat runner error: context deadline exceeded
[GIN] | 500 | 1m23s | POST "/api/v1/agent/chat"
```

**c1 · env-configurable AgentStepTimeout** — `Policy.StepTimeout` was
hard-coded to 60s in every profile. Refine flows inject ~21K tokens of
referenced summary content per turn (see `profile.go` `summary_refine`
comment), and a single LLM planning call on sonnet/kimi with that much
context reproducibly needs 60-100s+ to complete. The 60s stepCtx
tripped before the LLM finished streaming, `runner.RunWithHistory`
returned err, and the request 500'd. Introduce `AGENT_STEP_TIMEOUT`
env (default 240s in `config.go`) that overrides the profile's static
StepTimeout at `GetProfile` lookup time; both existing profiles
(`summary`, `summary_refine`) benefit uniformly and any future profile
inherits automatically. Setting to 0 or leaving unset keeps the
60s code default (useful for tests / fast-fail dev). See commit
`b01c4a3`.

**c2 · whitelisted `detail` on 500 responses** — Users hitting a 500
currently see only `{"code":50000,"message":"agent chat failed"}` in
F12 and must ping ops for the backend log line to distinguish "my
request timed out" (actionable — smaller channel, shorter window)
from "something broke server-side" (open an issue). Add a `Detail`
field on `apiResponse` (`omitempty`; success responses unchanged) and
populate on all 5 500 sites in `agent_chat.go`. Value goes through
`safeErrorDetail()`, a whitelist of five well-known runner failure
modes whose text is known to contain no URLs / IPs / tokens / stack
fragments (`context deadline exceeded`, `context canceled`,
`max steps exceeded`, `LLM returned empty response ... at final step`,
`unknown agent profile`); everything else collapses to `internal error`.
See commit `66b7ae6`.

## Linked Spec

No standalone brief. Both fixes trace to the reproducer above and to
the constraint already documented in the original 500-site comment
(`agent_chat.go`: "真实错误只记服务端日志,避免向调用方泄漏上游 LLM
地址/网络/内部细节"). c2's whitelist honors that intent while giving
clients enough signal to react.

## How verified

Local octo-alex compose stack + `im-test.deepminer.com.cn`:

- **c1 reproducer**: seeded 2000 msgs into a group, ran refine on
  claude-sonnet-4-6 with `AGENT_STEP_TIMEOUT=300`. Same scenario
  that failed at 83s on unmodified `main` now completes in ~100s.
  Reverted the env → back to 60s → failure reproduces at ~80s.
- **c2 sanity**: temporarily set `AGENT_STEP_TIMEOUT=5` to force
  `context deadline exceeded`; F12 shows
  `{"code":50000,"message":"agent chat failed","detail":"context deadline exceeded"}`
  instead of the opaque original.

Automated:

- `TestGetProfile_StepTimeoutOverride` (new) — 4 sub-cases across
  both profiles: unset env / 240 / 0 / invalid; all pass.
- `TestSafeErrorDetail` (new) — 11 sub-cases: nil / 5 whitelisted
  patterns (including wrapped `context.DeadlineExceeded` via
  `errors.New`) / 3 redaction cases (LLM gateway URL, internal IP,
  DB error) / generic-unknown fallback. All pass.
- Full pkg CGO=1 test:
  - `internal/agent` — 0.278s pass
  - `internal/api/handler` — 11.367s pass
  - `internal/config` — 0.002s pass
- `go vet ./internal/...` — clean.

## COMPREHENSION

1. **What this change actually does**: gives ops a knob
   (`AGENT_STEP_TIMEOUT`) to prevent single-turn LLM planning from
   tripping stepCtx on slow models with large context, and gives
   clients enough detail on 500s to self-triage without pulling
   backend logs.
2. **What could break**: the override can be set too small — but
   set-to-0 or unset preserves original 60s behavior, so rollback
   is a config change with no code redeploy. Detail field is
   `omitempty` on the JSON side, so old clients that don't know
   about it just ignore it; the SSE side changes the sink signature
   but the two-arg helper is preserved as a wrapper for the
   missing-auth 40100 case that has no err in scope.
3. **How you know it works**: reproducer failure at 83s ↔ success
   at ~100s with env=300 is a clean before/after on the exact
   production scenario. Detail whitelist is unit-tested for both
   pass-through (5 patterns) and redaction (3 sensitive shapes).
   Reverse-verified once by removing the c1 code change locally —
   test fails; restoring — passes.

### DIFF
--- a/CONFIGURATION.md
+++ b/CONFIGURATION.md
@@ -17,6 +17,7 @@ All configuration is done via environment variables.
 | `LLM_API_KEY` | API key for the LLM gateway | Yes (Worker) | — |
 | `LLM_MODEL` | Model identifier to use for summarization | Yes | — |
 | `LLM_TIMEOUT` | LLM request timeout in seconds | No | `180` |
+| `AGENT_STEP_TIMEOUT` | Maximum duration in seconds for one agent planning step; streaming requests remain capped by the 300-second request deadline | No | `240` |
 | `LLM_MAX_TOKENS` | Maximum tokens for LLM response | No | `4096` |
 | `LLM_TEMPERATURE` | Sampling temperature for LLM | No | `0.3` |
 | `LLM_ENABLE_THINKING` | Enable extended thinking mode | No | `false` |
--- a/internal/agent/profile.go
+++ b/internal/agent/profile.go
@@ -132,13 +132,11 @@ func BuildRegistry(toolNames []string) (*Registry, error) {
 
 // GetProfile 取一个场景定义。未知场景报错。
 //
-// StepTimeout override: the static profile values above are code defaults
-// (60s) that historically fit the pipeline flow. Interactive agent chat —
-// especially refine flows injecting ~21K tokens of referenced summary per
-// turn — can push a single LLM planning call well past 60s on slower
-// models (kimi / sonnet with large context). AGENT_STEP_TIMEOUT env
-// overrides the static 240s default. Setting
-// to 0 disables the override and keeps the code default.
+// StepTimeout: the static profile values above are the code default (240s —
+// raised from a historical 60s that tripped stepCtx on large-context refine
+// turns). AGENT_STEP_TIMEOUT env overrides it per-deploy; an unset / 0 /
+// invalid value keeps the 240s static default. The runner reads
+// Policy.StepTimeout from here — nothing reads config for it.
 func GetProfile(name string) (Profile, error) {
 	p, ok := profiles[name]
 	if !ok {
--- a/internal/agent/profile.go
+++ b/internal/agent/profile.go
@@ -54,12 +54,12 @@ var profiles = map[string]Profile{
 	"chat": {
 		PromptFile: "chat",
 		Tools:      []string{"get_current_time", "extract_time_range"},
-		Policy:     Policy{MaxSteps: 8, MaxTokens: 8000, StepTimeout: 240e9},
+		Policy:     Policy{MaxSteps: 8, MaxTokens: 8000, StepTimeout: 240 * time.Second},
 	},
 	"summary": {
 		PromptFile: "summary",
 		Tools:      []string{"get_current_time", "extract_time_range", "list_channels", "narrow_channels_by_topic", "find_shared_channels", "peek_channel", "fetch_channel", "search_messages", "filter_relevant", "summarize_chunk", "merge_summaries"},
-		Policy:     Policy{MaxSteps: 20, MaxTokens: 60000, StepTimeout: 240e9},
+		Policy:     Policy{MaxSteps: 20, MaxTokens: 60000, StepTimeout: 240 * time.Second},
 	},
 	"summary_refine": {
 		PromptFile: "summary_refine",
@@ -76,7 +76,7 @@ var profiles = map[string]Profile{
 		// the old 60s tripped stepCtx before the LLM finished streaming and
 		// surfaced as `[agent] chat runner error: context deadline exceeded`.
 		// AGENT_STEP_TIMEOUT env still overrides at GetProfile time (see below).
-		Policy: Policy{MaxSteps: 15, MaxTokens: 120000, StepTimeout: 240e9},
+		Policy: Policy{MaxSteps: 15, MaxTokens: 120000, StepTimeout: 240 * time.Second},
 	},
 }
 
@@ -137,7 +137,7 @@ func BuildRegistry(toolNames []string) (*Registry, error) {
 // especially refine flows injecting ~21K tokens of referenced summary per
 // turn — can push a single LLM planning call well past 60s on slower
 // models (kimi / sonnet with large context). AGENT_STEP_TIMEOUT env
-// overrides the static default; default 240s (see config.go). Setting
+// overrides the static 240s default. Setting
 // to 0 disables the override and keeps the code default.
 func GetProfile(name string) (Profile, error) {
 	p, ok := profiles[name]
@@ -159,9 +159,9 @@ func GetToolFactory(name string) (ToolFactory, bool) {
 // agentStepTimeoutOverride returns the AGENT_STEP_TIMEOUT env value as a
 // duration if it parses to a positive integer, else 0 (meaning: keep the
 // profile's static StepTimeout). Read directly from os.Getenv rather than
-// via config.Config so profile lookup does not require SetSummaryDeps —
-// this matters for unit tests that build a runner without initializing
-// the whole deps container.
+// via config.Config so this remains the single source of truth and profile
+// lookup does not require SetSummaryDeps. This matters for unit tests that
+// build a runner without initializing the whole deps container.
 func agentStepTimeoutOverride() time.Duration {
 	v := strings.TrimSpace(os.Getenv("AGENT_STEP_TIMEOUT"))
 	if v == "" {
--- a/internal/agent/profile.go
+++ b/internal/agent/profile.go
@@ -5,8 +5,10 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"strconv"
 	"strings"
 	"sync"
+	"time"
 )
 
 // 提示词与工具的可配置机制：
@@ -52,12 +54,12 @@ var profiles = map[string]Profile{
 	"chat": {
 		PromptFile: "chat",
 		Tools:      []string{"get_current_time", "extract_time_range"},
-		Policy:     Policy{MaxSteps: 8, MaxTokens: 8000, StepTimeout: 60e9},
+		Policy:     Policy{MaxSteps: 8, MaxTokens: 8000, StepTimeout: 240e9},
 	},
 	"summary": {
 		PromptFile: "summary",
 		Tools:      []string{"get_current_time", "extract_time_range", "list_channels", "narrow_channels_by_topic", "find_shared_channels", "peek_channel", "fetch_channel", "search_messages", "filter_relevant", "summarize_chunk", "merge_summaries"},
-		Policy:     Policy{MaxSteps: 20, MaxTokens: 60000, StepTimeout: 60e9},
+		Policy:     Policy{MaxSteps: 20, MaxTokens: 60000, StepTimeout: 240e9},
 	},
 	"summary_refine": {
 		PromptFile: "summary_refine",
@@ -68,7 +70,13 @@ var profiles = map[string]Profile{
 		// refine (get_time + fetch + answer) blew through 40K by step 2 and
 		// triggered "已达 token 预算" mid-flow. 120K gives room for 4-5 tool
 		// steps at ~25K each. See CHAT-REFERENCE-BASED-DESIGN-v1 diagnostic.
-		Policy: Policy{MaxSteps: 15, MaxTokens: 120000, StepTimeout: 60e9},
+		//
+		// StepTimeout 240s (was 60s): a single LLM planning call on
+		// sonnet/kimi with 20-40K tokens of context routinely takes 60-100s;
+		// the old 60s tripped stepCtx before the LLM finished streaming and
+		// surfaced as `[agent] chat runner error: context deadline exceeded`.
+		// AGENT_STEP_TIMEOUT env still overrides at GetProfile time (see below).
+		Policy: Policy{MaxSteps: 15, MaxTokens: 120000, StepTimeout: 240e9},
 	},
 }
 
@@ -123,10 +131,21 @@ func BuildRegistry(toolNames []string) (*Registry, error) {
 }
 
 // GetProfile 取一个场景定义。未知场景报错。
+//
+// StepTimeout override: the static profile values above are code defaults
+// (60s) that historically fit the pipeline flow. Interactive agent chat —
+// especially refine flows injecting ~21K tokens of referenced summary per
+// turn — can push a single LLM planning call well past 60s on slower
+// models (kimi / sonnet with large context). AGENT_STEP_TIMEOUT env
+// overrides the static default; default 240s (see config.go). Setting
+// to 0 disables the override and keeps the code default.
 func GetProfile(name string) (Profile, error) {
 	p, ok := profiles[name]
 	if !ok {
 		return Profile{}, fmt.Errorf("unknown agent profile %q", name)
+	}
+	if override := agentStepTimeoutOverride(); override > 0 {
+		p.Policy.StepTimeout = override
 	}
 	return p, nil
 }
@@ -135,4 +154,22 @@ func GetProfile(name string) (Profile, error) {
 func GetToolFactory(name string) (ToolFactory, bool) {
 	f, ok := toolFactories[name]
 	return f, ok
+}
+
+// agentStepTimeoutOverride returns the AGENT_STEP_TIMEOUT env value as a
+// duration if it parses to a positive integer, else 0 (meaning: keep the
+// profile's static StepTimeout). Read directly from os.Getenv rather than
+// via config.Config so profile lookup does not require SetSummaryDeps —
+// this matters for unit tests that build a runner without initializing
+// the whole deps container.
+func agentStepTimeoutOverride() time.Duration {
+	v := strings.TrimSpace(os.Getenv("AGENT_STEP_TIMEOUT"))
+	if v == "" {
+		return 0
+	}
+	secs, err := strconv.Atoi(v)
+	if err != nil || secs <= 0 {
+		return 0
+	}
+	return time.Duration(secs) * time.Second
 }
--- a/internal/agent/profile_step_timeout_test.go
+++ b/internal/agent/profile_step_timeout_test.go
@@ -12,8 +12,8 @@ import (
 //   - 设 env > 0 → override 生效,3 个 profile 都受影响
 //   - 设 env=0 或非法 → 保持 default(不 override)
 func TestGetProfile_StepTimeoutOverride(t *testing.T) {
-	// 2 个 profile 都要跑一次 · summary / summary_refine(仓库里只有这两个)
-	profileNames := []string{"summary", "summary_refine"}
+	// 所有 profile 都要覆盖，保证统一 override 语义不会漏掉 chat 场景。
+	profileNames := []string{"chat", "summary", "summary_refine"}
 
 	// case 1: 未设 env · 保持 240s 静态 default
 	t.Run("no_env_keeps_static_default", func(t *testing.T) {
--- a/internal/agent/profile_step_timeout_test.go
+++ b/internal/agent/profile_step_timeout_test.go
@@ -1,0 +1,74 @@
+package agent
+
+import (
+	"testing"
+	"time"
+)
+
+// TestGetProfile_StepTimeoutOverride 覆盖 AGENT_STEP_TIMEOUT env 覆盖 profile 静态
+// StepTimeout 的路径。修 refine 场景 60s 硬编码 stepCtx 挂 (agent chat runner
+// error: context deadline exceeded @ 83s) 的核心机制;要保证:
+//   - 未设 env → 保持 profile 静态 default(现在 240s · 之前 60s)
+//   - 设 env > 0 → override 生效,3 个 profile 都受影响
+//   - 设 env=0 或非法 → 保持 default(不 override)
+func TestGetProfile_StepTimeoutOverride(t *testing.T) {
+	// 2 个 profile 都要跑一次 · summary / summary_refine(仓库里只有这两个)
+	profileNames := []string{"summary", "summary_refine"}
+
+	// case 1: 未设 env · 保持 240s 静态 default
+	t.Run("no_env_keeps_static_default", func(t *testing.T) {
+		t.Setenv("AGENT_STEP_TIMEOUT", "")
+		for _, name := range profileNames {
+			p, err := GetProfile(name)
+			if err != nil {
+				t.Fatalf("GetProfile(%q): %v", name, err)
+			}
+			if p.Policy.StepTimeout != 240*time.Second {
+				t.Errorf("profile %q: unset env should keep static default 240s, got %v",
+					name, p.Policy.StepTimeout)
+			}
+		}
+	})
+
+	// case 2: env=120 · override 生效(选一个不同于 240 的值来区分 default vs override)
+	t.Run("env_120_overrides_all_profiles", func(t *testing.T) {
+		t.Setenv("AGENT_STEP_TIMEOUT", "120")
+		for _, name := range profileNames {
+			p, err := GetProfile(name)
+			if err != nil {
+				t.Fatalf("GetProfile(%q): %v", name, err)
+			}
+			if p.Policy.StepTimeout != 120*time.Second {
+				t.Errorf("profile %q: AGENT_STEP_TIMEOUT=120 should override to 120s, got %v",
+					name, p.Policy.StepTimeout)
+			}
+		}
+	})
+
+	// case 3: env=0 · 保持 default(0 视为"关闭 override")
+	t.Run("env_zero_keeps_static_default", func(t *testing.T) {
+		t.Setenv("AGENT_STEP_TIMEOUT", "0")
+		p, err := GetProfile("summary_refine")
+		if err != nil {
+			t.Fatalf("GetProfile: %v", err)
+		}
+		if p.Policy.StepTimeout != 240*time.Second {
+			t.Errorf("env=0 should keep default 240s, got %v", p.Policy.StepTimeout)
+		}
+	})
+
+	// case 4: env 非法(负值 / 非数字) · 保持 default
+	t.Run("env_invalid_keeps_static_default", func(t *testing.T) {
+		invalidVals := []string{"-1", "abc", "60s", "1.5"}
+		for _, v := range invalidVals {
+			t.Setenv("AGENT_STEP_TIMEOUT", v)
+			p, err := GetProfile("summary_refine")
+			if err != nil {
+				t.Fatalf("GetProfile: %v", err)
+			}
+			if p.Policy.StepTimeout != 240*time.Second {
+				t.Errorf("env=%q should keep default 240s, got %v", v, p.Policy.StepTimeout)
+			}
+		}
+	})
+}
--- a/internal/agent/profile_test.go
+++ b/internal/agent/profile_test.go
@@ -5,6 +5,7 @@ import (
 	"path/filepath"
 	"strings"
 	"testing"
+	"time"
 )
 
 // TestGetProfile_SummaryRefine 验证 summary_refine profile 能正确装配
@@ -49,8 +50,8 @@ func TestGetProfile_SummaryRefine(t *testing.T) {
 	if profile.Policy.MaxTokens != 120000 {
 		t.Errorf("Policy.MaxTokens = %d, want 120000", profile.Policy.MaxTokens)
 	}
-	if profile.Policy.StepTimeout != 240e9 {
-		t.Errorf("Policy.StepTimeout = %d, want 240e9", profile.Policy.StepTimeout)
+	if profile.Policy.StepTimeout != 240*time.Second {
+		t.Errorf("Policy.StepTimeout = %v, want 240s", profile.Policy.StepTimeout)
 	}
 }
 
--- a/internal/agent/profile_test.go
+++ b/internal/agent/profile_test.go
@@ -49,8 +49,8 @@ func TestGetProfile_SummaryRefine(t *testing.T) {
 	if profile.Policy.MaxTokens != 120000 {
 		t.Errorf("Policy.MaxTokens = %d, want 120000", profile.Policy.MaxTokens)
 	}
-	if profile.Policy.StepTimeout != 60e9 {
-		t.Errorf("Policy.StepTimeout = %d, want 60e9", profile.Policy.StepTimeout)
+	if profile.Policy.StepTimeout != 240e9 {
+		t.Errorf("Policy.StepTimeout = %d, want 240e9", profile.Policy.StepTimeout)
 	}
 }
 
--- a/internal/api/handler/agent_chat.go
+++ b/internal/api/handler/agent_chat.go
@@ -9,6 +9,7 @@ import (
 	"log"
 	"net/http"
 	"regexp"
+	"strings"
 	"time"
 
 	"github.com/Mininglamp-OSS/octo-smart-summary/internal/agent"
@@ -220,7 +221,7 @@ func (h *AgentChatHandler) Chat(c *gin.Context) {
 		runner, system, err = h.buildRunnerForProfile(profileName, uid, req.SessionID)
 		if err != nil {
 			log.Printf("[agent] build runner for profile %q: %v", profileName, err)
-			c.JSON(http.StatusInternalServerError, apiResponse{Code: 50000, Message: "failed to initialize agent"})
+			c.JSON(http.StatusInternalServerError, apiResponse{Code: 50000, Message: "failed to initialize agent", Detail: safeErrorDetail(err)})
 			return
 		}
 	}
@@ -230,7 +231,7 @@ func (h *AgentChatHandler) Chat(c *gin.Context) {
 	history, err := h.store.LoadHistory(ctx, req.SessionID, uid)
 	if err != nil {
 		log.Printf("[agent] load history error: %v", err)
-		c.JSON(http.StatusInternalServerError, apiResponse{Code: 50000, Message: "agent chat failed"})
+		c.JSON(http.StatusInternalServerError, apiResponse{Code: 50000, Message: "agent chat failed", Detail: safeErrorDetail(err)})
 		return
 	}
 
@@ -256,8 +257,11 @@ func (h *AgentChatHandler) Chat(c *gin.Context) {
 	reply, newMsgs, err := runner.RunWithHistory(ctx, system, history, req.Message)
 	if err != nil {
 		// 真实错误只记服务端日志，避免向调用方泄漏上游 LLM 地址/网络/内部细节。
+		// Detail 走白名单：仅 context deadline / max steps / empty response 等
+		// 明确不含内部地址/IP/token 的 error 会被透传给客户端，其它一律为
+		// "internal error"（safeErrorDetail 保证），前端可据此区分超时 vs 未知错。
 		log.Printf("[agent] chat runner error: %v", err)
-		c.JSON(http.StatusInternalServerError, apiResponse{Code: 50000, Message: "agent chat failed"})
+		c.JSON(http.StatusInternalServerError, apiResponse{Code: 50000, Message: "agent chat failed", Detail: safeErrorDetail(err)})
 		return
 	}
 
@@ -311,7 +315,7 @@ func (h *AgentChatHandler) History(c *gin.Context) {
 	history, err := h.store.LoadHistory(ctx, sessionID, uid)
 	if err != nil {
 		log.Printf("[agent] load history error: %v", err)
-		c.JSON(http.StatusInternalServerError, apiResponse{Code: 50000, Message: "agent chat history failed"})
+		c.JSON(http.StatusInternalServerError, apiResponse{Code: 50000, Message: "agent chat history failed", Detail: safeErrorDetail(err)})
 		return
 	}
 
@@ -431,7 +435,7 @@ func (h *AgentChatHandler) ChatStream(c *gin.Context) {
 		if err != nil {
 			log.Printf("[agent] build runner for profile %q: %v", profileName, err)
 			sink := &sseSink{w: c.Writer}
-			h.writeSSEErrorViaSink(sink, 50000, "failed to initialize agent")
+			h.writeSSEErrorViaSinkWithDetail(sink, 50000, "failed to initialize agent", safeErrorDetail(err))
 			return
 		}
 	}
@@ -457,7 +461,7 @@ func (h *AgentChatHandler) ChatStream(c *gin.Context) {
 	history, err := h.store.LoadHistory(ctx, req.SessionID, uid)
 	if err != nil {
 		log.Printf("[agent] load history error: %v", err)
-		h.writeSSEErrorViaSink(sink, 50000, "agent chat failed")
+		h.writeSSEErrorViaSinkWithDetail(sink, 50000, "agent chat failed", safeErrorDetail(err))
 		return
 	}
 
@@ -481,7 +485,7 @@ func (h *AgentChatHandler) ChatStream(c *gin.Context) {
 	reply, newMsgs, err := runner.RunWithHistory(ctx, system, history, req.Message)
 	if err != nil {
 		log.Printf("[agent] chat runner error: %v", err)
-		h.writeSSEErrorViaSink(sink, 50000, "agent chat failed")
+		h.writeSSEErrorViaSinkWithDetail(sink, 50000, "agent chat failed", safeErrorDetail(err))
 		return
 	}
 
@@ -536,9 +540,19 @@ func (h *AgentChatHandler) writeSSEDoneViaSink(sink *sseSink, reply, sessionID s
 
 // writeSSEErrorViaSink writes an error SSE event via the provided sink.
 func (h *AgentChatHandler) writeSSEErrorViaSink(sink *sseSink, code int, message string) {
+	h.writeSSEErrorViaSinkWithDetail(sink, code, message, "")
+}
+
+// writeSSEErrorViaSinkWithDetail is the detail-aware variant. Emits the
+// standard SSE error frame plus a safe-to-expose detail string when
+// non-empty. See safeErrorDetail below for the whitelist policy.
+func (h *AgentChatHandler) writeSSEErrorViaSinkWithDetail(sink *sseSink, code int, message string, detail string) {
 	data := map[string]interface{}{
 		"code":    code,
 		"message": message,
+	}
+	if detail != "" {
+		data["detail"] = detail
 	}
 	jsonData, err := json.Marshal(data)
 	if err != nil {
@@ -547,4 +561,42 @@ func (h *AgentChatHandler) writeSSEErrorViaSink(sink *sseSink, code int, message
 	}
 
 	sink.write("error", jsonData)
+}
+
+// safeErrorDetail returns a compact, safe-to-expose string derived from err
+// for inclusion in JSON/SSE error responses. Whitelist-only: we return the
+// raw err.Error() only for a small set of well-known agent runner failure
+// modes whose text is known not to embed URLs, IPs, tokens, or stack
+// fragments. Everything else collapses to "internal error", which lets the
+// client distinguish "your request timed out" (actionable) from "something
+// broke server-side" (open an issue / poll the ops channel) without
+// leaking backend geometry.
+//
+// Grow this whitelist deliberately: each new pattern must be traceable to
+// a specific errors.New / fmt.Errorf site in the runner or handler code
+// path whose format string contains no operator-supplied data.
+func safeErrorDetail(err error) string {
+	if err == nil {
+		return ""
+	}
+	msg := err.Error()
+	switch {
+	case strings.Contains(msg, "context deadline exceeded"):
+		// runner.go stepCtx timeout (single LLM planning call). Tune via
+		// AGENT_STEP_TIMEOUT env; see config.go / profile.go.
+		return "context deadline exceeded"
+	case strings.Contains(msg, "context canceled"):
+		// upstream cancellation (client disconnect or outer ctx timeout).
+		return "context canceled"
+	case strings.Contains(msg, "max steps exceeded"):
+		// runner.go MaxSteps loop guard. Model failed to converge.
+		return "max steps exceeded"
+	case strings.Contains(msg, "LLM returned empty response with no tool_calls"):
+		// runner.go final-step empty content guard (SUM-158 blocker follow-up).
+		return "LLM returned empty response with no tool_calls at final step"
+	case strings.Contains(msg, "unknown agent profile"):
+		// profile.go GetProfile lookup miss.
+		return "unknown agent profile"
+	}
+	return "internal error"
 }
--- a/internal/api/handler/agent_chat.go
+++ b/internal/api/handler/agent_chat.go
@@ -5,6 +5,7 @@ import "sync"
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"log"
 	"net/http"
@@ -500,8 +501,10 @@ func (h *AgentChatHandler) ChatStream(c *gin.Context) {
 
 // writeSSEProgressViaSink writes a progress SSE event via the provided sink.
 // Contract (stable with frontend): only abstract, non-leaking fields are emitted —
-//   phase (safe enum: understand|retrieve|filter|distill|compose|reply),
-//   step / ofSteps / elapsed_ms, and an optional integer count (omitted when 0).
+//
+//	phase (safe enum: understand|retrieve|filter|distill|compose|reply),
+//	step / ofSteps / elapsed_ms, and an optional integer count (omitted when 0).
+//
 // It intentionally does NOT emit the raw tool name, an internal label, or free-text detail.
 func (h *AgentChatHandler) writeSSEProgressViaSink(sink *sseSink, phase string, step, ofSteps, count int, elapsedMs int64) {
 	data := map[string]interface{}{
@@ -579,22 +582,21 @@ func safeErrorDetail(err error) string {
 	if err == nil {
 		return ""
 	}
-	msg := err.Error()
 	switch {
-	case strings.Contains(msg, "context deadline exceeded"):
+	case errors.Is(err, context.DeadlineExceeded):
 		// runner.go stepCtx timeout (single LLM planning call). Tune via
-		// AGENT_STEP_TIMEOUT env; see config.go / profile.go.
+		// AGENT_STEP_TIMEOUT env; see profile.go / CONFIGURATION.md.
 		return "context deadline exceeded"
-	case strings.Contains(msg, "context canceled"):
+	case errors.Is(err, context.Canceled):
 		// upstream cancellation (client disconnect or outer ctx timeout).
 		return "context canceled"
-	case strings.Contains(msg, "max steps exceeded"):
+	case strings.Contains(err.Error(), "max steps exceeded"):
 		// runner.go MaxSteps loop guard. Model failed to converge.
 		return "max steps exceeded"
-	case strings.Contains(msg, "LLM returned empty response with no tool_calls"):
+	case strings.Contains(err.Error(), "LLM returned empty response with no tool_calls"):
 		// runner.go final-step empty content guard (SUM-158 blocker follow-up).
 		return "LLM returned empty response with no tool_calls at final step"
-	case strings.Contains(msg, "unknown agent profile"):
+	case strings.Contains(err.Error(), "unknown agent profile"):
 		// profile.go GetProfile lookup miss.
 		return "unknown agent profile"
 	}
--- a/internal/api/handler/agent_chat_detail_test.go
+++ b/internal/api/handler/agent_chat_detail_test.go
@@ -1,0 +1,82 @@
+package handler
+
+import (
+	"context"
+	"errors"
+	"testing"
+)
+
+// TestSafeErrorDetail 覆盖 500 response Detail 字段的白名单规则(#agent_chat.go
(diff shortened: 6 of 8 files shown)


lml2468 (dismissed):
# PR Review — PR #168 (Mininglamp-OSS/octo-smart-summary)

Reviewed at head SHA `66b7ae655925` against merge-base (base `main`). Two related agent-chat resilience/observability fixes traced from a reproducible ~83s production 500. Security-relevant (error-detail exposure).

Intent: **X** = (c1) make `Policy.StepTimeout` env-configurable (`AGENT_STEP_TIMEOUT`, default 240s) so long refine-flow LLM planning calls don't trip the hard-coded 60s stepCtx and 500; (c2) expose a **whitelisted** `detail` on 500 responses so clients can distinguish timeout vs internal error — without leaking backend geometry. Affecting **Z** = `GetProfile`, `config`, agent-chat error paths.

- `Build`, `Test`, `Test (race, cgo)` (1m52s), `Vet` → **pass**; `check-sprint` (governance) → fail.
- Local: build/vet clean; `go test ./internal/agent ./internal/api/handler ./internal/config` → pass (incl. new timeout + detail tests).

## 1. Spec compliance — ✅
Both fixes match the described intent; scope is tight. Nothing missing/over-built.

## 2. Code quality — Approved

**c2 — whitelisted 500 detail (the security-sensitive part) — correct default-deny.**
`safeErrorDetail` is whitelist-only: it matches a small set of known-safe runner failure signatures and — critically — **returns a hardcoded literal for each, never the raw `err.Error()`**. So even if the underlying error embeds an IP/URL/token (e.g. `dial tcp 10.0.0.42:8080…`), only the sanitized literal or `"internal error"` reaches the client. I verified there is **no `return msg` / `return err.Error()`** path in the function — every branch is a literal, default → `"internal error"`. The new `Detail` field is `omitempty` (success responses omit it). This directly closes the "leak upstream LLM address" concern that the prior code comment worried about, while giving clients the timeout-vs-broke signal. The test asserts the exact adversarial case: `dial tcp 10.0.0.42…` → `"internal error"`, whitelisted patterns (incl. wrapped via `Contains`) pass through.

**c1 — env-configurable StepTimeout — correct, invariant preserved.**
`GetProfile` applies `agentStepTimeoutOverride()` (reads `AGENT_STEP_TIMEOUT` directly via `os.Getenv`, documented: avoids requiring `SetSummaryDeps` in unit tests) over the static 60s. Default 240s < the 300s ChatStream outer backstop (`agent_chat.go:410`), so stepCtx stays bounded by the wall-clock ceiling — the #158/#161 relationship (planning-call bound, not tool-exec bound) is preserved; `runTools` still uses the outer ctx. Parse guards: non-int / ≤0 / unset → 0 → keeps the static default. Applies uniformly to both profiles + any future one. Tested (unset→60s, env=240→override across all profiles).

- **No raw-error passthrough** in `safeErrorDetail` — every whitelist branch returns a constant string; leaky network errors collapse to `"internal error"` (test-asserted).
- **Timeout invariant:** 240s default < 300s backstop; tool execution unaffected (still outer ctx from #161).
- **No regression:** the empty-response / max-steps / uid-guard behavior from the #158/#161 arc is intact; full suite + race+cgo green.

### Non-blocking notes
- 🟡 **`Config.AgentStepTimeout` is loaded but never read** — the override path uses `os.Getenv` directly in `profile.go`, so the config field (and its `envInt("AGENT_STEP_TIMEOUT", 240)` default) is effectively dead. Worse, it creates a **default divergence**: with the env unset, `config.go` implies 240s but the runner actually keeps 60s (`agentStepTimeoutOverride()` returns 0 when unset). Effective default is 60s unless the env is explicitly set — despite config/docs claiming 240. Recommend either wiring the Config field through or dropping it and aligning the doc/default (make unset → 240 to match config). Behavioral, but not a correctness break: prod sets the env; tests want 60s fast-fail.
- 🟡 **`AGENT_STEP_TIMEOUT` > 300** would exceed the outer ChatStream backstop, so values above ~300 are silently capped by the outer ctx. Worth documenting the effective ceiling.
- 🟡 **Whitelist maintenance:** `safeErrorDetail` matches by substring on error text — future refactors that reword these runner errors would silently drop them to `"internal error"`. The comment already flags "grow deliberately"; a light coupling (shared sentinel errors via `errors.Is`) would be more robust. Minor.

## 3. Test assessment
- Both new tests are strong and targeted: `TestSafeErrorDetail` covers whitelist pass-through, wrapped-error matching, and the leaky-network→"internal error" security case; `TestGetProfile_StepTimeoutOverride` covers unset-default and override-all-profiles. They'd fail without the fixes.

## 4. Coverage / blind spots
- Executed: build, vet, agent+handler+config tests, **live CI** (race+cgo green), full read of `safeErrorDetail` (confirmed literal-only returns) and the timeout override path.

- Decide on the `Config.AgentStepTimeout` field: wire it or drop it, and reconcile the unset-default (60 vs 240).

## Verdict
**APPROVE** — both gates pass. The whitelisted 500-detail is a correct default-deny design (literal-only returns; leaky errors collapse to "internal error", test-asserted), and the env-configurable StepTimeout fixes the refine-flow 60s 500 while preserving the 240s < 300s backstop invariant. Functional CI (incl. race+cgo) is green. The one substantive note — `Config.AgentStepTimeout` loaded-but-unread with a 60-vs-240 unset-default divergence — is a config-tidiness/documentation issue, not a correctness break (prod sets the env).
Suggested decision: `Safe to merge after non-blocking follow-up` — reconcile the AgentStepTimeout config field/default; merge once check-sprint clears.

Jerry-Xin (dismissed):
The PR is relevant, but its advertised 240-second default is not connected to runtime behavior.

🔴 Blocking

- 🔴 Critical — `Config.AgentStepTimeout` is dead configuration. [`internal/config/config.go:168`](/Users/mlamp/.openclaw/tmp/pr-review-7oj226xf/repo/internal/config/config.go:168) defaults it to 240, but [`internal/agent/profile.go:136`](/Users/mlamp/.openclaw/tmp/pr-review-7oj226xf/repo/internal/agent/profile.go:136) bypasses `Config` and reads the raw environment directly. Consequently, when `AGENT_STEP_TIMEOUT` is unset, `Config` reports 240 while every runner still uses 60 seconds. Invalid values are similarly interpreted inconsistently: configuration falls back to 240, while profiles fall back to 60. Wire the loaded configuration into runner/profile construction, or remove the unused field and clearly retain 60 as the sole default. Add an integration-level test proving the value returned by `config.Load()` reaches the runner policy.

💬 Non-blocking

- 🔵 Suggestion — [`internal/agent/profile_step_timeout_test.go:17`](/Users/mlamp/.openclaw/tmp/pr-review-7oj226xf/repo/internal/agent/profile_step_timeout_test.go:17) omits the existing `chat` profile despite claiming all profiles are covered. Include it to protect the intended uniform override behavior.

✅ Highlights

- The error detail function returns fixed canonical strings rather than raw matched errors, preventing appended URLs, tokens, or internal addresses from leaking.
- JSON and SSE paths apply the same detail policy.
- Agent and configuration tests passed locally. Handler tests could not link because the local `tokenizers` library is unavailable.

Jerry-Xin (dismissed):
The PR is well-targeted and the security-sensitive `detail` whitelist is implemented correctly, but the advertised 240s default is not actually connected to runtime behavior.

🔴 Blocking

- 🔴 `Config.AgentStepTimeout` is dead configuration. `internal/config/config.go:168` defaults `AGENT_STEP_TIMEOUT` to 240 via `envInt(...,240)`, but the runner path never reads `Config`: `internal/agent/profile.go` `agentStepTimeoutOverride()` calls `os.Getenv("AGENT_STEP_TIMEOUT")` directly and returns 0 when unset, which keeps the profile's static 60s. Net effect: when the env is unset, `config.Load()` reports 240 while every profile lookup still uses 60s; invalid values diverge too (config falls back to 240, profile falls back to 60s). This contradicts the PR description ("default 240s in config.go"). Either wire the loaded config value into `GetProfile`/runner construction so the 240s default is real, or delete the unused `Config.AgentStepTimeout` field and document 60s as the sole code default when the env is unset. An integration-level assertion that the value from `config.Load()` reaches `Policy.StepTimeout` would lock this down.

💬 Non-blocking

- 🔵 `internal/agent/profile_step_timeout_test.go` covers only `summary` and `summary_refine` and its comment claims those are the only profiles in the repo, but a `chat` profile also exists (`internal/agent/profile.go`). Since the override is meant to apply uniformly to all profiles, add `chat` to the covered set (and fix the comment) to protect that intent.

✅ Highlights

- `safeErrorDetail` returns fixed canonical literals per matched case rather than the raw matched error, so an appended internal URL / IP / token / stack fragment can never ride along even when the underlying error embeds one (verified against the LLM client error path, which wraps `http do: Post "<url>": context deadline exceeded`). Everything unmatched collapses to `internal error`. This is stricter than the "passthrough of err.Error()" wording in the PR body — good.
- The same detail policy is applied consistently on both the JSON and SSE error paths; the two-arg SSE helper is preserved as a wrapper for the no-err 40100 case.
- `Detail` is `omitempty`, so success responses and old clients are unaffected.
- Both new unit test files are thorough: the detail test explicitly asserts redaction of an internal gateway URL, an internal IP, and a DB error shape.

yujiawei (changes requested):
# Code Review — PR #168 (octo-smart-summary)

Reviewed at head SHA `66b7ae655925d0c89e948bf3ff26884152976bcd` against merge-base `658ef95`. Build, `go vet`, and the full `internal/agent` + `internal/api/handler` + `internal/config` suites pass locally (incl. `-race`). This review used two independent analysis passes in addition to a line-level read; findings below are the reconciled result.

**Verdict: CHANGES_REQUESTED** — one functional (P1) blocker in c1. The c2 security work is sound (see below).

---

## Part 1 — Spec compliance

- **Missing:** none — both stated fixes are present.
- **Extra:** `config.AgentStepTimeout` field + `envInt("AGENT_STEP_TIMEOUT", 240)` in `config.go` is added but never consumed by any code path (dead field) — see P1 below.
- **Deviation:** the effective runtime default does **not** match what the code documents. `config.go:41-50` comments "Default 240s covers observed p99", but with the env unset the runner actually uses **60s**. The PR's own body carries the same contradiction ("default 240s in config.go" vs "leaving unset keeps the 60s code default").

**Spec: ❌** (deviation — advertised default diverges from actual behavior; extra dead config field).

## Part 2 — Code quality

### P1 — The advertised 240s default is inert; the fix is opt-in via an undocumented env var

`internal/config/config.go:50,168`
```go
AgentStepTimeout int
...
AgentStepTimeout: envInt("AGENT_STEP_TIMEOUT", 240),
```
Nothing reads `Config.AgentStepTimeout` — a repo-wide search for `.AgentStepTimeout` returns zero call sites. The timeout is driven **exclusively** by `agentStepTimeoutOverride()`, which reads `os.Getenv` directly and returns `0` when unset:

`internal/agent/profile.go:141,159-169`
```go
if override := agentStepTimeoutOverride(); override > 0 {
    p.Policy.StepTimeout = override
}
...
func agentStepTimeoutOverride() time.Duration {
    v := strings.TrimSpace(os.Getenv("AGENT_STEP_TIMEOUT"))
    if v == "" {
        return 0   // -> GetProfile keeps the static 60s
    }
    ...
}
```
Consequences:
1. **Default deploy stays broken.** With `AGENT_STEP_TIMEOUT` unset, `StepTimeout` remains 60s — the exact value that produced the production `context deadline exceeded @ 83s` this PR is meant to fix. The fix only activates if an operator remembers to set the env var, and this PR adds **no** deployment wiring for it (no `.env.example` / compose / manifest entry in this repo).
2. **Split-brain / misleading config.** `Config.AgentStepTimeout=240` with the "covers observed p99" comment reads as "production defaults to 240s" — which is false. The next person triaging a timeout incident will trust that comment and be wrong.

**Required before merge — pick one:**
- (a) Wire `Config.AgentStepTimeout` into the override so an unset env yields the intended **240s** (make the profile/runner builder read config, defaulting to 240); **or**
- (b) If "60s unless explicitly set" is genuinely intended, **remove** the dead `Config.AgentStepTimeout` field, fix the "Default 240s covers observed p99" comment to state the true default, and ensure the deployment sets `AGENT_STEP_TIMEOUT` (call it out in the PR / deploy repo).

Either way the code must stop advertising a 240s default it does not deliver. (Note the testability rationale in the `agentStepTimeoutOverride` comment — reading env directly to avoid `SetSummaryDeps` in unit tests — is reasonable; the problem is only that config's 240 default is orphaned. Option (a) can still keep an env-first read with a 240 fallback.)

### P2 (advisory, non-blocking)

- **Silent overflow on absurd values** — `profile.go:167` `time.Duration(secs) * time.Second`: values above `math.MaxInt64/1e9` (~9.2e9 s) wrap instead of being rejected. Practically unreachable, but an explicit upper bound would fail loud instead of silent.
- **`os.Getenv` on every `GetProfile` call** — negligible in practice (one lookup per request in `buildRunnerForProfile`, not a tight loop), noted for completeness.
- **Non-stream `Chat()` has no outer ctx backstop** — `agent_chat.go:201` uses `c.Request.Context()` directly (unlike `ChatStream`'s 300s `WithTimeout` at `agent_chat.go:410`), and `publicSrv` (`cmd/summary-api/main.go:121`) sets no `WriteTimeout`. Pre-existing, but raising `StepTimeout` to 240s widens the window in which a single non-stream request holds a connection bounded only by client patience. Consider a symmetric outer deadline on the non-stream path.

### c2 — whitelisted `detail` on 500s: ✅ verified safe

- `safeErrorDetail` (`agent_chat.go:578`) returns **hardcoded canonical constants**, never the raw `err.Error()`. The adversarial case — an error like `context deadline exceeded while dialing 10.0.0.42` — returns the clean `"context deadline exceeded"`, not the wrapped message. No raw-message return path exists; unknown errors (LLM gateway `http status 500: <body>`, `dial tcp 10.0.0.x`, gorm messages) all collapse to `"internal error"`.
- All error sinks are covered: 4 `c.JSON` 500 sites + 3 SSE 500 sites route through `safeErrorDetail`; the 40100 missing-auth SSE path correctly stays detail-less via the 2-arg wrapper. `Detail` is `omitempty` so success responses and old clients are unaffected.
- Redaction is unit-tested (URL / internal IP / DB fragment → `internal error`).

**Quality: Changes-Requested** (P1 above; c2 clean).

## Overall

**CHANGES_REQUESTED.** c2 (the security-sensitive part) is well-constructed and safe to ship as-is. c1's mechanism is correct *when the env is set*, but as written the primary outage fix is inert in a default deployment and the config layer advertises a 240s default it never applies — that must be reconciled before merge.

## Human-verify note (security-sensitive PR)
- Confirm `octo-deployment` (or the target environment) actually sets `AGENT_STEP_TIMEOUT`; without it, deploying this PR does not change the 60s production behavior.
- The c2 whitelist is safe for the 5 current patterns; when the whitelist grows, each new entry must remain a fixed canonical string (never pass `err.Error()` through) to preserve the redaction guarantee.

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# Code Review — PR #168 (octo-smart-summary)

## Summary

Two focused fixes for a production 500 observed on `claude-sonnet-4-6` refine flows: (c1) an `AGENT_STEP_TIMEOUT` env var that overrides the 60s hard-coded profile `StepTimeout`, and (c2) a whitelisted `Detail` field on 500 responses so clients can distinguish timeout from unknown error without pulling backend logs.

The whitelist design in `safeErrorDetail()` is solid — it returns fixed strings for five known-safe patterns and collapses everything else to `"internal error"`, which correctly prevents leakage of URLs, IPs, or stack fragments. The `writeSSEErrorViaSinkWithDetail` SSE variant and the backward-compatible `writeSSEErrorViaSink` wrapper are clean. All 7 error sites in `agent_chat.go` (across `Chat`, `ChatStream`, and `History` handlers) are covered consistently.

## Verification

Static analysis only at head `66b7ae6`; build and tests not executed in this environment.

- ✅ **Dual-path parity (C1)** — all three handler entry points (`Chat`/`ChatStream`/`History`) consistently apply `Detail: safeErrorDetail(err)` on every 500 site. The 401 sites correctly omit `Detail` (no `err` in scope).
- ✅ **Control-flow safety (C2)** — `GetProfile` returns a value copy of `Profile`; the `p.Policy.StepTimeout = override` assignment mutates the copy, not the global `profiles` map. No cross-request state pollution.
- ✅ **Whitelist returns fixed strings** — every `strings.Contains` branch in `safeErrorDetail` returns a compile-time constant, not a substring of `err.Error()`. Even if a whitelisted pattern appears alongside sensitive data in the same error, only the safe constant is emitted.
- ✅ **Backward-compatible JSON** — `apiResponse.Detail` uses `json:"detail,omitempty"`. Success responses and non-500 errors produce no `detail` key on the wire.
- ✅ **Input-shape edges** — `agentStepTimeoutOverride` handles empty, zero, negative, and non-numeric env values correctly (all return 0 → no override → keeps 60s default).
- ✅ **SSE wrapper backward compat** — `writeSSEErrorViaSink` delegates to `writeSSEErrorViaSinkWithDetail` with empty detail, preserving the 401 call site.

## Findings

### P2 — Dead `Config.AgentStepTimeout` field (`internal/config/config.go:50`)

`Config.AgentStepTimeout` is declared at `:50` and loaded from `AGENT_STEP_TIMEOUT` env at `:168` (default 240), but **no production code reads `Config.AgentStepTimeout`**. The actual override mechanism in `profile.go:agentStepTimeoutOverride()` reads `os.Getenv("AGENT_STEP_TIMEOUT")` directly — with no default (empty → 0 → no override → keeps 60s).

The runtime default when the env var is unset is therefore **60s** (the profile static value), not the 240s documented in the config.go field comment ("Default 240s covers observed p99"). Operators reading `config.go` may set up deployment configs expecting a 240s default that doesn't take effect.

Fix: either (a) wire `Config.AgentStepTimeout` into the profile lookup, or (b) remove the dead field from `Config` and document the env-only mechanism in `profile.go` alone. Option (b) matches the explicit design rationale in `agentStepTimeoutOverride`'s doc comment.

### P2 — Test misses `chat` profile (`internal/agent/profile_step_timeout_test.go:16`)

`profileNames` covers only `summary` and `summary_refine`, but `profile.go` defines three profiles (`chat`, `summary`, `summary_refine`). The `chat` profile also has `StepTimeout: 60e9` and should be included. The test comment ("仓库里只有这两个") is inaccurate.

- All 7 error sites in `agent_chat.go` (3 in `Chat`, 3 in `ChatStream`, 1 in `History`) correctly use `safeErrorDetail(err)`.
- The `Detail` field placement between `Message` and `Data` in `apiResponse` is safe — JSON marshaling uses tags, not field order.
- `agentStepTimeoutOverride` correctly uses `strconv.Atoi` (not `ParseInt`/`ParseDuration`), which rejects float/negative/suffixed values as intended.
- The `writeSSEErrorViaSinkWithDetail` function only adds `detail` to the JSON map when non-empty, preserving the existing SSE wire format for 401/non-detail error events.
- No authorization, visibility, or persistence contracts are changed.

## Verdict: APPROVED

No correctness, security, or blocking issues. The two P2 items (dead config field and test gap) are non-blocking quality improvements. The core mechanism — env-based step timeout override and whitelisted error detail on 500s — is well-designed and consistently applied.

[Octo-Q] verdict: APPROVED — no P0/P1 findings; two P2 items (dead config field, incomplete test coverage) are non-blocking.

yujiawei (dismissed):
# Code Review — PR #168 (octo-smart-summary) — re-review

Re-reviewed at head SHA `a291d7738a0a9ee6e4793e0c4cc39d5282b366f8` (force-push after my previous CHANGES_REQUESTED against `66b7ae6`). Reviewed against the current head with a fresh merge-base; the prior approval does not carry over automatically. Build, `go vet ./internal/...`, and the full `internal/agent` + `internal/api/handler` + `internal/config` suites pass locally, including `-race`.

**Verdict: APPROVED** — the P1 blocker from the previous round is fixed. Only non-blocking P2 cleanup remains.

---

## What changed since the last review

The new commit `e8b1ce6` ("raise StepTimeout default to 240s + env-configurable override") addresses the P1 directly:

`internal/agent/profile.go:57,62,79`
```go
"chat":           Policy{... StepTimeout: 240e9},
"summary":        Policy{... StepTimeout: 240e9},
"summary_refine": Policy{... StepTimeout: 240e9},
```
All three profile static `StepTimeout` values are raised 60s → **240s**. This is a valid resolution of the P1: the effective runtime default with `AGENT_STEP_TIMEOUT` unset is now genuinely 240s (the documented target), so a default deployment no longer reproduces the production `context deadline exceeded @ 83s`. The `AGENT_STEP_TIMEOUT` env override at `GetProfile` time is preserved. Tests were updated to assert the 240s default and now use env=120 (distinct from the default) to prove override vs default independently.

`c2` (whitelisted `Detail` on 500s) is byte-identical to the version I already cleared — no re-litigation needed. Re-confirmed on this head: `safeErrorDetail` returns fixed canonical constants only, all 7 error sinks covered, redaction unit-tested, `omitempty` intact.

## Part 1 — Spec compliance

- **Missing:** none.
- **Extra / deviation:** the previous deviation (advertised 240s default not delivered) is resolved. Static defaults now match the documented intent.

**Spec: ✅**

## Part 2 — Code quality

**Quality: Approved.** No P0/P1. Remaining items are all non-blocking:

- **P2 — `Config.AgentStepTimeout` is now dead code.** `internal/config/config.go:50,168` still declares the field and loads `envInt("AGENT_STEP_TIMEOUT", 240)`, but nothing reads `Config.AgentStepTimeout` (repo-wide `.AgentStepTimeout` = 0 reads); the override path uses `os.Getenv` directly in `profile.go`. This is now harmless (behavior no longer depends on it) but is orphaned config surface — recommend removing the field, or wiring it as the single source of truth so config and profile can't drift again. Non-blocking.
- **P2 — silent int64 overflow** on absurd `AGENT_STEP_TIMEOUT` values (`profile.go` `time.Duration(secs) * time.Second`, values > ~9.2e9 s wrap). Practically unreachable; an explicit upper bound would fail loud.
- **P2 — non-stream `Chat()` has no outer ctx backstop** (`agent_chat.go:201` uses `c.Request.Context()` directly, unlike `ChatStream`'s 300s `WithTimeout`), and `publicSrv` sets no `WriteTimeout`. Pre-existing, but raising `StepTimeout` to 240s widens the window a single non-stream request can hold a connection. Consider a symmetric outer deadline. Non-blocking.

## Overall

**APPROVED.** The blocking P1 (inert 240s default) is fixed by raising the static profile timeouts; the security-sensitive c2 whitelist remains sound. The three P2 items are quality follow-ups that do not block merge — the dead `Config.AgentStepTimeout` field is the most worth cleaning up to prevent future config/profile drift.

## Human-verify note (security-sensitive PR)
- c2 whitelist safe for the 5 current patterns; any future addition must remain a fixed canonical string (never pass `err.Error()` through) to keep the redaction guarantee.
- Default step timeout is now 240s even without env config — confirm this is acceptable against the outer `ChatStream` 300s backstop (it is: 240s < 300s) and any upstream LB/proxy idle timeouts in the deploy environment.

lml2468 (dismissed):
# PR Review — PR #168 (Mininglamp-OSS/octo-smart-summary)

🔄 Re-review at head SHA `95e7c10c5b33` against merge-base `658ef953` (base `main`). Re-anchored from my APPROVE at `66b7ae65` — this head **reconciles the timeout-config divergence** the panel (Steve/yujiawei 🔴, me 🟡) converged on. Base also advanced (#163/#167 merged in).

**Delta since `66b7ae65`: 7 files** — commit "reconcile timeout configuration review".

- `Build`, `Test`, `Test (race, cgo)` (1m29s), `Vet` → **pass**; `check-sprint` (governance) → fail.
- Local: build/vet clean; `go test ./internal/agent ./internal/api/handler ./internal/config` → pass.

## The finding that was escalated, now fixed
Prior head had `Config.AgentStepTimeout` loaded-but-unread (override read `os.Getenv` directly), so the runner's *effective* default with the env unset was 60s despite config/docs claiming 240 — a real config/behavior divergence contradicting the PR's stated "default 240s". I graded it 🟡 (config-tidiness, prod sets the env); Steve and yujiawei graded it 🔴. **Fair escalation on my part to acknowledge** — the docs↔behavior contradiction is more than cosmetic, and a "default 240s" that's actually 60s unset is a genuine correctness-of-intent bug. This head fixes it cleanly (the better of the two options I'd suggested):

- **Removed the dead `Config.AgentStepTimeout` field** and its `Load()` entry (`config.go`) — no more loaded-but-unread config.
- **Raised the static profile default to 240s** in all three profiles — `chat`, `summary`, `summary_refine` (`profile.go:57/62/79`). So the code default is now genuinely 240s everywhere; unset env → 240s, matching intent.
- `AGENT_STEP_TIMEOUT` env remains the single-source-of-truth override at `GetProfile`; comments updated to say "static 240s default".

## 1. Spec compliance — ✅
Now consistent with the PR's stated default (240s) across config, code, and docs (CONFIGURATION.md updated). c2 whitelist unchanged and correct.

## 2. Code quality — Approved
- ✅ **Timeout divergence resolved** — effective default is 240s in every profile with env unset; no dead config field. Invariant preserved: 240s ≤ 300s ChatStream backstop (equal-or-under; tool exec still on outer ctx per #161).
- ✅ **c2 whitelisted 500 detail** — unchanged from my prior review; `safeErrorDetail` still returns literal-only per whitelist branch, default `"internal error"`; leaky errors redacted (test-asserted). Correct default-deny.
- ✅ **Test coverage note fixed** — `profile_step_timeout_test.go` now iterates **all three** profiles including `chat` (the prior 🟡 gap), and the stale "only two profiles" comment is corrected. Cases: unset→240s, env override, env=0→240s, invalid (`-1`/`abc`/`60s`/`1.5`)→240s. Thorough.

- **No dead config remnant** — grep confirms `AgentStepTimeout` fully removed (field + Load).
- **Uniform 240s** across chat/summary/summary_refine; env override applies to all.
- **c2 safety intact** — literal-only returns, no raw-error passthrough; leaky network/DB errors → "internal error".

- 🟡 `AGENT_STEP_TIMEOUT` > 300 is silently capped by the outer ChatStream backstop — worth a doc note on the effective ceiling. yujiawei also suggested an explicit upper bound / overflow guard — reasonable hardening (a multi-billion-second value would overflow `time.Duration` math, though operationally implausible).
- 🟡 `safeErrorDetail` substring-matching is refactor-fragile (shared sentinel errors via `errors.Is` would be sturdier) — minor.
- 🟡 (yujiawei) non-streaming `Chat()` has no outer request-timeout equivalent to ChatStream's 300s backstop — worth confirming, but pre-existing and out of this PR's scope.

## 3. Test assessment
- The timeout test now covers all profiles + default/override/invalid paths; the detail test covers the security redaction cases. Both strong, would fail without the fixes.

## 4. Coverage / blind spots
- Executed: build, vet, agent+handler+config tests, **live CI** (race+cgo green), delta trace of the reconciliation + grep-confirmed dead-field removal.

- Optional: upper-bound guard on `AGENT_STEP_TIMEOUT`; outer timeout for non-streaming `Chat()`.

## Verdict
**APPROVE** — re-anchored to `95e7c10c`. The timeout-config divergence (dead `Config.AgentStepTimeout` + 60-vs-240 unset default) that the panel escalated to 🔴 is properly reconciled: the dead field is removed, the static default is genuinely 240s across all three profiles, and the test now covers all profiles + default/override/invalid. The c2 whitelisted-detail security design remains correct. Functional CI (incl. race+cgo) is green. I'm aligning up to the panel's 🔴 severity in retrospect — the docs↔behavior contradiction warranted a block, and it's now cleanly fixed. Credit to Steve and yujiawei for the escalation.
Suggested decision: `Safe to merge after non-blocking follow-up` — merge once check-sprint clears.

wanyiwang06:
@Jerry-Xin 你的 step-timeout blocking 现已完整解决,请复审当前 head `2d44c07`：

- ✅ **240s 默认已接上真实运行路径**：`95e7c10` 把 3 个 profile 的静态 `StepTimeout` 从 60e9 → 240e9,`GetProfile` 直接返回它,`AGENT_STEP_TIMEOUT` env 仍在 `GetProfile` 处覆盖。
- ✅ **dead `Config.AgentStepTimeout` 已删**：`95e7c10` 移除了该字段及其 `Load()` 赋值(config.go −17 行),消除了 config(240)/profile(60) 的分歧根源。
- ✅ **注释矛盾已修**(本次 `2d44c07`):`GetProfile` 文档此前同段既写 "code defaults (60s)" 又写 "static 240s default",现已改为一致——240s 是唯一静态默认,env 覆盖,runner 只读 `Policy.StepTimeout`(不读 config)。

`go build` / `go vet` / `go test ./internal/agent ./internal/config` 全绿。前两项由并行的 timeout reconcile 提交完成,我只在其上补了注释一致性这一处。

Jerry-Xin (dismissed):
This PR is relevant to the agent-chat functionality and is ready to merge; no blocking correctness or security issues found.

💬 Non-blocking

- 🔵 Suggestion — Add JSON and SSE integration tests verifying that `detail` is serialized for whitelisted errors and redacted otherwise. Current tests cover only the helper function ([agent_chat_detail_test.go:12](/Users/mlamp/.openclaw/tmp/pr-review-wbxy0ixw/repo/internal/api/handler/agent_chat_detail_test.go:12)).
- 🔵 Suggestion — Guard against `time.Duration` overflow when converting extremely large environment values ([profile.go:168](/Users/mlamp/.openclaw/tmp/pr-review-wbxy0ixw/repo/internal/agent/profile.go:168)).

✅ Highlights

- The whitelist returns fixed signatures rather than raw error strings, preserving the existing anti-leakage policy ([agent_chat.go:581](/Users/mlamp/.openclaw/tmp/pr-review-wbxy0ixw/repo/internal/api/handler/agent_chat.go:581)).
- The override is applied uniformly through `GetProfile` without mutating the shared profile map ([profile.go:140](/Users/mlamp/.openclaw/tmp/pr-review-wbxy0ixw/repo/internal/agent/profile.go:140)).
- Tests cover all profiles, invalid configuration, wrapped context errors, and sensitive-error redaction.
- `internal/agent` tests passed. Handler tests could not link locally because the external `tokenizers` library was unavailable, not because of a source failure.

yujiawei (approved):
# Code Review — PR #168 (octo-smart-summary)

Reviewed at head SHA `95e7c10c` against merge-base `658ef95` (which equals current `main` HEAD — the branch is up to date, no sibling-PR conflict risk). Build, `go vet ./internal/...`, and the full `internal/agent` + `internal/api/handler` + `internal/config` suites pass locally (incl. `-race`, CGO=1). Security-sensitive PR (error-detail exposure) — reviewed the redaction path with extra care.

**Verdict: APPROVED.** Both fixes are correct and tightly scoped. The prior blocking concern (an advertised 240s default that did not actually take effect) is resolved, and the dead config field that earlier rounds flagged has now been removed. Only non-blocking P2 items remain.

---

## Part 1 — Spec compliance

- **Missing:** none — both stated fixes (c1 env-configurable step timeout, c2 whitelisted 500 detail) are present.
- **Extra:** none — the previously-added-but-unread `Config.AgentStepTimeout` field has been removed in this head (`internal/config/config.go`), so there is no orphaned config surface anymore. `grep -rn AgentStepTimeout` across the repo now returns zero hits.
- **Deviation:** none — the code no longer advertises a default it does not deliver. All three profiles (`chat`, `summary`, `summary_refine`) statically declare `StepTimeout: 240 * time.Second`, so an unset `AGENT_STEP_TIMEOUT` yields a real 240s default. `CONFIGURATION.md` documents the env with default `240` and the 300s request-deadline ceiling.

**Spec: ✅**

## Part 2 — Code quality

**Quality: Approved.** No P0/P1.

### c1 — env-configurable StepTimeout — correct, invariant preserved

`internal/agent/profile.go:147-148` applies `agentStepTimeoutOverride()` over the static 240s at `GetProfile` time. `GetProfile` returns a value copy of `Profile`, so the `p.Policy.StepTimeout = override` assignment mutates the copy, not the shared `profiles` map — no cross-request state pollution. Parse guards are correct: empty / non-integer / `<= 0` all return `0` → no override → keeps the static default (unit-tested across all three profiles for unset / `120` / `0` / `-1` / `abc` / `60s` / `1.5`). The 240s default stays below the 300s outer `ChatStream` backstop (`agent_chat.go:411`), and `runTools` still runs on the outer request ctx (the #158/#161 planning-bound-not-tool-bound invariant is intact).

### c2 — whitelisted `detail` on 500s (the security-sensitive part) — correct default-deny

`safeErrorDetail` (`agent_chat.go:581`) is whitelist-only and returns a **hardcoded canonical constant** for each matched case — it never returns `err.Error()` or any substring of it. Everything unmatched collapses to `"internal error"`. I confirmed there is no raw-message return path in the function. Redaction is unit-tested against the exact adversarial shapes: an internal LLM gateway URL, an internal IP (`dial tcp 10.0.0.42:8080`), and a gorm/DB fragment all → `"internal error"`.

This head also upgrades the two context cases from substring matching to `errors.Is(err, context.DeadlineExceeded)` / `errors.Is(err, context.Canceled)`. I verified this is a strict improvement, not a regression: the timeout error chain preserves the sentinel end-to-end — `Client.doOnce` wraps `http.Do`'s `*url.Error` via `fmt.Errorf("http do: %w", …)` (`llm.go:115`), and the retry loop returns `ctx.Err()` directly (`llm.go:83`); `runner.go:100` returns the error verbatim. So a real production timeout (`context deadline exceeded while dialing <ip>`) matches the sentinel and is reported as the clean `"context deadline exceeded"` — the embedded IP can no longer ride along even in principle, whereas the old substring form could have matched a leaky wrapper. The remaining three cases (`max steps exceeded`, empty-response guard, `unknown agent profile`) are plain `errors.New` sentinels in the runner/profile code with no operator data, so substring matching there is fine.

Coverage of error sinks: all four `c.JSON` 500 sites (`Chat`, `History`) and three SSE 500 sites (`ChatStream`) route through `safeErrorDetail`; the two 40100 missing-auth sites correctly stay detail-less (no `err` in scope, via the 2-arg `writeSSEErrorViaSink` wrapper). `apiResponse.Detail` is `json:"detail,omitempty"`, so success responses and old clients are unaffected.

### Non-blocking (P2)

- **Silent int64 overflow on absurd `AGENT_STEP_TIMEOUT`** — `profile.go` `time.Duration(secs) * time.Second` wraps for values above ~9.2e9 seconds. Practically unreachable; an explicit upper bound would fail loud rather than silently.
- **Effective ceiling is the 300s outer backstop** — any `AGENT_STEP_TIMEOUT > 300` is silently capped by the `ChatStream` outer ctx. Worth a one-line note in `CONFIGURATION.md`.
- **Non-stream `Chat()` has no outer ctx backstop** — `agent_chat.go:203` uses `c.Request.Context()` directly (unlike `ChatStream`'s 300s `WithTimeout`), and `publicSrv` (`cmd/summary-api/main.go:121`) sets no `WriteTimeout`. Pre-existing, but raising `StepTimeout` to 240s widens the window a single non-stream request can hold a connection bounded only by client patience. Consider a symmetric outer deadline on the non-stream path.

## Test assessment

Both new test files are strong and would fail without the fixes: `TestSafeErrorDetail` covers whitelist pass-through, wrapped-sentinel matching (`fmt.Errorf("...: %w", context.DeadlineExceeded)`), and the three leaky-error → `"internal error"` security cases; `TestGetProfile_StepTimeoutOverride` now covers all three profiles (the earlier `chat`-profile gap is closed) for unset-default / override / zero / invalid.

## Coverage / blind spots

Executed: build, vet, `internal/agent` + `internal/api/handler` + `internal/config` (incl. `-race`), full read of `safeErrorDetail` (literal-only returns confirmed), the timeout override path, and the LLM-client error-wrapping chain that the `errors.Is` migration depends on. Not executed here: live end-to-end refine reproducer (the PR author reports before/after at 83s→~100s); `check-sprint` is a governance gate, not a code check.

## For a human to double-check (security-sensitive)

- The c2 whitelist is safe for the five current patterns; any future addition MUST remain a fixed canonical string (never pass `err.Error()` through) to preserve the redaction guarantee — the doc comment already says "grow deliberately."
- Default step timeout is now 240s even without env config. Confirm this sits within any upstream LB/proxy idle timeouts in the deploy environment (it is below the 300s `ChatStream` backstop).

Jerry-Xin (approved):
This PR is relevant to the agent-chat functionality and is ready to merge; no blocking correctness or security issues found.

💬 Non-blocking

- 🔵 Suggestion — Add JSON and SSE integration tests verifying that `detail` is serialized for whitelisted errors and redacted otherwise. Current tests cover only the helper function (`internal/api/handler/agent_chat_detail_test.go:12`).
- 🔵 Suggestion — Guard against `time.Duration` overflow when converting extremely large environment values (`internal/agent/profile.go:168`). Note: effective stepCtx is already bounded by the 300s ChatStream request deadline, so this is defense-in-depth, not a live-path risk.

✅ Highlights

- The whitelist returns fixed signatures rather than raw error strings, preserving the existing anti-leakage policy (`internal/api/handler/agent_chat.go:581`).
- The override is applied uniformly through `GetProfile` without mutating the shared profile map (`internal/agent/profile.go:140`).
- Tests cover all profiles, invalid configuration, wrapped context errors, and sensitive-error redaction.
- Prior blocker RESOLVED: the dead `Config.AgentStepTimeout` field was removed and all three profiles now default to 240s statically, so the documented `AGENT_STEP_TIMEOUT` default (240) matches the effective runtime default — config↔runtime now agree. The `chat` profile was added to the timeout test and the stale `60s` doc comment was corrected.

yujiawei (approved):
# Code Review — PR #168 (octo-smart-summary)

Re-review at head SHA `2d44c079` (a force-push landed mid-review over the previously-assigned `95e7c10c`). Confirmed the only delta between the two is a **documentation-comment fix** in `GetProfile` — it corrects a genuinely contradictory doc block that simultaneously said "static default (60s)" and "static 240s default", now stating the single 240s default plainly. **No functional code changed** (`git diff 95e7c10c..2d44c079` excluding comments/markdown is empty). My prior analysis carries over unchanged.

Reviewed against merge-base `658ef95` (== current `main` HEAD — branch is up to date, no sibling-PR conflict). Build, `go vet ./internal/...`, and the full `internal/agent` + `internal/api/handler` + `internal/config` suites pass locally (incl. `-race`, CGO=1). Security-sensitive PR (error-detail exposure) — reviewed the redaction path with extra care.

**Verdict: APPROVED.** Both fixes are correct and tightly scoped. The earlier blocking concern (an advertised 240s default that did not actually take effect) is resolved, and the dead `Config.AgentStepTimeout` field that earlier rounds flagged has been removed. Only non-blocking P2 items remain.

---

## Part 1 — Spec compliance

- **Missing:** none — both stated fixes (c1 env-configurable step timeout, c2 whitelisted 500 detail) are present.
- **Extra:** none — the previously-added-but-unread `Config.AgentStepTimeout` field has been removed. `grep -rn AgentStepTimeout` across the repo returns zero hits.
- **Deviation:** none — the code no longer advertises a default it does not deliver. All three profiles (`chat`, `summary`, `summary_refine`) statically declare `StepTimeout: 240 * time.Second`, so an unset `AGENT_STEP_TIMEOUT` yields a real 240s default; `CONFIGURATION.md` documents the env (default `240`, capped by the 300s request deadline). The `GetProfile` doc comment is now internally consistent after this head's fix.

**Spec: ✅**

## Part 2 — Code quality

**Quality: Approved.** No P0/P1.

### c1 — env-configurable StepTimeout — correct, invariant preserved

`internal/agent/profile.go` applies `agentStepTimeoutOverride()` over the static 240s at `GetProfile` time. `GetProfile` returns a value copy of `Profile`, so `p.Policy.StepTimeout = override` mutates the copy, not the shared `profiles` map — no cross-request state pollution. Parse guards are correct: empty / non-integer / `<= 0` return `0` → no override → keeps the static default (unit-tested across all three profiles for unset / `120` / `0` / `-1` / `abc` / `60s` / `1.5`). The 240s default stays below the 300s outer `ChatStream` backstop, and `runTools` still runs on the outer request ctx (the #158/#161 planning-bound-not-tool-bound invariant is intact).

### c2 — whitelisted `detail` on 500s (security-sensitive) — correct default-deny

`safeErrorDetail` is whitelist-only and returns a **hardcoded canonical constant** for each matched case — it never returns `err.Error()` or a substring of it; everything unmatched collapses to `"internal error"`. No raw-message return path exists. Redaction is unit-tested against the exact adversarial shapes: internal LLM gateway URL, internal IP (`dial tcp 10.0.0.42:8080`), and a gorm/DB fragment all → `"internal error"`.

The two context cases use `errors.Is(err, context.DeadlineExceeded)` / `errors.Is(err, context.Canceled)`. I verified this is a strict improvement, not a regression: the timeout error chain preserves the sentinel end-to-end — `Client.doOnce` wraps `http.Do`'s `*url.Error` via `fmt.Errorf("http do: %w", …)` and the retry loop returns `ctx.Err()` directly; `runner.go` returns the error verbatim. A real production timeout (`context deadline exceeded while dialing <ip>`) therefore matches the sentinel and is reported as the clean `"context deadline exceeded"` — the embedded IP cannot ride along even in principle. The three remaining cases (`max steps exceeded`, empty-response guard, `unknown agent profile`) are plain `errors.New` sentinels with no operator data, so substring matching is fine there.

Coverage: all four `c.JSON` 500 sites (`Chat`, `History`) and three SSE 500 sites (`ChatStream`) route through `safeErrorDetail`; the two 40100 missing-auth sites correctly stay detail-less (via the 2-arg `writeSSEErrorViaSink` wrapper). `apiResponse.Detail` is `json:"detail,omitempty"` — success responses and old clients unaffected.

### Non-blocking (P2)

- **Silent int64 overflow on absurd `AGENT_STEP_TIMEOUT`** — `time.Duration(secs) * time.Second` wraps for values above ~9.2e9 seconds. Practically unreachable; an explicit upper bound would fail loud.
- **Effective ceiling is the 300s outer backstop** — any `AGENT_STEP_TIMEOUT > 300` is silently capped by the `ChatStream` outer ctx. Worth a one-line note in `CONFIGURATION.md`.
- **Non-stream `Chat()` has no outer ctx backstop** — it uses `c.Request.Context()` directly (unlike `ChatStream`'s 300s `WithTimeout`), and `publicSrv` sets no `WriteTimeout`. Pre-existing, but raising `StepTimeout` to 240s widens the window a single non-stream request can hold a connection bounded only by client patience. Consider a symmetric outer deadline.

## Test assessment

Both new test files are strong and would fail without the fixes: `TestSafeErrorDetail` covers whitelist pass-through, wrapped-sentinel matching, and the three leaky-error → `"internal error"` security cases; `TestGetProfile_StepTimeoutOverride` covers all three profiles (the earlier `chat`-profile gap is closed) for unset / override / zero / invalid.

## For a human to double-check (security-sensitive)

- The c2 whitelist is safe for the five current patterns; any future addition MUST remain a fixed canonical string (never pass `err.Error()` through) to preserve the redaction guarantee.
- Default step timeout is now 240s even without env config. Confirm this sits within any upstream LB/proxy idle timeouts in the deploy environment (it is below the 300s `ChatStream` backstop).

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #168 (octo-smart-summary)

Reviewed at head `95e7c10c`. Scope: env-configurable `AGENT_STEP_TIMEOUT`
(raising the per-step planning ceiling from 60s to 240s, with an env override)
plus a whitelist-gated `Detail` field on the five 500 sites in
`agent_chat.go` / `agent_chat_detail_test.go` / `task.go`. PR is routed
`complexity=security_sensitive`, so the whitelist got extra attention.

## Summary

Two tightly-scoped fixes addressing a real production 500 (`[agent] chat
runner error: context deadline exceeded` at ~83s on
`claude-sonnet-4-6` refine load). The env override lands via
`agentStepTimeoutOverride()` at `internal/agent/profile.go:158`, reading
`os.Getenv("AGENT_STEP_TIMEOUT")` directly (bypasses `config.Config` so
unit tests don't need the deps container) — `>0` overrides, `0` /
invalid / unset keeps the code default. The detail field is added as
`Detail string \`json:"detail,omitempty"\`` on `apiResponse`
(`internal/api/handler/task.go:219`) and populated on the five 500
sites in `agent_chat.go` via `safeErrorDetail(err)`
(`internal/api/handler/agent_chat.go:581`), which gates on five
hardcoded patterns and collapses everything else to `"internal error"`.

Both fixes are conservative and well-tested. Three P2 cleanup items, the
loudest of which is a doc contradiction: the PR description and the new
godoc above `GetProfile` claim the code default stays at 60s, but the
diff actually bumps it to 240s across all three profiles — the
`CONFIGURATION.md` change gets this right, the prose does not.

## Verification

Local checks at head `95e7c10`:
- `go test ./internal/agent/...` — pass (0.290s), including new
  `TestGetProfile_StepTimeoutOverride` (4 subcases: unset / 120 / 0 /
  invalid).
- `go test ./internal/api/handler/...` — pass (11.002s), including new
  `TestSafeErrorDetail` (11 subcases covering nil, five whitelisted
  patterns, three redaction shapes, generic fallback).
- `go test ./internal/config/...` — pass (0.003s).
- `go vet ./...` — not executed; rely on test-suite pass above.

## Findings

No P0/P1 issues; three P2 items below.

### P2 — Godoc contradicts the new 240s code default (`internal/agent/profile.go:140`)

The new godoc above `GetProfile` reads:

> the static profile values above are code defaults (60s) …
> AGENT_STEP_TIMEOUT env overrides the static 240s default. Setting
> to 0 disables the override and keeps the code default.

But this PR also changed the static values at `internal/agent/profile.go:57`,
`:62`, and `:79` from `60e9` to `240 * time.Second`. So the *current*
code default is 240s, not 60s — the two sentences contradict each
other, and the same contradiction appears in the PR description
("Setting to 0 or leaving unset keeps the 60s code default"). The
`CONFIGURATION.md` change in this PR documents `240` correctly; the
prose around it does not. Suggest rewording the godoc to:

> The static profile values above are code defaults (240s; was 60s
> pre-PR#168). `AGENT_STEP_TIMEOUT` env overrides them at `GetProfile`
> time; setting to 0 disables the override and keeps the 240s code
> default.

### P2 — CONFIGURATION.md claim about the 300-second request deadline is scoped to ChatStream only (`CONFIGURATION.md:20`)

The new row reads:

> `AGENT_STEP_TIMEOUT` | Maximum duration in seconds for one agent
> planning step; streaming requests remain capped by the 300-second
> request deadline | No | `240`

The "300-second request deadline" only applies to `ChatStream` —
`internal/api/handler/agent_chat.go:411` sets
`context.WithTimeout(baseCtx, 300*time.Second)` on that path alone.
The non-streaming `Chat` handler (`POST /api/v1/agent/chat`,
`internal/api/handler/agent_chat.go:173`) uses
`c.Request.Context()` directly with no `WithTimeout`, and
`cmd/summary-api/main.go:121` constructs `http.Server` without
`ReadTimeout` / `WriteTimeout`. So non-streaming `Chat` has **no
outer request deadline at the Go server level** — a single step can
hold the connection for up to `AGENT_STEP_TIMEOUT` seconds, and
multi-step flows can hold it for `MaxSteps * AGENT_STEP_TIMEOUT` in
the worst case. Suggest:

> … Streaming requests (`/chat/stream`) have an additional 300-second
> outer request deadline; non-streaming `/chat` has no outer deadline
> at the Go server level (set one at the reverse proxy if desired).

### P2 — Whitelist couples to a literal runner.go substring (`internal/api/handler/agent_chat.go:598`)

`strings.Contains(err.Error(), "LLM returned empty response with no
tool_calls")` is tied to the exact prose at `internal/agent/runner.go:118`.
If anyone rephrases that runner error (adds a model name, trims
wording, i18n's it), the whitelist silently collapses to
`"internal error"` — fail-closed and safe, but the client-side triage
signal this PR is adding disappears, and no test would catch the
regression. Small refactor worth considering: export a sentinel
(`agent.ErrLLMEmptyFinalStep`) from `runner.go`, wrap with `%w`, and
match here via `errors.Is`. Low priority — safe as-is — but the
coupling is the kind of thing that bites on the next runner edit.

- **Whitelist conservative**: five hardcoded patterns, all derived
  from error sites whose format strings contain no operator-supplied
  data (`context deadline exceeded` / `context canceled` via
  `errors.Is`, `max steps exceeded` at `runner.go:191`, the empty
  response guard at `runner.go:118`, `unknown agent profile` at
  `profile.go:145`). No raw `err.Error()` passthrough.
- **Redaction tests cover the failure modes**: `TestSafeErrorDetail`
  asserts `"internal error"` for an internal URL, an internal IP,
  and a DB/gorm error shape.
- **`omitempty` preserves success-response wire format**: `Detail`
  drops out of the JSON on 200 responses; old clients see no change.
- **Five 500 sites consistent**: all five `c.JSON(500, …)` and
  `writeSSEErrorViaSink(50000, …)` call sites now route through
  `safeErrorDetail`. The two 40100 sites (`agent_chat.go:210`,
  `:311`, `:423`) correctly stay detail-free — no `err` is in scope.
- **`writeSSEErrorViaSink` wrapper preserved** as a two-arg shim
  over the new detail-aware helper (`agent_chat.go:545`), so the
  40100 SSE path keeps its old signature.
- **Env override logic correct**: `strconv.Atoi` + `secs <= 0` short
  circuits to 0 (no override); invalid strings (`"abc"`, `"60s"`,
  `"1.5"`, `"-1"`) all fall through; test at
  `profile_step_timeout_test.go:56` covers them explicitly.
- **No cross-request state pollution**: `GetProfile` operates on a
  value copy of `Profile` (not a pointer); the
  `p.Policy.StepTimeout = override` assignment does not mutate the
  global `profiles` map.
- **C1 double-path parity**: all three handler entry points
  (`Chat` / `ChatStream` / `History`) consistently apply
  `Detail: safeErrorDetail(err)` on every 500 site; the 401 sites
  correctly omit `Detail`.
- **C3 authorization boundary**: no change to auth, visibility,
  tenant scope, or persisted-history ownership contracts.

## Data-flow trace

- `AGENT_STEP_TIMEOUT` env → `agentStepTimeoutOverride()` (profile.go:158) →
  `p.Policy.StepTimeout` value copy returned by `GetProfile` → runner's
  `stepCtx, cancel := context.WithTimeout(ctx, r.policy.StepTimeout)` at
  `runner.go:96` → LLM client call. When env is unset/invalid/0, the
  copy keeps the 240s static value from `profile.go:57/62/79`. Verified
  end-to-end by `TestGetProfile_StepTimeoutOverride`.
- Runner error → `runner.RunWithHistory` return → `Chat`/`ChatStream`/
  `History` handler → `safeErrorDetail(err)` → `apiResponse.Detail`
  (`json:"detail,omitempty"`) on `c.JSON(500, …)` and
  `writeSSEErrorViaSinkWithDetail(sink, 50000, ..., detail)`. Each of
  the five whitelisted strings is produced from a hardcoded literal in
  `safeErrorDetail`, never from `err.Error()` passthrough; everything
  else collapses to `"internal error"`.

## Blind-spot checklist (security_sensitive)

- **C1 — double-path parity**: clear. All three handlers (Chat,
  ChatStream, History) apply `Detail: safeErrorDetail(err)` uniformly
  on every 500 site; 401 sites correctly omit Detail.
- **C2 — control-flow ordering**: clear. `safeErrorDetail` is pure;
  no nesting/reuse hazard; `GetProfile` value-copy prevents cross-
  request pollution.
- **C3 — authorization boundary**: N/A. No auth/visibility/tenant
  change.
- **C4 — authorization lifecycle**: N/A.
- **C5 — build/note ≠ runtime**: clear. Local `go test` on the three
  touched packages passes; `http.Server` ReadTimeout/WriteTimeout
  absence was verified directly at `cmd/summary-api/main.go:121`.
- **C6 — governance/doc self-consistency**: HIT — see P2 #2 above
  (CONFIGURATION.md claim about a 300s outer deadline is scoped to
  `ChatStream`, not the non-streaming `Chat`).

## Cross-round blocker re-check (R6)

N/A — first review pass on this PR.

## Verdict: APPROVED

Solid, well-tested fix for a real production failure mode. The
whitelist design is exactly the right call on a security-sensitive
surface, and the env override gives ops a real knob without
requiring a redeploy. Cleaning up the contradictory godoc (P2 #1)
would be nice before merge so operators don't ship under the wrong
assumption, but nothing in this diff blocks landing.

[Octo-Q] verdict: APPROVE — no P0/P1 findings; three P2 items (misleading godoc + PR-body default claim, CONFIGURATION.md scope wording, substring-coupled whitelist) are non-blocking.

Issue #144, opened 2026-07-08
智能总结：支持通过用户自己的 agent（龙虾）发起总结


Ranwanglc:
## 需求

让用户自己的 agent（龙虾 / OpenClaw agent）能代替用户发起并执行智能总结任务，实现 agent 驱动的自动化总结，而非仅限用户在前端手动操作。

## 场景

用户在与自己的 agent 对话中说"帮我总结某群最近的消息"，agent 调用 summary 能力完成总结，并把结果返回给用户。

## 待确认 / 待设计

1. **调用方式**：agent 侧以何种方式调用 summary（OpenClaw skill / MCP / REST API + 用户凭据）。
2. **身份与权限**：agent 代发时，用户的鉴权与群访问权限如何传递与校验（避免越权访问用户无权的群）。
3. **结果回传形态**：直接返回文本给用户，还是生成 task 供前端查看，或两者兼有。

## 验收标准

用户通过自己的 agent 发起一次智能总结 → agent 以用户身份鉴权 → 完成总结 → 结果正确回传给用户。

wanyiwang06:
## Review: MVP read-only Bot access plan

Reviewed the 544-line implementation plan (`octo-summary-agent-readonly-implementation-plan.md`, local draft) against this issue's three open design questions. Overall the plan's security model is sound (owner ACL inheritance + server-side space resolution via `/v1/auth/verify-bot`), but the scope has room to be tighter for a real MVP.

Not a personal opinion on the whole spec — just calling out what's mandatory vs deferrable so the first cut lands in weeks, not months.

### Mandatory (P0)

Cannot ship a working MVP without these:

- §4 / §5.1 / §5.2 — Bot Identity Resolver + StrictBotAuthMiddleware. Foundation of the whole scheme; without it the ACL story is empty.
- §5.3 — read-only route `/api/v1/bot/summaries*` with list / get / result endpoints.
- §5.6 — reference output that by default returns citation metadata only, **not** `context_before` / `context_after`. **Non-negotiable safety line** — otherwise the Bot leaks surrounding chat content to the Agent without any user consent.
- §6.1 / §6.3 — octo-cli OpenAPI spec + `octo-summary` Skill. Without these the Agent side has nothing to consume.
- §7 — octo-web nginx `/v1/bot/summaries` route.
- §8 Docker Compose bits — `nginx.conf.template` + `SUMMARY_API_URL` wiring.
[...]

PR #172, opened 2026-07-27
feat: expose owner-scoped summaries to personal agents

wanyiwang06:
## Summary
- add a read-only bot mount for listing and reading summaries
- resolve trusted bot, owner, and space identity through verify-bot
- preserve existing owner ACLs and strip surrounding citation context
- fall back to the owner's completed personal result when no team result exists

## API
- GET /api/v1/bot/summaries
- GET /api/v1/bot/summaries/:id
- GET /api/v1/bot/summaries/:id/result

## Safety
- accept only bf_ bearer tokens
- ignore client-supplied space headers
- expose no bot write routes
- keep the existing human result behavior unchanged

## Validation
- go test ./...
- CGO_ENABLED=0 go vet ./...
- CGO SQLite integration tests for owner fallback, non-personal rejection, and title/topic filtering
- local octo-alex CLI and chat-agent verification

Related to #144. This PR provides the read-only foundation; bot-triggered summary creation remains follow-up work.


lml2468 (dismissed):
# PR Review — PR #172 (Mininglamp-OSS/octo-smart-summary)

Reviewed at head SHA `6f5be7d334fb` against merge-base (base `main`). **Security-critical**: introduces a new bot-authenticated read surface exposing summaries to personal agents.

Intent: **X** = add a read-only `/api/v1/bot/summaries*` mount where a `bf_` bot token resolves (via verify-bot) to a trusted `(bot_uid, owner_uid, space_id)`, and the bot reads summaries **as its human owner** so that **Y** = personal agents can list/read the owner's summaries without any new ACL path, affecting **Z** = new bot middleware/resolver + reuse of the existing task handlers.

- `Build`, `Test`, `Test (race, cgo)` (1m22s), `Vet` → **pass**; `check-sprint` (governance) → fail.
- Local: build/vet clean; `go test ./internal/middleware ./internal/auth ./internal/api/router ./internal/api/handler` → pass.

## 1. Spec compliance — ✅
All four safety claims implemented and test-locked: `bf_`-only bearer, server-resolved space (client header ignored), no write routes, human path unchanged. Read-only mount (3 GETs). Nothing over-built.

## 2. Code quality — Approved

**The load-bearing property — bot inherits the owner's exact ACL, no new access path:**
- `StrictBotAuthMiddleware` requires `Bearer bf_…`, resolves identity via `verify-bot`, and sets `user_id = OwnerUID` + `space_id = <server-resolved>` in the gin context (`space.go:23-52`). All three identity fields must be non-empty or 401.
- The bot mount reuses the **same handlers** (`ListSummaries`/`GetSummary`/`GetResult`) and deliberately does **not** use `StrictSpaceMiddleware` — so there is no client-header space path. The handlers read `middleware.GetUserID/GetSpaceID`, which read the exact `user_id`/`space_id` keys the bot middleware set → the vetted `authorizeTaskAccess` (space-scoped load + owner ACL) runs unchanged with the owner identity.
- `authorizeTaskAccess` is fail-closed: empty userID → 401, **empty spaceID → 404** (never queries `space_id=''`), space-scoped task load. Since the bot middleware guarantees a non-empty server-resolved space, cross-space reads are structurally impossible.
- **GetResult bot fallback is inside the authorized block** — `authorizeTaskAccess` runs first (`task.go:1105`); only then, for a ByPerson task with no team result, does it return the **owner's** completed PersonalResult (`user_id = OwnerUID`). Can't reach a task the owner can't access.

**Data-minimization for bots:**
- `citationsForRequest` strips `ContextBefore`/`ContextAfter` (surrounding messages) from citations on bot requests — reduces context leakage to the agent while keeping the cited message. Applied on all bot result paths.

**Resolver hygiene:** `ResolveBot` treats 401/403 as empty identity (→ 401 at middleware), non-2xx as error (→ 500), and requires all three fields — no partial-identity acceptance.

- **Client space spoofing:** `TestStrictBotAuthUsesVerifiedOwnerAndSpace` sends `X-Space-Id: spoofed-space` and asserts the response uses the verify-bot's `trusted-space` — header ignored. ✅
- **Token whitelist / fail-closed:** `app_`-prefixed token → 401; missing owner/space → 401; resolver error → 500 (`TestStrictBotAuthRejectsMissingOrIncompleteIdentity`). ✅
- **Human token on bot mount / cross-space:** rejected (`TestBotSummaryRoutesRejectHumanTokenAndClientSpace`). ✅
- **Owner personal fallback + context strip:** `TestBotGetResultFallsBackToOwnersCompletedPersonalResult` asserts `ContextBefore` len 0 for the bot. ✅
- **No bleak into human path:** `TestHumanGetResultDoesNotUseBotPersonalFallback`. ✅
- **Read-only:** bot mount has only GETs; no write route exists.

### Non-blocking notes
- 🟡 The `keyword` filter change (`title LIKE ? OR topic LIKE ?`) is a reasonable improvement carried in this PR but is orthogonal to the bot feature — minor scope bundling; the added `topic` arg is placeholder-aligned (both count + page paths receive the extra arg via `args`, verified consistent with the #167 pattern). No misalignment.
- 🟡 `writeBotPersonalResult` duplicates a chunk of the human result-envelope shape; a shared builder would reduce drift risk over time. Cosmetic.

## 3. Test assessment
- Comprehensive across the new surface: middleware (space-spoof, token-whitelist, incomplete-identity), router (mount-behind-auth, human-token/client-space rejection), handler (owner personal fallback, context-strip, human-path non-regression), resolver (verify-bot decode). Each safety claim has a corresponding failing-without-the-fix test. Strong.

## 4. Coverage / blind spots
- Executed: build, vet, all four affected packages' tests, **live CI** (race+cgo green), full trace of the bot identity → context → authorizeTaskAccess → owner-scoped query path.
- Not verified: the real `verify-bot` upstream contract (mocked in tests) — recommend confirming the upstream returns exactly `{bot_uid, owner_uid, space_id}` and treats bad tokens as 401/403.

- Confirm the production `verify-bot` endpoint's response shape + that it never returns a space the owner isn't actually in (the whole model trusts verify-bot's space).

## Verdict
**APPROVE** — both gates pass. The bot read surface is correctly built as a thin identity shim over the existing, vetted owner ACL: `bf_`-only tokens, fully server-resolved space (client header provably ignored, test-locked), owner as effective reader through the same `authorizeTaskAccess` fail-closed gate, read-only routes, and citation-context stripped for bots. No new access path, no cross-space or cross-owner exposure, human path unchanged. Functional CI (incl. race+cgo) is green; test coverage locks every safety claim.
Suggested decision: `Safe to merge after non-blocking follow-up` — confirm the upstream verify-bot contract; merge once check-sprint clears.

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# Code Review — PR #172 (octo-smart-summary)

## Summary

This PR introduces a read-only bot-facing API surface (`/api/v1/bot/summaries*`) that exposes owner-scoped summaries to personal agents. The design is well-layered: a new `StrictBotAuthMiddleware` authenticates bot tokens (`bf_`-prefixed `Authorization: Bearer`) via a server-side `/v1/auth/verify-bot` endpoint, resolves the bot's human owner and space from the auth service (never from request headers), then delegates to the existing `ListSummaries`/`GetSummary`/`GetResult` handlers. Bot-specific behaviors — citation context stripping and a personal-result fallback for `ModeByPerson` tasks — are cleanly gated on the `bot_request` context flag.

The security architecture is sound. Identity is server-resolved, space isolation is enforced at both the middleware and query level, and the bot endpoints are strictly read-only. The PR ships with thorough test coverage across middleware, resolver, handler, and router levels.

## Verification

- ✅ **Auth boundary isolation** — Bot routes use `Authorization: Bearer bf_*` header; human routes use `Token` header. No cross-contamination possible. Verified at `internal/middleware/space.go:30-33` (bot middleware checks `strings.HasPrefix(parts[1], "bf_")`) and `:122` (human middleware reads `c.GetHeader("Token")`).
- ✅ **Server-resolved identity** — `StrictBotAuthMiddleware` sets `user_id`, `space_id`, `bot_id` exclusively from the `ResolveBot` response. The test `TestStrictBotAuthUsesVerifiedOwnerAndSpace` explicitly sends a spoofed `X-Space-Id` header and verifies the server-resolved space wins.
- ✅ **Space isolation at query level** — `ListSummaries` gates on `spaceID == ""` at `task.go:488`; `authorizeTaskAccess` gates on `spaceID == ""` at `:129`. Bot middleware rejects empty `SpaceID` at `space.go:43`. Triple-layered fail-closed.
- ✅ **Access scoping** — `canAccessTask` (creator or participant check) is applied identically for bot and human paths. Bot cannot escalate beyond the owner's permissions.
- ✅ **Citation privacy** — `citationsForRequest` strips `ContextBefore`/`ContextAfter` (raw chat messages) for all bot-reachable citation return paths: `GetSummary` (`:935`), `GetResult` (`:1141`), `writeBotPersonalResult` (`:1175`). Test `TestCitationsForBotDropsSurroundingContextWithoutMutatingSource` verifies non-mutation of source.
- ✅ **Read-only mount** — Bot group at `router.go:59-64` registers only GET endpoints. No POST/PUT/DELETE.
- ✅ **Personal fallback scoping** — Gated on `bot_request == true` AND `SummaryMode == ModeByPerson` AND `user_id` match AND `worker_status == completed` AND non-empty content. Test `TestBotGetResultDoesNotFallbackForNonPersonalMode` verifies the negative case.
- ✅ **Token resolver fail-closed** — Empty/missing fields in `ResolveBot` response return empty identity with nil error; middleware rejects empty fields with 401. HTTP errors return Go errors; middleware returns 500.

Static analysis only at head `6f5be7d334fb77ce5679ae488359356057c417ba`; build and tests not executed in this environment.

## Findings

### P2 — Personal-result fallback silently swallows database errors (`internal/api/handler/task.go:1117`)

`h.db.Where(...).First(&personal).Error` is assigned to `err` but only the success path (`err == nil`) is handled. A transient DB error (connection timeout, deadlock) is treated identically to "no personal result found," falling through to the 404 response. This is safe from a correctness standpoint (fail-closed), but it masks observability: an operator debugging why bots suddenly cannot see personal results would have no log signal.

Consider at minimum logging non-`gorm.ErrRecordNotFound` errors:

```go
if err != nil && !errors.Is(err, gorm.ErrRecordNotFound) {
    log.Printf("bot personal fallback query error: task=%d user=%s err=%v", taskID, userID, err)
}
```

### P2 — Bot token resolution has no caching layer (`cmd/summary-api/main.go:120`)

The human auth path uses `authResolver = NewCachedResolver(httpResolver, 30s, 10000)`, but `httpResolver` (the raw `HTTPTokenResolver`) is passed as the `botAuthResolver` because `CachedResolver` does not implement `BotTokenResolver`. Every bot request hits the upstream `/v1/auth/verify-bot` endpoint with no cache.

This is likely intentional — real-time revocation is more important for bot tokens than for human session tokens — but worth documenting as a deliberate design decision. Under high bot traffic, this creates sustained load on the auth service. If caching becomes desirable later, a separate `CachedBotResolver` or a `ResolveBot` method on `CachedResolver` would be needed.

## Data-Flow Trace

### Bot request lifecycle: token → identity → handler → response

1. **Token extraction**: `StrictBotAuthMiddleware` reads `Authorization` header, splits with `strings.Fields`, validates `Bearer` scheme + `bf_` prefix. Rejects malformed headers with 401.
2. **Server-side verification**: `ResolveBot` POSTs `{bot_token: "<token>"}` to `<octo-api>/v1/auth/verify-bot`. Returns `{bot_uid, owner_uid, space_id}`. 401/403 → empty identity → middleware 401. Non-200 → error → middleware 500.
3. **Context injection**: Middleware sets `actor_type=bot`, `actor_id=BotUID`, `bot_id=BotUID`, `user_id=OwnerUID`, `space_id=SpaceID`, `bot_request=true`.
4. **ListSummaries**: Reads `space_id` + `user_id` from context → WHERE `space_id=? AND (creator_id=? OR participant OR schedule_visibility)`. Keyword search now includes `topic` (additive, within same scope).
5. **GetSummary/GetResult**: `authorizeTaskAccess` → space_id gate → `canAccessTask(ownerUID, taskID, creatorID)` → creator or participant check.
6. **Personal fallback** (GetResult only): When no team result found AND `bot_request` AND `ModeByPerson` → queries `PersonalResult WHERE task_id=? AND user_id=ownerUID AND worker_status=completed`.
7. **Citation stripping**: `citationsForRequest` checks `bot_request` → copies citations → nils `ContextBefore`/`ContextAfter`. Applied in all three return paths.

### Upstream data verification

- **SpaceID**: Server-resolved from `/v1/auth/verify-bot`. Cannot be spoofed via headers. ✅
- **OwnerUID**: Server-resolved. Bot acts as the owner — access bounded by owner's permissions. ✅
- **PersonalResult.UserID**: Queried with `user_id=OwnerUID` (from middleware). Cannot access other users' personal results. ✅
- **Citation.ContextBefore/After**: Stripped before serialization for bot requests. Source slice not mutated (verified by test). ✅

## Blindspot Checklist (C1–C6)

- **C1 — Dual-path parity**: CLEAR. Bot routes reuse the same handlers as human routes. Bot-specific behavior (`bot_request` gate) is applied in exactly 2 places: personal fallback (`task.go:1114`) and citation stripping (`task.go:1192`). Both paths verified consistent across `GetSummary`, `GetResult`, and `writeBotPersonalResult`.
- **C2 — Control-flow ordering**: CLEAR. `citationsForRequest` is called after `callerPlainCitationsVisible`, so the privacy gate runs first (potentially emptying citations), then the bot strip runs on whatever remains. No ordering issue.
- **C3 — Authorization boundary**: CLEAR. Bot identity is fully server-resolved. The `bf_` prefix prevents accidental use of human tokens on bot routes. Human tokens use a different header (`Token` vs `Authorization`). Bot routes are read-only.
- **C4 — Authorization lifecycle**: N/A for this PR. Bot token revocation is handled by the upstream `/v1/auth/verify-bot` endpoint. If the owner is removed from the space, the bot's `space_id` would no longer match any tasks at query time.
- **C5 — Build/note**: N/A. No build/CI/deployment changes.
- **C6 — Governance/docs**: N/A. No governance or documentation policy changes.

## Cross-Round Blocker Review (R6)

N/A — first review round.

## Verdict: APPROVED

Well-designed security boundary for bot access. Server-resolved identity, fail-closed space isolation, read-only mount, and citation privacy stripping are all correctly implemented. The two P2 items are non-blocking observability and performance considerations.

[Octo-Q] verdict: APPROVE — No P0/P1 issues. The bot API surface correctly delegates to existing handlers with server-resolved identity and space isolation. Auth boundary between bot and human paths is clean (different headers, different middleware, different token prefixes). Citation privacy is enforced across all return paths. Test coverage is thorough across middleware, resolver, handler, and router levels.

Jerry-Xin (dismissed):
Summary: The PR is relevant to `octo-smart-summary` and the bot read-only summary access is implemented with the existing task ACLs, server-resolved owner/space identity, and no bot write routes. No blocking issues found.

💬 Non-blocking

🟡 Warning: `/api/v1/bot/summaries/:id` currently reuses the full human detail handler (`internal/api/router/router.go:62-64`), so bot callers also receive human UI-oriented fields such as `permissions`, `personal_result`, and `members` from `internal/api/handler/task.go:1032`, `internal/api/handler/task.go:1059`, and `internal/api/handler/task.go:1084`. This appears consistent with “owner-scoped” ACLs, but a dedicated bot DTO would reduce the chance of future human-only fields being exposed to agents accidentally.

🔵 Suggestion: `writeBotPersonalResult` returns `personal_version_id` but not the numeric `version` field that other result/version APIs expose (`internal/api/handler/task.go:1178-1180`). If bot clients need display/version history parity later, consider including it or documenting that the bot result fallback only exposes the current version id.

✅ Highlights

- Bot auth rejects non-`bf_` bearer tokens and ignores client-supplied space headers (`internal/middleware/space.go:28-49`).
- `verify-bot` identity resolution fails closed when required identity fields are missing (`internal/auth/http_resolver.go:78-106`).
- Existing human result behavior is preserved while bot fallback is limited to completed personal-mode results (`internal/api/handler/task.go:1113-1123`).
- Bot citation context stripping avoids mutating the source slice (`internal/api/handler/task.go:1191-1201`).

Validation: `CGO_ENABLED=0 go test ./...` passed, and `CGO_ENABLED=0 go vet ./internal/middleware ./internal/auth ./internal/api/router ./internal/api/handler` passed. A default CGO test run could not link locally because `libtokenizers` was missing.

yujiawei (dismissed):
# Code Review — PR #172 (octo-smart-summary)

Reviewed at head SHA `6f5be7d334fb77ce5679ae488359356057c417ba` against merge-base `a35cff5`.
Verified locally: `go build ./...` clean, `go vet ./...` clean, `go test ./...` and the CGO-gated SQLite suite all pass (10 new tests, all green).
Cross-repo: I read the upstream `POST /v1/auth/verify-bot` implementation in octo-server (`modules/user/api.go:4257-4327`) to check the wire contract this PR depends on.

---

## 1. Spec compliance

**Spec: ✅**

Scope is the read-only foundation described in the PR body (agent-driven *creation* is explicitly deferred). Checked against the linked issue's acceptance criteria for identity/permission passthrough.

- **Missing:** none for the stated scope. The three routes in the PR body all exist and are mounted (`internal/api/router/router.go:58-64`). Every "Safety" claim is implemented:
  - `bf_` bearer only — `internal/middleware/space.go:31`
  - client space header ignored — space comes only from `identity.SpaceID` (`internal/middleware/space.go:48`), and the negative test at `internal/middleware/bot_auth_test.go:31-40` sends `X-Space-Id: spoofed-space` and asserts the trusted value wins
  - no bot write routes — the group registers three `GET`s only
- **Extra:** one behavior change reaches beyond the bot mount — see P2-1. It is disclosed under "Validation" and covered by a test, and the bot list route is the *same* handler (so topic search cannot be enabled for agents without touching it), so I am not treating it as a scope violation — but it is a change to the existing human list API that the "Summary/API" sections do not mention.
- **Deviation:** none found. `ModeByPerson` is `2` (`internal/model/model.go:96`), so the `SummaryMode: 1` case in `TestBotGetResultDoesNotFallbackForNonPersonalMode` really is a non-personal mode — the guard is tested correctly, not tautologically.

### Authorization audit (the part that matters most here)

I traced every bot-reachable handler end-to-end for context keys the human middleware chain sets and the bot middleware does not.

- `GetSummary` / `GetResult` both gate through `authorizeTaskAccess` (`internal/api/handler/task.go:111-145`), which requires non-empty `user_id`, **rejects an empty space outright** (`:129`), scopes the load by `space_id` + `deleted_at IS NULL` (`:134`), and then requires creator-or-participant (`:139`). The bot middleware guarantees a non-empty, server-resolved space, so the fail-closed gate is never bypassed.
- `ListSummaries` has its own empty-space short-circuit (`:487`) and its visibility predicate is `creator_id = ? OR id IN (SELECT task_id FROM summary_participant WHERE user_id = ?) OR <schedule invite>` (`:510`) — owner-scoped, space-scoped.
- No context key used by these three handlers is left unset: they read only `user_id` and `space_id` (via `middleware.GetUserID` / `GetSpaceID`). `token` is set by the human chain but never read by these handlers (`grep` for `c.Get("token")` in `internal/` returns nothing).
- All three citation exit points on bot-reachable paths pass through `citationsForRequest` (`task.go:935`, `:1141`, `:1172`). `team_citations` is *not* stripped, but `model.TeamCitation` (`internal/model/model.go:246-252`) carries only `index/user_id/user_name/ids` — no message bodies — so that is correct as written.
- Upstream contract matches exactly: request field `bot_token` ✅, response fields `bot_uid` / `owner_uid` / `space_id` ✅, and `bf_` is genuinely the personal-bot prefix (`robot.bot_token`; app bots use `app_` in a separate table), so the prefix check is a correct extra guard rather than a mismatch. The upstream route also sits behind a rate limiter.

Two points raised during review that I checked and am **not** carrying as findings, with reasoning, since they look plausible on the diff alone:

1. *"`writeBotPersonalResult` skips `callerPlainCitationsVisible`, so it is fail-open."* — It is not. That helper's actual test is "does the caller own a `PersonalResult` for this task whose `citations_json` equals the displayed result's" (`task.go:196-214`). In the fallback the row served **is** the caller's own `PersonalResult` (`:1117` filters `user_id = <owner>`), so the check would pass by construction; it also takes a `*model.SummaryResult`, not a `PersonalResult`.
2. *"The bot group skipping `StrictSpaceMiddleware` drops a guardrail."* — `StrictSpaceMiddleware` (`internal/middleware/space.go:160-176`) only copies `X-Space-Id` into the context and rejects empty space on write methods. It performs no membership or revocation check. Replacing it with a server-resolved, mandatory-non-empty space is strictly stronger, and deliberately so.

---

## 2. Code quality

**Quality: Approved**

No P0/P1. The auth design is careful: server-resolved space, fail-closed on any missing identity field, read-only mount, ACL reuse instead of reimplementation, and non-mutating citation copy (`task.go:1191-1201`, asserted by `TestCitationsForBotDropsSurroundingContextWithoutMutatingSource`). The findings below are all advisory.

### P2-1 — The `keyword` widening also changes the existing human list API

`internal/api/handler/task.go:540-543`
```go
if s := c.Query("keyword"); s != "" {
    whereSQL += " AND (title LIKE ? OR topic LIKE ?)"
    args = append(args, "%"+s+"%", "%"+s+"%")
}
```
`ListSummaries` backs both `GET /api/v1/summaries` and the new bot route, so human clients' keyword searches now also match on `topic`, changing result sets, `total`, and pagination for an endpoint this PR otherwise claims not to touch. The SQL itself is fine (parameterized, `topic` exists on `summary_task`, no ambiguity inside the window subquery). Either declare this as an intentional product change in the PR description, or gate it on `bot_request` the way the citation strip already is.

### P2-2 — Bot token verification is uncached, unlike the human path

`cmd/summary-api/main.go:71-72`
```go
httpResolver := auth.NewHTTPTokenResolver(cfg.OctoAPIURL)
authResolver := auth.NewCachedResolver(httpResolver, 30*time.Second, 10000)
```
The cached resolver goes to the human chain; the **raw** resolver is passed as the bot resolver, so every bot API call makes a fresh `POST /v1/auth/verify-bot` (5s client timeout, `internal/auth/http_resolver.go:21-25`). A polling agent doing list → detail → result burns three upstream verifies per cycle. Upstream, `/v1/auth/verify` and `/v1/auth/verify-bot` share one per-IP limiter budget (tag `verify`, 1000 req/min, burst 100). Because summary-api's egress is a handful of pod IPs and human verifies are cached while bot verifies are not, agent traffic can consume the shared budget; a 429 there is a non-200, which `ResolveUID` turns into an error and `StrictAuthMiddleware` surfaces as **HTTP 500 for human users**. Wrapping the bot resolver in an equivalent short-TTL cache (or a bot-specific one) removes the amplification for a few lines.

### P2-3 — The bot data contract is half-specified; please make it explicit

The PR strips `context_before` / `context_after` for bots, but a bot caller still receives, via `GET /api/v1/bot/summaries/:id`:
- each citation's quoted message body, sender, channel id/type and sequence — `internal/model/model.go:217-228` (only the two context slices are nil'd at `task.go:1197-1200`);
- every other participant's **submitted** personal report — `task.go:1068-1083`, `member["content"] = pr.Content`;
- participant roster, source ids/names, `origin_channel_id` — `task.go:951-970`.

I do not think this is an authorization bug: the effective reader is the owner, and all of it is already visible to the owner in the product UI. There is also a coherent line behind the current strip (surrounding context is unvetted raw chat; a submitted report is text its author wrote knowing task members read it). But the boundary is undocumented, and the destination is now a third-party agent runtime rather than the owner's browser, so it deserves one explicit sentence in code or the PR description saying which of these fields are intentionally in the bot contract. Right now a future reader cannot tell the strip apart from an oversight.

### P2-4 — The personal fallback ignores `submitted_at`, so agents can surface pre-submit drafts

`internal/api/handler/task.go:1114-1121`
```go
if bot, _ := c.Get("bot_request"); bot == true && taskPtr.SummaryMode == model.ModeByPerson {
    err := h.db.Where("task_id = ? AND user_id = ? AND worker_status = ?", taskID, userID, model.PersonalStatusCompleted).First(&personal).Error
```
`worker_status == Completed` with `submitted_at IS NULL` is exactly the draft state that `PUT /summaries/:id/personal-draft` is designed around (draft edits deliberately do not trigger recompute or set `edited_at`). The content belongs to the owner, so this is not a leak — but an agent asked "what's my summary" can now read back an unsubmitted draft and forward it into a chat as if it were final. If that is intended, say so; otherwise add `AND submitted_at IS NOT NULL`.

### P2-5 — Two negative tests can pass without their fixtures

`internal/api/handler/bot_citation_test.go:71-72` and `:86-87`
```go
_ = db.Create(&task).Error
_ = db.Create(&model.PersonalResult{...}).Error
...
if w.Code != http.StatusNotFound {
```
Both "must not fall back" tests swallow the insert error and then assert 404. If a schema change made the insert fail, the task would not exist and the test would still pass — proving nothing. Check the `Create` errors like the first test does (`:21-23`).

Also worth adding, since this is the security-relevant behavior with no coverage today: an end-to-end case that drives the **real** `StrictBotAuthMiddleware` with a valid `bf_` token whose resolved space differs from the task's space, asserting 404. The current handler tests set `bot_request` / `space_id` by hand (`bot_citation_test.go:36-40`), so they exercise the handler but cannot prove the verified identity actually reaches the ACL. `bot_routes_test.go` covers only the 401 paths.

### P2-6 — `result_source` only appears on the fallback response

`internal/api/handler/task.go:1178` sets `"result_source": "personal"`, but the normal `GetResult` response (`:1143-1162`) has no `result_source` key at all, and the fallback omits `version`, `operation_type`, `operation_note`, `parent_result_id`. A client cannot branch on `result_source == "team"`; it has to treat "absent" as team. Emitting `"result_source": "team"` on the normal path makes the discriminator usable.

Relatedly, `GET /api/v1/bot/summaries/:id` has **no** personal fallback, so for the same task an agent sees `result: null` on the detail route while `/result` returns content. Intentional or not, it is a surprising asymmetry for an agent that reads detail first.

### Nits

- `internal/api/handler/task.go:1114` / `:1192` — `bot, _ := c.Get("bot_request"); bot == true` depends on the exact dynamic type. `c.GetBool("bot_request")` is idiomatic and type-safe. (Only this middleware sets the key today, so behavior is correct as written.)
- `internal/auth/http_resolver.go:11` — `internal/auth` importing `internal/middleware` for the `BotIdentity` DTO inverts the usual layering. It follows the existing `CachedResolver` precedent, so this is consistency, not a regression; a shared `internal/authz` types package would be cleaner if this grows.

---

## 3. Overall verdict

**APPROVED** — no P0/P1. Spec ✅ and Quality Approved; the six P2s are advisory and none of them blocks merge.

## 4. For manual verification before merge

This is a new authenticated read surface, so a few things cannot be settled from this repository:

1. **Revocation semantics.** `authVerifyBot` validates `robot.bot_token` with `status = 1` and looks up `creator_uid`, but performs **no check on the owner's own status**. Confirm that offboarding or disabling a user invalidates their `bf_` bot token; otherwise a departed employee's personal agent keeps owner-scoped read access to their summaries.
2. **Space resolution is "first active membership."** Upstream returns a single space via `SELECT space_id FROM space_member WHERE uid = <bot> AND status = 1 ORDER BY created_at LIMIT 1`. Nothing leaks (the owner still has to be creator/participant), but for a multi-space owner the agent silently sees only one space, and the choice shifts if the bot joins another space. Confirm the personal bot is guaranteed co-resident with its owner's working space.
3. **Live contract + rate limits.** Run one real request per route against a deployed octo-server, and check whether P2-2's shared per-IP `verify` budget is a real constraint in your topology.
4. **The data contract in P2-3** wants a product/security sign-off, not just an engineering one.

Keeping the `needs-human-review` label until 1, 2 and 4 have an owner seems right.

yujiawei:
One more P2 to add to my review above — an observability nit I missed, verdict unchanged (still APPROVED):

**`internal/api/handler/task.go:1117` — the personal-result fallback cannot distinguish "no row" from "DB error"**

```go
err := h.db.Where("task_id = ? AND user_id = ? AND worker_status = ?", taskID, userID, model.PersonalStatusCompleted).First(&personal).Error
if err == nil && strings.TrimSpace(personal.Content) != "" {
```

Only the success branch is handled, so a connection timeout, deadlock, or context cancellation falls through to the same `40008 暂无结果` as a genuinely absent row. Behavior is fail-closed, so this is not a correctness or security issue — but if agents suddenly stop seeing personal results there is no log signal to distinguish an infrastructure problem from expected emptiness, and this is a brand-new call path with no operational history.

`errors`, `log`, and `gorm` are already imported in this file, and the repo already has this idiom (e.g. `internal/api/handler/personal_refine.go:801,825` use `else if err != gorm.ErrRecordNotFound`), so it is a two-line addition:

```go
if err != nil && !errors.Is(err, gorm.ErrRecordNotFound) {
    log.Printf("[bot] personal fallback query failed task=%d user=%s: %v", taskID, userID, err)
}
```

Noting for fairness that the immediately surrounding code in this file uses the same `err == nil` shorthand (e.g. `task.go:1050`), so this is a local-convention question rather than a regression introduced here. Fine to fold into a follow-up.

Jerry-Xin (dismissed):
The PR is in scope for `octo-smart-summary` and the bot read-only surface preserves the existing owner/participant ACL path while sourcing identity and space from `verify-bot`.

💬 Non-blocking

🟡 Warning: I could not run the CGO-enabled handler/router test packages locally because the native `tokenizers` library is missing. `CGO_ENABLED=0 go test ./internal/api/handler ./internal/middleware ./internal/auth ./internal/api/router` passed, and `go test ./internal/middleware ./internal/auth` also passed with CGO enabled.

🔵 Suggestion: Consider adding one end-to-end successful bot route test through `SetupPublic` with a real test DB, not only direct handler middleware injection. Current coverage is good across layers, but a full `/api/v1/bot/summaries/:id/result` success path would lock the router, auth middleware, trusted space, ACL, and citation stripping together. Relevant files: [internal/api/router/router.go](/Users/mlamp/.openclaw/tmp/pr-review-evbh4ty5/repo/internal/api/router/router.go:59), [internal/api/handler/task.go](/Users/mlamp/.openclaw/tmp/pr-review-evbh4ty5/repo/internal/api/handler/task.go:1113).

✅ Highlights

🔵 The bot mount is read-only and isolated from human auth/space headers: [internal/api/router/router.go](/Users/mlamp/.openclaw/tmp/pr-review-evbh4ty5/repo/internal/api/router/router.go:57).

🔵 Bot auth fails closed, accepts only `Bearer bf_...`, and injects verified owner and space from the resolver: [internal/middleware/space.go](/Users/mlamp/.openclaw/tmp/pr-review-evbh4ty5/repo/internal/middleware/space.go:28).

🔵 Result fallback is scoped to bot requests and by-person tasks only, and requires the owner’s completed non-empty personal result: [internal/api/handler/task.go](/Users/mlamp/.openclaw/tmp/pr-review-evbh4ty5/repo/internal/api/handler/task.go:1113).

🔵 Bot citation context stripping copies before mutating, avoiding accidental source mutation: [internal/api/handler/task.go](/Users/mlamp/.openclaw/tmp/pr-review-evbh4ty5/repo/internal/api/handler/task.go:1191).

lml2468 (approved):
# PR Review — PR #172 (Mininglamp-OSS/octo-smart-summary)

🔄 Re-review at head SHA `67eec74c6e73` against merge-base `2225a804` (base `main`). Re-anchored from my APPROVE at `6f5be7d3`. **The head move is a merge-of-main, not a change to #172's feature.**

**Delta analysis (re-review protocol):**
- The head is a merge commit ("Merge origin/main into HEAD"). #172's **own** delta vs the current merge-base is unchanged in shape (483 lines, same 10 files) from what I approved.
- The two core bot-auth security files — `internal/middleware/space.go` (`StrictBotAuthMiddleware`) and `internal/auth/http_resolver.go` (`ResolveBot`) — are **byte-identical** to my approved `6f5be7d3` (verified via `git diff 6f5be7d3 67eec74`). `task.go` bot handling: no diff vs my approved head.
- The **only** file changed vs `6f5be7d3` is `router.go`, and the diff is **purely benign merge integration**: `llmFallbackModels []string` threaded into `SetupPublic`/`NewAgentChatHandler` (from #180 landing in main). The bot mount is fully intact — `botV1 := r.Group("/api/v1/bot")` behind `StrictBotAuthMiddleware`, 3 GETs (lines 59-64), unchanged.

- `Build`, `Test`, `Test (race, cgo)` (1m48s), `Vet`, **`check-sprint`** → **all pass** (governance now cleared).
- Local: build/vet clean; `go test ./internal/middleware ./internal/auth ./internal/api/router ./internal/api/handler` → pass.

## 1. Spec compliance — ✅
Unchanged from prior approval; the feature (bot read surface, owner-scoped, server-resolved space, read-only) is byte-identical. Nothing new introduced by the merge beyond the #180 fallback param on a signature.

## 2. Code quality — Approved
The full security model I verified at `6f5be7d3` is intact and re-confirmed:
- `bf_`-only bearer, identity + space server-resolved via verify-bot, client space header ignored, all identity fields required (space.go — byte-identical).
- Bot routes reuse the same handlers with `user_id = OwnerUID` → same `authorizeTaskAccess` fail-closed gate (empty space → 404); no new ACL path.
- Owner personal-result fallback inside the authorized block; citation context stripped for bots.
- Read-only mount; no write route.

- **Merge integrity:** #172's owned files unchanged; only router.go touched, and only for the #180 fallback param — bot mount + bot-auth middleware unchanged.
- **Security files byte-identical:** space.go / http_resolver.go diff-clean vs approval.
- **All bot tests still pass:** middleware (space-spoof/token-whitelist/incomplete-identity), router (mount-behind-auth, human-token/client-space rejection), handler (owner fallback, context-strip, human-path non-regression).
- **CI pulled directly** — race+cgo green; check-sprint now green too.

- 🟡 (Jerry-Xin) `/bot/summaries/:id` reuses the full human DTO (permissions/members) — a dedicated bot DTO would shrink the exposure surface. Cosmetic/hardening.
- 🟡 (mochashanyao) the personal-fallback DB error is unlogged — observability nit.

## 3. Test assessment
- Unchanged, comprehensive coverage of every safety claim (verified at prior head, all still passing at this head).

## 4. Coverage / blind spots
- Executed: byte-diff of #172's owned files vs approval, router-change scoping, build/vet, all four affected packages' tests, **live CI** (race+cgo + check-sprint green).
- Not verified: real verify-bot upstream contract (mocked) — same standing note.

- Nothing new — the feature is unchanged from the prior approval; the merge only integrated #180's fallback param. Confirm verify-bot contract (carried forward).

## Verdict
**APPROVE** — re-anchored to `67eec74c`. The head move is a clean merge-of-main: #172's bot read-surface feature (and its entire owner-ACL / server-resolved-space / read-only security model) is byte-identical to my prior approval, with the two core auth files unchanged and the bot mount intact; the only new diff is the #180 `llmFallbackModels` param threading. All functional CI is green (incl. race+cgo) and **check-sprint now passes**. No regression.
Suggested decision: `Safe to merge` — all gates green.

Jerry-Xin (approved):
The PR is in scope for `octo-smart-summary` and the bot read-only surface preserves the existing owner/participant ACL path while sourcing identity and space from `verify-bot`.

💬 Non-blocking

🟡 Warning: I could not run the CGO-enabled handler/router test packages locally because the native `tokenizers` library is missing. `CGO_ENABLED=0 go test ./internal/api/handler ./internal/middleware ./internal/auth ./internal/api/router` passed, and `go test ./internal/middleware ./internal/auth` also passed with CGO enabled.

🔵 Suggestion: Consider adding one end-to-end successful bot route test through `SetupPublic` with a real test DB, not only direct handler middleware injection. Current coverage is good across layers, but a full `/api/v1/bot/summaries/:id/result` success path would lock the router, auth middleware, trusted space, ACL, and citation stripping together. Relevant files: [internal/api/router/router.go](internal/api/router/router.go:59), [internal/api/handler/task.go](internal/api/handler/task.go:1113).

✅ Highlights

🔵 The bot mount is read-only and isolated from human auth/space headers: [internal/api/router/router.go](internal/api/router/router.go:57).

🔵 Bot auth fails closed, accepts only `Bearer bf_...`, and injects verified owner and space from the resolver: [internal/middleware/space.go](internal/middleware/space.go:28).

🔵 Result fallback is scoped to bot requests and by-person tasks only, and requires the owner’s completed non-empty personal result: [internal/api/handler/task.go](internal/api/handler/task.go:1113).

🔵 Bot citation context stripping copies before mutating, avoiding accidental source mutation: [internal/api/handler/task.go](internal/api/handler/task.go:1191).

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #172 (octo-smart-summary)
**PR:** Mininglamp-OSS/octo-smart-summary #172
**Head SHA:** `67eec74c6e739b74f89695714732b8b35e5636f0`
**Routing:** security_sensitive (automated review)

## Summary

This PR introduces a bot-facing read-only API surface (`/api/v1/bot/summaries`, `/summaries/:id`, `/summaries/:id/result`) for personal agents to access owner-scoped summary data. It adds a new `StrictBotAuthMiddleware` that verifies bot tokens via a server-side `/v1/auth/verify-bot` endpoint and injects the bot owner's identity and space — never derived from client headers. A personal-result fallback path lets bots retrieve individual summaries when no group result exists, and a `citationsForRequest` filter strips surrounding chat context (`ContextBefore`/`ContextAfter`) from citations in bot responses. The keyword search in `ListSummaries` is expanded to match both `title` and `topic` (affects both human and bot paths). The design is security-conscious throughout: space identity is exclusively server-resolved, the `bf_` token prefix gates non-bot traffic at the middleware layer, and citation privacy is properly layered on top of the existing `callerPlainCitationsVisible` gate. No P0/P1 issues found; two P2 observations on operational hardening below.

## Verification

Static analysis only at head `67eec74c6e739b74f89695714732b8b35e5636f0`; build and tests not executed in this environment.

- ✅ **Space isolation** — `authorizeTaskAccess` (`internal/api/handler/task.go:111`) queries `WHERE id = ? AND space_id = ?` using `middleware.GetSpaceID(c)`, which for bot requests is server-resolved by `StrictBotAuthMiddleware`. Empty space is rejected before any query (fail-closed). Same pattern in `ListSummaries` (`:481`).
- ✅ **User isolation** — `canAccessTask` (`:107`) requires creator or participant. Bot middleware sets `user_id` = `identity.OwnerUID`, so the bot can only see its owner's tasks.
- ✅ **Token-realm separation** — Bot middleware requires `bf_` prefix (`internal/middleware/space.go:33`). Human tokens (`Token` header) are ignored by the bot group; bot tokens (`Authorization: Bearer bf_*`) are ignored by the human group. No cross-realm leakage.
- ✅ **Citation context stripping** — `citationsForRequest` (`internal/api/handler/task.go:1196`) deep-copies citations and nils `ContextBefore`/`ContextAfter` for bot callers. Applied in both `GetSummary` (`:935`) and `GetResult` (`:1141`), plus the personal-result fallback (`:1175`). Privacy gate (`callerPlainCitationsVisible`) runs first; context stripping is an additional bot-specific filter on top.
- ✅ **Personal fallback scoping** — The bot-only fallback at `:1114` is gated on `bot_request == true` AND `SummaryMode == ModeByPerson`, and the DB query is `WHERE task_id = ? AND user_id = ? AND worker_status = ?`. A bot can only retrieve its own owner's completed personal result. Human requests always skip this fallback.
- ✅ **Fail-closed auth** — `ResolveBot` (`internal/auth/http_resolver.go:76`) returns empty identity (not error) on 401/403, which the middleware maps to 401. Incomplete identity (missing any of BotUID/OwnerUID/SpaceID) is also rejected. Resolver errors map to 500 (not a silent pass).

## Findings

No P0/P1 issues. Two P2 operational-hardening items below.

### P2 — Bot auth resolver is not cached (`internal/auth/http_resolver.go:76`)

The human-token path wraps the same `HTTPTokenResolver` in a `CachedResolver` (30 s TTL, 10 000 entries — see `cmd/summary-api/main.go:72`), giving it burst and DoS protection against the upstream auth service. Bot tokens bypass this: every request triggers a full HTTP round-trip to `/v1/auth/verify-bot` with no caching layer. Under bursty bot traffic, or a credential-stuffing attack with random `bf_*` strings that pass the middleware prefix check, the auth service sees one upstream call per inbound request. The middleware's `bf_` prefix check reduces noise from human/app tokens but does not rate-limit valid-format bot tokens. Consider wrapping `httpResolver` in a `CachedBotResolver` analogous to `CachedResolver`, or adding a short negative-cache for empty-identity results.

### P2 — Response body not drained before close (`internal/auth/http_resolver.go:92`)

`ResolveBot` uses `defer resp.Body.Close()` without reading the body to EOF. In Go's `net/http`, calling `Close` without draining prevents the underlying TCP connection from being returned to the pool. The existing `ResolveUID` has the same pattern (pre-existing), but given that bot auth is uncached and will see higher call volume per unique token, the connection churn is proportionally larger. Add `io.Copy(io.Discard, resp.Body)` before the deferred close, or use a `LimitReader` to cap the read.

## Data-Flow Tracing

| Consumed data | Upstream source | Runtime validation |
|---|---|---|
| `user_id` in handlers | `identity.OwnerUID` from `/v1/auth/verify-bot` | Set by `StrictBotAuthMiddleware`; used in `authorizeTaskAccess` (space+creator/participant gate) and `ListSummaries` (creator/participant WHERE clause). Data flows correctly. |
| `space_id` in handlers | `identity.SpaceID` from `/v1/auth/verify-bot` | Set by middleware; used in `authorizeTaskAccess` and `ListSummaries` as `WHERE space_id = ?`. Not derived from `X-Space-Id` header for bot requests. Correct. |
| `bot_request` context flag | Set by `StrictBotAuthMiddleware` only | Used as gate for personal fallback (`GetResult:1114`) and citation stripping (`citationsForRequest:1196`). Not reachable from human route group (separate gin group). Correct. |
| Citations in bot responses | `result.GetCitations()` or `personal.GetCitations()` | Privacy-gated by `callerPlainCitationsVisible` first, then `citationsForRequest` strips context. Deep-copy prevents source mutation. Correct. |
| Keyword param in list | `c.Query("keyword")` | Parameterized via GORM `LIKE ?` with `"%"+s+"%"`. No SQL injection. Now matches `title OR topic` (was `title` only). Intentional behavioral expansion, tested. |

## Blind-Spot Checklist (C1–C6)

- **C1 — Dual-path parity**: Bot group (`/api/v1/bot/*`) and human group (`/api/v1/*`) reuse the same handlers (`ListSummaries`, `GetSummary`, `GetResult`). Handlers correctly differentiate via `bot_request` context for: personal-result fallback, citation context stripping. No asymmetric authorization bypass. **Clear.**
- **C2 — Control-flow ordering**: `StrictBotAuthMiddleware` runs before handlers; citation stripping runs after `callerPlainCitationsVisible`; personal fallback runs after `authorizeTaskAccess` and before the 404 fall-through. No ordering issues. **Clear.**
- **C3 — Authorization boundary**: Bot routes are properly gated behind `StrictBotAuthMiddleware` with server-resolved identity. The `bf_` prefix check is defense-in-depth alongside upstream verification. No endpoint is reachable without bot auth. **Clear.**
- **C4 — Authorization lifecycle**: Bot tokens are verified per-request against the upstream auth service. No persistent session or cached authorization state. If a bot token is revoked upstream, the next request will fail verification. **Clear.**
- **C5 — Build/runtime**: N/A — no build, packaging, or deployment changes in this diff.
- **C6 — Governance/policy**: N/A — no policy or documentation changes.

## Cross-Round Blocker Recheck (R6)

## Verdict: APPROVED

Well-designed security boundary for bot access to owner-scoped summaries. Space and identity are exclusively server-resolved, the token prefix check provides defense-in-depth, and citation privacy is properly layered. The two P2 items (uncached bot auth and response-body drain) are operational-hardening suggestions, not merge blockers.





[Octo-Q] verdict: APPROVE — No P0/P1 issues found. Bot auth boundary is properly designed with server-resolved identity, fail-closed behavior, and layered citation privacy. Two P2 operational suggestions (bot auth caching, response body drain) are non-blocking.

yujiawei (approved):
# Code Review — PR #172 (octo-smart-summary)

Reviewed at head `67eec74c6e739b74f89695714732b8b35e5636f0`, merge-base `2225a80`, against a fresh checkout. This PR is flagged security-sensitive, so I also read the counterpart implementation in `octo-server`/`octo-lib` to check the cross-service contract that the unit tests mock out.

**Verification performed:** `go build ./...` clean · `CGO_ENABLED=0 go vet ./...` clean · `go test ./internal/middleware/... ./internal/auth/... ./internal/api/router/...` green · the six new cgo-tagged bot tests all pass individually.

---

## 1. Spec compliance

**Spec: ✅**

Every claim in the PR description checks out:

| Claim | Where | Verdict |
|---|---|---|
| read-only bot mount | `internal/api/router/router.go:57-66` — three `GET`s only | ✅ |
| identity via verify-bot | `internal/auth/http_resolver.go:76-107` | ✅ |
| preserve existing owner ACLs | reuses `authorizeTaskAccess` (`task.go:111-145`) and the `ListSummaries` visibility predicate (`task.go:509`) unchanged | ✅ |
| strip surrounding citation context | `task.go:1190-1200`, applied at all three emission points (`:935`, `:1141`, `:1187`) | ✅ |
| owner personal-result fallback | `task.go:1114-1122`, correctly scoped `user_id = <owner>` | ✅ |
| accept only `bf_` bearer tokens | `internal/middleware/space.go:31` | ✅ |
| ignore client-supplied space headers | `space.go:47` sets `space_id` from the verified identity; asserted at `bot_auth_test.go:522-541` | ✅ |
| expose no bot write routes | `router.go:63-65` | ✅ |
| keep existing human behavior unchanged | `bot_request` is set only by `StrictBotAuthMiddleware` (confirmed by grep — no other writer), so `citationsForRequest` is a no-op on `/api/v1`; regression-pinned at `bot_citation_test.go:123` | ✅ |

- **Missing:** none against the PR's stated scope. Issue #144's full acceptance criterion (agent *initiates* a summary) is not met, but the PR explicitly scopes itself to the read-only foundation and defers creation — that is a declared boundary, not a gap.
- **Over-build:** one, declared. The `keyword` filter widening at `task.go:541-542` (`title LIKE` → `(title LIKE OR topic LIKE)`) lands on the *shared* handler, so it also changes the existing human `GET /api/v1/summaries`. It is ACL-neutral, listed in the PR's Validation section, and plausibly needed for agent-side search — noting it for the record rather than gating on it.
- **Deviation:** none.

### Trust-boundary checks that pass

I specifically tried to break the auth boundary and could not:

1. **Bot token → human write routes.** `/api/v1` gates on `StrictAuthMiddleware` reading the `Token` header → `/v1/auth/verify` → Redis `token:<value>`. A `bf_` token lives in `robot.bot_token`, not that cache, so it resolves to an empty uid → 401. Bots cannot reach the write surface. This is the single most important property and it holds.
2. **Human token → bot mount.** Requires `Authorization: Bearer bf_…`; `verify-bot` (`octo-server modules/user/api.go:4287-4290`) looks the value up in `robot.bot_token`, which a session token never occupies → 401.
3. **Space spoofing.** `X-Space-Id` is never read on the bot group; `authorizeTaskAccess:129` and `ListSummaries:487` both hard-reject an empty space before any query, so a blank server-resolved space fails closed rather than matching `space_id=''` rows.
4. **Fail-closed on partial identity.** `http_resolver.go:104` and `space.go:39` both require all three of `bot_uid`/`owner_uid`/`space_id`; octo's `IFNULL(creator_uid,'')` therefore degrades to a clean 401 rather than an empty-uid read.
5. **Wire contract.** `wkhttp.Context.Response` is `c.JSON(200, data)` (`octo-lib pkg/wkhttp/http.go:130-132`) — a bare envelope, so `ResolveBot`'s direct decode into `{bot_uid, owner_uid, space_id}` matches `authVerifyBotResp` (`octo-server modules/user/api.go:4261-4267`) field-for-field. This is the part `roundTripFunc` cannot prove, and it is correct.
6. **Citation strip completeness.** `model.Citation` (`model/model.go:217-228`) has exactly two nested-context fields and both are nilled; `model.TeamCitation` (`:246-252`) carries no message bodies, so leaving `team_citations` unstripped is right, not an oversight.
7. **No read-state side effects.** `summary_user_read` is only written by `MarkSummaryRead`, which is not on the bot mount — bot reads don't consume the owner's unread badges.

---

## 2. Code quality

**Quality: Approved**

No P0 or P1. The auth design is sound and noticeably more careful than the surrounding code (server-resolved space, all-fields-required, no write routes). Everything below is P2 or nit.

### P2-1 — Fallback block cannot distinguish "not found" from "database broken"

`task.go:1113-1122` contains two error collapses in eight lines:

- `pickDisplayResult` (`task.go:147-153`) maps **every** error from `queryDisplayResult` to `found=false`. So a transient failure on the team-result lookup routes a bot into the personal fallback *even when a team result exists* — the bot then reports the owner's personal report as the answer. Wrong data, not just a wrong status.
- `task.go:1118`'s `if err == nil` guard likewise turns a connection timeout or deadlock on the personal query into `40008 "暂无结果"` / 404, so clients can't retry on 5xx and the failure is invisible in error-rate dashboards.

Gate both on `errors.Is(err, gorm.ErrRecordNotFound)` and return 500 otherwise. This file already uses that idiom at `task.go:1525`, `:1851`, `:1922`, so it's a local-style fix.

### P2-2 — Fallback ignores `submitted_at`

`task.go:1117` selects on `worker_status = Completed` alone. But the schema documents that column as `NULL=草稿/未提交` (`migrations/legacy/002_p2_personal_summary.sql:16`), `ListSummaries` treats `worker_status=Completed AND submitted_at IS NULL AND participantCount>1` as *"you still owe a submission"* (`task.go:601`), and `GetSummary` only exposes a member's content when `SubmittedAt != nil` (`task.go:1077`). So on a multi-participant personal-mode task the bot can present the owner's **unsubmitted draft** as the finished result.

For the single-participant case — the common one — returning it is clearly right, and `TestBotGetResultFallsBackToOwnersCompletedPersonalResult` pins that intent. Consider narrowing to `submitted_at IS NOT NULL OR participantCount <= 1` so the multi-person draft case is explicit rather than incidental.

### P2-3 — Bot auth is uncached while human auth is cached; octo 429 surfaces as 500

`cmd/summary-api/main.go:71-72` wraps the human resolver in `NewCachedResolver(…, 30s, 10000)` but passes the **raw** `httpResolver` as the bot leg at `:120`. Every bot request is therefore an uncached POST to octo, and `/v1/auth/verify-bot` sits behind `StrictIPRateLimitMiddleware(…, "verify", 1000/60, burst 100)` — **per-IP**, sharing the `"verify"` keyspace with `/auth/verify` and `/auth/verify-api-key` (`octo-server modules/user/api.go:203, 308`). summary-api is one egress IP, so chatty agent traffic competes with human token cache-misses for one bucket. On 429, `http_resolver.go:100-102` returns a generic error and `space.go:35-38` maps it to **500**, not 503.

To be clear, uncached is a defensible choice — it gives instant token revocation, which matters given bot tokens are long-lived. My asks are: state that tradeoff in a comment so it isn't "fixed" later by accident, and map non-2xx/non-401/403 to 503 so the failure is honestly classified.

### P2-4 — Privacy stripping covers citations but not the other content the reused handler returns

`GET /api/v1/bot/summaries/:id` reuses `GetSummary` unchanged, which returns `members[].content` — **every other participant's submitted personal report** (`task.go:1076-1081`) — plus `personal_result.content` (`:1051-1058`). Neither passes through `citationsForRequest`.

There is no privilege escalation here: the owner can already read both via the human API under the same ACL, and these are *generated summaries* rather than raw chat, so the "strip raw uncited messages, keep entitled summaries" line is coherent. But the same payload now flows into an LLM context automatically instead of onto a screen a human chose to open, which is a different exfiltration surface than the one the ACL was designed for. Given the `needs-human-review` label, this deserves an explicit decision rather than inheriting one.

### P2-5 — No audit trail for bot reads, and the hooks for it are dead code

`space.go:43-45` sets `actor_type`, `actor_id`, and `bot_id`, and **nothing reads any of them** (grep finds no non-test consumer). `gin.Logger()` records path and status but not which bot acted. On a security-sensitive read path standing in for a human, "which agent read whose summary" is exactly the question an incident would ask. Either wire `bot_id` into a structured log line for the bot group, or drop the three unused keys so they don't read as an implemented feature.

Relatedly, `GetSummary` still returns a `permissions` block advertising `can_edit_personal` / `can_add_member` / `can_edit_team` (`task.go:1024-1032`) on a mount with no write routes — harmless, but it will mislead a bot client into attempting calls that 404.

### P2-6 — Response-shape asymmetry between the two `GetResult` branches

`writeBotPersonalResult` (`task.go:1163-1187`) emits `result_source:"personal"`, but the team branch (`:1145-1161`) emits **no** `result_source` at all, so a client must infer "team" from absence. The personal branch also drops `version`, `operation_type`, `operation_note`, and `parent_result_id`, and can emit `generated_at: null` (`PersonalResult.GeneratedAt` is `*time.Time`) where the team branch always emits an RFC3339 string. A client doing `new Date(generated_at)` or reading `.version` breaks on the personal shape. Add `result_source:"team"` and align the key set.

### P2-7 — Fallback bypasses `callerPlainCitationsVisible` on an unstated invariant

`writeBotPersonalResult` calls `citationsForRequest` directly, skipping the `callerPlainCitationsVisible` gate that both other emission points use. This is safe *today* only because `task.go:1117` pins `user_id = <owner>`, making the owner the producer by construction — but that reasoning is invisible at the call site, so a future widening of the query silently un-gates raw citations. Add a comment, or route through the shared gate.

### P2-8 — `bf_` prefix gate rejects legacy user-bot tokens that octo still honours

`space.go:31` requires `HasPrefix(token, "bf_")`. `octo-server modules/bot_api/auth.go:26,39` documents User Bot auth as *"bf_ prefix **or legacy tokens**"*, and `authVerifyBot` matches `robot.bot_token` with no prefix constraint. A legacy bot therefore gets an opaque `401 "authentication required"` here while working everywhere else in the platform.

The prefix check is otherwise good defense-in-depth and I want to record why: `app_` bots live in a **separate table** (`app_bot`, per `bot_api/auth.go:12-13`) that `verify-bot` never queries, so app bots are already excluded server-side — the prefix is belt-and-braces rather than the sole boundary. Worth confirming the legacy-token exclusion is intended.

### P2-9 — `space_id` is whatever single row octo happens to pick

`octo-server modules/user/api.go:4313-4318` takes the bot's first active `space_member` row (`ORDER BY created_at DESC LIMIT 1`, error discarded). A bot in more than one space is silently pinned to one of them, and the owner's summaries in the other spaces become invisible with no distinguishable error — the agent just returns an empty list, which reads as a bug rather than a scope limit. Worth documenting as a single-space constraint.

### Nits

- **`internal/auth` now imports `internal/middleware`** (`http_resolver.go:11`) so the octo client depends on the HTTP layer for a DTO. The existing `TokenResolver` avoided this by trafficking in primitives. It compiles only because `middleware` doesn't import `auth`; any future edge in that direction is an import cycle. Consider moving `BotIdentity`/`BotTokenResolver` to a neutral package.
- **`bot_routes_test.go:335-345`** doesn't test what its name says. `fixedBotResolver{}` is a zero value, so the 401 comes from the missing `Authorization` header — the `Token` and `X-Space-Id` headers are inert and the test passes identically without them. The real assertion does exist at `middleware/bot_auth_test.go:522-541`; consider renaming this one or driving it with a resolver that succeeds.
- **`First()` without `ORDER BY`** at `task.go:1117` returns the lowest-id row. The code relies on `(task_id, user_id)` uniqueness, but the schema only enforces `UNIQUE KEY uk_task_participant (task_id, participant_ref_id)` (`migrations/sql/20260101-00-baseline.sql:65`). I tried to construct a live path producing two rows for one user and could not — `RemoveMember` hard-deletes both participant and personal rows (`personal.go:1598-1607`) — so this is hardening, not a live bug. An explicit `Order("id DESC")` would remove the dependence on an unenforced invariant.
- **`http_resolver.go:81-83`** discards the `json.Marshal` error and concatenates `baseURL` without normalizing a trailing slash. Both mirror `ResolveUID` at `:47,49` exactly, so the new method is consistent with the file rather than newly wrong — fix both together or neither.
- **`ListSummaries` N+1** (`task.go:677`, `:688`, `:706` — per-item source, display-result and creator lookups) and the router-level `Access-Control-Allow-Origin: *` with `Authorization` in `Allow-Headers` (`router.go:24-33`) are both pre-existing, but an automated poller will exercise the N+1 far harder than a human list view.

---

## 3. Overall verdict

**APPROVE**

Spec ✅ and Quality Approved. The auth boundary is correct on every path I could attack, the cross-service contract matches the real `octo-server` implementation, and the new tests pin the properties that matter (space-spoof rejection, human-path non-regression, non-personal-mode non-fallback). The P2s are worth a follow-up but none of them should hold up merge.

## 4. Suggested follow-ups, in priority order

1. Gate the fallback on `errors.Is(err, gorm.ErrRecordNotFound)` in both places (P2-1) — this is the only finding that can return *wrong data* rather than a wrong status.
2. Decide explicitly whether the multi-participant unsubmitted-draft case should be reachable (P2-2).
3. Map octo non-2xx to 503 and comment the deliberate no-cache choice (P2-3).
4. Add `result_source:"team"` and align the two response shapes (P2-6).
5. Either log `bot_id` on the bot group or delete the unused `actor_*` keys (P2-5).

## 5. For a human to verify before/after merge

These are outside what static review can settle:

1. **Confirm a `bf_` bot token is not resolvable via `/v1/auth/verify`.** My reading of `authVerifyToken` says it only consults the Redis `token:` cache, which bot tokens never occupy — so bots cannot reach the human *write* routes. This is the load-bearing assumption of the whole design and deserves one empirical check rather than my inference.
2. **Confirm bots are provisioned into exactly one space in production** (P2-9), otherwise the "empty list" failure mode will generate support noise.
3. **Confirm transport security to octo** — whether `OCTO_API_URL` is HTTPS or in-cluster only. `bf_` tokens are long-lived bearer secrets now crossing this hop on every request.
4. **Confirm the intended agent call volume** against the shared 1000 req/min `"verify"` bucket (P2-3). My concern there is analytical, not measured.
5. **Confirm the local `octo-alex` CLI / chat-agent validation covered the multi-participant personal-mode path** (P2-2) — the automated tests only cover the single-participant shape.

## Coverage note

**Covered:** full diff at `67eec74`; the three reused handlers read end-to-end including every response field reachable on the bot mount; `Citation`/`TeamCitation`/`PersonalResult` models and the `summary_personal_result` DDL; build, vet, and the new tests executed. Cross-repo: `octo-server` `authVerifyBot` + its rate limiter + `bot_api/auth.go` token realms, and `octo-lib` `wkhttp.Context.Response` — i.e. the service contract the mocked `roundTripFunc` test cannot reach.

**Not covered:** production `space_member` cardinality for bots; transport security on the summary-api → octo hop; real agent request volume; behavior under MySQL (all tests run on SQLite, and `ListSummaries` branches on `Dialector.Name()`, so the MySQL arg-count path is untested here); the manual `octo-alex` verification claimed in the description; and whether `robot.bot_token` values ever reach the Redis `token:` cache by some path I did not find.

yujiawei:
One more P2 nit I missed in my review above — non-blocking, does not change the verdict.

**`ResolveBot` discards a pooled connection on every non-200 (`internal/auth/http_resolver.go:92-98`)**

`defer resp.Body.Close()` at `:92` fires on all paths, but the two early returns leave the body entirely unread:

- `:93-95` — the 401/403 branch returns without touching the body. This is the *common* path for any invalid or expired token.
- `:96-98` — the non-200 branch likewise.

Go's transport won't return a persistent connection to the pool when the body hasn't been consumed to EOF, so each of these costs a fresh TCP (and TLS) handshake on the next call. That compounds with the uncached bot-auth path in P2-3 of my review: uncached means one upstream call per request, and the 401 branch means each rejected request also burns a connection. Under a burst of invalid tokens you get connection churn and ephemeral-port pressure rather than pool reuse.

Even the success path leaves a remainder — `json.NewDecoder(...).Decode()` stops at the end of the JSON value, so a trailing newline is left unread.

Suggested fix, applied before the early returns:

```go
defer func() {
    _, _ = io.Copy(io.Discard, io.LimitReader(resp.Body, 4<<10))
    resp.Body.Close()
}()
```

`ResolveUID` has the same shape at `:59,65`, so it's worth fixing both together or neither — consistent with how I framed the other shared-pattern nits.

PR #178, opened 2026-07-28
feat(summary): generate and surface a short AI abstract (Option B)

wanyiwang06:
## Summary
给总结新增「AI 摘要」(详情页顶部 callout)。生成走 **方案 B**——对**定稿后的交付内容**做**一次轻量 LLM 调用**压成 2-4 句摘要。不用"主 prompt 搭车"(方案 A)的原因:**Agent 总结的交付物是被捕获的会话消息、没有可控主 prompt**,只有"content→摘要"这条能同时覆盖传统与 agent 两条链。

## 改动
- **迁移**:`summary_result` + `summary_personal_result` 各加 `abstract VARCHAR(300)`(默认空)。
- `service.GenerateAbstract`:best-effort——nil/空/LLM 失败均返回空,**绝不阻断或失败总结**;输出裁剪 ≤120 字。
- **惰性生成+缓存(单一路径)**:两个读取点(`GetSummary` 团队结果、`GetPersonal` 个人/agent 结果)首次读到无摘要且有正文时生成一次并回写。**一套路径同时覆盖 传统 / agent / 存量回填**,不动 worker 写入事务。
- 响应新增 `abstract` 字段。

## How verified
容器内(golang:1.25-bookworm + libtokenizers.a,与 CI 一致):`go build ./...` ✅ · `go vet ./internal/...` ✅ · `go test ./internal/service/ ./internal/api/handler/` 全部 `ok`(含新 abstract 测试 + 既有 handler 无回归)。

## 备注
前端渲染 callout 是配套 octo-web PR(读 `result.abstract`,空则隐藏)。惰性生成取舍:首次查看多一次 LLM 调用(之后缓存),换来一处接入 + 自动回填 + 不碰 worker 事务。

PR #180, opened 2026-07-28
feat(agent): cross-model fallback via LLM_FALLBACK_MODELS

### LINKED ISSUE #179: fix(agent): 提升模型不稳定场景的容错与可观测性
## 问题背景

线上 Agent 总结请求会因底层模型服务不稳定而失败。PR #168 已提高单步超时时间并为部分已知错误增加安全的 `detail`，但模型服务返回非白名单错误时，客户端仍只能收到统一的 `internal error`，难以定位和恢复。

## 复现信息

- 接口：`POST /api/v1/agent/chat/stream`（非流式 `/api/v1/agent/chat` 也可能受影响）
- session_id：`7222e9bb-579f-4582-9cb8-ab32fad0932a`
- message：`总结一下我最近的未读消息`
- profile：`summary_refine`
- referenced_task_ids：`[1120]`
- 响应：

```json
{
  "code": 50000,
  "message": "agent chat failed",
  "detail": "internal error"
}
```

已确认此次失败由 Agent 使用的模型不稳定导致，不是 PR #168 已处理的 60 秒 step timeout。

## 当前问题

1. 模型服务的瞬时 5xx、429、网络错误、异常响应或空 `choices` 最终都会使整个 Agent 请求返回 500。
2. 非白名单错误统一映射为 `internal error`，符合防泄漏要求，但服务端日志缺少 `session_id`、`profile`、step、模型和错误分类，跨 Pod 排查困难。
3. 前端无法判断错误是否可重试，也无法给用户提供有意义的恢复操作。

## 期望改进

- 为模型调用建立稳定的错误分类，例如：`rate_limited`、`upstream_5xx`、`network_error`、`invalid_response`、`timeout`；客户端只返回安全、固定的错误码，不透传上游响应正文或内部地址。
- 服务端结构化记录 `session_id`、`profile`、模型名、step、attempt、耗时、上游状态码和安全错误分类，并确保 stdout/stderr 能被日志系统采集。
- 对明确可恢复的模型瞬时错误实施有上限的重试；评估重试耗尽后的模型 fallback/降级策略。
- SSE 与非流式接口保持一致的错误分类和响应行为。
- 增加模型 429、5xx、网络失败、非法 JSON、空 `choices` 等场景的单元测试或集成测试。

## 验收标准

- [ ] 模型瞬时错误不会未经分类地直接变成不可诊断的 500。
- [ ] 可通过 `session_id` 在服务端日志中定位到对应请求及安全错误分类。
- [ ] 客户端响应不泄漏 URL、IP、token、上游响应正文或堆栈。
- [ ] 可恢复错误按配置进行有限重试，重试次数和最终结果可观测。
- [ ] SSE 和 JSON 两条请求路径均有测试覆盖。

## 关联

- #168

wanyiwang06:
> **Revision (2026-07-28, in response to yujiawei review at head 1e05d278):** the P1 (fallback unreachable under default timeouts) and every P2 raised are now addressed at head 2903f46. See the second commit and the 'Response to review' section at the bottom for the point-by-point breakdown.

## Summary

Adds `LLM_FALLBACK_MODELS` as a comma-separated environment variable that lets the agent Chat client switch to a fallback model when the primary `LLM_MODEL` exhausts its per-model retry budget on retryable failures. Scope-limited to just the fallback knob from #179's five expected improvements; structured logging, unified error taxonomy, and SSE/JSON parity are deliberately left for follow-up PRs so this stays easy to review and revert.

## What changes

**Configuration (`internal/config/config.go`)**
- New `LLMFallbackModels []string` field on `Config`.
- Sourced from `LLM_FALLBACK_MODELS` via a new `envStrList` helper (comma-separated, trimmed, empty entries dropped).
- Empty by default: preserves current single-model behavior byte-for-byte.

**Agent LLM client (`internal/agent/llm.go`)**
- `NewClient` gains a `fallbackModels []string` parameter. The constructor dedups the primary model and empty strings out of the list at construction time so the retry budget is never wasted on the same model twice.
- `Chat` split into an outer loop over models and an inner `chatOneModel` per-model retry loop. Semantics:
  - Per-model: unchanged — up to 3 attempts, exponential backoff (1s, 2s).
  - Across models: only escalate to the next model when the current model exhausted its budget on **retryable** errors (network / 5xx / 429). Terminal errors (4xx non-429, decode failure, empty `choices`) short-circuit — those indicate a caller-side or contract problem that the next model would hit too.
  - Every fallback transition emits one log line: `[llm] falling back from "A" to "B" after retries exhausted: <last err>` so silent quality drift is observable during incidents.

**Handler / router wiring**
- `AgentChatHandler` gains an `llmFallbackModels` field; propagated through `NewAgentChatHandler` -> `buildRunnerForProfile` -> `agent.NewClient`.
- `router.SetupPublic` and `cmd/summary-api/main.go` propagate `cfg.LLMFallbackModels`.
- Shared `buildRunner` helper signature updated to match; no in-tree caller uses it yet, keeping the change surface small.

**Docs**
- `CONFIGURATION.md` documents `LLM_FALLBACK_MODELS` with an example (`claude-sonnet-4-6,qwen3.6-max`).

## Test coverage

`internal/agent/llm_fallback_test.go` (new, 4 tests):

- `TestChat_FailoverToFallbackAfterRetriesExhausted` — primary returns 503 continuously; verifies exactly 3 primary attempts followed by a fallback attempt that succeeds, and the returned content came from the fallback model.
- `TestChat_NoFallbackOnTerminalError` — primary returns 400; verifies Chat surfaces the terminal error immediately after **one** request, never touching the fallback.
- `TestChat_NoFallbackConfiguredKeepsSingleModelBehavior` — persistent 503 with `nil` fallback list; verifies exactly 3 primary attempts and no unexpected behavior. Locks in backward compatibility.
- `TestNewClient_DedupesFallbackList` — constructor drops empty strings, duplicates within the list, and any entry equal to the primary model.

## Verification

Docker `golang:1.25-alpine`, `CGO_ENABLED=0`:

- `go build ./...` — OK
- `go vet ./...` — OK
- `go test -count=1 ./internal/agent/... ./internal/api/handler/... ./internal/config/...` — OK (all packages green; the four new tests pass in ~6 s)

## Non-goals (deliberately out of scope for this PR)

The other four items from #179 are separate concerns that each deserve their own reviewable diff:

1. Stable error taxonomy (`rate_limited` / `upstream_5xx` / `network_error` / `invalid_response` / `timeout`) surfaced to clients.
2. Structured logging with `session_id`, `profile`, model, step, attempt, upstream status.
3. SSE / non-streaming response parity for error classification.
4. Test coverage for 429 / 5xx / network / bad JSON / empty `choices` at the handler layer.

Happy to open follow-up PRs for those once this lands.

## Rollout

Zero-config default is unchanged (`LLM_FALLBACK_MODELS` empty -> single-model retry, same as before). To enable, set e.g. `LLM_FALLBACK_MODELS=claude-sonnet-4-6,qwen3.6-max` in the environment. No database migrations, no other config changes required.

Refs #179 (partial: fallback knob only; error taxonomy, structured logging, SSE parity and handler-level failure-mode tests remain follow-ups)

Also references #168 (previous timeout work).

---

## Response to review (yujiawei, 2026-07-28 @ 1e05d278)

**P1 — Fallback unreachable under default budgets:** fixed at 2903f46 via deadline-aware early escalation in `chatOneModel`. Before every RETRY (never the first attempt) the loop compares `time.Until(deadline)` with `c.timeout`; if there is not enough remaining budget for another full-timeout attempt AND a fallback model is available, it escalates `retry=true` immediately so the fallback gets real budget. New `TestChat_HangingPrimaryEscalatesEarlyToFallback` covers this exact scenario with a hanging upstream, ratio-preserved 2s/4s budgets, and asserts primary is attempted exactly once then the fallback receives the request.

**P2 — `Chat` escalates after parent ctx is dead:** fixed at 2903f46. `Chat` now re-checks `ctx.Err()` before each outer-loop iteration, and `doOnce` distinguishes parent-context cancellation from transport errors before returning `retry`. New `TestChat_ParentContextCancelledIsNotMasqueradedAsFallback` locks this in.

**P2 — Log line leaks upstream body:** fixed at 2903f46. Fallback log line now uses `truncateForLog(lastErr.Error(), 200)` instead of `%v` on the full wrapped error, keeping gateway response bodies out of stdout on the success path. New `TestTruncateForLog` pins the helper.

**P2 — Same gateway / same credential:** now stated explicitly in CONFIGURATION.md.

**P2 — No stickiness across steps:** deliberately deferred. Not a correctness regression; per-client sticky current-model or a short-lived breaker deserves its own PR (with its own tests around the state lifecycle). Happy to open the follow-up.

**P2 — Single-model error contract changed:** fixed at 2903f46. When `LLM_FALLBACK_MODELS` is empty, `Chat` returns `chatOneModel`'s error verbatim, without the `chat failed after exhausting N model(s)` wrapper. Existing log alerts / dashboards on "chat failed after N attempts" continue to match.

**P2 — buildRunner dead code:** removed at 2903f46 (`internal/api/handler/agent_runner_builder.go` deleted; `buildSummaryRegistryWithUID` is defined and used inside `agent_chat.go` on the `AgentChatHandler` receiver, so nothing else was affected).

**P2 — `Closes #179` → `Refs #179`:** updated in this PR body. #179 has five expected improvements; this PR delivers one (the fallback knob). The remaining four (error taxonomy, structured logging, SSE/JSON parity, handler-layer failure-mode tests) will land as follow-up PRs. Closing #179 automatically would overstate delivery.

**Coverage note:** `-race` requires cgo, which is not available in the dev alpine sandbox used for verification, so the race sweep the reviewer ran on the previous head is not re-run here. Test suite is otherwise the same as before, plus the three new tests above. CI in this repo runs `Test (race, cgo)` as a separate check — the earlier revision passed it, and this revision changes only two functions and their tests; happy to re-run once CI reports.


lml2468 (dismissed):
# PR Review — PR #180 (Mininglamp-OSS/octo-smart-summary)

Reviewed at head SHA `1e05d2786ad1` against merge-base (base `main`).

Intent: **X** = add `LLM_FALLBACK_MODELS` (comma-separated env) so the agent Chat client escalates to fallback models after the primary exhausts its per-model retry budget on **retryable** failures, so that **Y** = transient primary-model outages don't fail agent chat, affecting **Z** = `agent.Client` retry loop + config/handler/router wiring. Scoped to just the fallback knob from #179. Provider-agnostic (models flow through the repo's OpenAI-compatible `/chat/completions` client, not any vendor SDK).

- `Build`, `Test`, `Test (race, cgo)` (1m46s), `Vet` → **pass**; `check-sprint` (governance) → fail.
- Local: build/vet clean; `go test ./internal/agent` → pass.

## 1. Spec compliance — ✅
Exactly the described scope: config field + `envStrList` helper + two-level retry/fallback loop + wiring + docs. Empty default preserves single-model behavior. Deliberately defers structured logging / error taxonomy / SSE parity to follow-ups — clean, revertible surface. Nothing over-built.

## 2. Code quality — Approved

**Retry/fallback semantics are correct — the load-bearing logic:**
- Clean split: `Chat` = outer loop over `[primary, ...fallbacks]`; `chatOneModel` = the (unchanged) 3-attempt exponential-backoff per-model loop returning `(turn, retry, err)`.
- **Escalate only on retryable exhaustion:** `chatOneModel` returns `retry=true` only after exhausting its budget on retryable errors (network / 5xx / 429). Terminal errors (4xx non-429, decode failure, empty `choices`) return `retry=false` → outer loop short-circuits and does **not** try the next model. Correct: a contract/payload error would hit every model identically; escalating would waste budget and muddy the error.
- **Constructor dedup:** empty entries and any entry equal to the primary are dropped (`seen := {model:true}`), and the caller's slice is copied — no budget waste, no aliasing.
- **Backward-compatible:** empty fallback list → `models = [primary]` → byte-for-byte the prior single-model path.
- Per-model budget (3 attempts) unchanged; a fresh budget per fallback model.
- One log line per transition — observability without silent quality drift.

**ctx-cancellation is handled safely (the subtle edge I checked):**
- An http-do failure (incl. ctx cancel mid-request) is `retry=true`, but on the next attempt the backoff `select { case <-ctx.Done(): return …, false, ctx.Err() }` returns **terminal**, so a cancelled ctx short-circuits the per-model loop → the outer loop stops **without** firing wasted fallback-model requests. Bounded to at most one fast-failing extra attempt. No fallback storm on client disconnect / step timeout.

**Config:** `envStrList` trims, drops empties, returns `def` on unset/blank — treats "unset" and "empty list" identically (documented, matches the consumer's needs).

- **Terminal-error no-escalate:** `TestChat_NoFallbackOnTerminalError` — 400 → exactly 1 request, fallback untouched. ✅
- **Failover on retryable exhaustion:** `TestChat_FailoverToFallbackAfterRetriesExhausted` — 503×∞ → 3 primary + fallback succeeds. ✅
- **No-config regression:** `TestChat_NoFallbackConfiguredKeepsSingleModelBehavior` — exactly 3 primary attempts, no behavior change. ✅
- **Dedup:** `TestNewClient_DedupesFallbackList`. ✅
- **ctx cancel:** traced — terminates without wasted fallback attempts.

### Non-blocking notes
- 🟡 The outer `Chat` loop doesn't itself re-check `ctx.Done()` before the next model; it relies on `chatOneModel`'s backoff-guard to surface cancellation as terminal. Works correctly (verified), but an explicit `ctx.Err()` check at the top of the outer loop would make the intent obvious and shave the one residual fast-failing attempt. Cosmetic.
- 🟡 Total worst-case latency now scales with `(1+len(fallbacks)) × per-model-budget` (3 attempts + backoff each). With the #168 stepCtx (240s) as the outer bound this is fine, but operators configuring many fallbacks should be aware the wall-clock budget multiplies. Doc mentions the knob; a note on cumulative latency vs StepTimeout would help.

## 3. Test assessment
- Four targeted tests cover the full decision matrix (failover / terminal-no-escalate / no-config-regression / dedup) using a fake server that counts per-model requests — strong, would fail on any semantic regression. The httptest-based approach exercises real retry timing (kept short).

## 4. Coverage / blind spots
- Executed: build, vet, agent tests, **live CI** (race+cgo green), full trace of the two-level loop + ctx-cancel classification.
- Not verified: real multi-model production failover (mocked) — immaterial to the control-flow correctness.

- Operational: cumulative retry latency across many fallbacks vs the StepTimeout ceiling.

## Verdict
**APPROVE** — both gates pass. The cross-model fallback has correct semantics: escalate only on retryable-budget exhaustion, short-circuit on terminal errors (test-locked), dedup at construction, and byte-for-byte single-model behavior when unset. ctx-cancellation terminates cleanly without wasted fallback attempts. Backward-compatible, well-tested, tightly scoped. Functional CI (incl. race+cgo) is green.
Suggested decision: `Safe to merge` — pending `check-sprint`; optional cosmetic ctx-recheck + a latency note.

Jerry-Xin (dismissed):
The PR is in scope for `octo-smart-summary` and cleanly adds fallback model support to the agent LLM chat path without introducing blocking correctness, security, or architecture issues.

💬 Non-blocking

🟡 Warning: `LLM_FALLBACK_MODELS` is documented as applying to “the agent client,” but LLM-backed tools still instantiate `service.NewLLMClient` directly and will not use this fallback path, for example `summarizeMessagesChunk` and `mergeSummariesWithLLM`. Consider clarifying that this only applies to the agent planning/chat client, not map/reduce tool LLM calls. References: `CONFIGURATION.md:19`, `internal/agent/tool_summarize_chunk.go:252-254`, `internal/agent/tool_merge_summaries.go:69-71`.

🔵 Suggestion: The fallback tests exercise 503 and terminal 400 behavior, but not 429 or network errors even though those are part of the advertised retry/fallback contract. A small table test around `doOnce`/`Chat` would make the retry classification harder to regress. References: `internal/agent/llm.go:181-192`, `internal/agent/llm_fallback_test.go:54-165`.

🔵 Suggestion: The new tests intentionally wait through real 1s/2s backoff, making `internal/agent` take about 7.4s locally. This is acceptable, but future retry tests would benefit from an injectable sleeper/backoff helper to keep the suite fast. References: `internal/agent/llm.go:143-150`, `internal/agent/llm_fallback_test.go:67-70`.

✅ Highlights

The fallback loop is scoped well: terminal errors stop immediately, retryable failures exhaust a per-model budget before switching, and fallback configuration is deduped before use. References: `internal/agent/llm.go:31-43`, `internal/agent/llm.go:92-117`, `internal/agent/llm.go:125-166`.

The config parser trims comma-separated entries and drops empty values, preserving the default single-model behavior when unset or blank. References: `internal/config/config.go:159`, `internal/config/config.go:253-270`.

Verification: `go test -count=1 ./internal/agent/... ./internal/config/...` passed, as did `go test -count=1 ./internal/api/router` and `go test -count=1 ./cmd/summary-api`. Full handler/all-package tests could not link in this local environment because native library `tokenizers` was unavailable (`ld: library 'tokenizers' not found`).

yujiawei (changes requested):
# Code Review — PR #180 (octo-smart-summary)

Reviewed at head `1e05d2786ad18352d045f43d88d20a039228c817` against `main` (`a35cff5`).

Verified locally in the checkout at that exact SHA:

- `go build ./...` — OK
- `go vet ./internal/agent/... ./internal/config/... ./internal/api/...` — OK
- `go test -shuffle=on -count=1 ./internal/agent/... ./internal/config/... ./internal/api/...` — all green
- `go test -race -count=5 -run 'TestChat_|TestNewClient_' ./internal/agent/...` — green, no race reported

The shape of the change is good: the outer-model-loop / inner-retry-loop split is clean, the tri-state `(turn, retry, err)` contract is documented at each hop, terminal-vs-retryable classification is deliberate, the constructor copies and dedups the caller's slice, and the zero-config path is genuinely byte-compatible. The four new tests are readable and assert on the actual request sequence rather than just the return value.

I am requesting changes on one functional issue: **under the shipped default timeouts, the fallback model is unreachable for the timeout/unresponsive-provider failure mode.** Details below.

---

## 1. Specification compliance

The diff faithfully implements the scope the PR body declares — config field + `envStrList`, client-side fallback loop, handler/router/`main` wiring, docs, tests. I found no over-build: nothing is added beyond that knob, and I confirmed the four deferred items are genuinely absent rather than half-wired.

One mismatch:

**The PR body says `Closes #179`, but deliberately implements 1 of that issue's 5 expected improvements.** Merging will auto-close #179 while most of its acceptance criteria remain unmet — no stable error taxonomy (`rate_limited` / `upstream_5xx` / `network_error` / `invalid_response` / `timeout`), no structured logging keyed by `session_id` / `profile` / step / attempt, no SSE-vs-JSON error parity, and no test coverage for the 429 / network / bad-JSON / empty-`choices` classes that issue enumerates. The PR body is admirably explicit that these are out of scope, which makes the closing keyword look like an oversight rather than an overclaim.

Please change `Closes #179` to `Refs #179` (or `Part of #179`) so the tracking issue survives until the remaining four items land.

**Spec: ❌** (scope implemented faithfully; issue-closing keyword overstates delivery)

---

## 2. Code quality

### P1 — Fallback is unreachable under the default timeout budget

`internal/agent/llm.go:141-166` combined with `internal/agent/runner.go:96` and `internal/agent/llm.go:171`.

Every agent step is wrapped in the step budget, and that context is what `Chat` receives:

```go
// internal/agent/runner.go:96
stepCtx, cancel := context.WithTimeout(ctx, r.policy.StepTimeout)
turn, err := r.client.Chat(stepCtx, msgs, r.reg.Schemas())
```

Each attempt then takes its own, *longer-lived* sub-budget:

```go
// internal/agent/llm.go:171
reqCtx, cancel := context.WithTimeout(ctx, c.timeout)
```

With the documented defaults — `AGENT_STEP_TIMEOUT=240s`, `LLM_TIMEOUT=180s`, `maxAttempts = 3`, backoff 1s + 2s — the primary model's worst case is `180 + 1 + 180 + 2 + 180 = 543s`, which is 2.3× the entire step budget. So when the primary is *slow* rather than fast-failing, the timeline is:

- `t=0` attempt 1 → `reqCtx` = min(240, 180) → fails at `t=180`
- backoff 1s → `t=181`; attempt 2 → `reqCtx` = min(59, 180) → fails at `t=240` when the step budget expires
- backoff 2s → the `select` at `internal/agent/llm.go:147` observes the already-cancelled parent and returns `retry=false`:

```go
select {
case <-ctx.Done():
    return AssistantTurn{}, false, ctx.Err()
case <-time.After(backoff):
}
```

`Chat` sees `retry=false` at `internal/agent/llm.go:113` and returns immediately. **The fallback model is never contacted.**

I confirmed this empirically rather than inferring it. Using a hanging upstream and the same timeout *ordering* as production (per-attempt 2s, step budget 4s, i.e. `2 < 4 < 6` mirroring `180 < 240 < 543`), with one fallback configured:

```
primary_attempts=2  fallback_attempts=0  err=context deadline exceeded  seq=[primary primary]
```

Widening the per-attempt timeout relative to the budget makes it worse, not better — at the true 240/180 ratio a hanging primary can consume the whole step on a single attempt.

Why this matters: `doOnce` explicitly classifies transport/timeout failures as retryable (`internal/agent/llm.go:184`), and `CONFIGURATION.md` advertises fallback for "network/5xx/429". A slow or unresponsive provider is the most common form of the instability this knob exists to mitigate, yet for that class the knob silently does nothing — and an operator enabling it mid-incident gets no signal explaining why, because the fallback log line at `internal/agent/llm.go:98` never fires either. Fallback *does* work for fast 5xx/429, so this is a partial gap, not a dead feature.

None of the four new tests exercises a hanging request, a per-attempt timeout, or a parent deadline — all four use fast HTTP status responses — which is why CI is green on a path that does not work.

Possible directions (any one is sufficient):

1. Make the per-model loop deadline-aware: before each attempt, if `time.Until(deadline) < c.timeout` and unused fallbacks remain, stop retrying the current model and escalate now, reserving budget for the fallback.
2. Partition the remaining budget across the remaining models rather than giving every attempt the full `c.timeout`.
3. At minimum, enforce and document the invariant that fallback requires `maxAttempts*LLM_TIMEOUT + 3s <= AGENT_STEP_TIMEOUT`, log a startup warning when `LLM_FALLBACK_MODELS` is set but the budgets cannot accommodate it, and add a test that drives a hanging upstream.

Please also add a client-layer test for the timeout path regardless of which direction you pick.

### P2 — `Chat` escalates to the next model after the parent context is already dead

`internal/agent/llm.go:184` marks *every* transport error retryable, including parent cancellation:

```go
resp, err := c.http.Do(req)
if err != nil {
    // 网络层错误值得重试。
    return AssistantTurn{}, true, fmt.Errorf("http do: %w", err)
}
```

If cancellation lands on the final attempt, `chatOneModel` returns `retry=true` (`internal/agent/llm.go:166`), and `Chat` escalates — because it never re-checks `ctx.Err()` before the next iteration. Observed with a fast-failing upstream, two fallbacks, and a deadline landing on the third primary attempt:

```
err=context deadline exceeded  seq=[primary primary primary fb1]
```

One wasted upstream request against a doomed context, plus a log line claiming `falling back from "primary" to "fb1" after retries exhausted` when the real cause was the caller's deadline. Two cheap fixes, ideally both: distinguish `context.Canceled` / `context.DeadlineExceeded` from transient transport failures before setting `retry=true`, and add `if ctx.Err() != nil { return AssistantTurn{}, ctx.Err() }` before advancing to the next model in `Chat`.

### P2 — Fallback log line emits the raw upstream response body, now on successful requests

`internal/agent/llm.go:98`:

```go
log.Printf("[llm] falling back from %q to %q after retries exhausted: %v", models[modelIdx-1], model, lastErr)
```

`lastErr` wraps `fmt.Errorf("http status %d: %s", resp.StatusCode, string(body))` (`internal/agent/llm.go:191`) — the complete, unbounded upstream body. Previously that body only reached the log when the whole request failed (`internal/api/handler/agent_chat.go:366` / `:591`). Now it is logged on the **success** path too: primary fails, fallback succeeds, the client gets a 200, and the gateway's raw error payload has been written to stdout. Given #179's own criterion about not exposing upstream response bodies, consider logging status code + a truncated/classified error instead of `%v` on the full wrapped error.

### P2 — Fallback is confined to one gateway and one credential

`chatOneModel` varies only the `model` field; `c.apiURL` and `c.apiKey` are fixed for every model (`internal/agent/llm.go:171-177`). The documented example `claude-sonnet-4-6,qwen3.6-max` therefore only works if a single OpenAI-compatible gateway fronts both families under the same key. That is a real architectural constraint on how far this can degrade — falling back to a genuinely independent provider (a different endpoint, different key) is out of reach, so a gateway-level outage takes down primary and all fallbacks together. Worth stating explicitly in `CONFIGURATION.md` so operators do not plan capacity around cross-provider redundancy this cannot deliver.

### P2 — No stickiness: every step re-probes a known-dead primary

`Chat` always starts at `c.model`, and the client is rebuilt per request (`internal/api/handler/agent_chat.go:110`), so nothing remembers that the primary just failed. With `MaxSteps: 8` (and 15/20 for other profiles, `internal/agent/profile.go:57-79`), a persistently failing primary costs 3 wasted upstream calls plus 3s of pure backoff *per step* before each step re-discovers the fallback. Not a correctness bug, but on a 20-step profile that is 60 doomed requests and a minute of sleeping. A per-client sticky "current model" (or a short-lived breaker) would make the feature much cheaper during a real outage.

### P2 — Error-string contract changed on the single-model path

`internal/agent/llm.go:117` and `:166` add two new wrappers, so retry exhaustion with no fallback configured now reads `chat failed after exhausting 1 model(s): chat failed after 3 attempts on model "x": ...` where it previously read `chat failed after 3 attempts: ...`. Behaviour is compatible, but anything matching on that text (log alerts, dashboards) will regress silently. `TestChat_NoFallbackConfiguredKeepsSingleModelBehavior` asserts the request count but not the error contract, so it would not catch this. Mentioning it so the change is a decision rather than a surprise; the extra `1 model(s)` wrapper is arguably worth suppressing when no fallback is configured.

### P2 — `buildRunner`'s stale docstring, and dead code the PR grew

`internal/api/handler/agent_runner_builder.go:13` claims:

```go
// This is a shared helper used by both AgentChatHandler and AgentSummaryHandler.
```

Neither uses it. `buildRunner` has zero callers (`AgentChatHandler` has its own `buildRunnerForProfile` at `internal/api/handler/agent_chat.go:90`), and `AgentSummaryHandler.runnerFactory` / `refineRunner` (`internal/api/handler/agent_summary.go:53-61`) are likewise declared and never called. Pre-existing, not introduced here — but this PR extends the dead function's signature and adds documentation to it. Prefer deleting it, or at least correcting the comment, rather than growing it.

### Checked and clean

For the record, these were examined and found sound — no action needed:

- `models := append([]string{c.model}, c.fallbackModels...)` (`internal/agent/llm.go:93`) allocates a fresh backing array (len==cap==1), so it cannot alias or clobber `c.fallbackModels`. `Chat` mutates no client state, so the client stays safe for concurrent use.
- `envStrList` (`internal/config/config.go:253`) handles unset, `""`, `" "`, `","` and `"a,,b"` correctly, returning `def` for the effectively-empty cases and preserving order otherwise. Only `nil` is ever passed as `def` today, making that parameter speculative, but the logic is right.
- Constructor dedup (`internal/agent/llm.go:31-49`) correctly drops empty entries, intra-list duplicates, and the primary model, and copies the caller's slice.
- Both production `agent.NewClient` call sites receive the fallback list; `AgentSummaryHandler` omitting it is inert because it makes no LLM call on its active route.
- The `seen` slices in the new tests are unsynchronized, but accesses are ordered by the request/response cycle — `-race -count=5` reported nothing, so I am not treating this as a defect.

---

## 3. Verdict

**CHANGES_REQUESTED** — the P1 above (fallback unreachable for the timeout failure mode under shipped defaults, and untested on that path) should be resolved or explicitly documented as a known limitation before this lands. The P2 items are advisory; the `Closes #179` → `Refs #179` edit is a one-word fix worth folding into the same push.

To be clear about what is *not* being asked: the scope decision is right, the deferral of the other four #179 items is right, and the implementation of the declared scope is solid.

---

## 4. Coverage note

What I could not verify, so it should not be read as cleared:

- **Deployment reality of the timeout budgets.** The P1 arithmetic uses the documented defaults. If production overrides `LLM_TIMEOUT` low enough that `3*LLM_TIMEOUT + 3s <= AGENT_STEP_TIMEOUT` (e.g. `LLM_TIMEOUT=60`), fallback becomes reachable for timeouts too. I have no visibility into the deployed env, which is exactly why I suggest making the invariant explicit rather than implicit.
- **Whether one gateway actually serves both documented example models** under a single key — the basis of the same-gateway P2.
- **Tool-calling fidelity across models.** Nothing checks that a fallback model supports the OpenAI `tools` / `tool_choice` shape this client sends (`internal/agent/llm.go:121-137`). A fallback that ignores tools returns content with no `tool_calls`, which the runner may read as a final answer — silent quality degradation of a different kind than the log line is designed to surface. Not exercised by any test here.
- **The SSE / streaming path end to end.** I reasoned about it from the shared `Chat` entry point but drove no live streaming request; #179's SSE-vs-JSON parity item is deferred anyway.
- **Behaviour against a real unstable gateway** (429 with `Retry-After`, partial responses, mid-stream resets). All my evidence comes from local `httptest` upstreams.
- Files outside the eight touched by this diff were read only where the data flow led (`runner.go`, `profile.go`, `agent_summary.go`, `service/llm.go`, worker startup); I did not audit the repository broadly.

wanyiwang06:
@yujiawei — thanks for the deep review at `1e05d278`. All P1 and P2 items are now addressed at head `2903f46`. Point-by-point:

- **P1 (fallback unreachable under tight step budgets):** fixed via deadline-aware early escalation in `chatOneModel`. Before every RETRY (never the first attempt) the loop compares `time.Until(deadline)` with `c.timeout` and escalates `retry=true` when the remaining budget cannot fit another full-timeout attempt AND a fallback is configured. New `TestChat_HangingPrimaryEscalatesEarlyToFallback` reproduces your hanging-upstream scenario with 4s parent deadline / 2s per-attempt timeout and asserts exactly one primary attempt then a successful fallback contact.
- **P2 (`Chat` escalates after parent ctx dead):** fixed. `Chat` now re-checks `ctx.Err()` before each outer-loop iteration; `doOnce` distinguishes parent-context cancellation from transport errors before returning `retry=true`. New `TestChat_ParentContextCancelledIsNotMasqueradedAsFallback` covers this.
- **P2 (log line leaks upstream body):** fixed via `truncateForLog(lastErr.Error(), 200)` in the fallback log statement. Response bodies no longer land in stdout on the success path.
- **P2 (same gateway / same credential constraint):** now documented in CONFIGURATION.md.
- **P2 (single-model error contract):** fixed. When `LLM_FALLBACK_MODELS` is empty, `Chat` returns `chatOneModel`'s error verbatim (no `exhausting N model(s)` wrapper). Existing log alerts that match on `chat failed after N attempts` continue to work.
- **P2 (`buildRunner` dead code + docstring):** file `internal/api/handler/agent_runner_builder.go` deleted; `buildSummaryRegistryWithUID` lives on `AgentChatHandler` in `agent_chat.go` and remains the only used copy.
- **P2 (`Closes #179` → `Refs #179`):** PR body updated. The four other #179 items (error taxonomy, structured logging, SSE parity, handler-layer failure-mode tests) will land as follow-up PRs so this one stays reviewable.
- **P2 (no stickiness across steps):** deliberately deferred. Not a correctness regression, and a sticky "current model" (or short-lived breaker) has its own state-lifecycle test surface. Happy to open the follow-up separately.

Please re-review `2903f46`. Verified locally: `go build ./...`, `go vet ./...`, and the full agent + handler + config test suite are all green. Race sweep requires cgo which is not available in my local sandbox, but CI runs `Test (race, cgo)` — the previous head passed it, and this revision only touches two functions and their tests. Happy to iterate if the race sweep on CI catches anything.

lml2468 (approved):
# PR Review — PR #180 (Mininglamp-OSS/octo-smart-summary)

🔄 Re-review at head SHA `2903f468f372` against merge-base `a35cff54` (base `main`, now includes #170). Re-anchored from my APPROVE at `1e05d278` — this head fixes the **fallback-reachability 🔴 yujiawei raised, which I missed**.

**Delta since `1e05d278`: 4 files** — commit "make fallback reachable under tight step budgets" (llm.go +92, +174 tests, agent_runner_builder.go deleted, doc).

- `Build`, `Test`, `Test (race, cgo)` (1m47s), `Vet` → **pass**; `check-sprint` (governance) → fail.
- Local: build/vet clean; `go test ./internal/agent ./internal/api/handler` → pass.

## Correcting the record
At `1e05d278` I verified the retry/fallback *control flow* in isolation and approved — but yujiawei correctly caught that under the shipped defaults it was **structurally unreachable**: each primary attempt can consume up to `LLM_TIMEOUT=180s`, and the outer `AGENT_STEP_TIMEOUT=240s` stepCtx (#168) would cancel mid-budget, so the primary never reached "retryable-budget-exhausted" (`retry=true`) — ctx-cancel made it terminal → fallback never fired. Worse, that defeats the exact *hanging-upstream/timeout* case (#178's prod 500 was "context deadline exceeded") the feature exists to mitigate. I'd even noted the latency-multiplication as a 🟡 but missed the inverse consequence — the outer bound *starving* the fallback. Fair, sharp catch by yujiawei.

## 2. Code quality — Approved

The fix is correct and does not reintroduce the cancel-storm risk:

- **Deadline-aware early escalation (the P1 fix):** before each *retry* (never the first attempt — one good-faith primary try guaranteed), if `hasFallback && time.Until(deadline) < c.timeout`, `chatOneModel` bails with `retry=true` so the outer loop moves to the fallback while budget remains (`llm.go:206-216`). The fallback then runs with whatever budget is left (bounded by the outer stepCtx) — it can't manufacture more time, but it's now **reachable**, which was the defect.
- **Cancel-storm still prevented:** the outer `Chat` loop now checks `ctx.Err()` **at the top, before each model** (`llm.go:104-111`). So even though the backoff `select` now returns `retry=true` on `ctx.Done()`, a *truly* cancelled parent short-circuits at the top of the next iteration → no wasted fallback request. I traced both cancel paths (backoff-cancel and http-do-cancel) — both terminate cleanly via this guard.
- **Parent-cancel vs local-timeout disambiguated:** `doOnce` now returns `retry=false` (terminal, surfaces cancel verbatim) when `ctx.Err() != nil`, but a per-attempt `c.timeout` expiry on the child ctx (parent alive) stays `retry=true` (`llm.go:255-262`). Correct: a local timeout is retryable, a caller cancel is not masqueraded as an upstream retry.
- **Single-model contract preserved:** `len(models)==1` returns `lastErr` unwrapped (log/dashboard patterns matching "chat failed after N attempts" keep working); multi-model adds the wrapper. `hasFallback=false` skips the deadline check → single-model behavior byte-for-byte unchanged.
- **Dead-code removal:** `agent_runner_builder.go` (the unused `buildRunner`/`buildSummaryRegistryWithUID`) deleted wholesale — grep confirms no live caller; build passes.
- **Log hygiene:** `truncateForLog` caps upstream error bodies at 200 runes in the fallback log line (avoids gateway-payload leakage into stdout).

- **Reachability:** `TestChat_HangingPrimaryEscalatesEarlyToFallback` — a slow primary now escalates to the fallback within budget (the P1 fix, test-locked).
- **Cancel not masqueraded:** `TestChat_ParentContextCancelledIsNotMasqueradedAsFallback` — a genuine parent cancel surfaces as cancel, not a fallback attempt.
- **No storm on true cancel:** outer-loop `ctx.Err()` guard traced across both cancel entry points.
- **Prior matrix intact:** failover-on-exhaustion, terminal-no-escalate, no-config-regression, dedup all still pass.

### Non-blocking notes
- 🟡 **Stale comment:** `agent_summary.go:47` still says "newRunner falls back to buildRunner" after `buildRunner` was deleted — comment-only, no compile impact; update it.
- 🟡 The fallback receiving only *remaining* budget (not a fresh full `c.timeout`) means a fallback can still be time-starved if the primary consumed most of the step budget on its one good-faith attempt. This is inherent to the outer-ctx bound; the real lever is tuning `AGENT_STEP_TIMEOUT` ≥ `(1 + n_fallbacks) × per-attempt` if operators want each model a full budget. Worth a doc note (yujiawei's #179 error-taxonomy follow-up territory).

## 3. Test assessment
- The two new tests target exactly the fixed behaviors (early escalation under tight budget; cancel-not-masqueraded), and `TestTruncateForLog` covers the log-hygiene helper. Combined with the retained decision-matrix tests, coverage is strong and would fail against the pre-fix reachability gap.

## 4. Coverage / blind spots
- Executed: build, vet, agent+handler tests, **live CI** (race+cgo green), full trace of the deadline-escalation + both ctx-cancel classifications + the outer guard, dead-code-removal confirmation.

- Operational: set `AGENT_STEP_TIMEOUT` with headroom for at least primary + one fallback attempt if fallback success (not just reachability) matters under sustained primary hangs.
- `check-sprint` gate; and per yujiawei, change #179 from "Closes" to a reference since only 1 of its 5 items shipped.

## Verdict
**APPROVE** — re-anchored to `2903f468`. yujiawei's reachability 🔴 (fallback starved by the 240s step / 180s per-attempt hierarchy, never triggered on the timeout case it targets) is correctly fixed via deadline-aware early escalation, while the outer-loop `ctx.Err()` guard preserves cancel-safety (no fallback storm) and parent-cancel is no longer masqueraded as a retry. Single-model behavior unchanged; dead code removed; new tests lock the fix. Functional CI (incl. race+cgo) green. Credit to yujiawei — I verified the control flow but missed its interaction with the timeout hierarchy.
Suggested decision: `Safe to merge after non-blocking follow-up` — fix the stale comment + change #179 Closes→ref; merge once check-sprint clears.

Jerry-Xin (approved):
Re-review at head `2903f46` after the CHANGES_REQUESTED at `1e05d278`. Both blocking items raised in the prior review are resolved, verified against the byte-level delta and the new tests. Approving.

**✅ Blocking items from the previous review — resolved**

- 🟢 P1 (fallback unreachable under default step budget) — **FIXED**. `chatOneModel` now takes a `hasFallback` flag and, before every *retry* (never the first attempt), checks `ctx.Deadline()`: if `time.Until(deadline) < c.timeout` and a fallback model is still waiting, it escalates early with `retry=true` so the fallback gets a real request window instead of being starved by an already-dead parent context. Arithmetic under shipped defaults (`AGENT_STEP_TIMEOUT=240s`, `LLM_TIMEOUT=180s`, 3 attempts, backoff 1s/2s): a hanging primary consumes one full 180s attempt, then on the next retry the remaining ~59s budget (< 180s) trips the early-escalation guard, handing the fallback a genuine attempt window. `TestChat_HangingPrimaryEscalatesEarlyToFallback` locks this in (parent deadline 4s / per-attempt 2s): the primary is contacted exactly once, then the fallback receives the request and succeeds. This is a real functional fix, not a doc-only note.

- 🟢 P2 (`Closes #179` overstates delivery) — **FIXED**. The effective linkage in the PR body is now `Refs #179 (partial: fallback knob only ...)`. GitHub's `closingIssuesReferences` for this PR is now empty, so merging will **not** auto-close #179. The remaining "#179" occurrences in the body are prose / a quoted change-log entry, not active closing keywords.

**Also re-confirmed the original semantics still hold at the new head**
- Escalation only on retryable-exhaustion (5xx/429/network); terminal 4xx/decode/empty-choices short-circuit without touching the fallback (`TestChat_NoFallbackOnTerminalError`).
- Constructor dedups the primary model, empty strings, and intra-list dupes (`TestNewClient_DedupesFallbackList`).
- Empty `LLM_FALLBACK_MODELS` preserves single-model behavior and now also preserves the single-model error contract verbatim, no multi-model wrapper (`TestChat_NoFallbackConfiguredKeepsSingleModelBehavior`).
- Parent-context cancellation is surfaced verbatim and no longer masquerades as a fallback transition (`TestChat_ParentContextCancelledIsNotMasqueradedAsFallback`).
- Dead `buildRunner` / `agent_runner_builder.go` removed cleanly.

**💬 Non-blocking**

- 🟡 The fallback transition log at `internal/agent/llm.go:126` uses `truncateForLog(lastErr.Error(), 200)`. For HTTP failures, `lastErr` is built as `fmt.Errorf("http status %d: %s", status, string(body))` at `internal/agent/llm.go:271`, so a short (≤200-char) upstream gateway body can still reach stdout on the fallback path. `truncateForLog` bounds length but does not redact content. This is strictly better than the prior head (which logged the full `%v`), and it fires only on a fallback transition, so I'm not treating it as a merge blocker — but a follow-up that logs a structured summary (status code + retry category + model + attempt count) instead of the raw error string would fully close the body-leakage concern.
- 🔵 The `LLM_FALLBACK_MODELS` budgeting note in `CONFIGURATION.md` is slightly hard to parse; a one-line example of the deadline math (attempts × timeout + backoffs ≤ step timeout) would help operators size it.

**Note on CI:** `check-sprint` is failing because the repo policy wants a `Closes #<issue>` linked to a board sprint — that is a direct consequence of (correctly) switching to `Refs #179`, and is a maintainer/board action, not a code defect. `Test (race, cgo)` was still pending at review time; the non-race `Build`/`Test`/`Vet`/`Lint` checks are green, and `go test ./internal/agent/...` passes locally (7 tests, including the two new ones).

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #180 (octo-smart-summary)

## Summary

This PR introduces cross-model fallback for the agent LLM client via a new `LLM_FALLBACK_MODELS` environment variable. When the primary model exhausts its per-model retry budget (3 attempts with exponential backoff on network/5xx/429 errors), `Chat` now iterates through an ordered list of fallback models, each getting a fresh retry budget. A deadline-aware escalation mechanism ensures the fallback gets a real time budget instead of being starved by a hanging primary — the P1 fix from issue #179. Terminal errors (4xx, decode failures, empty choices) correctly skip fallback since they signal caller-side problems. The wiring chain is complete from `LLM_FALLBACK_MODELS` env through `config.Load()` → `SetupPublic` → `NewAgentChatHandler` → `agent.NewClient` → `Chat`. The deleted `internal/api/handler/agent_runner_builder.go` had no remaining callers.

## Verification

Local checks at head `2903f468`:
- `go build ./...` — clean
- `go test ./internal/agent/ -run "TestChat_Failover|TestChat_NoFallback|TestNewClient_Dedupes|TestChat_Hanging|TestChat_ParentContext|TestTruncateForLog" -v` — 7/7 pass (9.5s)

- ✅ **Wiring chain complete** — `cfg.LLMFallbackModels` flows from `cmd/summary-api/main.go:120` → `router.SetupPublic` → `handler.NewAgentChatHandler` → `agent.NewClient`. Grep confirms exactly one caller at each hop; no stale call sites.
- ✅ **Deleted file safe** — `agent_runner_builder.go` had `buildRunner` (package-level) and `buildSummaryRegistryWithUID` (package-level). Zero remaining callers for `buildRunner(`; `buildSummaryRegistryWithUID` survives only as a method on `AgentChatHandler` at `internal/api/handler/agent_chat.go:117`.
- ✅ **Deadline-aware escalation correct under shipped defaults** — With AGENT_STEP_TIMEOUT=240s and LLM_TIMEOUT=180s: first attempt consumes ~180s, remaining ~60s < 180s triggers escalation before attempt 2. Fallback gets ~60s of real budget. Verified by `TestChat_HangingPrimaryEscalatesEarlyToFallback` (parent deadline 4s, timeout 2s → 1 primary attempt then escalate).
- ✅ **Context cancellation surfaced cleanly** — `Chat` re-checks `ctx.Err()` at the top of each model iteration before firing any HTTP request or log line. `doOnce` distinguishes parent ctx cancel (non-retryable) from local timeout (retryable) via `ctx.Err()` vs `reqCtx` cancellation. Verified by `TestChat_ParentContextCancelledIsNotMasqueradedAsFallback`.
- ✅ **Terminal errors skip fallback** — 4xx (non-429), JSON decode errors, and empty choices all return `retry=false` from `doOnce`, which propagates through `chatOneModel` → `Chat` as an immediate return without fallback. Verified by `TestChat_NoFallbackOnTerminalError`.
- ✅ **Backward compatibility preserved** — nil/empty `fallbackModels` produces an empty internal slice; `Chat` iterates only `[c.model]`. Error format for single-model case preserves the original `"chat failed after 3 attempts"` shape. Verified by `TestChat_NoFallbackConfiguredKeepsSingleModelBehavior`.
- ✅ **`envStrList` input shape edges** — handles unset env (returns `nil`), empty string (returns `nil`), whitespace-only entries (trimmed and dropped), comma-only `" , , "` (returns `nil`). `NewClient` further dedupes and drops primary-model duplicates.
- ✅ **CONFIGURATION.md accurate** — documents comma-separated format, same-gateway constraint, budget math (`maxAttempts*LLM_TIMEOUT + backoffs <= AGENT_STEP_TIMEOUT`), and empty-preserves-single-model. All match code behavior.
- ✅ **`log.Printf` consistent** — project uses stdlib `log` throughout (47 files import it). New log line at `internal/agent/llm.go:127` follows the `[llm]` prefix convention used elsewhere.

## Findings

No P0/P1 issues. One P2 item below.

### P2 — No multi-fallback chain test (`internal/agent/llm_fallback_test.go:58`)

All six fallback tests exercise exactly one fallback model. No test covers the production contract when `LLM_FALLBACK_MODELS=model-a,model-b` is configured and both the primary and first fallback exhaust their budgets before the second fallback succeeds. The outer `Chat` loop is straightforward, but the per-model fresh-budget contract is non-trivial — a future refactor could accidentally share the retry budget globally (e.g. moving `maxAttempts` to `Chat` instead of `chatOneModel`) without any existing test catching it.

Suggested addition: a `TestChat_MultiFallbackChain` where primary 503s forever, fallback1 503s forever, fallback2 succeeds, asserting the exact request sequence is 3×primary, 3×fallback1, 1×fallback2.

## Data-Flow Trace

- `LLM_FALLBACK_MODELS` env → `config.envStrList("LLM_FALLBACK_MODELS", nil)` at `internal/config/config.go:159` → `Config.LLMFallbackModels []string`
- `cfg.LLMFallbackModels` → `router.SetupPublic(...)` parameter at `internal/api/router/router.go:19` → `handler.NewAgentChatHandler(...)` at `:142` → stored in `AgentChatHandler.llmFallbackModels` at `internal/api/handler/agent_chat.go:74`
- Per-request: `buildRunnerForProfile` at `:110` → `agent.NewClient(..., h.llmFallbackModels)` → constructor dedupes into `Client.fallbackModels` at `internal/agent/llm.go:31`
- `Chat` at `:115` builds `models := append([]string{c.model}, c.fallbackModels...)`, iterates with `chatOneModel` per model, each with fresh 3-attempt budget
- Step context: `runner.go:96` wraps `ctx` with `context.WithTimeout(ctx, r.policy.StepTimeout)` → passed as parent ctx to `Chat` → `chatOneModel` uses this deadline for early escalation

## Blind-Spot Checklist

- **C1 (dual-path parity)** — N/A. No symmetric add/remove or subscribe/unsubscribe paths changed. The fallback feature is additive; single-model path is preserved.
- **C2 (control-flow ordering)** — Clear. `chatOneModel` is called only from `Chat`'s linear loop. No nesting, re-entry, or double-firing. Each model iteration is independent.
- **C3 (auth boundary)** — Clear. All fallback models use the same `c.apiURL` and `c.apiKey`. No new endpoints or auth paths are exposed.
- **C4 (auth lifecycle)** — N/A. No auth changes.
- **C5 (build ≠ runtime correct)** — Clear. Build passes; test suite validates runtime behavior including timing-sensitive deadline escalation. Wiring chain verified end-to-end.
- **C6 (governance/docs self-consistency)** — Clear. CONFIGURATION.md accurately describes the constraint (same gateway, budget math). No contradiction with other docs.

## Cross-Round Blocker Check

## Verdict: APPROVED

Well-designed, well-tested implementation of cross-model fallback. The deadline-aware escalation correctly addresses the P1 from issue #179 (hanging primary starving fallback of budget). The wiring chain is complete, backward compatibility is preserved, and the 7 new tests cover the key behavioral contracts. The only gap is a multi-fallback chain test (P2), which is low-risk given the straightforward loop iteration. Not a blocker.

[Octo-Q] verdict: APPROVE — no P0/P1 found; one P2 test-coverage gap is non-blocking.

yujiawei (approved):
# Code Review — PR #180 (octo-smart-summary)

Re-review at head `2903f468f3724cde98441c7704fe9d907409252a` against merge-base `a35cff5`. This is round 2; the previous round requested changes at `1e05d278`.

Verified locally in a fresh checkout at this exact SHA:

- `go build ./...` — OK
- `go vet ./...` — OK
- `go test -count=1 -race ./internal/agent/...` — ok (10.8s)
- `go test -count=1 -race ./internal/api/handler/... ./internal/config/...` — ok
- `go test -count=5 -race -run 'TestChat_HangingPrimaryEscalatesEarlyToFallback'` — 5/5 pass, 3.00s each

CI at this head: `Build`, `Lint`, `Test`, `Test (race, cgo)`, `Vet`, dependency/secret/osv scans all pass. `check-sprint` fails (governance gate, not code).

This closes the coverage gap the PR body flagged — `-race` was not runnable in the author's alpine sandbox, but it is green both locally and in CI at this revision, so that caveat can be dropped.

---

## 1. Specification compliance

Every item from the previous round is accounted for:

| Previous finding | Status at this head |
|---|---|
| **P1** fallback unreachable under default budgets | **Fixed** — deadline-aware early escalation, `internal/agent/llm.go:205-224`, covered by `internal/agent/llm_fallback_test.go:198` |
| P2 `Chat` escalates after parent ctx is dead | **Fixed** — `ctx.Err()` re-check at `internal/agent/llm.go:118-124`; `doOnce` now classifies parent cancellation as terminal at `internal/agent/llm.go:259-263`; test at `internal/agent/llm_fallback_test.go:281` |
| P2 log line leaks upstream body | **Partial** — see Q1 |
| P2 same gateway / same credential constraint | **Fixed** — `CONFIGURATION.md:19` |
| P2 no stickiness across steps | **Deferred with rationale** — accepted, it was advisory |
| P2 single-model error contract changed | **Partial** — see Q2 |
| P2 `buildRunner` dead code | **Fixed** — file deleted; build/vet/tests green, see the rejected-claims section below |
| P2 `Closes #179` → `Refs #179` | **Fixed** |

The chosen P1 fix is option 1 of the three offered, implemented as described — no deviation. I checked for over-build specifically: no new env knob, struct field, endpoint, or exported symbol appears beyond what the review asked for. `truncateForLog` is the only new helper and it is directly traceable to the leakage item.

**Spec: ✅**

---

## 2. Code quality

No P0 or P1. Eight advisory items, roughly in the order I would fix them.

### Q1 — Truncation caps the leak's length but does not remove the leak

`internal/agent/llm.go:126-128`:

```go
log.Printf("[llm] falling back from %q to %q after retries exhausted: %s",
    models[modelIdx-1], model, truncateForLog(lastErr.Error(), 200))
```

`lastErr` is built at `internal/agent/llm.go:272` as `fmt.Errorf("http status %d: %s", resp.StatusCode, string(body))` — the full upstream body. 200 runes of `http status 503: {"error":{...}}` is mostly body. And the transport-error variant carries the gateway URL unconditionally; this is the actual line my test run emitted:

```
[llm] falling back from "hanging-primary" to "healthy-fallback" after retries exhausted: http do: Post "http://127.0.0.1:35937/chat/completions": context deadline exceeded
```

If the goal is keeping gateway payloads out of stdout, length-capping is the wrong instrument. Log the status code plus a classified reason — which is #179's first bullet anyway, so this converges with planned work rather than adding scope.

### Q2 — The single-model error string still changed, and nothing pins it

The outer `exhausting 1 model(s)` wrapper is correctly suppressed (`internal/agent/llm.go:153`), but the inner string at `internal/agent/llm.go:238` is:

```go
return AssistantTurn{}, true, fmt.Errorf("chat failed after %d attempts on model %q: %w", maxAttempts, model, lastErr)
```

Pre-PR it was `chat failed after %d attempts: %w`. So `on model "x"` is now inserted before the colon. The PR body states existing alerts on `chat failed after N attempts` "continue to match" — true for an unanchored substring, false for anything anchored as `chat failed after \d+ attempts:`. `TestChat_NoFallbackConfiguredKeepsSingleModelBehavior` (`internal/agent/llm_fallback_test.go:136-154`) asserts the attempt count and `err != nil` but never the string, so nothing catches this. Either restore the exact prefix or assert the intended one in the test so it becomes a decision rather than drift.

### Q3 — Under shipped defaults the fallback's budget may be too thin to use, and the documented invariant is unsatisfiable

Two coupled problems.

The deadline check at `internal/agent/llm.go:216` sits *after* the backoff sleep at `internal/agent/llm.go:193-198`, so the loop pays 1s, then decides it cannot afford another attempt. Timeline at `AGENT_STEP_TIMEOUT=240s` / `LLM_TIMEOUT=180s` with a hanging primary: attempt 1 burns 180s, backoff burns 1s, escalation fires at t≈181s, and the fallback inherits ~59s — less than a third of its own configured 180s timeout. A healthy fallback that answers in 60s is cancelled by the parent.

Meanwhile `CONFIGURATION.md:19` tells operators to keep `maxAttempts*LLM_TIMEOUT + backoffs (3s) <= AGENT_STEP_TIMEOUT`. At the shipped defaults that is `3*180 + 3 = 543 <= 240`, which is false. Every default deployment violates the invariant this revision just introduced, and the doc offers no compliant worked example.

Please fix the doc half in this PR — restate it as the actual code behavior plus a concrete compliant setting (e.g. `LLM_TIMEOUT=90` with `AGENT_STEP_TIMEOUT=240` leaves ~150s for one fallback). Moving the deadline check above the `select` is a two-line change that recovers 1–3s of the fallback's budget for free.

**Severity note, recorded rather than smoothed over:** one of the two independent analysis passes I cross-checked against rated this blocking. I disagree and am calling it advisory, on this basis: before this PR the fallback was reached in 0% of hanging-primary cases and it is now reached with ~59s, so no budget is *lost* relative to `main`, and the zero-config path is byte-identical. The code cannot manufacture budget that `AGENT_STEP_TIMEOUT` does not contain — the real fixes are operator config or a budget-partitioning redesign, which is follow-up material. But an incomplete improvement shipping with a doc whose arithmetic does not hold is worth correcting now.

### Q4 — The all-slow path returns a bare context error with no diagnostic content

`internal/agent/llm.go:259-263` returns `ctx.Err()` unwrapped when the parent is dead; `chatOneModel` passes it through at `:232` and `Chat` at `:150`, and no log line fires on that path. So the most likely production failure — primary slow, fallback also unable to finish inside the remainder — surfaces as bare `context deadline exceeded` with no record of which models were attempted or how many attempts each received. Pre-PR the comparable scenario produced `chat failed after 3 attempts: ...`.

Since #179's second acceptance criterion is exactly server-side diagnosability, this moves the wrong way on the axis the issue cares about. Wrapping with model and attempt number fixes it at zero client-facing cost: `safeErrorDetail` matches via `errors.Is` (`internal/api/handler/agent_chat.go:690`), so the response detail stays `context deadline exceeded`.

### Q5 — Dead branch means the escalation reason is never observable

`internal/agent/llm.go:217-219`:

```go
if lastErr == nil {
    lastErr = fmt.Errorf("insufficient budget for another %s attempt on model %q", c.timeout, model)
}
```

Unreachable: this block only runs at `attempt > 0`, and reaching `attempt > 0` requires a prior `doOnce` failure that set `lastErr`. The consequence is not cosmetic — that message never appears anywhere, so the fallback log line at `:127` renders identically whether the primary exhausted all 3 attempts or escalated early after 1. During an incident those are different diagnoses and there is no way to tell them apart. Wrap `lastErr` instead of conditionally replacing it.

### Q6 — Fallback coverage stops at the planning hop

`agent.NewClient` is reached only from `internal/api/handler/agent_chat.go:110`. The LLM-backed tools each construct their own client with `cfg.LLMModel` and no fallback list:

- `internal/agent/tool_summarize_chunk.go:254`
- `internal/agent/tool_merge_summaries.go:71`
- `internal/agent/tool_narrow_channels.go:54`

For profile `summary_refine` — the exact profile in #179's reproduction — most LLM wall-clock and therefore most model-instability exposure lives in those tool calls, not in the planning hop. `CONFIGURATION.md:19` describes the knob as "models the agent client tries", which an operator will read as all agent LLM traffic. Depending on where #179's failure actually occurred, this knob may not mitigate the reported incident at all.

This was also raised against the previous head, and `CONFIGURATION.md` was edited in this revision without adding the clarification. One sentence closes it: the knob covers the agent planning call only, not the map/reduce tool LLM calls.

### Q7 — Nothing checks tool-calling fidelity of a fallback model

`chatOneModel` sends `tools` and `tool_choice: "auto"` (`internal/agent/llm.go:170-180`) and `chatResponse` reads `choices[0].message.tool_calls` (`internal/agent/llm.go:63-72`). A fallback that ignores tools, or shapes tool calls differently, unmarshals to a nil `ToolCalls`. `internal/agent/runner.go:110` then reads that as a stuck model and nudges, or the loop runs out to `max steps exceeded`.

That is silent quality degradation, and the switch log line at `internal/agent/llm.go:127` fires *before* it, so the one observability hook this feature added does not surface it. The doc's own example mixes model families. Worth stating that fallbacks must honor the same OpenAI `tools` / `tool_choice` contract.

### Q8 — Two tests assert less than their names claim

Both also flagged by the independent pass.

`internal/agent/llm_fallback_test.go:312` — the cancellation test checks only `err == nil` and that the fallback was not contacted. Its comment says Chat "must surface the cancellation", but any non-nil error satisfies the assertion, including a plain retry-exhaustion error. Add `errors.Is(err, context.DeadlineExceeded)`.

`internal/agent/llm_fallback_test.go:232` — the hanging-primary test proves the fallback was *contacted*, not that it received usable budget. With a 4s parent and 2s per-attempt timeout the fallback window is ~1s; I measured 3.00s elapsed of the 4s budget across 5 race-enabled runs. So this test structurally cannot catch Q3, and ~1s of slack is thin on a loaded runner. Widening the parent deadline and asserting on remaining budget would cover both concerns.

**Quality: Approved**

---

## 3. Overall verdict

**APPROVE**

The P1 from the previous round is genuinely fixed, not papered over: I traced the escalation arithmetic at the real 240/180 ratio and confirmed the primary is cut at t≈181s so the fallback actually runs, and the new test drives a hanging upstream rather than a fast status code — which was the specific reason CI was green on a broken path last round. Every other previous item is either resolved or explicitly deferred with a reason I accept. Zero-config behavior is unchanged, so the blast radius is limited to deployments that opt in.

Everything above is advisory. The two I would actually fold in before merge are the `CONFIGURATION.md` arithmetic in Q3 and the one-sentence scope clarification in Q6 — both are doc-only and both concern statements this revision introduced.

---

## 4. Suggestions

1. Fix `CONFIGURATION.md:19`: the invariant does not hold at defaults. Replace it with the actual behavior plus one compliant example (Q3).
2. Add one sentence to the same row: the knob covers the agent planning call, not the map/reduce tool LLM calls (Q6), and fallbacks must support the same `tools` / `tool_choice` contract (Q7).
3. Move the deadline check above the backoff `select` (Q3), and wrap `lastErr` rather than conditionally replacing it (Q5) — together roughly five lines, and they make early escalation both cheaper and visible.
4. Wrap the parent-cancellation error with model and attempt (Q4). `errors.Is` keeps the client-facing detail identical.
5. Strengthen the two assertions in Q8 while the tests are still fresh.
6. When the follow-ups land, the error taxonomy from #179's first bullet subsumes Q1, Q4, and Q5 — worth sequencing that next rather than patching the log line three times.

---

## 5. Additional findings outside the diff

**Merge-order collision with #172.** That PR also rewrites `SetupPublic`'s signature, inserting `botAuthResolver middleware.BotTokenResolver` before `workerTriggerURL`, and adds test callers of the form `SetupPublic(nil, nil, nil, nil, fixedBotResolver{}, "", 0, false, 0, nil, "", "", "", 0, 0)`. Once #180 lands, those callers are missing the `llmFallbackModels` positional argument and will not compile. Both PRs report `MERGEABLE` independently, which is not the same as "main is green after both" — whichever lands second needs updating. Not a defect in this PR.

The underlying cause is worth naming: `SetupPublic` is now a 16-positional-parameter function with a trailing variadic, so every feature that needs wiring collides with every other one. A config struct would end this class of conflict. Out of scope here.

**`check-sprint` fails at this head.** Governance gate, unrelated to the code; all functional checks pass.

---

## 6. Hypotheses I tested and rejected

Recording these so they are not re-litigated:

- **"Deleting `agent_runner_builder.go` breaks `AgentSummaryHandler` / fails to compile."** It does not. `buildRunner` had zero callers — its "shared helper used by both `AgentChatHandler` and `AgentSummaryHandler`" docstring was stale, which is what the previous round flagged. And `buildSummaryRegistryWithUID` survives as a method on `*AgentChatHandler` at `internal/api/handler/agent_chat.go:117`; a method and a package-level function may share a name in Go, which is why both coexisted before the deletion. `go build ./...`, `go vet ./...`, `-race` tests across all three affected packages, and CI `Build` are all green at this head.
- **"The backoff is 2s/4s, not the documented 1s/2s, and that destabilizes the hanging test."** It is not. `math.Pow(2, float64(attempt-1))` at `internal/agent/llm.go:193` with `attempt ∈ {1,2}` yields 1s and 2s; I isolated the expression to confirm. Empirically the hanging test measures exactly 3.00s (2s attempt + 1s backoff) on 5/5 race-enabled runs — it would be 4.00s if the backoff were 2s.

---

## 7. Coverage note

**Verified:** full diff read against merge-base `a35cff5`; build and vet; `-race` across `internal/agent`, `internal/api/handler`, `internal/config`; 5× stress of the timing-sensitive test; CI checks pulled at this head; data flow traced both directions (`Chat` → `chatOneModel` → `doOnce`, and `runner.go` `stepCtx` → `Chat` → `safeErrorDetail`); every `agent.NewClient`, `SetupPublic`, and `NewAgentChatHandler` call site enumerated; all open sibling PRs screened for file overlap on the wiring surface.

**Not verified, so not to be read as cleared:** behavior against a real unstable gateway (429 with `Retry-After`, partial responses, mid-stream resets) — all evidence comes from `httptest` upstreams; whether one gateway actually serves both example models under a single key; the *deployed* values of `AGENT_STEP_TIMEOUT` and `LLM_TIMEOUT`, which is precisely what decides how hard Q3 bites in practice; tool-calling fidelity of any real fallback model (Q7 is reasoned, not exercised); the SSE path end to end — I reasoned from the shared `Chat` entry point but drove no live streaming request, and #179's SSE parity item is deferred anyway; the contents of the failing `check-sprint` gate.

On the two independent analysis passes I cross-checked against: one was truncated mid-output by a token limit, so its later findings are unknown and I have not treated that silence as agreement; the other could not execute tests in its sandbox, so its findings are static-analysis only and I re-verified each against a live run. Both of the blocking claims they raised are the two rejected in section 6.

yujiawei:
### Addendum — correcting my own suggestion #3, plus evidence on the multi-fallback path

Verdict unchanged (**approve**). Two things I got wrong or understated, and one confirmed gap. All of this is from driving the code at head `2903f46`, not from reading it.

**1. My suggestion #3 was wrong to call the reorder "free."** I recommended moving the deadline check above the backoff `select` to "recover 1–3s of the fallback's budget for free." I implemented it and it is not free — it breaks an existing test:

```
--- FAIL: TestChat_ParentContextCancelledIsNotMasqueradedAsFallback (0.50s)
    llm_fallback_test.go:326: fallback "would-be-fallback" was contacted after
    parent context cancel (seen=[flaky-primary would-be-fallback])
```

With the check first, escalation fires on the *first* retry whenever `time.Until(deadline) < c.timeout`. In that test the remaining budget is 495ms against a 5s `c.timeout`, so it escalates at t≈5ms — while the parent is still alive — and the fallback is contacted, which the test forbids. The current ordering masks this only because the 1s backoff outlives the 500ms deadline, so `ctx.Done()` wins the `select` first.

So the reorder needs pairing with a minimum-useful-budget guard, or with comparing remaining budget against a *measured* attempt duration rather than the full `c.timeout`. Please don't take suggestion #3 as a mechanical two-line change; I'm sorry for putting it that way.

**2. There is a configuration band where the fallback is completely inert** — this sharpens Q3 from "the budget is thin" to "the escalation never runs." When `AGENT_STEP_TIMEOUT - LLM_TIMEOUT` is at or below the 1s first backoff, attempt 0 consumes the budget and the backoff sleep consumes the remainder, so the check at `internal/agent/llm.go:216` never executes. Probed with `c.timeout=3s` / parent `4s` (same ordering as 180s/240s), three models configured:

```
seq=[A]   err=context deadline exceeded (last: context deadline exceeded)
```

Only the primary is ever contacted; neither fallback runs. With `AGENT_STEP_TIMEOUT == LLM_TIMEOUT` — a natural "one timeout value" choice — attempt 0 alone consumes everything and it fails even earlier.

This is **not** triggered at the shipped defaults: 240−180 = 60s comfortably exceeds the 1s backoff, so the fallback does get its ~59s, exactly as the PR intends. It stays advisory for that reason. But it compounds the doc problem in Q3 — `CONFIGURATION.md:19` asks operators to satisfy `maxAttempts*LLM_TIMEOUT + backoffs <= AGENT_STEP_TIMEOUT`, which is `543 <= 240` at defaults and therefore unsatisfiable, so operators *will* be adjusting these two knobs, and one plausible adjustment lands them in the inert band with no startup warning and no log line. A startup check when `LLM_FALLBACK_MODELS` is non-empty would close this cheaply.

**3. Confirmed: the reorder does fix the starvation case, and the multi-fallback contract is currently correct.** With the reorder applied the same probe gives `seq=[A B]` instead of `seq=[A]` — so the concern is real and worth solving, just not by that edit alone.

Separately, I drove the multi-fallback chain that no shipped test covers (primary and first fallback both 503 forever, second fallback healthy):

```
seq=[A A A B B B C]   err=<nil>   content="reply from C"
```

Exactly 3 + 3 + 1, succeeding on the second fallback. The per-model fresh-budget contract is correct today. A `TestChat_MultiFallbackChain` asserting that sequence would be a worthwhile regression lock — every current `Chat`-driving test configures exactly one fallback (`internal/agent/llm_fallback_test.go` lines 68, 118, 147, 226, 302; only line 174 has two, and it never calls `Chat`), so a refactor that hoisted `maxAttempts` from `chatOneModel` up into `Chat` would pass the whole suite.

PR #181, opened 2026-07-29
feat(summary): allow personal bots to create owner-scoped summaries

wanyiwang06:
## Summary

- add an owner-scoped `POST /api/v1/bot/summaries` endpoint for personal Agent bots
- resolve owner, bot identity, and Space exclusively from the authenticated `bf_*` token
- require explicit authorized sources, a bounded time range, and a stable `Idempotency-Key`
- persist `trigger_type=bot` ownership metadata and dispatch through the existing Summary worker workflow
- add the schema migration and handler coverage for authorization, idempotency, validation, and replay

## Relationship to the read-only PR

Includes the complete latest head of #172, including its main-conflict resolution commit `67eec74`, and supersedes that PR. This PR can be squash-merged directly without merging #172 first.

## Safety

- does not accept caller-supplied owner, participant, bot, or Space identity
- rejects sources outside the resolved owner's permissions/Space
- does not implicitly search all owner channels
- feature remains gated by `BOT_SUMMARY_CREATE_ENABLED`

## Verification

```text
go test ./internal/api/handler ./internal/api/router ./internal/model
```

Also exercised locally through the authenticated CLI and OpenClaw against the traditional Summary worker workflow.

Closes #144

PR #151, opened 2026-07-13
feat(notify): append summary detail deep link to completed notifications (#150)

Ranwanglc:
## What & why
`Closes #150` (worker side).

The completed-summary notification now appends a browser-openable detail deep link so users can jump straight to the summary. Pairs with the octo-web landing handler (Mininglamp-OSS/octo-web #150) that consumes `${origin}/summary/detail?taskId=<id>` on boot.

## Change
- `internal/notify/notify.go`: add `Config.WebBaseURL`; `buildText` appends `查看详情：${WebBaseURL trimmed}/summary/detail?taskId=<task.ID>` after the meta lines. **Only** for `completed`; `failed` never gets it; empty `WebBaseURL` => no link (plain-text fallback).
- `cmd/summary-worker/main.go`: wire `WebBaseURL: cfg.SummaryWebBaseURL` into the notifier config.
- `internal/notify/notify_test.go`: assert empty-base omits the link; add `TestBuildText_CompletedIncludesDetailLink` and `TestBuildText_FailedOmitsDetailLink`.

## Deploy prerequisite (env vars) — worker container
- **`SUMMARY_WEB_BASE_URL=<web origin>`** (default empty). Empty => completed notifications carry **no** detail link (safe degrade, non-fatal). Set it to the web deployment origin for the link to appear and open the detail page. Trailing slash is trimmed automatically.
- (unchanged, still required for notifications at all) `SUMMARY_NOTIFY_ENABLED=true`, `SUMMARY_NOTIFY_TOKEN`, `OCTO_API_URL`.

## Tests
`go test ./internal/notify/... ./internal/config/...` → ok.

PR #185, opened 2026-08-03
refactor(pipeline): use model.Source* and model.ChannelType* constants in fetch.go

wanyiwang06:
## Summary

`#181` introduced the two source-side and IM-side enums as named constants
in `internal/model/model.go`:

```go
const (
    SourceGroup  = 1  // smart-summary API frontend
    SourceThread = 2
    SourceDirect = 3
)

const (
    ChannelTypeDM     = 1  // WuKongIM storage
    ChannelTypeGroup  = 2
    ChannelTypeThread = 5  // WuKongIM reserves 3/4 so thread jumps to 5
)
```

The bot-summary-create handler adopted these constants consistently, but
the older helpers in `internal/pipeline/fetch.go` still spelled the numeric
values inline. This PR finishes the migration inside that one file so a
reader never has to flip between "which numeric enum am I looking at right
now".

## Changes

- `mapFrontendSourceType`: `case 1/2/3` → `case model.SourceGroup/SourceThread/SourceDirect`;
  returns switched from bare 1/2/5 to `model.ChannelTypeGroup/Thread/DM`.
- `NormalizeDMChannelID`: `if channelType != 1` → `if channelType != model.ChannelTypeDM`.
- `selectedThreadChannelIDs`: `!= 5` → `!= model.ChannelTypeThread`.
- Comment cleanup in the three affected functions so the doc names the
  constants each parameter carries instead of restating the digits.

## Why this is safe — and why there is no DB migration

The constant values are unchanged. The compiled instructions are the same
switch on the same integer values. In particular:

1. **DB records are untouched.** `summary_task_sources.source_type` still
   holds `1 / 2 / 3` for group / thread / DM. Every existing row in every
   deployment (local dev, alex-test, prod) is byte-identical after this PR.
2. **API contract is untouched.** Bot POST / GET request bodies and
   responses continue to send `source_type: 1 / 2 / 3`. `octo-cli`
   summary spec and the octo-summary SKILL.md still document `1=group,
   2=thread, 3=DM` — no client update needed.
3. **octo-server side is not involved.** The WuKongIM `channel_type` enum
   `1=DM, 2=Group, 5=Thread` was already the target of the mapper; we now
   name it via `ChannelTypeDM / ChannelTypeGroup / ChannelTypeThread`
   instead of the digits, but the wire values octo-server receives are the
   same as before.
4. **Behaviour is verified byte-identical** locally with
   `go test ./internal/pipeline/... ./internal/api/handler/...`.

Migrating the DB (i.e. renumbering `source_type` in existing rows to align
with WuKongIM's `channel_type`) was explicitly considered and rejected for
this PR: it would be a breaking API change requiring coordinated updates
to `octo-web`, `octo-cli`, the `octo-summary` SKILL.md, and a data
migration across every environment, none of which delivers new user-facing
value. If we ever decide to run that unification, this PR makes the code
side considerably simpler — only the constant values change; every call
site already reads through them.

## How verified

- `go build ./...` clean
- `go vet ./...` clean
- `go test ./internal/pipeline/... ./internal/api/handler/...` all pass
- Diff limited to `internal/pipeline/fetch.go` (+23 / -17)

## Linked Spec

N/A — pure refactor, no spec or handler surface changes.

## Comprehension

The gap this closes surfaced when reading `bot_summary_create.go` (post-#181)
side-by-side with `fetch.go`: the handler used named constants, and the
pipeline helpers used bare digits, so any reader spent time confirming
"is this 1 the smart-summary group, or the WuKongIM DM?" every time.
With this PR, `fetch.go` reads exactly like the handler.

### DIFF
--- a/internal/pipeline/fetch.go
+++ b/internal/pipeline/fetch.go
@@ -254,11 +254,13 @@ func getPeerUID(channelID, selfUID string) string {
 	return parts[0]
 }
 
-// NormalizeDMChannelID converts a logical DM channel id (peerUID or peer@self)
-// into the WuKongIM storage format: the UID with the larger CRC32 hash comes first.
-// For non-DM channels (channelType != 1), returns input unchanged.
+// NormalizeDMChannelID canonicalises the WuKongIM DM channel_id ("uid_a@uid_b")
+// into the WuKongIM storage format: the UID with the larger CRC32 hash comes
+// first. For non-DM channels (channelType != model.ChannelTypeDM), returns
+// input unchanged. The channel id may be logical (peerUID or peer@self) on
+// input.
 func NormalizeDMChannelID(channelID string, selfUID string, channelType int) string {
-	if channelType != 1 {
+	if channelType != model.ChannelTypeDM {
 		return channelID
 	}
 	var a, b string
@@ -275,16 +277,20 @@ func NormalizeDMChannelID(channelID string, selfUID string, channelType int) str
 	return b + "@" + a
 }
 
-// mapFrontendSourceType maps frontend source_type to backend channelType.
-// Frontend: 1=group, 3=DM; Backend IM server: 1=DM, 2=group
+// mapFrontendSourceType maps smart-summary's frontend source_type
+// (see model.SourceGroup / SourceThread / SourceDirect) to octo-server's
+// WuKongIM channel_type (see model.ChannelTypeDM / ChannelTypeGroup /
+// ChannelTypeThread). The two enums intentionally use different numbers
+// — see the discussion in each const block in model/model.go and the
+// PR that introduced these constants (#181).
 func mapFrontendSourceType(frontendType int) int {
 	switch frontendType {
-	case 1: // frontend group -> backend group
-		return 2
-	case 2: // frontend thread -> backend thread
-		return 5
-	case 3: // frontend DM -> backend DM
-		return 1
+	case model.SourceGroup:
+		return model.ChannelTypeGroup
+	case model.SourceThread:
+		return model.ChannelTypeThread
+	case model.SourceDirect:
+		return model.ChannelTypeDM
 	default:
 		return frontendType
 	}
@@ -303,14 +309,14 @@ func sourceType(s map[string]interface{}) int {
 }
 
 // selectedThreadChannelIDs returns the channel ids of explicitly-selected thread
-// sources (frontend source_type=2). These scope the archived-thread relaxation in
-// GetUserChannels so that only threads the user actually picked can be archived;
-// auto/background summaries (no explicit sources) get an empty slice and never
-// surface archived threads.
+// sources (frontend source_type == model.SourceThread). These scope the
+// archived-thread relaxation in GetUserChannels so that only threads the user
+// actually picked can be archived; auto/background summaries (no explicit
+// sources) get an empty slice and never surface archived threads.
 func selectedThreadChannelIDs(specifiedSources []map[string]interface{}) []string {
 	var ids []string
 	for _, s := range specifiedSources {
-		if mapFrontendSourceType(sourceType(s)) != 5 {
+		if mapFrontendSourceType(sourceType(s)) != model.ChannelTypeThread {
 			continue
 		}
 		if id, ok := s["source_id"].(string); ok && id != "" {


Jerry-Xin (dismissed):
In scope for `octo-smart-summary`; the PR is a safe readability refactor with no behavior change detected.

💬 Non-blocking

🟡 Warning: The migration is not fully consistent within `fetch.go`: `GetUserChannels` still uses raw WuKongIM channel-type literals in SQL/result construction, including `channel_type = 1`, `NormalizeDMChannelID(..., 1)`, `ChannelType: 1`, `5 AS channel_type`, and `ChannelType: 5` at [internal/pipeline/fetch.go](/Users/mlamp/.openclaw/tmp/pr-review-tz357l27/repo/internal/pipeline/fetch.go:153), [internal/pipeline/fetch.go](/Users/mlamp/.openclaw/tmp/pr-review-tz357l27/repo/internal/pipeline/fetch.go:163), [internal/pipeline/fetch.go](/Users/mlamp/.openclaw/tmp/pr-review-tz357l27/repo/internal/pipeline/fetch.go:166), [internal/pipeline/fetch.go](/Users/mlamp/.openclaw/tmp/pr-review-tz357l27/repo/internal/pipeline/fetch.go:196), and [internal/pipeline/fetch.go](/Users/mlamp/.openclaw/tmp/pr-review-tz357l27/repo/internal/pipeline/fetch.go:220). This does not create a bug, but it undercuts the PR’s stated goal of removing enum ambiguity from the file.

✅ Highlights

🔵 The changed mappings in `mapFrontendSourceType` correctly preserve the existing frontend-to-storage translation: group `1 -> 2`, thread `2 -> 5`, DM `3 -> 1`.

🔵 `NormalizeDMChannelID` and `selectedThreadChannelIDs` now read through the shared model constants, matching the newer handler-side pattern and reducing the chance of confusing source enums with storage channel enums.

Verification: `go test ./internal/pipeline/...` passed. `go test ./internal/pipeline/... ./internal/api/handler/...` could not complete locally because the handler package failed to link due to missing native library `tokenizers`, unrelated to this PR.

Jerry-Xin (approved):
In scope for `octo-smart-summary`; the PR is a safe readability refactor with no behavior change detected.

💬 Non-blocking

🟡 Warning: The migration is not fully consistent within `fetch.go`: `GetUserChannels` still uses raw WuKongIM channel-type literals in SQL/result construction, including `channel_type = 1`, `NormalizeDMChannelID(..., 1)`, `ChannelType: 1`, `5 AS channel_type`, and `ChannelType: 5` at `internal/pipeline/fetch.go:153`, `:163`, `:166`, `:196`, and `:220`. This does not create a bug, but it undercuts the PR's stated goal of removing enum ambiguity from the file.

✅ Highlights

🔵 The changed mappings in `mapFrontendSourceType` correctly preserve the existing frontend-to-storage translation: group `1 -> 2`, thread `2 -> 5`, DM `3 -> 1` (verified byte-identical against the const definitions in `internal/model/model.go`: SourceGroup=1/SourceThread=2/SourceDirect=3, ChannelTypeDM=1/ChannelTypeGroup=2/ChannelTypeThread=5).

🔵 `NormalizeDMChannelID` (`!= ChannelTypeDM`) and `selectedThreadChannelIDs` (`!= ChannelTypeThread`) now read through the shared model constants, matching the newer handler-side pattern and reducing the chance of confusing source enums with storage channel enums. The guards preserve the exact prior integer comparisons (`!= 1`, `!= 5`).

Verification: `go test ./internal/pipeline/...` passed. The combined `go test ./internal/pipeline/... ./internal/api/handler/...` could not complete locally because the handler package failed to link due to a missing native `tokenizers` library, unrelated to this PR.

lml2468 (approved):
# PR Review — PR #185 (Mininglamp-OSS/octo-smart-summary)

Reviewed at head SHA `852bd9661c1f` against merge-base `1d1333d7` (base `main`). Mode: STANDARD (pure refactor).

**Intent:** replace bare integer literals in `internal/pipeline/fetch.go` with the named `model.Source*` / `model.ChannelType*` constants (introduced in #181) **so that** the pipeline helpers read like `bot_summary_create.go` and a reader no longer has to guess whether a `1` means the frontend group or the WuKongIM DM, **affecting** only `fetch.go` (+23/-17, comments + 5 literal sites). No behavior change intended.

**Verification at this head:**
- Local: `go build ./internal/pipeline/`, `go vet ./internal/pipeline/`, `go test ./internal/pipeline/` → **all pass** (1.24s). `model` already imported (line 14).
- CI: `check-sprint`, `osv-scan`, `secret-scan`, `dependency-review`, `history`, `pr-title-lint` → pass. Note: the heavy compile/test jobs (`Build`, `Test`, `Vet`, `Lint`, `Test (race, cgo)`) show **`skipping`** — the repo's changed-path filter didn't trigger them for this diff — so local build/vet/test on the affected package carries the verification weight here (done, green).

## 1. Spec compliance — ✅
Faithful, focused refactor. No spec/handler surface touched; no over-build; no deviation. The migrated comments accurately describe the two distinct enums.

## 2. Code quality — Approved
The single correctness question a literal→constant swap must answer is **do the constants equal the literals exactly?** Verified every site against `model/model.go` (SourceGroup=1, SourceThread=2, SourceDirect=3; ChannelTypeDM=1, ChannelTypeGroup=2, ChannelTypeThread=5):

| Original | Replacement | Value | ✓ |
|---|---|---|---|
| `channelType != 1` | `!= model.ChannelTypeDM` | 1 | ✅ |
| `case 1 → return 2` | `SourceGroup → ChannelTypeGroup` | 1→2 | ✅ |
| `case 2 → return 5` | `SourceThread → ChannelTypeThread` | 2→5 | ✅ |
| `case 3 → return 1` | `SourceDirect → ChannelTypeDM` | 3→1 | ✅ |
| `!= 5` | `!= model.ChannelTypeThread` | 5 | ✅ |

All five are numerically identical — **zero behavior change**. `default: return frontendType` unchanged; switch cases still match the same frontend values. Notably, `ChannelTypeThread = 5` (WuKongIM reserves 3/4) is preserved, which is the exact off-by-nothing trap this refactor de-risks for future readers.

- **No literal↔constant mismatch:** each of the 5 sites cross-checked against the const block.
- **Compiles:** `model` already imported; build/vet clean — no unused-import or undefined-symbol.
- **No logic drift:** pure value-preserving substitution; `go test ./internal/pipeline/` passes.
- **Residual literals are pre-existing, not regressions:** `NormalizeDMChannelID(..., 1)` and the `ChannelType: 1` / `ChannelType: 5` struct assignments + `5 AS channel_type` SQL (lines ~163–220) still use bare digits, but git confirms **this PR did not touch those lines** — they carry the correct values and are out of this PR's chosen scope.

### Non-blocking note
- 🟡 (consistency, optional) The channel-construction block still hardcodes `ChannelType: 1` (DM, line 166), `ChannelType: 5` (thread, line 220), and `NormalizeDMChannelID(d.ChannelID, uid, 1)` (line 163). Migrating these to `model.ChannelTypeDM` / `model.ChannelTypeThread` would fully realize the "reads like the handler" goal. Not required — the SQL `5 AS channel_type` can't take a Go const, and the current values are correct; a reasonable scope boundary.

## 3. Test assessment
- No new tests needed — behavior is provably unchanged (identical constant values), and the existing `internal/pipeline` suite passes at this head, exercising `NormalizeDMChannelID` / `mapFrontendSourceType` / `selectedThreadChannelIDs`.

## 4. Coverage / blind spots
- Executed: constant-value cross-check (model.go), build/vet/test on `internal/pipeline`, blame on the residual literals, CI status pull.
- Not run: the heavy CI jobs were path-filter-skipped; no DB-backed integration (local test uses the package's own fixtures).

- Confirm the changed-path CI filter skipping `Build`/`Test`/`Vet`/`race-cgo` for a core `internal/pipeline` `.go` change is intended repo policy — a compile-affecting file arguably should trigger them. (Not a defect in this PR; a CI-config observation.)

## Verdict
**APPROVE** — both gates pass. This is a clean, value-preserving refactor: all five migrated literals map to `model.*` constants with byte-identical values (verified against `model/model.go`), it compiles and the `internal/pipeline` tests pass locally, and no behavior changes. The one nit — a few sibling literals in the channel-construction paths left un-migrated — is pre-existing, correct, and out of scope.
Suggested decision: `Safe to merge` — pure refactor, no behavior change; just confirm the CI path-filter skip is the intended policy.

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #185 (octo-smart-summary)

## Summary

Pure mechanical refactor: replaces inline magic numbers in three helper
functions in `internal/pipeline/fetch.go` with the named constants introduced
by #181 (`model.SourceGroup`, `model.SourceThread`, `model.SourceDirect`,
`model.ChannelTypeDM`, `model.ChannelTypeGroup`, `model.ChannelTypeThread`).
No behavioral change — the compiled switch and comparison instructions are
byte-identical because the constant values match the digits they replace.

## Verification

Static analysis only at head `852bd966`; build and tests not executed in this
environment.

- ✅ **Constant-value equivalence** — verified in `internal/model/model.go`:
  `SourceGroup=1`, `SourceThread=2`, `SourceDirect=3`, `ChannelTypeDM=1`,
  `ChannelTypeGroup=2`, `ChannelTypeThread=5`. Every replacement in the diff
  maps to the same integer the old code used.
- ✅ **No residual magic numbers in changed functions** — grepped
  `internal/pipeline/fetch.go` for `case \d+:` and
  `channelType\s*[!=]=\s*\d`; zero hits inside the three refactored
  functions. The `model` import was already present.
- ✅ **Data-flow preserved** — `mapFrontendSourceType` receives frontend
  `source_type` integers (from JSON-decoded spec maps via `sourceType()`);
  the switch arms match the same inputs and return the same outputs. Callers
  at `:349` and `:784` are untouched and continue to consume the return
  value identically.
- ✅ **Comment accuracy** — updated doc-comments correctly name the
  constant enums each parameter carries; no stale references to digits.

## Findings

No P0/P1 issues. No correctness, security, or build concerns.

### Nit — Nearby call sites still use bare `1` (`internal/pipeline/fetch.go:163-166`)

The DM-fetch loop at `:163` passes `1` as `channelType` to
`NormalizeDMChannelID` and sets `ChannelType: 1` on the `ChannelInfo` struct.
These are pre-existing and outside this PR's diff scope, so not a blocker.
Worth cleaning up in a follow-up pass for full file-wide consistency.

## Data-Flow Trace

| Consumer | Data consumed | Upstream source | Verified |
|---|---|---|---|
| `mapFrontendSourceType` switch | `frontendType int` | `sourceType(s)` extracts `s["source_type"]` from JSON-decoded spec map | ✅ same int values (1/2/3) match `model.Source*` constants |
| `NormalizeDMChannelID` guard | `channelType int` | caller passes `backendChType` from `mapFrontendSourceType` or literal `1` | ✅ `model.ChannelTypeDM == 1` |
| `selectedThreadChannelIDs` filter | `!= model.ChannelTypeThread` | `mapFrontendSourceType(sourceType(s))` return value | ✅ `model.ChannelTypeThread == 5` |

## Blind-Spot Checklist

- **C1 (dual-path parity)** — N/A. No paired add/remove paths touched.
- **C2 (control-flow ordering)** — N/A. No nesting or reuse-order changes.
- **C3 (auth boundary)** — N/A. No auth/permission changes.
- **C4 (auth lifecycle)** — N/A. No auth changes.
- **C5 (build/runtime)** — N/A. Pure constant substitution; Go compiler validates type correctness.
- **C6 (governance/docs)** — N/A. No governance or policy changes.

## Cross-Round Blocker Recheck

N/A — first review pass.

## Verdict: APPROVED

Clean, well-scoped refactoring that does exactly what the title claims.
Constant values verified byte-identical. No behavioral change, no API
surface change, no DB impact.

[Octo-Q] verdict: APPROVE — pure constant rename with verified value equivalence, no behavioral change, no P1/P2 defects.

yujiawei (approved):
# Code Review — PR #185 (octo-smart-summary)

Reviewed at head SHA `852bd9661c1fbe87b9c5317ee2698546a9a331e5`, merge base `1d1333d` (current `main` tip, #181). Diff: `internal/pipeline/fetch.go` only, +23 / −17.

---

## 1. Spec compliance

**Spec: ✅**

- **Missing:** none relative to the enumerated *Changes* list — all four bullets landed exactly as described (`mapFrontendSourceType` case labels + returns, `NormalizeDMChannelID` DM gate, `selectedThreadChannelIDs` thread comparison, doc comments in the three functions).
- **Extra:** none. No new flags, fields, exported symbols, or behaviour. `model` was already imported at `internal/pipeline/fetch.go:14`, so not even an import was added.
- **Deviation:** none in the code. One inaccuracy in the PR description: the body claims *"This PR finishes the migration inside that one file"*, which is not achieved — see **P2-1**. Every enumerated deliverable did land, so this is a description accuracy issue rather than a missing deliverable, but the sentence should be softened before merge so future readers don't trust it.

### Constant-substitution audit (the thing that actually matters here)

Each replaced literal was checked against its definition:

| Site | Before | After | Constant value | Match |
|---|---|---|---|---|
| `fetch.go:288-289` | `case 1 → return 2` | `SourceGroup → ChannelTypeGroup` | `1 → 2` (`model.go:104`, `:123`) | ✅ |
| `fetch.go:290-291` | `case 2 → return 5` | `SourceThread → ChannelTypeThread` | `2 → 5` (`model.go:105`, `:124`) | ✅ |
| `fetch.go:292-293` | `case 3 → return 1` | `SourceDirect → ChannelTypeDM` | `3 → 1` (`model.go:106`, `:122`) | ✅ |
| `fetch.go:263` | `channelType != 1` | `!= ChannelTypeDM` | `1` | ✅ |
| `fetch.go:319` | `!= 5` | `!= ChannelTypeThread` | `5` | ✅ |

The `3 → 1` row is where this class of refactor usually breaks (the two enums collide on the value `1`: `SourceGroup = 1` and `ChannelTypeDM = 1`). It is correct here — `SourceDirect` maps to `ChannelTypeDM`, not to `SourceGroup`.

### Byte-identical behaviour: verified mechanically, not taken on trust

The PR asserts codegen is unchanged. I confirmed it rather than assuming it:

```
# built the pipeline test binary at origin/main and at the PR head, both with -trimpath
go tool objdump -s 'pipeline\.(mapFrontendSourceType|NormalizeDMChannelID|selectedThreadChannelIDs|GetUserChannels)$'
# diff of the two instruction streams → IDENTICAL
```

All four affected functions compile to identical machine instructions before and after. Combined with the constant table above, the "no DB migration / no API contract change" reasoning in the PR body holds: `summary_task_sources.source_type` rows and the wire values sent to octo-server are provably untouched.

Also independently re-ran: `go build ./...` clean, `go vet ./...` clean, `go test ./internal/pipeline/... ./internal/api/handler/...` pass.

### Merge-safety checks

- No other open PR touches `internal/pipeline/fetch.go` (the rest of the open set is CI dependency bumps plus #151 on the notify path), so there is no post-merge semantic conflict risk on this file.
- The rewritten `NormalizeDMChannelID` doc comment claims the input may be a bare `peerUID`. Confirmed accurate — `fetch.go:270-273` handles the no-`@` case by pairing against `selfUID`. Doc and code agree.

---

## 2. Code quality

**Quality: Approved** — no P0/P1. Three P2s, all non-blocking.

### P2-1 — The migration stops short of the file, contradicting the PR body

Four Go-level numeric enum literals survive in the same file:

- `internal/pipeline/fetch.go:879` — `if sourceType(s) != 3 { // not DM` → should be `model.SourceDirect`
- `internal/pipeline/fetch.go:163` — `NormalizeDMChannelID(d.ChannelID, uid, 1)` → should be `model.ChannelTypeDM`
- `internal/pipeline/fetch.go:166` — `ChannelType: 1,` → should be `model.ChannelTypeDM`
- `internal/pipeline/fetch.go:220` — `ChannelType: 5,` → should be `model.ChannelTypeThread`

Line 879 is the one I would actually fix. It is the *same* comparison class the PR just migrated 560 lines earlier, so the file is now internally inconsistent with itself:

```go
// fetch.go:319 — after this PR
if mapFrontendSourceType(sourceType(s)) != model.ChannelTypeThread {

// fetch.go:879 — untouched
if sourceType(s) != 3 { // not DM
```

A reader landing on 879 still has to stop and ask "is this `3` the smart-summary source enum or the WuKongIM one?" — precisely the cost the PR set out to remove. It needs a trailing `// not DM` comment to be legible, which is the tell.

Line 163 is worth a second look too: the PR rewrote `NormalizeDMChannelID`'s doc comment to describe its gate in terms of `model.ChannelTypeDM`, while the in-file caller one screen above still passes a bare `1`.

The SQL literals (`:123`, `:153`, `:196`) are a different situation — they live inside raw query strings and can't reference Go constants without `fmt`-splicing, which I would not trade for this. Leaving them is the right call; the stale comment at `:172` (`// Thread channels (channelType=5)`) could name the constant though.

### P2-2 — `default:` passthrough now reads more clearly wrong than it did before (pre-existing)

```go
// fetch.go:294-295
default:
    return frontendType
```

The new doc comment states the two enums "intentionally use different numbers", which makes returning an unmapped frontend value into a channel-type slot visibly incoherent. Tracing it end to end:

1. `POST /api/v1/summaries` (`internal/api/handler/task.go:389`) persists `s.SourceType` **with no range validation**.
2. The bot path does validate — `bot_summary_create.go:282` gates on `validFrontendSourceType` (`1..3`, `:289`).
3. So a client-supplied `source_type: 5` can reach the worker, and `mapFrontendSourceType(5)` falls through `default` returning `5`, which equals `ChannelTypeThread`.
4. `selectedThreadChannelIDs` (`fetch.go:319`) therefore treats it as an explicitly-selected thread and relaxes the archived-thread filter in `GetUserChannels`.

Blast radius is bounded: the thread query at `fetch.go:196-211` still enforces `group_member.uid = ? AND gm.is_deleted = 0`, so at worst a user surfaces an **archived** thread in a group they already belong to. No cross-tenant or cross-user exposure, and `status = 3` (deleted) remains excluded by construction.

**Not introduced by this PR and not blocking it** — the codegen is unchanged, as proven above. Raising it because this PR is the moment the code starts *claiming* the two enums are deliberately distinct. The cheap fix, in a follow-up: align the default with the handler's sibling (`default: return 0`) and/or add `validFrontendSourceType` to the `task.go` create path.

### P2-3 — Two copies of one enum mapping

After this PR, `pipeline.mapFrontendSourceType` (`fetch.go:286-296`) and `handler.frontendToIMChannelType` (`bot_summary_create.go:291-302`) are character-for-character identical except the default branch (`frontendType` vs `0`). Duplicating an enum mapping is the failure mode named constants exist to prevent, so the refactor is arguably one step short of its own goal. Consider hoisting a single `model.SourceTypeToChannelType` and having both call it — but resolve the default-branch divergence deliberately rather than silently picking one, since P2-2 shows the two defaults are not equivalent in effect.

### Nit

The commit headline says `model.SourceType*`; the constants are actually `model.Source*`. The PR title has it right.

---

## 3. Overall verdict

**APPROVE**

Spec ✅ and Quality Approved. This is a genuinely low-risk change: the substitution is value-for-value correct, and the four affected functions were verified to compile to identical instructions against `main`. The "no DB migration, no API change" argument in the description is sound and I confirmed it independently rather than accepting it.

The reasoning for *rejecting* the DB renumbering in this PR is also the right call, and the stated payoff is real — `internal/api/handler/agent_reference_context.go:256-260` documents SUM-158, where raw storage `channel_type` (`1/2/5`) was written straight into `origin_channel_type`, mis-storing DM sessions as Group and pushing Thread outside the `1..3` validation window. That is exactly the bug class bare digits produce, which is the strongest argument for landing this.

## 4. Suggested changes

Optional, none blocking:

1. Migrate `fetch.go:879` to `model.SourceDirect` — highest value per character, and it removes the need for the `// not DM` comment.
2. Migrate `fetch.go:163`, `:166`, `:220` to `model.ChannelTypeDM` / `ChannelTypeThread`, and update the `:172` comment to name the constant.
3. Soften the "finishes the migration inside that one file" sentence in the PR body, or land (1)+(2) so it becomes true.
4. Follow-up (separate PR): unify the two mappers (P2-3) and decide the `default:` contract (P2-2).

## 5. For a human to verify

This PR carries the security-sensitive label. Two items, neither caused by this diff:

- **Input-validation asymmetry (P2-2):** `task.go:389` persists a client-supplied `source_type` without the `1..3` check that `bot_summary_create.go:282` applies. Someone should decide whether to close that gap. Bounded by the group-membership check, so low severity — but it's the kind of asymmetry that gets worse quietly.
- **Nothing in this diff touches auth, tokens, credentials, or middleware.** The security label appears to be keyword-triggered rather than reflecting the change's content; the diff is a value-preserving constant substitution with machine-verified identical codegen.

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #185 (octo-smart-summary)

## Summary

This PR introduces a bot-facing API for creating personal summaries (`POST /api/v1/bot/summaries`), with a complete authorization chain: server-resolved bot identity via `/v1/auth/verify-bot`, owner-anchored source authorization (including DM peer-in-space checks), and idempotency with request-hash replay detection. The PR also refactors magic numbers in `internal/pipeline/fetch.go` to use `model.Source*` and `model.ChannelType*` constants, widens the keyword search to cover both title and topic, and strips citation context (ContextBefore/ContextAfter) for bot responses as a defense-in-depth measure.

The overall design is solid — defense-in-depth authorization, feature-gated rollout (default off), conservative idempotency replay semantics, and proper separation of migrations. Three P2 observations below; no P1 blockers.

## Verification

Static analysis only at head `852bd9661c1fbe87b9c5317ee2698546a9a331e5`; build and tests not executed in this environment.

- ✅ **Bot auth chain** — `StrictBotAuthMiddleware` requires `bf_` prefix, calls `ResolveBot`, rejects empty identity fields. Server-resolved space bypasses client header. `*auth.HTTPTokenResolver` satisfies `middleware.BotTokenResolver` interface via the new `ResolveBot` method.
- ✅ **Task access authorization** — `GetSummary` (`:884`) and `GetResult` (`:1106`) both route through `authorizeTaskAccess`, which enforces space_id match + creator/participant check. Bot routes inherit this gate via the same handler.
- ✅ **DM source authorization** — `resolveAuthorizedBotSources` derives peer UID from canonical channel ID, verifies peer holds active `space_member` row. Self-DM case handled correctly (peer == owner).
- ✅ **Idempotency** — Transaction uses `ON CONFLICT DO NOTHING` + `RowsAffected==0` for race detection. Pre-existing bindings (empty `request_hash`) are conservatively treated as mismatched (409). Re-fetch after conflict honours hash check.
- ✅ **Constant migration** — `mapFrontendSourceType` mapping verified: SourceGroup(1)→ChannelTypeGroup(2), SourceThread(2)→ChannelTypeThread(5), SourceDirect(3)→ChannelTypeDM(1). All three switch cases equivalent to the original magic numbers.
- ✅ **Citation stripping** — `citationsForRequest` applied in both `GetSummary` (`:935`) and `GetResult` (`:1141`) paths for bot requests. ContextBefore/ContextAfter nil'd to prevent raw message content leaking through citations.
- ✅ **Migration safety** — Two separate migrations (`20260728-01` table + column, `20260729-01` request_hash). Split is intentional and documented to avoid sql-migrate filename-key issues. Both have proper Down migrations.
- ✅ **Input validation** — `validateBotSummaryRequest` covers: sources required + max count, title/topic rune limits, time range validity + max duration, origin channel type/id consistency, per-source type validation. JSON body is `LimitReader`-capped (1MB), `DisallowUnknownFields`, and `ensureJSONEOF`-checked.

## Findings

No P0/P1 issues; three P2 items below.

### P2 — Fire-and-forget worker dispatch (`internal/api/handler/bot_summary_create.go:234`)

`go h.triggerWorker(...)` runs in a detached goroutine outside gin's `Recovery()` middleware. If `triggerWorker` panics (e.g. nil pointer in the HTTP call to the worker URL), the worker dispatch is silently dropped with no retry, DLQ entry, or log. The task row is persisted with `status=pending`, so it *could* be recovered by a future scheduled sweep or operator intervention — but no such recovery mechanism is visible in this PR. Consider wrapping the goroutine with a `recover()` + log, or adding a comment documenting the expected recovery path.

### P2 — Keyword search scope widened for all callers (`internal/api/handler/task.go:541`)

The change from `title LIKE ?` to `(title LIKE ? OR topic LIKE ?)` affects both the human route (`GET /api/v1/summaries`) and the bot route (`GET /api/v1/bot/summaries`) since they share `ListSummaries`. Existing API consumers that rely on title-only matching may see unexpected results (e.g. a keyword matching only a topic now returns a task it previously wouldn't). This is a behavior change on an existing endpoint, not just new functionality.

**Diff-scope check**: new behavior introduced by this PR (amplified scope on existing endpoint). Not pre-existing.

### P2 — Bot list endpoint returns all owner-visible summaries (`internal/api/router/router.go:63`)

`botV1.GET("/summaries", taskH.ListSummaries)` uses the same handler as the human route with no additional filtering. A bot calling `GET /api/v1/bot/summaries` sees every task where the owner is creator, participant, or pending invitee — not just tasks created through this bot API. If the product intent is "list my bot's summaries," consider adding a `trigger_type=4` or `creator_bot_id` filter on bot-mounted list requests. If the intent is "list everything the owner can see," the current behavior is correct but the API surface name (`/bot/summaries`) may be misleading.

## Blind Spot Checklist

- **C1 (Dual path parity)**: Bot and human route groups share `ListSummaries`, `GetSummary`, `GetResult` handlers. Both paths are gated by their respective middleware (`StrictBotAuthMiddleware` vs `StrictAuthMiddleware`+`StrictSpaceMiddleware`). Bot-specific behavior is correctly differentiated via the `bot_request` context flag. ✅ Clear.
- **C2 (Control flow ordering)**: No middleware overlap between bot and human route groups. `StrictBotAuthMiddleware` and `StrictAuthMiddleware` are never stacked. ✅ Clear.
- **C3 (Authorization boundary)**: `CreateBotSummary` enforces authorization at multiple layers: middleware (bot identity), handler (owner/space/bot context), source resolution (pipeline.GetUserChannels + peerInSpace for DMs), origin channel containment. ✅ Clear.
- **C4 (Authorization lifecycle)**: `peerInSpace` checks `space_member.status = 1` (active). If a member is deactivated, their DM sources are correctly rejected. ✅ Clear.
- **C5 (Build/runtime)**: Static analysis only; not verifiable from this environment. N/A.
- **C6 (Governance/docs)**: Not applicable — no governance/policy files changed.

## Data Flow Trace

- `ownerUID` / `spaceID` / `botUID` → server-resolved via `ResolveBot` → set in gin context by `StrictBotAuthMiddleware` → consumed by `CreateBotSummary` via `middleware.GetUserID(c)` / `middleware.GetSpaceID(c)` / `c.Get("bot_id")`. ✅ Data flows correctly.
- `sources` → `resolveAuthorizedBotSources` → `pipeline.GetUserChannels(ownerUID)` → per-source channel match + space/DM peer check → canonical IDs. ✅ Each source is verified against the owner's actual channel membership.
- `requestHash` → `canonicalBotCreateRequestHash` → deterministic from resolved sources + origin + title + topic + time range + include_archived. Stored in idempotency binding. ✅ Replay detection is sound.
- `triggerWorker` → detached goroutine with `task.ID` + `participantID` from persisted transaction. ⚠️ No failure propagation (see P2 above).

## Verdict: APPROVED

Well-structured PR with defense-in-depth authorization, conservative idempotency semantics, and clean constant migration. The three P2 items are non-blocking observations — the fire-and-forget goroutine is the most actionable of the three but doesn't rise to P1 given the persisted task row provides an operator recovery path.





[Octo-Q] verdict: APPROVE — Three P2 non-blocking observations; no P0/P1 blockers found. Defense-in-depth authorization is solid.

yujiawei (approved):
# Code Review — PR #185 (octo-smart-summary)

Reviewed at head SHA `852bd9661c1fbe87b9c5317ee2698546a9a331e5`, merge-base `1d1333d` (current `main` tip, #181). Diff: `internal/pipeline/fetch.go` only, +23 / −17.

This is an independent second pass. I did not inherit the conclusions of the existing reviews on this PR — I re-derived the constant table and independently machine-verified the "no behaviour change" claim, then went looking for what a literal→constant refactor tends to leave behind. Three of the notes below (§2.1, §2.3, §5) are new relative to what is already on the PR.

---

## 1. Spec compliance

**Spec: ✅**

- **Missing:** none. All four bullets in the *Changes* list landed exactly as described — `mapFrontendSourceType` case labels and returns, the `NormalizeDMChannelID` DM gate, the `selectedThreadChannelIDs` thread comparison, and the doc comments on those three functions.
- **Extra:** none. No new flags, fields, exported symbols, or behaviour. `model` was already imported at `internal/pipeline/fetch.go:14`, so not even an import was added.
- **Deviation:** none in the code. One prose inaccuracy in the description — see §2.2 — but every enumerated deliverable landed, so this is a description-accuracy issue rather than an incomplete spec.

### Constant-substitution audit

The single question a literal→constant swap has to answer is *do the constants equal the literals exactly?* Every site checked against `internal/model/model.go:104-106` and `:122-124`:

| Site | Before | After | Constant value | Match |
|---|---|---|---|---|
| `fetch.go:263` | `channelType != 1` | `!= model.ChannelTypeDM` | `1` | ✅ |
| `fetch.go:288-289` | `case 1 → return 2` | `SourceGroup → ChannelTypeGroup` | `1 → 2` | ✅ |
| `fetch.go:290-291` | `case 2 → return 5` | `SourceThread → ChannelTypeThread` | `2 → 5` | ✅ |
| `fetch.go:292-293` | `case 3 → return 1` | `SourceDirect → ChannelTypeDM` | `3 → 1` | ✅ |
| `fetch.go:319` | `!= 5` | `!= model.ChannelTypeThread` | `5` | ✅ |

The `3 → 1` row is where this class of refactor normally breaks, because the two enums collide on the value `1` (`SourceGroup = 1` and `ChannelTypeDM = 1`). It is correct here: `SourceDirect` maps to `ChannelTypeDM`, not to `SourceGroup`. `ChannelTypeThread = 5` (the WuKongIM 3/4 reservation) is likewise preserved.

### Behaviour: machine-verified, not taken on trust

The description asserts identical codegen. I confirmed it rather than assuming it — built the `internal/pipeline` test binary twice with `-trimpath`, once with `fetch.go` at the merge-base and once at this head, and diffed the disassembly of every affected function:

```
go tool objdump -s 'pipeline\.(mapFrontendSourceType|NormalizeDMChannelID|selectedThreadChannelIDs|GetUserChannels|ApplySourceConstraints)$'
# 1222 instruction lines each side → diff is empty
```

Instruction streams are identical. Combined with the constant table, the "no DB migration / no API contract change" reasoning holds structurally, not just by assertion: `summary_task_sources.source_type` rows and the wire values sent to octo-server are provably untouched.

Also re-ran locally at this head: `go build ./...` clean, `go vet ./internal/pipeline/` clean, `go test ./internal/pipeline/...` pass.

### Doc-comment accuracy

The rewritten `NormalizeDMChannelID` comment claims the input may be a bare `peerUID` and that the larger-CRC32 UID comes first. Both check out against `fetch.go:266-277` — the no-`@` branch pairs against `selfUID`, and the comparison is `>`. Doc and code agree.

### Merge safety

No other open PR touches `internal/pipeline/fetch.go` or `internal/model/model.go` (the rest of the open set is CI dependency bumps plus #151 on the notify path), so there is no post-merge semantic conflict on this file.

---

## 2. Code quality

**Quality: Approved** — no P0/P1. Four P2s, none blocking.

### 2.1 P2 — This PR's new comment points readers down a doc trail that dead-ends (new finding)

`fetch.go:284-285` adds:

```go
// — see the discussion in each const block in model/model.go and the
// PR that introduced these constants (#181).
```

Following that pointer:

- The `Source*` block (`model.go:102-107`) has **no discussion** — its entire comment is `// Source type constants.` There is nothing there explaining why the numbering differs from `ChannelType*`.
- The `ChannelType*` block (`model.go:118-125`) does have a discussion, and it ends with `see appToStorageChannelType() for mapping`. **That function does not exist anywhere in the repository.** A case-insensitive grep for `apptostorage` across the whole tree returns exactly one hit: the comment itself. The real function is `appOriginToStorageChannelType` (`internal/api/handler/agent_reference_context.go:229`).

The dangling name is pre-existing (it came in with #181, and this diff does not touch `model.go`), so it is not a regression. But this PR is what starts *directing readers into it* — the comment is new here. Since the stated purpose of the PR is "a reader never has to flip between which numeric enum am I looking at", sending that reader to a const block with no discussion, whose sibling block cites a non-existent function, undercuts the goal. Cheapest fix: correct the name to `appOriginToStorageChannelType` in `model.go:120` and either add a one-line rationale to the `Source*` block or drop the "each const block" half of the new pointer.

### 2.2 P2 — "finishes the migration inside that one file" is not accurate

Four Go-level numeric enum literals survive in `fetch.go`:

- `fetch.go:879` — `if sourceType(s) != 3 { // not DM` → should be `model.SourceDirect`
- `fetch.go:163` — `NormalizeDMChannelID(d.ChannelID, uid, 1)` → should be `model.ChannelTypeDM`
- `fetch.go:166` — `ChannelType: 1,` → should be `model.ChannelTypeDM`
- `fetch.go:220` — `ChannelType: 5,` → should be `model.ChannelTypeThread`

Line 879 is the one I would actually fix. It is the *same* comparison class the PR just migrated 560 lines earlier, so the file is now inconsistent with itself:

```go
// fetch.go:319 — after this PR
if mapFrontendSourceType(sourceType(s)) != model.ChannelTypeThread {

// fetch.go:879 — untouched
if sourceType(s) != 3 { // not DM
```

That it needs a trailing `// not DM` to be legible is the tell. Line 163 is worth a look too: the PR rewrote `NormalizeDMChannelID`'s doc comment to describe its gate as `model.ChannelTypeDM`, while the in-file caller one screen above still passes a bare `1`.

The SQL literals (`:153` `channel_type = 1`, `:123` `2 AS channel_type`, `:196` `5 AS channel_type`) are a different situation — they live inside raw query strings and cannot reference Go constants without `fmt`-splicing, which is not a trade worth making. Leaving them is the right call. The stale comment at `:172` (`// Thread channels (channelType=5)`) could name the constant, though.

Either land the four Go-level sites or soften the sentence in the description.

### 2.3 P2 — The same mapping now exists three times, and two enums are held equal only by coincidence (new finding)

After this PR, three functions compute the identical `1→2, 2→5, 3→1` mapping:

| Function | Input enum | Default branch |
|---|---|---|
| `pipeline.mapFrontendSourceType` (`fetch.go:286`) | `Source*` | `return frontendType` |
| `handler.frontendToIMChannelType` (`bot_summary_create.go:291`) | `Source*` | `return 0` |
| `handler.appOriginToStorageChannelType` (`agent_reference_context.go:229`) | `OriginChannel*` | `return 0` |

The first two are character-for-character identical except the default. The third takes a *different* enum — and the reason it computes the same arithmetic is that `OriginChannel*` (`1=Group, 2=Thread, 3=DM`, `model.go:112-116`) and `Source*` (`1=Group, 2=Thread, 3=Direct`) are **numerically identical enums with different names**.

That coincidence is load-bearing in at least three places where one enum is passed where the other is declared:

- `task.go:306` — `SourceType: req.OriginChannelType` assigns an `OriginChannel*` value straight into a `Source*` field.
- `bot_summary_create.go:173` — `canonicalBotSourceID(sourceReq{SourceType: req.OriginChannelType, ...})`, same substitution.
- `bot_summary_create.go:278` — validates an `OriginChannel*` value with `validFrontendSourceType`, a `Source*`-range check.

All three are correct today and all are pre-existing. The reason to raise it here is that the description's forward-looking claim —

> If we ever decide to run that unification, this PR makes the code side considerably simpler — only the constant values change; every call site already reads through them.

— is **overstated**. Renumbering `Source*` would silently break those three cross-enum assignments with no compile error, and would still miss `fetch.go:879/163/166/220` and the three SQL literals. The refactor is a real improvement in local readability; it does not yet make the renumbering safe. Worth adjusting the claim so a future author doesn't act on it.

### 2.4 P2 — `default: return frontendType` reads more clearly wrong than it did before (pre-existing)

```go
// fetch.go:294-295
default:
    return frontendType
```

The new doc comment states the two enums "intentionally use different numbers", which makes returning an unmapped *frontend* value into a *channel-type* slot visibly incoherent. Traced end to end:

1. `POST /api/v1/summaries` (`task.go:386-391`) persists `s.SourceType` with **no range validation** — the handler validates `title`, `topic`, `origin_channel_type`, and the time range, but never `req.Sources[].SourceType`.
2. The bot path *does* validate — `bot_summary_create.go:282` gates on `validFrontendSourceType` (`1..3`).
3. So a client-supplied `source_type: 5` reaches the worker, and `mapFrontendSourceType(5)` falls through `default` returning `5`, which equals `ChannelTypeThread`.
4. `selectedThreadChannelIDs` (`fetch.go:319`) then treats it as an explicitly-selected thread and relaxes the archived-thread filter in `GetUserChannels`.

Blast radius is bounded: the thread query at `fetch.go:196-211` still enforces `gm.uid = ? AND gm.is_deleted = 0` and `g.status = 1`, so the worst outcome is a user surfacing an **archived** thread in a group they already belong to. No cross-tenant or cross-user exposure, and `status = 3` (deleted) stays excluded by construction.

**Not introduced by this PR and not blocking it** — the codegen is provably unchanged. Flagging it because this PR is the moment the code starts *claiming* the two enums are deliberately distinct. Follow-up fix: align the default with its sibling (`default: return 0`) and add the `1..3` check to the `task.go` create path.

### Verified clean (adversarial checks that did *not* find a defect)

- No literal↔constant mismatch at any of the five sites.
- No unused-import or undefined-symbol; `model` was already imported.
- No logic drift — disassembly identical across five functions including the two untouched callers (`GetUserChannels`, `ApplySourceConstraints`).
- Residual literals confirmed by `git` to be outside this diff, carrying correct values.
- Rewritten doc comments checked line-by-line against the code they describe; all accurate.

---

## 3. Test assessment

No new tests needed. Behaviour is provably unchanged at the instruction level, and the existing `internal/pipeline` suite passes at this head, exercising `NormalizeDMChannelID`, `mapFrontendSourceType`, and `selectedThreadChannelIDs`.

---

## 4. CI state at this head

Contrary to what an earlier pass on this PR observed, the heavy jobs are **no longer path-filter-skipped** — `Build`, `Lint`, `Test`, `Test (race, cgo)`, and `Vet` all pass at `852bd96`. `osv-scan`, `secret-scan`, `dependency-review`, `history`, and `pr-title-lint` also pass.

⚠️ **`check-sprint` is failing**, and it is a code-independent process gate:

```
No linked issue found. Please add a 'Closes #<issue>' reference to
the PR description, then re-run this check.
```

The description says `Linked Spec: N/A`. This is not a code defect and does not affect my verdict, but it will block merge on its own until a `Closes #<issue>` reference is added and the check re-run. Calling it out so it isn't mistaken for noise.

---

## 5. Overall verdict

**APPROVE**

Spec ✅ and Quality Approved. This is a genuinely low-risk change: the substitution is value-for-value correct at all five sites, and the affected functions were verified to compile to identical machine instructions against the merge-base. The "no DB migration, no API change" argument in the description is sound and I confirmed it independently rather than accepting it.

The decision to *reject* the DB renumbering in this PR is also right, and the stated payoff is real — `agent_reference_context.go:252-260` documents SUM-158, where raw storage `channel_type` (`1/2/5`) was written straight into `origin_channel_type`, mis-storing DM sessions as Group and pushing Thread outside the `1..3` validation window. That is precisely the bug class bare digits produce, and it is the strongest argument for landing this.

## 6. Suggested changes

Optional, none blocking:

1. Fix `model.go:120` — `appToStorageChannelType` → `appOriginToStorageChannelType`, and give the `Source*` const block the one-line rationale this PR's new comment promises the reader (§2.1).
2. Migrate `fetch.go:879` to `model.SourceDirect` — highest value per character, and it removes the need for the `// not DM` comment.
3. Migrate `fetch.go:163`, `:166`, `:220`, and name the constant in the `:172` comment.
4. Soften the "finishes the migration inside that one file" sentence and the "only the constant values change; every call site already reads through them" claim, or land (2)+(3) so the first becomes true (§2.2, §2.3).
5. Add `Closes #<issue>` so `check-sprint` goes green.
6. Follow-up PR: unify the three mappers and decide the `default:` contract deliberately — the two `Source*` mappers' defaults are *not* equivalent in effect (§2.3, §2.4).

## 7. For a human to verify

This PR carries the `needs-human-review` label. **Nothing in this diff touches auth, tokens, credentials, sessions, or middleware** — the label appears to be keyword-triggered rather than reflecting the change's content, and the diff is a value-preserving constant substitution with machine-verified identical codegen. Two items to look at, neither caused by this diff:

- **Input-validation asymmetry (§2.4):** `task.go:386-391` persists a client-supplied `source_type` without the `1..3` check `bot_summary_create.go:282` applies. Someone should decide whether to close that gap. Bounded by the group-membership predicate in the thread query, so low severity — but it is the kind of asymmetry that worsens quietly.
- **Cross-enum coincidence (§2.3):** `OriginChannel*` and `Source*` are numerically identical and are assigned across each other in three places with no compile-time protection. Worth a decision on whether to make them distinct Go types before anyone attempts the renumbering the description contemplates.

Issue #96, opened 2026-06-13
feat: notification mechanism (scheduled-task & task-completion notifications)


Jerry-Xin:
## Background
Smart summary currently lacks a notification mechanism. Users have no proactive feedback for asynchronous summary work — they must manually poll the list/detail to learn whether a scheduled run or a long-running task has finished.

## Scope
Build a notification mechanism covering at least:

1. **Scheduled task notifications** — when a scheduled summary run fires (success or failure), notify the relevant user/channel.
2. **Task-completion notifications** — when an on-demand / long-running summary task completes, notify the requester (success with a link to the result, or failure with the reason).

## Suggested considerations
- Notification channels: in-app (Octo message), and/or push to the originating conversation (group/thread/DM).
- Delivery target resolution: route the notification back to the requester or the configured destination.
- Failure notifications should include actionable error context.
- Idempotency / dedup so a single task does not spam repeated notifications.
- Respect quiet windows if applicable.

## Acceptance
- A finished scheduled run produces a notification to the correct target.
- A completed/failed on-demand task notifies the requester with result link or error.

## Related issues (octo-web)
- **octo-web#289** (open) — `feat: 智能总结完成后向被总结的群发送 tip 通知消息` — frontend/product side of the task-completion notification; this backend issue should provide the trigger/delivery hook it consumes.
- **octo-web#290** (open) — `bug: 智能总结完成后左侧总结列表不自动刷新` — related "completion → user feedback" path (list refresh vs. notification).
- **octo-web#192** (closed) — `智能总结任务完成后 session 列表不自动刷新 + 缺少红点提醒` — prior art on completion feedback / red-dot reminder.

Jerry-Xin:
Linked related octo-web issues: #289 (send tip notification to summarized group on completion), #290 (summary list auto-refresh on completion), and closed #192 (completion refresh + red-dot reminder). This backend issue should expose the trigger/delivery hook those frontend issues consume.

lml2468:
## 🔀 Leader Dispatch

| Field | Value |
|---|---|
| **assigned_to** | `product-analyst` |
| **stage** | `spec` |
| **reason** | `type:feature` without `spec-ready` label — needs spec written before any design or implementation. |
| **next_step** | product-analyst: analyze the feature request and produce a spec. Once spec is ready, add `spec-ready` label and human approval is required before architect dispatch. |
| **waiting_for** | product-analyst to produce spec |

[...]

PR #186, opened 2026-08-07
feat(notify): server-driven group tip on task completion

wanyiwang06:
## Summary

Server-driven successor of the abandoned web-only [octo-web#1234](https://github.com/Mininglamp-OSS/octo-web/pull/1234) attempt at delivering "{creator} 总结了群聊内容" as a passive grey system tip in every group source of a completed non-by-person task. Client-side best-effort delivery couldn't reach the correctness bar (see #1234 round-6 architecture escalation). This PR moves the tip generation to the server so exactly-once delivery comes from the authoritative source.

## Linked Spec

- **Product requirement**: [octo-web#289](https://github.com/Mininglamp-OSS/octo-web/issues/289)
- **Superseded**: [octo-web#1234](https://github.com/Mininglamp-OSS/octo-web/pull/1234) (closed after round-6)
- **Companion web renderer PR**: [wanyiwang06/octo-web feat/summary-notify-renderer](https://github.com/wanyiwang06/octo-web/tree/feat/summary-notify-renderer) (renderer for `contentType=21` messages; will be opened in Mininglamp-OSS/octo-web separately)

## Behavior

On every terminal-state transition of a summary task, `processor.go` calls `notifier.OnGroupTip(task)` alongside the existing `OnTaskTerminal` fan-out. `OnGroupTip` is scope-guarded and no-ops for anything outside its intent:

- Only `Status == Completed` (`Failed` / `Cancelled` / mid-flight never emit).
- Only **non-by-person** tasks (by-person's fan-out already goes to participants individually via DMs; layering a group broadcast on top is out of scope for #289).
- Only sources with `SourceType == SourceGroup` (`SourceThread` / `SourceDirect` skipped — the tip is a group-scale signal by design).
- Only tasks with a `Creator` (empty creator renders as "someone" and defeats the purpose).

For each eligible group source, the deliverer posts a payload `{content_type: 21, from_uid, task_id, task_no, source_id}` to the group channel via `WireChannelGroup` — no `EnsureFriend` hop (unlike DM), `SpaceID` is threaded through so octo-server can enforce space membership.

## Idempotency

Reuses the existing `summary_notification` state machine with a new kind:

- `model.NotifyKindGroupTip = "group_tip"`
- `recipient_uid` slot is repurposed as the group's `ChannelID`
- unique index `(task_id, kind, recipient_uid)` unchanged, so each `(task, group)` converges on one row; a second/third trigger of the same completion (status event + fallback poll + retry) finds the claim already made and no-ops.

Distinct kind means the `group_tip` row does NOT collide with the existing `completed` / `failed` rows for the same task, so the creator-DM completion notification and the group tips dedup independently.

## Isolation

Each group source runs the claim / deliver / mark state machine on its own `(task, kind, source_id)` row so one group's delivery failure never blocks the others — matches `OnTaskTerminal`'s per-recipient isolation.

## How verified

- `go build ./...` → passes
- `go vet ./...` → clean
- New `internal/notify/group_tip_test.go` (8 cases, CGO-gated to match existing `notify_test.go`):
  - Fan-out to every group source, thread/dm sources filtered out
  - Idempotency across two triggers of the same completion
  - Failed status skipped
  - by-person mode skipped
  - Disabled config skipped
  - One group's failure does not block others (attempt_count reflects retry)
  - No group sources → no delivery attempts
  - Group delivery skips `EnsureFriend` and uses `Payload` not `Card`
- Local `go test` requires `gcc` for the sqlite driver (`//go:build cgo`); CI provides it.

Diff stat: **+409 / -0** across 4 files — fully additive, zero existing code paths modified.

## COMPREHENSION

- 本 PR 让服务器接管「总结完成后向群里发 tip」这一动作,取代 web-only 那条已废弃路径。所有 dedup / retry / space membership 校验都复用现有 `summary_notification` 状态机,只加一个新 kind = `group_tip`。
- 保持向后兼容:`OnTaskTerminal` (creator DM 通知)行为不变,`OnGroupTip` 是独立 method,两者互相独立 dedup,不会重发。
- Failed / by-person / 非 group source 全都显式 short-circuit,任何一条 guard 失效都独立触发相同的 no-op,不会误发。
- Payload 里 `content_type=21` 对应 `octo-web` renderer PR 里 `MessageContentTypeConst.summaryNotify = 21`,contract 是纯 additive。
- 依然 respect 现有 `Config.Enabled` 开关:功能可以在 config 侧灰度关闭而无需回滚代码。

### DIFF
--- a/internal/notify/notify.go
+++ b/internal/notify/notify.go
@@ -197,6 +197,130 @@ func kindForStatus(status int) (string, bool) {
 	}
 }
 
+// OnGroupTip fans out a passive "{creator} 总结了群聊内容" system tip to every
+// group source of a completed non-by-person task. Server-driven successor of
+// the abandoned web-only #1234 attempt — server is now the authoritative
+// exactly-once source for the tip.
+//
+// Scope guards (each an early return, all defensive):
+//   - only Status == Completed (Failed / Cancelled / mid-flight never emit).
+//   - only NON-by-person tasks (by-person's fan-out already goes to
+//     participants via personal DMs; layering a group tip on top is out of
+//     scope for #289).
+//   - only sources with SourceType == SourceGroup (SourceThread / SourceDirect
+//     never receive this tip).
+//   - only tasks with a Creator (empty creator → nothing to render).
+//
+// Idempotency: reuses the summary_notification state machine with
+// notify_kind = NotifyKindGroupTip and recipient_uid = source.ChannelID (the
+// GROUP channel_id), so retries and multiple triggers of the same completion
+// converge on one row per (task, group).
+func (n *Notifier) OnGroupTip(task model.SummaryTask) {
+	if n == nil || !n.cfg.Enabled || n.deliverer == nil || n.db == nil {
+		return
+	}
+	if task.Status != model.StatusCompleted {
+		return
+	}
+	if task.SummaryMode == model.ModeByPerson {
+		// By-person tasks fan out to participants individually — the group
+		// tip is orthogonal to that flow. Keeping this scope guard explicit
+		// so a future by-person change doesn't accidentally start double-
+		// broadcasting.
+		return
+	}
+	if task.CreatorID == "" {
+		return
+	}
+
+	// Resolve group sources. A source resolution error is not fatal — we log
+	// and skip; on the next trigger/sweep the resolution runs again.
+	var sources []model.SummarySource
+	if err := n.db.Where("task_id = ? AND source_type = ?", task.ID, model.SourceGroup).
+		Find(&sources).Error; err != nil {
+		log.Printf("[notify] task=%d kind=%s: resolve group sources failed: %v",
+			task.ID, model.NotifyKindGroupTip, err)
+		return
+	}
+	if len(sources) == 0 {
+		return
+	}
+
+	// Each group source runs the claim/deliver/mark state machine on its own
+	// (task, kind, source_channel_id) row so one group's delivery failure
+	// never blocks the others (matches OnTaskTerminal's fan-out semantics).
+	for _, src := range sources {
+		target := deliveryTarget{
+			ChannelID:   src.SourceID,
+			ChannelType: WireChannelGroup,
+			// TargetUID intentionally empty for group回发: no EnsureFriend
+			// is needed and octo-server does not build a whitelist DM channel
+			// for a group destination.
+			SpaceID: task.SpaceID,
+		}
+		n.deliverGroupTipToSource(task, target, src)
+	}
+}
+
+// deliverGroupTipToSource runs the per-source claim/deliver/mark state machine
+// for one group source. Extracted so OnGroupTip's loop stays flat and each
+// source's failure isolates from the others.
+func (n *Notifier) deliverGroupTipToSource(task model.SummaryTask, target deliveryTarget, src model.SummarySource) {
+	kind := model.NotifyKindGroupTip
+	uid := src.SourceID // recipient_uid slot repurposed for the group channel id
+
+	claimed, _, err := n.claim(task.ID, kind, uid)
+	if err != nil {
+		log.Printf("[notify] task=%d kind=%s group=%s: claim failed: %v", task.ID, kind, uid, err)
+		return
+	}
+	if !claimed {
+		// Already delivered / in-flight / attempts exhausted — dedup row
+		// carries the terminal state.
+		return
+	}
+
+	if err := n.deliverGroupTip(task, target, src); err != nil {
+		n.markFailed(task.ID, kind, uid, err)
+		log.Printf("[notify] task=%d kind=%s group=%s: delivery failed: %v",
+			task.ID, kind, uid, sanitize(err.Error()))
+		return
+	}
+	n.markSent(task.ID, kind, uid)
+	log.Printf("[notify] task=%d kind=%s group=%s: delivered", task.ID, kind, uid)
+}
+
+// deliverGroupTip posts the SummaryNotifyContent (contentType=21) system tip
+// to one group channel. It intentionally does NOT reuse `deliver` because:
+//   - group delivery skips EnsureFriend (no per-user relationship needed).
+//   - the payload is not a card but a raw content-type-21 system message
+//     whose renderer lives in octo-web packages/dmworkbase/Messages/SummaryNotify.
+func (n *Notifier) deliverGroupTip(task model.SummaryTask, target deliveryTarget, src model.SummarySource) error {
+	ctx, cancel := context.WithTimeout(context.Background(), 15*time.Second)
+	defer cancel()
+
+	payload := map[string]any{
+		// content_type=21 mirrors packages/dmworkbase/src/Service/Const.ts
+		// (MessageContentTypeConst.summaryNotify) and is decoded by the
+		// SummaryNotifyContent class in the web/mobile renderer PR.
+		"content_type": 21,
+		"from_uid":     task.CreatorID,
+		"task_id":      task.ID,
+		"task_no":      task.TaskNo,
+		"source_id":    src.SourceID,
+	}
+
+	msg := SendMessageRequest{
+		ChannelID:   target.ChannelID,
+		ChannelType: target.ChannelType,
+		Payload:     payload,
+	}
+	if err := n.deliverer.SendMessage(ctx, target.SpaceID, msg); err != nil {
+		return fmt.Errorf("sendMessage: %w", err)
+	}
+	return nil
+}
+
 // claim attempts the preemptive INSERT. Returns claimed=true when this run won
 // the (task, kind) slot. When the row already exists, claimed=false and the
 // existing row is returned so the caller can decide whether retry budget remains.
--- a/internal/model/notification.go
+++ b/internal/model/notification.go
@@ -8,6 +8,12 @@ import "time"
 const (
 	NotifyKindCompleted = "completed"
 	NotifyKindFailed    = "failed"
+	// NotifyKindGroupTip is the passive grey system tip fanned out to every
+	// group source of a Completed non-by-person task. Server-driven successor
+	// of the abandoned web-only #1234 attempt (see #289). Distinct kind so
+	// its per-(task, kind, source_channel_id) dedup row does not collide with
+	// the creator-DM Completed notification.
+	NotifyKindGroupTip = "group_tip"
 )
 
 // Notification status constants (summary_notification.status).
--- a/internal/notify/group_tip_test.go
+++ b/internal/notify/group_tip_test.go
@@ -1,0 +1,272 @@
+//go:build cgo
+
+package notify
+
+import (
+	"context"
+	"errors"
+	"testing"
+
+	"github.com/Mininglamp-OSS/octo-smart-summary/internal/model"
+	"gorm.io/driver/sqlite"
+	"gorm.io/gorm"
+)
+
+// setupGroupTipTestDB builds an in-memory sqlite fixture with the tables
+// OnGroupTip touches (notifications for the dedup rows, sources for the fan-out
+// query). Kept separate from setupNotifyTestDB so a schema drift on either
+// side surfaces here rather than silently in existing tests.
+func setupGroupTipTestDB(t *testing.T) *gorm.DB {
+	t.Helper()
+	db, err := gorm.Open(sqlite.Open(":memory:"), &gorm.Config{})
+	if err != nil {
+		t.Fatalf("open db: %v", err)
+	}
+	if err := db.AutoMigrate(&model.SummaryNotification{}, &model.SummarySource{}); err != nil {
+		t.Fatalf("automigrate: %v", err)
+	}
+	return db
+}
+
+func groupTipTask(id int64, mode int, status int) model.SummaryTask {
+	return model.SummaryTask{
+		ID:          id,
+		TaskNo:      "GT-1",
+		Title:       "群聊总结",
+		SpaceID:     "space-9",
+		CreatorID:   "user-creator",
+		SummaryMode: mode,
+		Status:      status,
+	}
+}
+
+// mustAddSources inserts the given (source_type, source_id) pairs against a
+// task so OnGroupTip's fan-out has something to iterate over.
+func mustAddSources(t *testing.T, db *gorm.DB, taskID int64, srcs []model.SummarySource) {
+	t.Helper()
+	for i := range srcs {
+		srcs[i].TaskID = taskID
+		if err := db.Create(&srcs[i]).Error; err != nil {
+			t.Fatalf("create source: %v", err)
+		}
+	}
+}
+
+func TestOnGroupTip_FansOutOneTipPerGroupSource(t *testing.T) {
+	db := setupGroupTipTestDB(t)
+	fake := &fakeDeliverer{}
+	n := newTestNotifier(db, fake, Config{Enabled: true, MaxAttempts: 3})
+
+	task := groupTipTask(1, model.ModeByPerson-1 /* non-by-person */, model.StatusCompleted)
+	mustAddSources(t, db, task.ID, []model.SummarySource{
+		{SourceType: model.SourceGroup, SourceID: "group-a"},
+		{SourceType: model.SourceGroup, SourceID: "group-b"},
+		{SourceType: model.SourceThread, SourceID: "thread-x"}, // must be skipped
+		{SourceType: model.SourceDirect, SourceID: "dm-y"},     // must be skipped
+	})
+
+	n.OnGroupTip(task)
+
+	if len(fake.sendCalls) != 2 {
+		t.Fatalf("expected 2 group sends, got %d", len(fake.sendCalls))
+	}
+	channels := map[string]bool{}
+	for _, msg := range fake.sendCalls {
+		if msg.ChannelType != WireChannelGroup {
+			t.Errorf("expected WireChannelGroup=%d, got %d", WireChannelGroup, msg.ChannelType)
+		}
+		if ct, _ := msg.Payload["content_type"].(int); ct != 21 {
+			t.Errorf("expected payload content_type=21 (summaryNotify), got %v", msg.Payload["content_type"])
+		}
+		if fu, _ := msg.Payload["from_uid"].(string); fu != task.CreatorID {
+			t.Errorf("expected payload from_uid=%s, got %v", task.CreatorID, msg.Payload["from_uid"])
+		}
+		channels[msg.ChannelID] = true
+	}
+	if !channels["group-a"] || !channels["group-b"] {
+		t.Errorf("expected both group-a and group-b to receive tip, got %v", channels)
+	}
+	if channels["thread-x"] || channels["dm-y"] {
+		t.Errorf("thread/dm sources must not receive group tip; got %v", channels)
+	}
+
+	// Each group source should have a Sent dedup row.
+	var count int64
+	db.Model(&model.SummaryNotification{}).
+		Where("task_id = ? AND notify_kind = ? AND status = ?",
+			task.ID, model.NotifyKindGroupTip, model.NotifyStatusSent).
+		Count(&count)
+	if count != 2 {
+		t.Errorf("expected 2 sent notification rows, got %d", count)
+	}
+}
+
+func TestOnGroupTip_IsIdempotentAcrossRetriggers(t *testing.T) {
+	db := setupGroupTipTestDB(t)
+	fake := &fakeDeliverer{}
+	n := newTestNotifier(db, fake, Config{Enabled: true, MaxAttempts: 3})
+
+	task := groupTipTask(1, 1 /* non-by-person */, model.StatusCompleted)
+	mustAddSources(t, db, task.ID, []model.SummarySource{
+		{SourceType: model.SourceGroup, SourceID: "group-a"},
+	})
+
+	// Two triggers of the same completion (status event + fallback poll
+	// pattern) must land at most one delivery per (task, group).
+	n.OnGroupTip(task)
+	n.OnGroupTip(task)
+
+	if len(fake.sendCalls) != 1 {
+		t.Fatalf("idempotency violated: expected exactly 1 send, got %d", len(fake.sendCalls))
+	}
+}
+
+func TestOnGroupTip_SkipsFailedTask(t *testing.T) {
+	db := setupGroupTipTestDB(t)
+	fake := &fakeDeliverer{}
+	n := newTestNotifier(db, fake, Config{Enabled: true, MaxAttempts: 3})
+
+	// Failed → group tip must NOT fire (would broadcast failure to the group).
+	task := groupTipTask(1, 1 /* non-by-person */, model.StatusFailed)
+	mustAddSources(t, db, task.ID, []model.SummarySource{
+		{SourceType: model.SourceGroup, SourceID: "group-a"},
+	})
+
+	n.OnGroupTip(task)
+	if len(fake.sendCalls) != 0 {
+		t.Errorf("failed task must not emit group tip, got %d sends", len(fake.sendCalls))
+	}
+}
+
+func TestOnGroupTip_SkipsByPersonMode(t *testing.T) {
+	db := setupGroupTipTestDB(t)
+	fake := &fakeDeliverer{}
+	n := newTestNotifier(db, fake, Config{Enabled: true, MaxAttempts: 3})
+
+	// by-person mode's fan-out already goes to participants via DMs; layering
+	// group tips on top would double-broadcast.
+	task := groupTipTask(1, model.ModeByPerson, model.StatusCompleted)
+	mustAddSources(t, db, task.ID, []model.SummarySource{
+		{SourceType: model.SourceGroup, SourceID: "group-a"},
+	})
+
+	n.OnGroupTip(task)
+	if len(fake.sendCalls) != 0 {
+		t.Errorf("by-person task must not emit group tip, got %d sends", len(fake.sendCalls))
+	}
+}
+
+func TestOnGroupTip_SkipsWhenNotEnabled(t *testing.T) {
+	db := setupGroupTipTestDB(t)
+	fake := &fakeDeliverer{}
+	n := newTestNotifier(db, fake, Config{Enabled: false, MaxAttempts: 3})
+
+	task := groupTipTask(1, 1, model.StatusCompleted)
+	mustAddSources(t, db, task.ID, []model.SummarySource{
+		{SourceType: model.SourceGroup, SourceID: "group-a"},
+	})
+
+	n.OnGroupTip(task)
+	if len(fake.sendCalls) != 0 {
+		t.Errorf("disabled notifier must not emit; got %d sends", len(fake.sendCalls))
+	}
+}
+
+func TestOnGroupTip_OneGroupFailureDoesNotBlockOthers(t *testing.T) {
+	db := setupGroupTipTestDB(t)
+	// Fail the very first delivery attempt (any group; we assert isolation by
+	// counting total deliveries across both groups).
+	fake := &fakeDeliverer{sendErrOnce: errors.New("transient network")}
+	n := newTestNotifier(db, fake, Config{Enabled: true, MaxAttempts: 3})
+
+	task := groupTipTask(1, 1, model.StatusCompleted)
+	mustAddSources(t, db, task.ID, []model.SummarySource{
+		{SourceType: model.SourceGroup, SourceID: "group-a"},
+		{SourceType: model.SourceGroup, SourceID: "group-b"},
+	})
+
+	n.OnGroupTip(task)
+
+	// Both groups must have been attempted (one failed, one succeeded) — one
+	// group's failure never blocks another.
+	if len(fake.sendCalls) != 2 {
+		t.Fatalf("expected 2 delivery attempts across groups, got %d", len(fake.sendCalls))
+	}
+
+	// One row Sent, one row Failed → attempt_count=1, retry-eligible on sweep.
+	var sent, failed int64
+	db.Model(&model.SummaryNotification{}).
+		Where("task_id = ? AND notify_kind = ? AND status = ?",
+			task.ID, model.NotifyKindGroupTip, model.NotifyStatusSent).Count(&sent)
+	db.Model(&model.SummaryNotification{}).
+		Where("task_id = ? AND notify_kind = ? AND status = ?",
+			task.ID, model.NotifyKindGroupTip, model.NotifyStatusFailed).Count(&failed)
+	if sent != 1 || failed != 1 {
+		t.Errorf("expected sent=1 failed=1, got sent=%d failed=%d", sent, failed)
+	}
+}
+
+func TestOnGroupTip_SkipsWhenNoGroupSources(t *testing.T) {
+	db := setupGroupTipTestDB(t)
+	fake := &fakeDeliverer{}
+	n := newTestNotifier(db, fake, Config{Enabled: true, MaxAttempts: 3})
+
+	// Only DM / thread sources — nothing to fan out to.
+	task := groupTipTask(1, 1, model.StatusCompleted)
+	mustAddSources(t, db, task.ID, []model.SummarySource{
+		{SourceType: model.SourceDirect, SourceID: "dm-y"},
+		{SourceType: model.SourceThread, SourceID: "thread-x"},
+	})
+
+	n.OnGroupTip(task)
+	if len(fake.sendCalls) != 0 {
+		t.Errorf("no group sources → no sends; got %d", len(fake.sendCalls))
+	}
+}
+
+func TestOnGroupTip_UsesGroupChannelWithoutEnsureFriend(t *testing.T) {
+	db := setupGroupTipTestDB(t)
+	fake := &fakeDeliverer{}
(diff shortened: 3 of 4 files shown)


internal/notify/notify.go
```diff
@@ -197,6 +197,130 @@ func kindForStatus(status int) (string, bool) {
...
+		"task_no":      task.TaskNo,
+		"source_id":    src.SourceID,
+	}
+
+	msg := SendMessageRequest{
+		ChannelID:   target.ChannelID,
+		ChannelType: target.ChannelType,
+		Payload:     payload,
+	}
+	if err := n.deliverer.SendMessage(ctx, target.SpaceID, msg); err != nil {
```
lml2468, line 318: [P0] This cannot address a group through the production deliverer. `InternalNotifyDeliverer.SendMessage` drops `msg.ChannelType` and serializes only `Targets: []{msg.ChannelID}`; the current octo-server `/v1/internal/notify` contract validates every target as a space-member UID and calls `NewPersonalMsgSendReq`. A group ID is therefore filtered as a non-member (or misrouted as a DM on an ID collision), never posted to a group. The fake test observes `SendMessageRequest` before this lossy serialization, so it cannot catch the failure. Please add a supported server-side group destination contract, serialize it here, and cover the actual HTTP body/integration path.

internal/notify/notify.go
```diff
@@ -197,6 +197,130 @@ func kindForStatus(status int) (string, bool) {
...
+//     whose renderer lives in octo-web packages/dmworkbase/Messages/SummaryNotify.
+func (n *Notifier) deliverGroupTip(task model.SummaryTask, target deliveryTarget, src model.SummarySource) error {
+	ctx, cancel := context.WithTimeout(context.Background(), 15*time.Second)
+	defer cancel()
+
+	payload := map[string]any{
+		// content_type=21 mirrors packages/dmworkbase/src/Service/Const.ts
+		// (MessageContentTypeConst.summaryNotify) and is decoded by the
+		// SummaryNotifyContent class in the web/mobile renderer PR.
+		"content_type": 21,
```
lml2468, line 306: [P1] WuKong raw message payloads use `type` as the content discriminator, not `content_type`. octo-server copies this payload unchanged and its existing notify contract uses `{type: 1, content: ...}`; the client dispatcher likewise selects `MessageContent.contentType` from payload `type`. With this shape the type-21 renderer is never selected even after group routing is added. Please send the wire discriminator expected by the IM contract and add a production-body test.

internal/notify/notify.go
```diff
@@ -197,6 +197,130 @@ func kindForStatus(status int) (string, bool) {
...
+		"from_uid":     task.CreatorID,
```
lml2468, line 307: [P1] The linked renderer derives the displayed creator from the authenticated message envelope and intentionally does not trust payload `from_uid`, but `/v1/internal/notify` sends legacy payloads from `NotifyBotUID()` and this client leaves `ActorUID` empty. Consequently the tip would attribute the summary to the notification bot, not `task.CreatorID`. This needs an explicit, trusted server-owned attribution contract (and a cross-repo contract test), rather than an unauthenticated payload field.

internal/notify/notify.go
```diff
@@ -197,6 +197,130 @@ func kindForStatus(status int) (string, bool) {
...
+		log.Printf("[notify] task=%d kind=%s group=%s: claim failed: %v", task.ID, kind, uid, err)
+		return
+	}
+	if !claimed {
+		// Already delivered / in-flight / attempts exhausted — dedup row
+		// carries the terminal state.
+		return
+	}
+
+	if err := n.deliverGroupTip(task, target, src); err != nil {
```
lml2468, line 283: [P1] Marking this row failed makes it eligible for the generic sweep, but `redeliver` always rebuilds `targetForUID` (DM), calls `buildCard`, and uses `deliver`. A `group_tip` retry therefore turns the group ID into a DM recipient and replaces the raw group payload with a card; direct `OnGroupTip` retriggers also skip the existing failed row. Please make redelivery kind-aware (re-resolve the group source and call the group transport) and add a fail → `Sweep` → successful group-send regression test.

internal/notify/notify.go
```diff
@@ -197,6 +197,130 @@ func kindForStatus(status int) (string, bool) {
...
+		return
+	}
+	if task.CreatorID == "" {
+		return
+	}
+
+	// Resolve group sources. A source resolution error is not fatal — we log
+	// and skip; on the next trigger/sweep the resolution runs again.
+	var sources []model.SummarySource
+	if err := n.db.Where("task_id = ? AND source_type = ?", task.ID, model.SourceGroup).
```
lml2468, line 239: [P2] There is no durable recovery when source discovery fails or the worker exits before the first per-group claim. No `group_tip` row exists in either case, while `Sweep` only scans existing failed/pending rows, so the one-shot terminal callback is permanently lost; the comment about retrying on the next sweep is not true. Persist/claim the outbox rows with the terminal transition, or add a sweep that discovers completed eligible tasks missing group-tip rows.

lml2468 (changes requested):
Request changes. The group-tip behavior is not reachable through the current production transport, and the payload/retry/attribution contracts are inconsistent. Also, the PR describes exactly-once delivery, but claim → send → markSent without a transport idempotency key is at-least-once across crash or ambiguous-timeout recovery. Verification: internal/notify tests and race tests passed; CGO-disabled full tests, vet, and formatting passed. The worker race binary could not link locally because libtokenizers is unavailable (environmental). CI Build/Test/Vet/Lint/Race all pass; check-sprint is the only failing workflow.

mochashanyao (changes requested):
[Octo-Q · automated review]

---

# Code Review — PR #186 (octo-smart-summary)

## Summary

This PR adds a server-driven "group tip" notification: when a non-by-person summary task completes, a passive system message (content_type=21) is fanned out to every group source of the task. The implementation reuses the existing `summary_notification` dedup state machine with a new `NotifyKindGroupTip` kind, and wires the call in `notifyTaskTerminal` right after `OnTaskTerminal`. The synchronous delivery path is well-structured with clear scope guards, per-source isolation, and idempotent claim semantics. However, the background sweep/redeliver retry path — which handles transient delivery failures — is not aware of the new kind and will produce incorrect deliveries.

## Verification

- ✅ **Scope guards are exhaustive** — `OnGroupTip` correctly filters on status=Completed, non-by-person mode, non-empty CreatorID, and group-type sources only. Each guard is an early return, tested individually in `group_tip_test.go`.
- ✅ **Per-source failure isolation** — `deliverGroupTipToSource` runs independently per group source. A failure in one group does not block others (verified by `TestOnGroupTip_OneGroupFailureDoesNotBlockOthers`).
- ✅ **Idempotency** — claim/INSERT UNIQUE on `(task_id, notify_kind, recipient_uid)` ensures exactly-once delivery per (task, group). `TestOnGroupTip_IsIdempotentAcrossRetriggers` confirms.
- ✅ **Channel type correctness (sync path)** — Group tips use `WireChannelGroup` (2) and skip `EnsureFriend` (verified by `TestOnGroupTip_UsesGroupChannelWithoutEnsureFriend`).
- ✅ **Payload shape** — content_type=21, from_uid, task_id, task_no, source_id all present and tested.
- ✅ **Data flow: task reload in `notifyTaskTerminal`** — `p.db.First(&task, taskID)` loads all columns including CreatorID, SummaryMode, Status, SpaceID — all fields consumed by `OnGroupTip` are populated.
- Static analysis only at head `93b933d01f2339546cb9158d8639b15aef860705`; build and tests not executed in this environment.

## Findings

One P1 blocker described below; one P2.

### P1 — Sweep/redeliver path uses wrong delivery mechanism for `group_tip` (`internal/notify/notify.go:798`)

`redeliver()` (called by both `sweepRetryFailed` at `:741` and `sweepStalePending` at `:772`) is kind-agnostic: it calls `targetForUID(task, uid)` → `buildCard(task, kind, errMsg)` → `deliver(taskID, target, card)` for every notification row it picks up. The sweep queries at `:722` and `:752` have no `notify_kind` filter, so `group_tip` rows are included.

For a `group_tip` row, `recipient_uid` holds the **group channel ID** (e.g. `"group-a"`), not a user UID. The redeliver path then:

1. `targetForUID()` (`target.go:107`) returns `{ChannelID: "group-a", ChannelType: WireChannelDM, TargetUID: "group-a"}` — wrong channel type (DM instead of Group) and wrong TargetUID (group ID treated as user).
2. `deliver()` (`:492`) calls `EnsureFriend(ctx, spaceID, "group-a")` — attempts to create a bot-friend relationship with a group channel ID as a user.
3. `buildCard()` (`:501`) produces a `notifyCard` struct — wrong payload format. Group tips need `content_type=21` payload, not a card.

**Consequence**: any group_tip delivery that fails on first attempt (transient octo-server 5xx, network blip) enters the sweep retry loop where every retry attempt also fails with wrong-format delivery. After `MaxAttempts` exhausts, the row reaches terminal `failed` state and the group tip is permanently lost — the user never sees it.

**Diff-scope**: new (introduced by this PR). The `redeliver` function predates this PR and handles `completed`/`failed` kinds correctly, but adding `group_tip` without kind-aware branching in the sweep path amplifies the incompatibility.

```go
// Fix sketch — add kind-aware branching in redeliver():
func (n *Notifier) redeliver(taskID int64, kind, uid string) {
    // ... existing task reload ...
    if kind == model.NotifyKindGroupTip {
        src := model.SummarySource{SourceID: uid, SourceType: model.SourceGroup}
        target := deliveryTarget{
            ChannelID: uid, ChannelType: WireChannelGroup, SpaceID: task.SpaceID,
        }
        if err := n.deliverGroupTip(task, target, src); err != nil {
            n.markFailed(taskID, kind, uid, err)
            return
        }
        n.markSent(taskID, kind, uid)
        return
    }
    // ... existing DM redeliver logic ...
}
```

### P2 — No sweep/redeliver test coverage for `group_tip` (`internal/notify/group_tip_test.go:1`)

The test suite thoroughly covers the synchronous `OnGroupTip` paths (fan-out, idempotency, scope guards, failure isolation, channel correctness) but has no test that seeds a `group_tip` notification row and calls `Sweep()`. The analogous `TestSweep_RetriesFailedRowWithBudget` in `notify_test.go:418` covers the completed/failed sweep path. Without a group_tip sweep test, the P1 above would pass CI undetected.

- **notify_kind column width**: `varchar(16)` — `"group_tip"` is 9 chars, fits comfortably.
- **UNIQUE constraint coverage**: `(task_id, notify_kind, recipient_uid)` correctly dedups group_tip per (task, group_channel_id) without colliding with completed/failed rows (different `notify_kind`).
- **`deliverGroupTipToSource` missing `claimRetry`**: Unlike `deliverToRecipient`, the group tip path does not attempt in-process retry of previously-failed rows. This is acceptable — group tips are best-effort, and the sweep handles retries.
- **AUTH/VISIBILITY**: No new API endpoints or auth surface. Group tips are emitted server-side only, to groups the task already has sources for.
- **SOURCE-COVERAGE**: The fan-out query `WHERE task_id = ? AND source_type = SourceGroup` correctly selects only group sources. Thread/DM sources are excluded at the DB level, not just filtered in Go.

## Verdict: CHANGES_REQUESTED

The P1 (sweep/redeliver incompatibility with group_tip) must be fixed before merge. Without it, any transient delivery failure for a group tip becomes a permanent loss. The fix is localized to `redeliver()` and needs one new sweep test in `group_tip_test.go`.

Jerry-Xin (dismissed):
The feature belongs in this repository, but group-tip retry handling is broken and can send an incorrect direct message.

### 🔴 Blocking

- 🔴 Critical — Group-tip retries use the existing DM redelivery path. After a group send fails, `Sweep` claims the `group_tip` row and calls `redeliver` ([internal/notify/notify.go:741](/Users/mlamp/.openclaw/tmp/pr-review-70mz6w_s/repo/internal/notify/notify.go:741)). That function treats the stored group channel ID as a user UID, constructs a DM target, calls `EnsureFriend`, and sends a card with no valid `group_tip` content ([internal/notify/notify.go:798](/Users/mlamp/.openclaw/tmp/pr-review-70mz6w_s/repo/internal/notify/notify.go:798), [internal/notify/target.go:105](/Users/mlamp/.openclaw/tmp/pr-review-70mz6w_s/repo/internal/notify/target.go:105)). This may attempt a malformed DM to a value that is actually a group ID instead of retrying the group payload.

  `redeliver` must dispatch by notification kind and rebuild the group target and content-type-21 payload for `NotifyKindGroupTip`. Add tests covering both failed-row sweep retry and stale-pending recovery.

- 🔴 Critical — Immediate retriggers never retry failed group tips. `deliverGroupTipToSource` discards the existing row returned by `claim` and returns whenever `claimed == false` ([internal/notify/notify.go:272](/Users/mlamp/.openclaw/tmp/pr-review-70mz6w_s/repo/internal/notify/notify.go:272)). Unlike `deliverToRecipient`, it never calls `claimRetry` for a failed row with remaining budget. This contradicts the documented retry state machine and leaves retry entirely to the currently incorrect sweep path.

### ✅ Highlights

- Scope guards correctly restrict delivery to completed, non-by-person tasks with group sources.
- Per-group deduplication and failure isolation are structured clearly.
- The group path correctly skips `EnsureFriend` on its initial delivery.

`go test ./internal/notify` passed. Worker tests could not link locally because the native `tokenizers` library was unavailable.

Jerry-Xin (changes requested):
Request changes. The feature belongs in this repository and the synchronous fan-out is well-structured, but the group tip is not deliverable through the production transport, and the retry state machine is inconsistent for the new kind.

### 🔴 Blocking

- 🔴 Critical — The group tip cannot address a group through the production deliverer. `InternalNotifyDeliverer.SendMessage` (`internal/notify/client.go:136`) serializes into `notifyReq`, which has **no `channel_type` field** and only carries `Targets: []string{msg.ChannelID}` (`internal/notify/client.go:160`). `msg.ChannelType = WireChannelGroup` is silently dropped, so the group `source_id` is sent to `/v1/internal/notify` as if it were a user target. octo-server resolves targets as space members and returns 200 with the value in `filtered` (see `notifyResp`), so the tip never reaches the group. The unit tests only pass because `fakeDeliverer` inspects `msg.ChannelType`/`msg.ChannelID` directly — the production wire shape is not exercised. This makes the feature non-functional in production.

- 🔴 Critical — Group-tip retries route through the DM redelivery path. On a failed initial send the row is marked `failed`, and the generic `Sweep`/`sweepRetryFailed` picks it up regardless of kind and calls `redeliver` (`internal/notify/notify.go:798`). `redeliver` unconditionally rebuilds a DM target via `targetForUID` (`internal/notify/target.go:104`, hardcoded `ChannelType: WireChannelDM`, `TargetUID: uid`), calls `EnsureFriend`, and sends a `buildCard` card through `deliver`. For a `group_tip` row `recipient_uid` is the **group channel id**, so the retry attempts a malformed DM/`EnsureFriend` against a group id and sends a card instead of the content-type-21 payload — the exact inverse of the intended group path. `redeliver` must dispatch by `notify_kind` and reconstruct the group target + content-type-21 payload for `NotifyKindGroupTip`, with tests for both failed-row sweep retry and stale-pending recovery.

- 🔴 Critical — Immediate re-triggers never retry a failed group tip. `deliverGroupTipToSource` (`internal/notify/notify.go:268`) discards the existing row returned by `claim` and returns whenever `claimed == false`. Unlike `deliverToRecipient` (`internal/notify/notify.go:143`), it never calls `claimRetry` for a `failed` row with remaining budget. Combined with the broken sweep path above, a transiently failed group tip has no correct retry route at all, contradicting the PR's retry/state-machine narrative and the `TestOnGroupTip_OneGroupFailureDoesNotBlockOthers` "retry-eligible on sweep" comment.

### 💬 Non-blocking

- Payload discriminator: the existing notify contract uses `{type: 1, content: ...}` on the raw payload; this PR emits `content_type: 21`. Confirm octo-server / the renderer keys off `content_type` and not `type`, otherwise the message will not be decoded as `summaryNotify`.
- Attribution: verify the renderer actually reads `from_uid` from the payload rather than deriving the sender from the authenticated envelope; if it ignores payload `from_uid`, the "{creator} 总结了群聊内容" attribution will be wrong.
- Exactly-once claim: `claim → send → markSent` without a transport-level idempotency key is at-least-once across a crash or ambiguous-timeout between send and `markSent`; the PR body's "exactly-once" wording is stronger than the implementation guarantees.

### ✅ Highlights

- Scope guards correctly restrict delivery to Completed, non-by-person tasks with group sources; each guard is an explicit early return.
- Per-group `(task, kind, source_channel_id)` dedup and per-source failure isolation are clearly structured and mirror `OnTaskTerminal` semantics.
- The initial group delivery path correctly skips `EnsureFriend`.
- Constant/enum mappings verified: `NotifyKindGroupTip="group_tip"` fits `varchar(16)`; `WireChannelGroup=2`, `StatusCompleted=3`, `ModeByPerson=2`, `SourceGroup=1`.

### Verification

- `go build ./internal/...` compiles clean; `go vet ./...` clean; `go test ./internal/notify/...` (cgo) passes including the new 9-case `group_tip_test.go`. The `cmd/summary-*` link failure is environmental (native `tokenizers` lib unavailable locally), unrelated to this PR.
- CI: Build / Test / Test(race,cgo) / Vet / Lint all pass; `check-sprint` is a process gate, not code.

yujiawei (changes requested):
# Code Review — PR #186 (octo-smart-summary)

Reviewed at head `93b933d0`. I read the full diff, then traced each claim in the PR body through the surrounding code and through the two external contracts this change depends on (octo-server `/v1/internal/notify`, octo-web message rendering). `go build ./...`, `go vet ./...` and `go test ./internal/notify/...` all pass locally with cgo available.

The Go code is clean, well-commented, and the scope guards are correct in isolation. My concern is not the code style — it is that **the delivery mechanism this PR builds cannot reach a group channel, and the payload it emits cannot be decoded as content type 21**. Both are verifiable from the contracts on either side of the wire, and neither is exercised by the new tests because the tests stop at an in-memory fake.

---

## 1. Spec compliance

**Spec: ❌**

Linked spec is octo-web#289 (plus the closing plan of the superseded octo-web#1234).

### Deviations

**S1 — The governance gate that was explicitly transferred to this PR is not satisfied.**
The closing comment of #1234 states that the outstanding governance requirement — *"real maintainer sign-off on #289 + `spec-approved` label"* — is carried by the backend PR, i.e. this one. As of this review, #289 carries labels `type:feature`, `priority:P3`, `needs-human-review`, `area:frontend`. There is no `spec-approved` label, and `needs-human-review` is still set. The product-form authorization comment on #289 (2026-08-05) also lists *"加 `spec-approved` label"* as an open item. This repo does gate features this way (cf. issue #96, label `stage:spec-pending-approval`).

**S2 — No tracking issue exists for the backend form.**
The #289 authorization comment explicitly scopes the server-driven form as *"值得单独立项(建议后续开一个 octo-server / octo-smart-summary 的 tracking issue),但不作为本 PR 前置"*, and #1234's closing plan says *"新起 tracking issue + PR"*. Searching this repo's issues (open and closed) returns no tracking issue for this feature. The PR links only the frontend issue, whose own scope section reads *"❌ 后端改动（octo-server / octo-smart-summary 不需要改）"*. The direction is endorsed; the paperwork that authorizes it landing **here** is missing.

**S3 — Idempotency key drops `result_id`, changing agreed re-generate semantics.**
The accepted design in #1234's closing comment is *"幂等(task_id, result_id, source_id)"*, and #289's authorization describes the backend form as *"按 completion run 权威生成一次通知"*. This PR keys on `(task_id, kind, recipient_uid)` with no `result_id`/run discriminator (`internal/notify/notify.go:270`). Consequence: a manual re-generate of the same task (the `allowCompletedTask` path, and the meta re-run path at `internal/worker/meta_processor.go:343`) produces **no** new tip, whereas the superseded implementation deliberately re-sent on a new run. Scheduled tasks happen to re-fire only because `internal/worker/scheduler.go:190` deletes the task's `summary_notification` rows on requeue — i.e. the correct behaviour there is incidental, not designed.

**S4 — Sender identity deviates from #289 AC1.**
AC1 requires *"发送身份为发起总结者的 UID"*. Server-driven delivery posts as the notify bot (`fromUID := NotifyBotUID()` in octo-server `modules/notify/api.go`); `from_uid` in the payload is display metadata only. This may well be the right trade for a server-driven form, but it is a spec change, not an implementation detail, and belongs in the amendment S1/S2 call for.

**S5 — Rollout order is inverted: no client can render this message.**
octo-web `main` (at `c096bbde`) contains no `summaryNotify` const and no `SummaryNotifyContent` class — I grepped `packages/` for both. The PR body states the renderer *"will be opened in Mininglamp-OSS/octo-web separately"* and currently lives on a fork branch. #1234's own disclosure records that an unknown content type on web falls through the factory `default:` and **renders blank**, and that iOS/Android behaviour is unconfirmed. Merging the emitter first therefore writes messages into real group channels that no shipped client can display. Combined with P1-3 below (no independent kill switch) there is no clean way to stop it.

### Not findings

No missing work and no unrequested feature surface: the four scope guards, per-source isolation, and reuse of the notification state machine all match what the PR body promises. Diff is genuinely additive.

---

## 2. Code quality

### P0-1 — The payload never sets a content type; it sets a key nothing reads

`internal/notify/notify.go:306`

```go
payload := map[string]any{
    "content_type": 21,
    "from_uid":     task.CreatorID,
```

The WuKongIM content-type discriminator on the wire is `type`, not `content_type`. Evidence on both sides:

- octo-server: `pkg/cardmsg.IsCardPayload` dispatches on `payload["type"]`, and every payload literal in the notify module is `{"type": 1, "content": ...}` / `{"type": 17, ...}` (`modules/notify/card.go:286`, `modules/notify/space_welcome.go:545`, `modules/notify/action_finalizer.go:364`).
- octo-web: `packages/dmworkbase/src/bridge/globalSearch/GlobalSearchVM.ts:215` reads `const contentType = v.payload.type`; the merge-forward tests assert `payload.type` equals the content type. Every `content_type` occurrence in `packages/` is an HTTP MIME field or a search filter — never a message discriminator.

So this message arrives with **no** content type. The renderer registration (`registerCurrentImMessageContent(MessageContentTypeConst.summaryNotify, ...)`) keys on the resolved content type and will never match, no matter when the web PR lands. The comment on `internal/notify/notify.go:303` correctly names the web const as the source of truth, but the key that carries it is wrong.

### P0-2 — The production transport discards `ChannelType`; group IDs are sent as user targets

`internal/notify/notify.go:255` sets `ChannelType: WireChannelGroup`, and `internal/notify/notify.go:313-317` puts it on `SendMessageRequest`. `InternalNotifyDeliverer.SendMessage` (`internal/notify/client.go:136-166`) never reads `msg.ChannelType`. It builds `notifyReq` (`internal/notify/client.go:81-89`), which has no channel-type field at all, and puts the channel ID into `Targets`.

On the server side, `Targets` are **user UIDs**: `modules/notify/api.go:509-530` dedups them, verifies each against space membership (`n.memberCache.verify`), and delivers via `config.NewPersonalMsgSendReq(uid, fromUID, ...)` — a 1:1 DM per target. The only routes on this API are `/notify`, `/notify/batch`, `/cards/mutate` (`modules/notify/api.go:201-203`); none accepts a channel type. `internal/notify/client.go:15-16` already says as much: *"The state machine only ever builds DM targets, but the enum is kept for target resolution parity."*

Net effect in production: a group channel ID is submitted as a target UID, fails membership verification, comes back in `filtered`, and `post()` (`internal/notify/client.go:204-212`) correctly turns that into an error. **Every group tip fails, 100% of the time.** In the (unlikely) event a group ID collided with a real member UID, the message would be delivered as a private DM to that user — not to the group.

There is no group-send capability to reuse here. Adding one is a real piece of work (new server-side request shape, plus bot-membership/permission handling per group), and it is the load-bearing assumption of the whole PR.

### P1-1 — `group_tip` rows are retried through the DM path

`internal/notify/notify.go:270` stores a group channel ID in `recipient_uid`. The recovery machinery is kind-agnostic:

- `sweepRetryFailed` (`internal/notify/notify.go:719-743`) and `sweepStalePending` (`internal/notify/notify.go:747-774`) select on `status`/`attempt_count` only — `notify_kind` is never filtered.
- Both call `redeliver(row.TaskID, row.NotifyKind, row.RecipientUID)` (`internal/notify/notify.go:798`), which calls `targetForUID` (`internal/notify/target.go:104-114`). That unconditionally returns `ChannelType: WireChannelDM, TargetUID: uid`.
- `redeliver` then builds a **card** via `buildCard(task, "group_tip", errMsg)` — which matches neither the `completed` nor the `failed` branch, so it emits a near-empty card — and calls `deliver` (`internal/notify/notify.go:374`), the DM path with `EnsureFriend`.

With `imDB` wired (which it is in production — `cmd/summary-worker/main.go:113`), `recipientIsActiveMember` proves the group ID is not a space member and fails the attempt with the misleading error *"recipient <group-id> is not an active member of space …"*. The row burns its retry budget and the tip is permanently lost. With a nil `imDB` the pre-check fail-opens and the DM path is actually attempted against a group ID.

Note also that `deliverGroupTipToSource` (`internal/notify/notify.go:277-281`) has no `claimRetry` branch, unlike `deliverToRecipient` (`internal/notify/notify.go:160-172`). So a failed group tip has no in-path retry either. Between the two, the test comment *"attempt_count=1, retry-eligible on sweep"* (`internal/notify/group_tip_test.go`) asserts a recovery property the system does not have.

### P1-2 — Payload omits `from_name`, so the tip renders with an undefined name

`SummaryNotifyContent` is specified in #289 with `fromUID` **and** `fromName`, mirroring `ScreenshotContent`. That class decodes exactly two keys (`packages/dmworkbase/src/Messages/Screenshot/index.tsx:30-32`):

```ts
decodeJSON(content: any): void {
    this.fromUID = content["from_uid"]
    this.fromName = content["from_name"]
}
```

and its `tip` getter falls back to `this.fromName` whenever the local channel-info lookup misses. The payload here sends `from_uid` (correct key) but no `from_name`, so for any recipient whose client has not cached the creator's profile the tip degrades to an undefined/empty name. The server has the data it needs — the notifier already resolves display names from the IM DB elsewhere (`resolveSpaceName`, `internal/notify/notify.go:597`).

### P1-3 — No independent kill switch; the PR body's gradual-rollout claim does not hold

`internal/notify/notify.go:219` gates on the shared `n.cfg.Enabled`, which is `SUMMARY_NOTIFY_ENABLED` (`internal/config/config.go:217`, `cmd/summary-worker/main.go:98-118`) — the master switch for *all* terminal-state notifications. So:

- turning the feature on for the existing creator DMs now also starts broadcasting into every group source, with no separate opt-in;
- turning it off to stop a bad group rollout also kills completed/failed creator DMs **and** the sweep that retries them.

The PR body states *"功能可以在 config 侧灰度关闭而无需回滚代码"*. For a change whose blast radius is "a message in every group that was ever summarized", that claim needs a real per-feature flag behind it.

### P2-1 — The channel-type and payload assertions are tautological

`internal/notify/group_tip_test.go` asserts `msg.ChannelType == WireChannelGroup` and `msg.Payload["content_type"] == 21` against `fakeDeliverer`, which just appends the unmodified struct (`internal/notify/notify_test.go:59-68`). Both assertions pass by construction and would keep passing regardless of P0-1 and P0-2. The isolation test likewise never invokes `Sweep`, so the retry claim in its comment is untested. Also:

```go
if _, hasCard := any(msg.Card).(*notifyCard); hasCard && msg.Card != nil {
```

`msg.Card` is already `*notifyCard`; the assertion is always true and the condition reduces to `msg.Card != nil`. Please simplify.

A test that would have caught P0-2 is one over `InternalNotifyDeliverer` (or its `notifyReq` output) asserting the group destination survives serialization — e.g. extend the existing `httptest` harness in `notify_test.go`.

### P2-2 — `at most once` is overstated across the send/mark crash window

A crash (or a failed `markSent`) between a successful `SendMessage` and the status write leaves a `pending` row that `sweepStalePending` reclaims and re-delivers, and `notifyReq` carries no idempotency key. Today this is masked by P1-1 (the resend takes the DM path and fails), so fixing P1-1 without adding a transport-level idempotency key would surface a genuine duplicate group broadcast. Worth designing for now rather than discovering later.

### P2-3 — Source-resolution failure silently drops every tip for that completion

`internal/notify/notify.go:239-244`: if the source query errors, no dedup row has been created yet, so there is nothing for `Sweep` to find, and terminal callbacks are one-shot. The comment *"on the next trigger/sweep the resolution runs again"* is therefore inaccurate. I rate this P2 rather than blocking: `OnTaskTerminal` has the identical accepted behaviour for `resolveTargets` (`internal/notify/notify.go:122-126`), and for a passive tip dropping on a transient DB error is a reasonable trade. The comment should say so instead of promising recovery.

### P2-4 — Serial fan-out with a 15s timeout per group blocks the worker

`OnGroupTip` iterates sources synchronously (`internal/notify/notify.go:252-262`) and each `deliverGroupTip` allows 15s (`internal/notify/notify.go:299`), while `notifyTaskTerminal` (`internal/worker/processor.go:158`) is on the worker's completion path. A task with many group sources against a degraded downstream can hold the worker for `len(sources) × 15s`. `OnTaskTerminal`'s per-recipient loop has the same shape, so this is a pre-existing pattern rather than a regression — but this PR adds a second serial fan-out on the same path, so it compounds. Consider a shorter per-source timeout or an overall budget.

### P2-5 — Stale schema comment

`migrations/sql/20260626-01-add-summary-notification.sql:5` still documents `notify_kind` as `'completed | failed'`, and `recipient_uid` as a per-recipient UID. Both are now overloaded. A follow-up migration comment (or at least a note in `internal/model/notification.go`) would keep the column semantics discoverable — the `recipient_uid` overload is precisely what makes P1-1 possible.

---

## 3. Overall verdict

**CHANGES_REQUESTED**

Spec section is ❌ (S1–S5) and quality section carries two P0s, so both gates fail.

The two P0s are not polish items: as written, this feature emits a message that no client can classify, over a transport that cannot address a group. I would not iterate on the Go code until the transport question is settled, because the answer changes the shape of `deliverGroupTip` and probably of `SendMessageRequest`.

## 4. Suggested direction

1. **Settle the transport first.** Decide whether octo-server grows a group-capable internal send (new request shape + per-group bot membership/permission checks), or whether the tip is emitted through a path that already reaches group channels. Everything else depends on this answer.
2. **Fix the payload contract.** Emit `type: 21` (not `content_type`) and include `from_name`. Pin it with a test that asserts the *serialized* request, not the in-process struct.
3. **Make the retry machinery kind-aware.** Either give `redeliver` a per-kind rebuild (so `group_tip` reconstructs a group target and the type-21 payload), or filter `group_tip` out of the sweep and give it its own recovery. Add a transport idempotency key while you are in there (P2-2).
4. **Add a dedicated flag** (e.g. `SUMMARY_GROUP_TIP_ENABLED`) independent of `SUMMARY_NOTIFY_ENABLED`, defaulting off.
5. **Restore `result_id`/run to the dedup key** to match the agreed per-completion-run semantics, or get the change of semantics recorded on the spec.
6. **Close the governance items** (`spec-approved` on #289, a tracking issue in this repo) and **land the renderer before the emitter** — or ship the emitter dark behind the flag from (4).

## 5. Additional observations

- `internal/worker/scheduler.go:190` deletes `summary_notification` rows on scheduled requeue, which contradicts the *"A row is NEVER DELETEd"* invariant documented at `internal/model/notification.go`. Not introduced here, but this PR's dedup correctness for scheduled tasks now depends on that deletion, so the invariant comment is actively misleading. Worth a separate cleanup.
- `OnGroupTip` gates on `task.Status` while `OnTaskTerminal` gates on the `status` argument passed by the caller. Every current call site writes the terminal status durably before notifying, so the two agree today. A future call site that notifies before committing would make them diverge — with the group tip firing off a stale `Completed`. Reading the same `status` argument would remove the latent hazard.

## 6. Review process note

The feature itself has now been through six review rounds on octo-web#1234 plus this first round on the re-architected backend PR. The pattern the round-6 note identified — incremental hardening of a delivery mechanism — repeats here, one layer down: the state machine is sound but the transport underneath it cannot carry the message. Recommend closing the transport decision and the governance items (S1, S2) *before* the next code round, so round 7 is not spent polishing an emitter that cannot deliver.

wanyiwang06:
@Jerry-Xin @lml2468 @mochashanyao 感谢 3 位 round-1 深度 review。3 位一致锁定的 P1 完全正确 —— `/v1/internal/notify` 契约是 DM-only(`notifyReq{Targets []string}`, 无 channel_type),group 投递从 wire 层就走不通,不是 sweep/redeliver 单点问题。

**采纳架构判断 · Close 本 PR**。经进一步查证参考了截屏 tip (contentType=20) 的实现姿势:

```objc
// octo-ios · WKConversationView.m
-(void) userDidTakeScreenshot {
    [self.conversationContext sendMessage:WKScreenshotContent.new];
}
```

—— **client-side send 用当前用户自己的 IM connection** 直接投,天然带 group send permission(用户已在群里),无需 backend 转发通道。summary tip 采用同一姿势,但**极简化** dedup(采纳 @yujiawei round-6 的 "retire the persistent dedup" 建议):去掉 localStorage + Web Locks,仅保留 in-memory 单飞标记,接受 multi-tab 罕见重复的 accepted trade-off。

**转向**:
- 关闭本 PR(server-driven 姿势)
- 在 octo-web 侧 [PR #1283](https://github.com/Mininglamp-OSS/octo-web/pull/1283) 上加 client-side send + 极简 dedup,与 renderer 一体
- 该姿势与截屏 tip 完全对齐,是产品早已验证的架构

再次感谢深度 review · P1 分析帮我看清了架构方向 🙏

PR #187, opened 2026-08-10
fix(summary): require space membership for member-candidates (close full-user leak)

wanyiwang06:
## 背景(渗透测试 高危)
`GET /api/v1/summary-member-candidates` 泄露**全量组织架构**:`SearchCandidates` 从客户端 query 取 `space_id`,**为空时兜底查询全部真人用户**;且**从不校验调用者是否为该空间成员**(传任意 `space_id` 即可跨空间枚举成员)。

## 修复(照抄 docs / Matter 的空间成员鉴权)
- **必须有可解析的空间**(query `space_id` 或 `X-Space-Id` 头);解析不出 → **403**,不再回退全量;
- **调用者必须是目标空间的在册成员**(`space_member.status=1`),非成员 → **403**;
- `space_member` INNER JOIN 改为**恒定生效**,结果始终按空间收窄。

`SearchChatCandidates` **无需改**:它本就按调用者自己的 `group_member`(`gm.uid = 调用者`)收窄,只返回你所在的群/子区,不会泄露全量、也无跨空间枚举。

## 零回归(已按前端代码验证)
前端 `getMemberCandidates` 只传 `keyword`、从不传 `space_id`,空间靠请求层注入的 `X-Space-Id`(= `currentSpaceId`)带;合法 UI 调用**必带 X-Space-Id 且用户必是本空间成员** → 两道校验自然通过,选参与者功能照旧。全量兜底从来不是给前端用的。

## 测试
`candidates_test.go` 新增(cgo/sqlite):
- 成员 → 200,返回本空间名单(排除自己);
- 非成员 → 403;
- 无空间(无 X-Space-Id / 无 query)→ 403,且不泄露任何用户。

本地 `go build ./internal/api/handler/` ✅、`gofmt` ✅;cgo 单测本机无 gcc,交 CI 运行。

涉及文件:`internal/api/handler/candidates.go`、`internal/api/handler/candidates_test.go`。

### DIFF
--- a/internal/api/handler/candidates.go
+++ b/internal/api/handler/candidates.go
@@ -36,23 +36,51 @@ func (imUser) TableName() string { return "user" }
 
 // SearchCandidates handles GET /api/v1/summary-member-candidates
 // Returns human members of the current Space only (excludes bots).
-// Falls back to all human users if no space_id is available.
+//
+// Security (summary member-candidates full-user leak, pentest 高危): the space
+// scope is REQUIRED and the caller MUST be an active member of that space.
+// Previously an empty space_id fell back to querying ALL human users (org-wide
+// roster leak) and a client-supplied space_id was trusted without a membership
+// check (cross-space enumeration). Both are closed here — mirrors the docs /
+// Matter space-membership authorization.
 func (h *CandidateHandler) SearchCandidates(c *gin.Context) {
 	if h.imDB == nil {
 		c.JSON(http.StatusOK, gin.H{"code": 0, "data": []interface{}{}})
 		return
 	}
 	keyword := c.Query("keyword")
-	// Space ID: prefer explicit query param (sent by frontend), fallback to middleware context.
+
+	// Resolve current user (set by AuthMiddleware via token).
+	currentUID, _ := c.Get("user_id")
+	currentUIDStr, _ := currentUID.(string)
+	if currentUIDStr == "" {
+		c.JSON(http.StatusUnauthorized, gin.H{"code": 40100, "message": "authentication required"})
+		return
+	}
+
+	// Space ID: prefer explicit query param (sent by frontend), fallback to
+	// middleware context (X-Space-Id).
 	spaceIDStr := c.Query("space_id")
 	if spaceIDStr == "" {
 		v, _ := c.Get("space_id")
 		spaceIDStr, _ = v.(string)
 	}
-
-	// Resolve current user (set by AuthMiddleware via token)
-	currentUID, _ := c.Get("user_id")
-	currentUIDStr, _ := currentUID.(string)
+	// Never fall back to "all users": without a resolvable space we cannot scope
+	// the roster, so deny instead of leaking every user.
+	if spaceIDStr == "" {
+		c.JSON(http.StatusForbidden, gin.H{"code": 40300, "message": "space membership required"})
+		return
+	}
+	// The caller must be an active member of the target space; otherwise a
+	// non-member could enumerate that space's roster by supplying its space_id.
+	var memberCnt int64
+	h.imDB.Table("space_member").
+		Where("space_id = ? AND uid = ? AND status = 1", spaceIDStr, currentUIDStr).
+		Count(&memberCnt)
+	if memberCnt == 0 {
+		c.JSON(http.StatusForbidden, gin.H{"code": 40300, "message": "not a member of this space"})
+		return
+	}
 
 	var users []imUser
 
@@ -60,22 +88,16 @@ func (h *CandidateHandler) SearchCandidates(c *gin.Context) {
 	// 1. user.robot = 1 (flag on user row)
 	// 2. uid in robot table (some system bots have robot=0, e.g. BotFather)
 	// 3. uid is not a 32-char hex string (system accounts like fileHelper/botfather)
+	//
+	// The space_member join is now UNCONDITIONAL (space is guaranteed non-empty
+	// and the caller is a verified member), so results are always space-scoped.
 	q := h.imDB.Table("user u").Select("u.uid, u.name").
 		Where("u.robot = 0").
 		Where("u.uid NOT IN (SELECT robot_id FROM robot)").
-		Where("LENGTH(u.uid) = 32")
-
-	if spaceIDStr != "" {
-		// Filter by Space members
-		q = q.
-			Joins("INNER JOIN space_member sm ON sm.uid = u.uid").
-			Where("sm.space_id = ? AND sm.status = 1", spaceIDStr)
-	}
-
-	// Exclude the currently logged-in user (task creator doesn't add themselves)
-	if currentUIDStr != "" {
-		q = q.Where("u.uid != ?", currentUIDStr)
-	}
+		Where("LENGTH(u.uid) = 32").
+		Joins("INNER JOIN space_member sm ON sm.uid = u.uid").
+		Where("sm.space_id = ? AND sm.status = 1", spaceIDStr).
+		Where("u.uid != ?", currentUIDStr)
 
 	if keyword != "" {
 		q = q.Where("u.name LIKE ? OR u.username LIKE ?", "%"+keyword+"%", "%"+keyword+"%")
@@ -187,7 +209,7 @@ func (h *CandidateHandler) SearchChatCandidates(c *gin.Context) {
 		if currentUIDStr != "" {
 			// Use group_member instead of thread_member so that all threads in the
 			// user's groups are returned, not just threads the user has posted in.
-			q = q.Joins("INNER JOIN group_member gm ON gm.group_no" + h.collate + " = t.group_no").
+			q = q.Joins("INNER JOIN group_member gm ON gm.group_no"+h.collate+" = t.group_no").
 				Where("gm.uid = ? AND gm.is_deleted = 0", currentUIDStr)
 		}
 		if spaceIDStr != "" {
--- a/internal/api/handler/candidates_test.go
+++ b/internal/api/handler/candidates_test.go
@@ -6,6 +6,7 @@ import (
 	"encoding/json"
 	"net/http"
 	"net/http/httptest"
+	"strings"
 	"testing"
 
 	"github.com/Mininglamp-OSS/octo-smart-summary/internal/middleware"
@@ -23,6 +24,10 @@ func setupCandidateImDB(t *testing.T) *gorm.DB {
 	imDB.Exec(`CREATE TABLE "group" (group_no TEXT NOT NULL, name TEXT, space_id TEXT, status INTEGER DEFAULT 1)`)
 	imDB.Exec(`CREATE TABLE thread (id INTEGER PRIMARY KEY, short_id TEXT, name TEXT, group_no TEXT, status INTEGER DEFAULT 1, message_count INTEGER DEFAULT 0)`)
 	imDB.Exec(`CREATE TABLE group_member (group_no TEXT NOT NULL, uid TEXT NOT NULL, is_deleted INTEGER DEFAULT 0)`)
+	// Tables for SearchCandidates (member-candidates) space-membership authz.
+	imDB.Exec(`CREATE TABLE user (uid TEXT, name TEXT, username TEXT, robot INTEGER DEFAULT 0)`)
+	imDB.Exec(`CREATE TABLE robot (robot_id TEXT)`)
+	imDB.Exec(`CREATE TABLE space_member (space_id TEXT, uid TEXT, status INTEGER DEFAULT 1)`)
 	return imDB
 }
 
@@ -356,5 +361,93 @@ func TestSearchChatCandidates_GroupAndDirectHaveIsArchivedFalse(t *testing.T) {
 	}
 	if v != false {
 		t.Errorf("group is_archived should be false, got %v", v)
+	}
+}
+
+// --- SearchCandidates (member-candidates) space-membership authorization ---
+
+func setupMemberRouter(h *CandidateHandler) *gin.Engine {
+	gin.SetMode(gin.TestMode)
+	r := gin.New()
+	r.Use(middleware.AuthMiddleware(&mockTokenResolver{}), middleware.SpaceMiddleware())
+	r.GET("/api/v1/summary-member-candidates", h.SearchCandidates)
+	return r
+}
+
+// doMemberRequest issues a member-candidates request. When spaceHeader is empty
+// the X-Space-Id header is omitted (simulating a request with no space scope).
+func doMemberRequest(r *gin.Engine, userID, spaceHeader string) *httptest.ResponseRecorder {
+	w := httptest.NewRecorder()
+	req := httptest.NewRequest("GET", "/api/v1/summary-member-candidates", nil)
+	if userID != "" {
+		req.Header.Set("Token", userID)
+	}
+	if spaceHeader != "" {
+		req.Header.Set("X-Space-Id", spaceHeader)
+	}
+	r.ServeHTTP(w, req)
+	return w
+}
+
+const (
+	memCaller   = "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa" // 32-hex; space1 member
+	memPeer     = "bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb" // 32-hex; space1 member
+	memOutsider = "cccccccccccccccccccccccccccccccc" // 32-hex; NOT in space1
+)
+
+func seedMemberSpace(imDB *gorm.DB) {
+	imDB.Exec(`INSERT INTO user (uid, name, username, robot) VALUES (?, 'Caller', 'caller', 0)`, memCaller)
+	imDB.Exec(`INSERT INTO user (uid, name, username, robot) VALUES (?, 'Peer', 'peer', 0)`, memPeer)
+	imDB.Exec(`INSERT INTO user (uid, name, username, robot) VALUES (?, 'Outsider', 'outsider', 0)`, memOutsider)
+	// space1 members: caller + peer (outsider intentionally absent).
+	imDB.Exec(`INSERT INTO space_member (space_id, uid, status) VALUES ('space1', ?, 1)`, memCaller)
+	imDB.Exec(`INSERT INTO space_member (space_id, uid, status) VALUES ('space1', ?, 1)`, memPeer)
+}
+
+func TestSearchCandidates_MemberGetsSpaceRoster(t *testing.T) {
+	imDB := setupCandidateImDB(t)
+	seedMemberSpace(imDB)
+	r := setupMemberRouter(NewCandidateHandler(imDB, -1))
+
+	w := doMemberRequest(r, memCaller, "space1")
+	if w.Code != http.StatusOK {
+		t.Fatalf("member should get 200, got %d body=%s", w.Code, w.Body.String())
+	}
+	body := w.Body.String()
+	if !strings.Contains(body, memPeer) {
+		t.Fatalf("expected peer in roster, body=%s", body)
+	}
+	if strings.Contains(body, memCaller) {
+		t.Fatalf("caller should be excluded from own roster, body=%s", body)
+	}
+}
+
+func TestSearchCandidates_NonMemberForbidden(t *testing.T) {
+	imDB := setupCandidateImDB(t)
+	seedMemberSpace(imDB)
+	r := setupMemberRouter(NewCandidateHandler(imDB, -1))
+
+	// Outsider is authenticated but not a member of space1 → 403, no roster.
+	w := doMemberRequest(r, memOutsider, "space1")
+	if w.Code != http.StatusForbidden {
+		t.Fatalf("non-member should get 403, got %d body=%s", w.Code, w.Body.String())
+	}
+	if strings.Contains(w.Body.String(), memPeer) {
+		t.Fatalf("non-member must not see any roster, body=%s", w.Body.String())
+	}
+}
+
+func TestSearchCandidates_NoSpaceNoFullDump(t *testing.T) {
+	imDB := setupCandidateImDB(t)
+	seedMemberSpace(imDB)
+	r := setupMemberRouter(NewCandidateHandler(imDB, -1))
+
+	// No X-Space-Id header and no space_id query → must NOT fall back to all users.
+	w := doMemberRequest(r, memCaller, "")
+	if w.Code != http.StatusForbidden {
+		t.Fatalf("missing space must be 403 (no full-user dump), got %d body=%s", w.Code, w.Body.String())
+	}
+	if strings.Contains(w.Body.String(), memPeer) || strings.Contains(w.Body.String(), memOutsider) {
+		t.Fatalf("no users must be leaked without a space, body=%s", w.Body.String())
 	}
 }


Jerry-Xin (dismissed):
The PR is relevant to `octo-smart-summary` and closes the described member-candidate roster leak by requiring authentication, requiring a space scope, checking active `space_member` membership, and always space-scoping the result query.

💬 Non-blocking

- 🟡 Warning: [internal/api/handler/candidates.go:77](/Users/mlamp/.openclaw/tmp/pr-review-6vi4vauu/repo/internal/api/handler/candidates.go:77) checks only `space_member.status = 1`. Nearby authorization code such as `ShareHandler.activeSpaceMember` also joins `space` and requires `space.status = 1`. If inactive/deleted spaces can retain active membership rows, this endpoint could still return that inactive space roster. Consider matching the stricter pattern if inactive spaces must be hidden.

- 🔵 Suggestion: [internal/api/handler/candidates_test.go:407](/Users/mlamp/.openclaw/tmp/pr-review-6vi4vauu/repo/internal/api/handler/candidates_test.go:407) should also assert `memOutsider` is absent in the successful member response. The current test proves `memPeer` is present and the caller is excluded, but it would not catch an accidental “all users except self” regression.

- 🔵 Suggestion: [internal/api/handler/candidates_test.go:369](/Users/mlamp/.openclaw/tmp/pr-review-6vi4vauu/repo/internal/api/handler/candidates_test.go:369) uses permissive test middleware while production mounts this route under `StrictAuthMiddleware` and `StrictSpaceMiddleware`. Handler-level fail-closed behavior is covered, but one production-stack-style test would better lock the real route behavior.

✅ Highlights

- The endpoint now fails closed for missing auth and missing space scope.
- The caller membership check closes the cross-space enumeration path for non-members.
- The unconditional `space_member` join keeps returned candidates space-scoped.
- Queries remain parameterized; no SQL injection concern found in the changed code.

Verification caveat: I attempted `go test ./internal/api/handler -run 'TestSearchCandidates|TestSearchChatCandidates'`, but this checkout cannot link locally because `libtokenizers` is missing.

Jerry-Xin (approved):
The PR is relevant to `octo-smart-summary` and closes the described member-candidate roster leak by requiring authentication, requiring a space scope, checking active `space_member` membership, and always space-scoping the result query.

💬 Non-blocking

- 🟡 Warning: `internal/api/handler/candidates.go:77` checks only `space_member.status = 1`. Nearby authorization code such as `ShareHandler.activeSpaceMember` also joins `space` and requires `space.status = 1`. If inactive/deleted spaces can retain active membership rows, this endpoint could still return that inactive space roster. Consider matching the stricter pattern if inactive spaces must be hidden.

- 🔵 Suggestion: `internal/api/handler/candidates_test.go:407` (`TestSearchCandidates_MemberGetsSpaceRoster`) should also assert `memOutsider` is absent in the successful member response. The current test proves `memPeer` is present and the caller is excluded, but it would not catch an accidental "all users except self" regression.

- 🔵 Suggestion: the new tests (`internal/api/handler/candidates_test.go`) use permissive test middleware, while production mounts this route under `StrictAuthMiddleware` and `StrictSpaceMiddleware`. Handler-level fail-closed behavior is covered, but one production-stack-style test would better lock the real route behavior.

✅ Highlights

- The endpoint now fails closed for missing auth and missing space scope.
- The caller membership check closes the cross-space enumeration path for non-members.
- The unconditional `space_member` join keeps returned candidates space-scoped.
- Queries remain parameterized; no SQL injection concern found in the changed code.
- Single call-site (router `/summary-member-candidates` → `SearchCandidates`); the census confirms no unguarded path bypasses the new checks. The membership `Count` also fails closed (a DB error leaves `memberCnt == 0` → 403).

Verification caveat: a local `go test ./internal/api/handler` run could not link in this checkout because `libtokenizers` is missing; cgo/sqlite tests are left to CI.

yujiawei (approved):
# Code Review — PR #187 (octo-smart-summary)

Reviewed at head `d8f0750`, merge-base `62a2d15`. I checked out the branch, read the surrounding handler/middleware/router code, and ran the test suite locally.

## 1. Specification compliance

**Spec: ✅**

The description states three changes plus one explicit non-change. All four hold:

| Claim | Verified |
|---|---|
| A resolvable space is required (`space_id` query **or** `X-Space-Id`); otherwise 403, no full-user fallback | `candidates.go:63-73` ✅ |
| Caller must be an active member of the target space (`space_member.status=1`); otherwise 403 | `candidates.go:74-83` ✅ |
| `space_member` INNER JOIN is now unconditional | `candidates.go:98-99` ✅ |
| `SearchChatCandidates` needs no change | ✅ — see below |

- **Missing work:** none.
- **Out-of-scope additions:** none functional. `candidates.go:56-59` adds a 401 guard that is unreachable in production (the route is mounted on the `v1` group behind `StrictAuthMiddleware`, `router.go:70` / `router.go:113`, which already rejects an empty `user_id`). It is harmless defence-in-depth and is reachable under the lenient test middleware, so I am not counting it as scope creep — just noting it is untested dead code in the deployed path.
- **Deviations:** none against the enumerated bullets. One imprecision in the description is discussed as P2-1.

**On the `SearchChatCandidates` no-change claim — confirmed.** I traced all three branches:
- Groups (`candidates.go:165-168`): `INNER JOIN group_member gm ... WHERE gm.uid = <caller>`.
- Threads (`candidates.go:209-213`): same `group_member` anchor on the caller.
- Direct (`candidates.go:242-245`): `WHERE ce.uid = <caller>`.

Every branch is anchored to the caller's own membership rows, so a supplied `space_id` can only *narrow* the result set. There is no unscoped fallback and no cross-space enumeration. Leaving it untouched is correct, and the reasoning in the description is sound rather than merely asserted.

**On the pentest vector being genuinely closed — verified by mutation.** I reverted `candidates.go` to `origin/main`, kept the new tests, and re-ran them:

```
--- PASS: TestSearchCandidates_MemberGetsSpaceRoster
--- FAIL: TestSearchCandidates_NonMemberForbidden
      non-member should get 403, got 200 body={"code":0,"data":[{...,"name":"Caller"},{...,"name":"Peer"}]}
--- FAIL: TestSearchCandidates_NoSpaceNoFullDump
      missing space must be 403 (no full-user dump), got 200 body={"code":0,"data":[{...,"name":"Peer"},{...,"name":"Outsider"}]}
```

Both security tests fail on `main` and reproduce exactly the two reported vectors (cross-space enumeration and the org-wide dump). These are real regression tests, not tests that pass either way. `go test ./internal/api/handler/` is green on this branch (11.2s, full package), so the three new schema tables in `setupCandidateImDB` (`candidates_test.go:27-29`) do not disturb the existing `SearchChatCandidates` tests.

Note that the handler check is load-bearing and not redundant with middleware: `StrictSpaceMiddleware` only rejects a missing `X-Space-Id` for POST/PUT/DELETE (`middleware/space.go:160-176`, `isWriteMethod`), so a GET reaches the handler with an empty space. This matches the "fail-closed hard gate" comments already at `task.go:482-486` and `personal.go:155-160`.

## 2. Code quality

**Quality: Approved** — no P0/P1. Four P2s, none blocking.

### P2-1 — Membership probe omits the `space.status = 1` join that `share.go` requires

`candidates.go:76-79`:
```go
h.imDB.Table("space_member").
    Where("space_id = ? AND uid = ? AND status = 1", spaceIDStr, currentUIDStr).
    Count(&memberCnt)
```
The repo's existing membership predicate, `activeSpaceMember` at `share.go:93-102`, additionally requires the space itself to be active:
```sql
SELECT COUNT(*) FROM space_member sm
INNER JOIN space s ON s.space_id = sm.space_id AND s.status = 1
WHERE sm.space_id = ? AND sm.uid = ? AND sm.status = 1
```
`notify.go:305-312` uses the same shape. Because `space_member` rows are retained when a space is deactivated (that is precisely why those two call sites join `space`), a member of a **deactivated** space can still pull that space's active roster here.

**Why I rate this P2 and not blocking**, despite it being an authorization gap: the repo is not actually uniform on this clause. `peerInSpace` at `bot_summary_create.go:427-441` is also an authorization primitive — its own comment ties it to issue #181 P1-1, "without this check a space-scoped bot token can reach the owner's DMs with peers outside that space" — and it omits the `space` join too, using the identical `Table("space_member").Where(... status = 1).Count()` shape this PR adopts. So the new code matches an accepted in-repo pattern; the gap is a codebase-wide consistency issue rather than something this PR introduces. The residual exposure is also confined to callers holding a genuine active membership row for the target space (stale access after deactivation), not to unauthorized third parties, and it does not reopen either reported vector. Blocking an active org-wide-leak fix on it would be the wrong trade.

Recommendation: extract one shared `activeSpaceMember(spaceID, uid) (bool, error)` helper and route `candidates.go`, `share.go`, and `peerInSpace` through it, in a follow-up. That converges the three variants and removes the ambiguity permanently. The description's framing — "mirrors the docs / Matter space-membership authorization" — is a little generous given the clause difference; the three enumerated bullets are what was actually implemented.

### P2-2 — `Count(...).Error` is dropped, so a DB failure is reported as 403

`candidates.go:77-79` ignores the error. The behaviour is **fail-closed**, which is the right default: GORM leaves `memberCnt` at 0 on error, so a failure denies rather than leaks. The cost is observability and diagnosability — a `space_member` outage, a timeout, or a missing-table deploy skew surfaces to every user as `403 not a member of this space`, which reads as an authorization decision and will send whoever is on call chasing a permissions bug instead of a database one.

Both existing call sites handle the error explicitly and document their choice: `share.go:99-101` (`return err == nil && count > 0`, fail-closed) and `notify.go:313-316` (logs, then fails **open**, with a comment explaining why a transient error must not become a dropped notification). Suggest matching that discipline — keep fail-closed, but capture the error and either log it or return 500:

```go
if err := h.imDB.Table("space_member").
    Where("space_id = ? AND uid = ? AND status = 1", spaceIDStr, currentUIDStr).
    Limit(1).Count(&memberCnt).Error; err != nil {
    log.Printf("[candidates] membership probe failed space=%s uid=%s: %v", spaceIDStr, currentUIDStr, err)
    c.JSON(http.StatusInternalServerError, gin.H{"code": 5001, "message": "membership check failed"})
    return
}
```
(`peerInSpace` also adds `Limit(1)` to the probe — worth copying.)

### P2-3 — Client-supplied `?space_id=` still overrides the server-injected header

`candidates.go:61-67` prefers the query parameter over the middleware-resolved space. Now that membership is enforced this is no longer a leak, so it is not blocking. Two residual concerns:

1. **Tenant-context mixing.** A user who belongs to both space A and space B can be operating in A (`X-Space-Id: A`) and still fetch B's roster with `?space_id=B`, then seed an A-scoped summary with B's members. Whether that is contained depends entirely on the participant validation on `POST /summaries` and `POST /summaries/:id/members`; the picker itself no longer guards it.
2. **Divergence from the established pattern.** Every other space-scoped read in this codebase resolves the scope server-side only, via `middleware.GetSpaceID(c)` — `task.go:482`, `personal.go:154`, `personal.go:1073` — each with a "fail-closed hard gate" comment, and the bot realm goes further (`router.go:56-58`: "Identity and space both come from verify-bot"). This handler is now the one place where the client can redirect the space scope.

Since the description states the frontend never sends `space_id` and relies on the injected header, dropping the query parameter entirely and using `middleware.GetSpaceID(c)` would tighten this to match the rest of the package at no behavioural cost. Worth confirming no other caller depends on it first.

### P2-4 — Test coverage gaps (three, all minor)

The three new tests are well-targeted and mutation-verified above. Gaps worth a follow-up:

- **The `?space_id=` vector has no test.** `doMemberRequest` (`candidates_test.go:379-390`) only ever sets the `X-Space-Id` header, yet the arbitrary-`space_id` parameter is one of the two vectors in the report. The handler funnels both into the same `spaceIDStr`, so coverage is functionally equivalent — but a test that passes `?space_id=space1` as a non-member would pin the query-parameter path against future refactors, which matters most for exactly the code path P2-3 proposes changing.
- **The new 401 branch (`candidates.go:56-59`) is untested.** It is reachable under the test router's lenient `AuthMiddleware`, so `doMemberRequest(r, "", "space1")` would cover it in one line.
- **Router fidelity.** `setupMemberRouter` (`candidates_test.go:369-375`) mounts `AuthMiddleware` + `SpaceMiddleware`, not the production `StrictAuthMiddleware` + `StrictSpaceMiddleware` pair from `router.go:70`. This is the established convention across the package (`auth_test.go`, `task_batch_status_test.go`, and others do the same) so I am not asking for a change, but it does mean the tests do not prove the middleware/handler composition, only the handler.

### Nits (non-blocking)

- `candidates.go:94-100`: `Select("u.uid, u.name")` has no `DISTINCT`. If duplicate active `space_member` rows exist for a `(space_id, uid)` pair the join will duplicate users in the response. Pre-existing on the space-scoped path, and the sibling thread query at `candidates.go:198` does use `SELECT DISTINCT`.
- `candidates.go:98`: this join is now the only code path, and it omits the `h.collate` clause that the same file applies to cross-table joins because of a MySQL collation mismatch (`candidates.go:15`, `candidates.go:212`). The existing `uid` comparisons repo-wide are collate-free, so the columns presumably agree and this is likely fine — but since the fallback that used to work without the join is gone, a collation mismatch here would now mean an empty picker for everyone rather than a degraded one. Cheap to confirm against the prod schema.
- `candidates.go:57` / `:71` / `:81`: new error codes `40100` and `40300`. The package currently uses `401`, `4010`, and `40100` for unauthorized across three sites, so there is no single convention to violate — but confirm the clients key off HTTP status rather than `code`, since these two `code` values are new to the API surface.
- `bot_summary_create.go:332` documents the DM space check as "mirroring the DM path candidates.go:231 already uses". This diff shifts that block to roughly `candidates.go:250-258`, so the cross-reference is now stale.

## 3. Verdict

**APPROVE** — spec ✅ and quality Approved (no P0/P1).

This closes a real high-severity leak, the regression tests demonstrably fail without the fix, the fail-closed direction is correct, the decision not to touch `SearchChatCandidates` is justified rather than assumed, and the reasoning is documented in the code rather than only in the PR description. The four P2s are hardening and consistency work that should not gate a security fix.

## 4. Recommended follow-ups

1. Extract a single `activeSpaceMember(spaceID, uid) (bool, error)` helper including the `space.status = 1` join, and migrate `candidates.go:76`, `share.go:93`, and `bot_summary_create.go:427` onto it (P2-1, P2-2 together).
2. Drop the `?space_id=` query parameter in favour of `middleware.GetSpaceID(c)` so this handler matches the server-resolved-scope discipline used everywhere else (P2-3).
3. Add the two one-line tests from P2-4 (query-parameter vector, 401 branch).

## 5. Additional observations outside this diff

- **Client coordination before release.** The description verifies the web client only, but iOS (`Modules/OctoContext/.../OctoSummaryAPI.m`) and Android (`wkbase/.../summary/api/SummaryApiService.kt`) both call `summary-member-candidates`. Both inject `X-Space-Id` globally via an interceptor, so the normal path is fine — but the iOS injection in `WKApp.m` sends an **empty** `X-Space-Id` when no space is selected (`currentSpaceId.length > 0 ? currentSpaceId : @""`), which lands on the new 403. That is the intended fail-closed outcome, not a bug in this PR; the question is whether the mobile pickers degrade gracefully on 403 or surface a raw error, and whether a legitimate "no space selected" state exists that product wants served differently. Worth a heads-up to the client owners rather than a change here.
- **`SUMMARY_CHAT_CANDIDATE_LIMIT` defaults to `-1`** (`config.go:190`), i.e. no `LIMIT` on the roster query. Pre-existing and untouched by this diff, but now that every request is space-scoped and joined, a large space returns its entire member list in one unpaginated payload.
- **`gm.status = 1`.** `share.go:117` gates group membership on `gm.status = 1 AND gm.is_deleted = 0`, whereas the `SearchChatCandidates` group/thread joins (`candidates.go:166`, `:212`) check only `gm.is_deleted = 0`. Same class of predicate drift as P2-1, entirely outside this diff.

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #187 (octo-smart-summary)

## Summary

This PR closes the pentest-reported full-user leak in `GET /api/v1/summary-member-candidates`. The handler now (1) rejects unauthenticated callers with 401, (2) rejects requests with no resolvable `space_id` with 403 instead of falling back to an org-wide user dump, and (3) verifies the caller is an active member of the target space (`space_member.status = 1`) before running the roster query, which now joins `space_member` unconditionally. Three new sqlite-based tests cover the member / non-member / no-space paths. The fix lands its stated security goal: both the empty-space full dump and cross-space roster enumeration by non-members are closed. The remaining items below are hardening completeness, not a reintroduction of the leak.

## Verification

Local checks at head `d8f07506` (merge-base `62a2d15`):
- `go build ./...` — clean
- `go vet ./internal/api/handler/` — clean
- `go test ./internal/api/... ./internal/middleware/...` — pass, including the three new `TestSearchCandidates_*` tests

Static tracing at head `d8f07506`:
- ✅ **Production middleware chain** — `internal/api/router/router.go:69-70` mounts the v1 group behind `StrictAuthMiddleware` + `StrictSpaceMiddleware`; strict auth aborts unauthenticated requests, while strict space allows an empty `space_id` on GET, so the new in-handler 403 gate is exactly where the fix must live.
- ✅ **Context value types** — `user_id` / `space_id` are set as strings by every middleware that sets them (`internal/middleware/space.go`), so the `.(string)` assertions at `internal/api/handler/candidates.go:54` and `:63` cannot mistype.
- ✅ **Fail-closed** — empty space → 403 (`:71`); membership probe failure keeps memberCnt=0 → 403 (`:80`); no path returns roster data without a verified space.
- ✅ **SQL injection** — all new predicates are parameterized; keyword LIKE behavior is unchanged from base.

## Findings

No P0/P1 issues; three P2 items and two nits below.

### P2 — Membership probe skips the parent-space status cascade (`internal/api/handler/candidates.go:77`)

The repo's canonical active-member checks — `internal/api/handler/share.go:99-103` (`activeSpaceMember`) and `internal/notify/notify.go:307` — both require `space_member sm INNER JOIN space s ON s.space_id = sm.space_id AND s.status = 1` in addition to `sm.status = 1`. The new probe checks only the member-row status, so a caller holding an active member row in a space whose `space.status != 1` (disabled/archived) can still enumerate that space's roster here, while share creation for the same space is denied — inconsistent authorization between sibling space-scoped paths (container disable does not cascade to member authorization). Diff-scope: newly introduced by this PR, but strictly narrower than pre-fix behavior (org dump and cross-space enumeration by non-members are closed; residual exposure is existing members of a disabled space, uid+name only), hence P2 rather than P1. Fix direction: reuse the canonical join query for the probe (or add the `space` join to both probe and roster query) so the endpoint matches the "mirrors the docs / Matter space-membership authorization" claim in its own comment.

### P2 — Membership probe ignores query errors; DB outage masquerades as 403 (`internal/api/handler/candidates.go:79`)

`.Count(&memberCnt)` is consumed without checking `.Error`. On an IM DB failure memberCnt stays 0 and every request silently returns 403 "not a member of this space" — fail-closed and leak-safe, but indistinguishable from genuine non-membership and invisible in logs. Compare `internal/api/handler/share.go:103` (`err == nil && count > 0`) and `peerInSpace` (`internal/api/handler/bot_summary_create.go:423`), which propagates the error. Diff-scope: new code introduced by this PR. Fix direction: check `.Error` and return 500 (and/or log) so an outage is not masked as an authorization decision.

### P2 — New gate branches without regression tests (`internal/api/handler/candidates_test.go:407`)

The three new tests cover member roster, non-member 403, and no-space 403, but leave gate branches untested: (a) unauthenticated request → the new 401 at `internal/api/handler/candidates.go:56` is reachable under the tests' lenient `AuthMiddleware` yet unverified; (b) the `?space_id=` query-param precedence path (now part of the authorization surface) is never exercised — only the X-Space-Id header path is; (c) no test pins the disabled-space semantics from the first finding. Add these cases so every branch of the gate is regression-protected. Diff-scope: new tests introduced by this PR.

### Nit — Response-code style drift (`internal/api/handler/candidates.go:57`)

The handler introduces 5-digit codes (40100/40300) while the middleware uses 4010/401/40001. Cosmetic; align if clients key on these codes.

### Nit — Unrelated whitespace change in SearchChatCandidates (`internal/api/handler/candidates.go:212`)

Removing the spaces around `+` in the `Joins` string is formatting-only (the generated SQL is byte-identical) but unrelated to the security fix; consider dropping it to keep the diff focused.

## Human-verify

1. The P2 at `:77` hinges on product semantics of `space.status != 1` in the IM system (the `space` table lives in the external IM DB, not in this repo's migrations). If a disabled space must hide its roster even from its own members, that finding should be escalated; if space-disable only blocks new activity, the current gate is acceptable. **Not a merge blocker for this PR as filed.**
2. Frontend callers (octo-web) may send `space_id` as query param or `X-Space-Id` header — both remain supported and both are now membership-checked; a quick composer-UI smoke test is worth doing. **Not a merge blocker.**

- **Sibling endpoint parity (C1):** `GET /api/v1/summary-chat-candidates` sits behind the same `StrictAuthMiddleware`; with a guaranteed-non-empty `user_id` in production, its group/thread branches always join `group_member gm.uid = <caller>` and its direct branch requires auth, so it has no analogous unauthenticated org-wide dump in production. The `if currentUIDStr != ""` guards there are defense-in-depth reachable only under the lenient test middleware.
- **Bot realm:** the `/api/v1/bot` mount uses `StrictBotAuthMiddleware` and does not register either candidates endpoint; no new capability surface for bot tokens.
- **Roster query scoping:** the `space_member` join is now unconditional with `sm.status = 1` and the caller excluded; bot-exclusion predicates unchanged from base.
- **Data flow:** `user_id`/`space_id` traced from the middleware `c.Set` sites through the type assertions to every consumer; no empty-value/short-circuit surprises — the gate returns before any roster work when either is missing.
- **PERSISTENCE/CACHE & SOURCE-COVERAGE families:** the change is read-only (no cache/state writes); the removed "all users" fallback source is the intentional security fix, not accidental data loss.

## Verdict: COMMENT

The PR achieves its stated goal — the full-user leak and cross-space enumeration are closed, fail-closed, with passing regression tests — and introduces no P1. The three P2 items are hardening completeness (canonical space-status cascade, error visibility, test branches) and are best addressed in this PR or an immediate follow-up.

[Octo-Q] verdict: APPROVE — 无 P0/P1；pentest 全量用户泄露（空 space 全量 dump + 跨 space 枚举）在 head d8f0750 已验证关闭（fail-closed，测试通过）。三个 P2 为加固完整性问题：成员校验漏 space.status 级联（对齐 share.go/notify.go canonical 查询）、Count 错误未检查、新门控分支缺测试。请 Opus 终审确认 space.status!=1 的产品语义（Human-verify #1）后拍板。

yujiawei:
Follow-up to my approval — one correction to the automated first-pass review above, so the author does not act on it.

**The "unrelated whitespace change" nit at `candidates.go:212` is not unrelated, and reverting it would break `gofmt`.**

The nit suggests dropping the change from `"…group_no" + h.collate + " = t.group_no"` to `"…group_no"+h.collate+" = t.group_no"` "to keep the diff focused". That change is mandated by `gofmt`, not a stylistic choice. `origin/main` was gofmt-dirty on exactly this line, and this PR fixes it:

```
$ git show origin/main:internal/api/handler/candidates.go > /tmp/main.go && gofmt -l /tmp/main.go
/tmp/main.go
$ gofmt -d /tmp/main.go
-  q = q.Joins("INNER JOIN group_member gm ON gm.group_no" + h.collate + " = t.group_no").
+  q = q.Joins("INNER JOIN group_member gm ON gm.group_no"+h.collate+" = t.group_no").
```

And at this PR's head the file is clean:

```
$ gofmt -l internal/api/handler/candidates.go     # (no output)
```

Reverting the line reintroduces the violation — `gofmt -l` flags the file again and `gofmt` rewrites it straight back. (`gofmt` drops the spaces because the `+` expression sits at greater depth inside the chained call, which is why the sibling `Joins` on the line above never had them.)

So: keep line 212 as it is. It is a one-line drive-by gofmt fix in a file this PR already touches, which is the right place for it.

Everything else in the first-pass review matches my own findings — the `space.status` cascade, the dropped `Count(...).Error`, and the untested gate branches are the same three P2s I listed, all non-blocking. My **APPROVE** stands unchanged.

PR #188, opened 2026-08-10
feat(summary): expose creator_bot_id + creator_bot_name on list/detail/create

wanyiwang06:
## 目的
bot 创建的总结在数据层已可区分(`trigger_type=TriggerBot` + `creator_bot_id`,`creator_id`=owner),但 API 响应没透出 bot 身份,前端无法标注"由 <bot> 创建"。本 PR 把 bot 身份透出。

## 改动
- `GetSummaries`(列表)与 `GetSummary`(详情)用 `service.ResolveUserName(creator_bot_id)` 解析 bot 显示名,响应新增 `creator_bot_id` + `creator_bot_name`;详情另加 `creator_name`(owner),供前端展示"由 <bot> 代 <owner> 创建"。
- `botCreateResponse` 增加 `creator_bot_name`,便于刚建完即时展示。
- 无迁移;worker/生成链不动。

## 测试
`bot_summary_create_test.go` 新增 `TestCreateBotSummaryResponseIncludesBotIdentity`:bot 创建响应含非空 `creator_bot_id`/`creator_bot_name`,且 `creator_id=owner`。

本地 `go build ./internal/api/handler/` ✅、`gofmt` ✅;cgo 单测交 CI。

配套前端 PR:octo-web 卡片"由 <bot> 创建"+ Bot 小标、详情"由 <bot> 代 <owner> 创建"。

### DIFF
--- a/internal/api/handler/bot_summary_create.go
+++ b/internal/api/handler/bot_summary_create.go
@@ -65,17 +65,17 @@ func canonicalBotCreateRequestHash(req createBotSummaryReq, sources []canonicalB
 		}
 	}
 	payload := struct {
-		Title            string               `json:"title"`
-		Topic            string               `json:"topic"`
-		TimeRangeStart   string               `json:"time_range_start"`
-		TimeRangeEnd     string               `json:"time_range_end"`
-		Sources          []canonicalBotSource `json:"sources"`
-		OriginChannelID  string               `json:"origin_channel_id"`
-		OriginChannelTp  int                  `json:"origin_channel_type"`
-		IncludeArchived  bool                 `json:"include_archived"`
+		Title           string               `json:"title"`
+		Topic           string               `json:"topic"`
+		TimeRangeStart  string               `json:"time_range_start"`
+		TimeRangeEnd    string               `json:"time_range_end"`
+		Sources         []canonicalBotSource `json:"sources"`
+		OriginChannelID string               `json:"origin_channel_id"`
+		OriginChannelTp int                  `json:"origin_channel_type"`
+		IncludeArchived bool                 `json:"include_archived"`
 	}{
-		Title:           strings.TrimSpace(req.Title),
-		Topic:           strings.TrimSpace(req.Topic),
+		Title: strings.TrimSpace(req.Title),
+		Topic: strings.TrimSpace(req.Topic),
 		// Round the time range to whole minutes so a normal client retry
 		// that recomputes a relative window ("summarize the last hour")
 		// replays cleanly instead of colliding on nanosecond drift. A
@@ -542,5 +542,9 @@ func (h *TaskHandler) findBotIdempotentTaskWithHash(spaceID, botID, key, request
 }
 
 func botCreateResponse(task model.SummaryTask) gin.H {
-	return gin.H{"task_id": task.ID, "task_no": task.TaskNo, "status": task.Status, "trigger_type": task.TriggerType, "creator_id": task.CreatorID, "creator_bot_id": task.CreatorBotID, "created_at": task.CreatedAt.Format(time.RFC3339)}
+	creatorBotName := ""
+	if task.CreatorBotID != "" {
+		creatorBotName = service.ResolveUserName(task.CreatorBotID)
+	}
+	return gin.H{"task_id": task.ID, "task_no": task.TaskNo, "status": task.Status, "trigger_type": task.TriggerType, "creator_id": task.CreatorID, "creator_bot_id": task.CreatorBotID, "creator_bot_name": creatorBotName, "created_at": task.CreatedAt.Format(time.RFC3339)}
 }
--- a/internal/api/handler/bot_summary_create_test.go
+++ b/internal/api/handler/bot_summary_create_test.go
@@ -13,6 +13,7 @@ import (
 	"time"
 
 	"github.com/Mininglamp-OSS/octo-smart-summary/internal/model"
+	"github.com/Mininglamp-OSS/octo-smart-summary/internal/service"
 	"github.com/gin-gonic/gin"
 	"gorm.io/driver/sqlite"
 	"gorm.io/gorm"
@@ -97,8 +98,24 @@ func requestBotCreate(r *gin.Engine, key string, body []byte) *httptest.Response
 func TestCreateBotSummaryResponseIncludesBotIdentity(t *testing.T) {
 	t.Setenv("BOT_SUMMARY_CREATE_ENABLED", "true")
 	_, r, _ := setupBotCreateTest(t)
+	// Install a temporary name resolver and restore the default (identity)
+	// afterwards. Under test no main() runs, so without this the stub
+	// resolver echoes the uid and the creator_bot_name assertion below
+	// could never fail (review P2-3).
+	service.SetUserNameResolver(func(uid string) string {
+		if uid == "bot-1" {
+			return "Bot Display Name"
+		}
+		return uid
+	})
+	t.Cleanup(func() {
+		service.SetUserNameResolver(func(uid string) string { return uid })
+	})
 	w := requestBotCreate(r, "botname-key", botCreateBody("group-a"))
-	if w.Code != http.StatusOK {
+	// A fresh create returns 201 Created; 200 OK is reserved for idempotent
+	// replays (bot_summary_create.go:198/:219). Every sibling test in this
+	// file already expects StatusCreated for a fresh create (review P0-1).
+	if w.Code != http.StatusCreated {
 		t.Fatalf("status=%d body=%s", w.Code, w.Body.String())
 	}
 	var resp struct {
@@ -117,11 +134,37 @@ func TestCreateBotSummaryResponseIncludesBotIdentity(t *testing.T) {
 	if resp.Data.CreatorBotID != "bot-1" {
 		t.Fatalf("creator_bot_id=%q, want bot-1", resp.Data.CreatorBotID)
 	}
-	if resp.Data.CreatorBotName == "" {
-		t.Fatalf("creator_bot_name is empty; want resolved bot name")
+	// Assert the exact resolved display name, not just non-empty: with the
+	// resolver installed above this fails if the handler stops resolving or
+	// resolves the wrong uid.
+	if resp.Data.CreatorBotName != "Bot Display Name" {
+		t.Fatalf("creator_bot_name=%q, want %q", resp.Data.CreatorBotName, "Bot Display Name")
 	}
 	if resp.Data.CreatorID != "owner" {
 		t.Fatalf("creator_id=%q, want owner (the human on whose behalf the bot acts)", resp.Data.CreatorID)
+	}
+}
+
+func TestBotCreateResponseNonBotTaskHasEmptyBotName(t *testing.T) {
+	// Non-bot tasks must not carry a bot label: the CreatorBotID guard keeps
+	// creator_bot_name empty instead of resolving a bogus name for uid ""
+	// (review P2-3 non-bot case). Install a resolver that would poison an
+	// empty uid so the assertion can actually fail if the guard is removed.
+	service.SetUserNameResolver(func(uid string) string {
+		if uid == "" {
+			return "BOGUS"
+		}
+		return uid
+	})
+	t.Cleanup(func() {
+		service.SetUserNameResolver(func(uid string) string { return uid })
+	})
+	resp := botCreateResponse(model.SummaryTask{ID: 1, TaskNo: "T-NONBOT", CreatorID: "owner"})
+	if got, _ := resp["creator_bot_id"].(string); got != "" {
+		t.Fatalf("creator_bot_id=%q, want empty for non-bot task", got)
+	}
+	if got, _ := resp["creator_bot_name"].(string); got != "" {
+		t.Fatalf("creator_bot_name=%q, want empty for non-bot task", got)
 	}
 }
 
--- a/internal/api/handler/bot_summary_create_test.go
+++ b/internal/api/handler/bot_summary_create_test.go
@@ -94,6 +94,37 @@ func requestBotCreate(r *gin.Engine, key string, body []byte) *httptest.Response
 	return w
 }
 
+func TestCreateBotSummaryResponseIncludesBotIdentity(t *testing.T) {
+	t.Setenv("BOT_SUMMARY_CREATE_ENABLED", "true")
+	_, r, _ := setupBotCreateTest(t)
+	w := requestBotCreate(r, "botname-key", botCreateBody("group-a"))
+	if w.Code != http.StatusOK {
+		t.Fatalf("status=%d body=%s", w.Code, w.Body.String())
+	}
+	var resp struct {
+		Data struct {
+			TriggerType    int    `json:"trigger_type"`
+			CreatorID      string `json:"creator_id"`
+			CreatorBotID   string `json:"creator_bot_id"`
+			CreatorBotName string `json:"creator_bot_name"`
+		} `json:"data"`
+	}
+	if err := json.Unmarshal(w.Body.Bytes(), &resp); err != nil {
+		t.Fatalf("unmarshal: %v body=%s", err, w.Body.String())
+	}
+	// Bot-created summaries must expose the acting bot so the frontend can mark
+	// the card as "由 <bot> 创建".
+	if resp.Data.CreatorBotID != "bot-1" {
+		t.Fatalf("creator_bot_id=%q, want bot-1", resp.Data.CreatorBotID)
+	}
+	if resp.Data.CreatorBotName == "" {
+		t.Fatalf("creator_bot_name is empty; want resolved bot name")
+	}
+	if resp.Data.CreatorID != "owner" {
+		t.Fatalf("creator_id=%q, want owner (the human on whose behalf the bot acts)", resp.Data.CreatorID)
+	}
+}
+
 func TestCreateBotSummaryFeatureDisabled(t *testing.T) {
 	t.Setenv("BOT_SUMMARY_CREATE_ENABLED", "false")
 	_, r, _ := setupBotCreateTest(t)
--- a/internal/api/handler/task.go
+++ b/internal/api/handler/task.go
@@ -803,6 +803,14 @@ func (h *TaskHandler) ListSummaries(c *gin.Context) {
 			creatorName = service.ResolveUserName(t.CreatorID)
 		}
 
+		// Bot-created summaries (trigger_type=TriggerBot) carry the acting bot's
+		// uid in creator_bot_id; resolve its display name so the frontend can
+		// mark the card as "由 <bot> 创建".
+		creatorBotName := ""
+		if t.CreatorBotID != "" {
+			creatorBotName = service.ResolveUserName(t.CreatorBotID)
+		}
+
 		// Expose schedule_id on list items so the frontend can detect a scheduled
 		// task by its bound schedule (the authoritative signal) instead of relying
 		// on trigger_type. A manual task that later gets a scheduled-update added
@@ -856,6 +864,8 @@ func (h *TaskHandler) ListSummaries(c *gin.Context) {
 			"trigger_type":     t.TriggerType,
 			"schedule_id":      scheduleIDOut,
 			"creator_id":       t.CreatorID,
+			"creator_bot_id":   t.CreatorBotID,
+			"creator_bot_name": creatorBotName,
 			"participants":     parts,
 			"time_range_start": t.TimeRangeStart.Format(time.RFC3339),
 			"time_range_end":   t.TimeRangeEnd.Format(time.RFC3339),
@@ -1081,6 +1091,14 @@ func (h *TaskHandler) GetSummary(c *gin.Context) {
 		}
 	}
 
+	// Owner + acting-bot display names, so the detail page can show
+	// "由 <bot> 代 <owner> 创建" for bot-created summaries (trigger_type=TriggerBot).
+	creatorName := service.ResolveUserName(task.CreatorID)
+	creatorBotName := ""
+	if task.CreatorBotID != "" {
+		creatorBotName = service.ResolveUserName(task.CreatorBotID)
+	}
+
 	resp := gin.H{
 		"task_id":          task.ID,
 		"task_no":          task.TaskNo,
@@ -1089,6 +1107,9 @@ func (h *TaskHandler) GetSummary(c *gin.Context) {
 		"summary_mode":     task.SummaryMode,
 		"status":           task.Status,
 		"creator_id":       task.CreatorID,
+		"creator_name":     creatorName,
+		"creator_bot_id":   task.CreatorBotID,
+		"creator_bot_name": creatorBotName,
 		"trigger_type":     task.TriggerType,
 		"time_range_start": task.TimeRangeStart.Format(time.RFC3339),
 		"time_range_end":   task.TimeRangeEnd.Format(time.RFC3339),


Jerry-Xin (changes requested):
The PR is in scope for this repository, but it currently introduces a failing test expectation around the bot create response status.

**🔴 Blocking**
- 🔴 Critical: `TestCreateBotSummaryResponseIncludesBotIdentity` expects `http.StatusOK` for a first-time create at `internal/api/handler/bot_summary_create_test.go:101`, but `CreateBotSummary` returns `http.StatusCreated` for new tasks at `internal/api/handler/bot_summary_create.go:229`. `200 OK` is only used for idempotent replays at lines 198 and 219. Update the test to accept/expect `201 Created`, or explicitly perform a replay if the test intends to validate the idempotent response path.

**💬 Non-blocking**
- 🔵 Suggestion: The new test only asserts `creator_bot_name` is non-empty at `internal/api/handler/bot_summary_create_test.go:120`. Since the default `service.ResolveUserName` fallback returns the UID unchanged at `internal/service/summary.go:166-174`, this test would pass even if no real display-name resolution is happening. Consider installing a temporary resolver in the test and asserting the exact bot display name, then restoring it after the test.

**✅ Highlights**
- The new `creator_bot_id` / `creator_bot_name` fields are added consistently across list, detail, and bot-create responses.
- Access control is not weakened; the new fields are only exposed after the existing list/detail authorization paths have already selected visible tasks.

I could not execute the focused Go test locally because linking failed due to missing `libtokenizers`, but the status-code mismatch is visible directly from the checked-out source.

yujiawei (changes requested):
# Code Review — PR #188 (octo-smart-summary)

Reviewed at head `a908a22d31c1787836d8da6c8431e5c69be14b26` against merge-base `62a2d15`. I checked out the branch, built it, and ran the affected test package locally.

Two verdicts below: first whether the diff implements exactly what it says it does, then implementation quality.

## 1. Spec compliance

**Spec: ❌**

The stated goal of this PR is: *"API 响应没透出 bot 身份,前端无法标注『由 &lt;bot&gt; 创建』。本 PR 把 bot 身份透出"* — i.e. give the frontend a **bot display name**.

- **Missing:** none. All three surfaces described in the PR body are wired: list (`internal/api/handler/task.go:747-749`), detail (`internal/api/handler/task.go:976-979`), and bot-create (`internal/api/handler/bot_summary_create.go:549`).
- **Extra:** the detail endpoint's new `creator_name` is declared in the PR body, so it is in scope — but it duplicates data the same response already carries in `participants[].user_name`, and it is computed by a weaker code path than the list endpoint uses (see P2-1). The `gofmt` realignment in `canonicalBotCreateRequestHash` and in the `apiResponse` struct is unrelated cosmetic churn; harmless, but it is not part of the stated change.
- **Deviation (blocking):** the value shipped in `creator_bot_name` / `creator_name` is **not a display name — it is the raw uid**. `service.ResolveUserName` is a stub whose only implementation returns its argument unchanged, and the injection point that would replace it is never called anywhere in the module (see P1-1). So the frontend receives `creator_bot_name == creator_bot_id` and will render `由 bot-1 创建` rather than a bot name. The feature as described is not delivered by this diff.
- **Test claim does not hold:** the PR body states the new test asserts the bot identity fields and that "cgo 单测交 CI". That test currently **fails**, and CI agrees — `Test (race, cgo)` is red on this head (see P0-1).

## 2. Code quality

### 🔴 P0-1 — the new test fails; `Test (race, cgo)` is red on this head

`internal/api/handler/bot_summary_create_test.go:101` requires `http.StatusOK`, but a *first-time* create returns `http.StatusCreated`:

```go
// bot_summary_create_test.go:101
if w.Code != http.StatusOK {
    t.Fatalf("status=%d body=%s", w.Code, w.Body.String())
}
```

```go
// bot_summary_create.go:229 — fresh create
c.JSON(http.StatusCreated, apiResponse{Code: 0, Message: "ok", Data: botCreateResponse(task)})
```

`200 OK` is reserved for idempotent replays (`bot_summary_create.go:198` and `:219`). Every sibling test in the same file already expects `StatusCreated` for a fresh create (lines 167, 228, 244, 279, 327, 347, 387) — this new test is the only one that does not.

Reproduced locally:

```
--- FAIL: TestCreateBotSummaryResponseIncludesBotIdentity (0.00s)
    bot_summary_create_test.go:102: status=201 body={"code":0,"message":"ok","data":{
      "created_at":"2026-08-10T06:52:52Z","creator_bot_id":"bot-1","creator_bot_name":"bot-1",
      "creator_id":"owner","status":0,"task_id":1,"task_no":"ST2026...","trigger_type":4}}
```

CI runs `go test ./...` on the cgo job (`.github/workflows/ci.yml:142`), which is why `Test (race, cgo)` fails while the `CGO_ENABLED=0` job passes. **Fix:** expect `http.StatusCreated`, or issue a second (replay) request if the intent was to exercise the idempotent path.

### 🔴 P1-1 — `creator_bot_name` can never be a name: the resolver has no implementation

```go
// internal/service/summary.go:166-179
func ResolveUserName(uid string) string { return resolveUserNameFn(uid) }

var resolveUserNameFn = func(uid string) string {
    return uid            // <- the only implementation that exists
}

func SetUserNameResolver(fn func(uid string) string) { resolveUserNameFn = fn }
```

`SetUserNameResolver` is **never called** — not in `cmd/summary-api`, not in `cmd/summary-worker`, not in any `init()`, not in any test. And because it lives under `internal/`, no code outside this module can call it either. So `service.ResolveUserName(x) == x`, unconditionally and permanently.

That makes all three new call sites inert as display-name lookups:

- `internal/api/handler/task.go:720` (list) → `creator_bot_name` = bot uid
- `internal/api/handler/task.go:963` and `:966` (detail) → `creator_name` = owner uid, `creator_bot_name` = bot uid
- `internal/api/handler/bot_summary_create.go:547` (create) → `creator_bot_name` = bot uid

The failing test's own response body above is the proof: `"creator_bot_id":"bot-1","creator_bot_name":"bot-1"`.

The codebase already treats this fallback as "unresolved" elsewhere — `internal/api/handler/personal.go:1117` guards with `if resolved := service.ResolveUserName(p.UserID); resolved != p.UserID` precisely because the returned value is often just the uid.

**Fix — use the IM DB handle the handler already holds.** `TaskHandler` has `imDB` (`task.go:42`), and the established pattern for name resolution in this package is to pass it explicitly, exactly as the source-name resolver does:

```go
// internal/service/summary.go:43
func ResolveSourceNameWithType(sourceID string, sourceType int, imDB *gorm.DB) string
// …:108  case 3 (DM): imDB.Table("user").Where("uid = ?", sourceID).Pluck("name", &name)
```

Add the symmetric `ResolveUserNameWithDB(uid string, imDB *gorm.DB) string` (or wire `SetUserNameResolver` once at API/worker startup) and call it from the three new sites. For the list path, batch the unique uids in one query — `internal/worker/processor.go:860` (`batchResolveUserNames`) already implements `SELECT uid, name FROM user WHERE uid IN ?` and can be reused rather than reinvented.

One thing I could not verify from source: whether personal-bot uids are rows in the IM `user` table at all. If they are not, an `imDB` lookup will also come back empty and the bot name needs a different source. Worth confirming against a real environment before settling on the fix.

### 🟡 P2-1 — detail's `creator_name` is weaker than list's, so the two can disagree

The list path prefers the persisted participant name and only then falls back:

```go
// task.go:706-713
creatorName := ""
var creatorParticipant model.SummaryParticipant
if err := h.db.Where("task_id = ? AND user_id = ?", t.ID, t.CreatorID).First(&creatorParticipant).Error; err == nil {
    creatorName = creatorParticipant.UserName
}
if creatorName == "" {
    creatorName = service.ResolveUserName(t.CreatorID)
}
```

The new detail path skips that entirely (`task.go:963`). Since `summary_participant.UserName` is populated with a client-supplied real name for invited members (`task.go:440-445`) and can be refreshed by the worker, the same task can show a real name on the card and a raw uid on the detail page.

`GetSummary` already loads every participant at `task.go:912-913`, so the correct value is in memory — pick the row where `UserID == task.CreatorID` and use its `UserName`, falling back to the resolver. Zero extra queries, and it removes the divergence.

### 🟡 P2-2 — the new assertion is a tautology

```go
// bot_summary_create_test.go:120
if resp.Data.CreatorBotName == "" {
    t.Fatalf("creator_bot_name is empty; want resolved bot name")
}
```

Under the uid fallback this can never fail whenever `creator_bot_id` is non-empty, so it does not test resolution at all — it is the reason P1-1 shipped unnoticed. Install a resolver (or an `imDB` fixture) in the test, assert the **exact** expected display name, and restore it afterwards. Please also add a negative case: a non-bot task must return an empty `creator_bot_name` and must not gain a spurious bot label.

### 🟡 P2-3 — per-row lookup in the list loop (latent, not yet a cost)

`task.go:719-721` sits inside the `ListSummaries` item loop. Today it is free because the resolver is an identity function, but the moment P1-1 is fixed with a naive per-uid DB call this becomes N+1 on every list page — on top of the existing per-row participant query at `task.go:708`. Batch the unique creator and bot uids once before the loop.

### 🟡 P2-4 — I/O inside a response-serialization helper

`botCreateResponse` (`bot_summary_create.go:545-549`) is a pure `model → gin.H` mapper; it now performs a name lookup inside it. Once that lookup actually touches the DB, a formatting helper becomes I/O-bound and harder to test. Prefer resolving in the handler and passing the name in.

### ✅ What is right

- No schema change needed — `creator_bot_id` already exists (`internal/model/model.go:133`, added by #181) and the worker/generation chain is correctly left alone.
- Authorization is not weakened. Detail still goes through `authorizeTaskAccess` (`task.go:111-145`), which is space-scoped and fail-closed on an empty `X-Space-Id`; list is already filtered to tasks the caller can see. The new fields are exposed only after those gates.
- The `creator_bot_id != ""` guards mean non-bot tasks get an empty string rather than a bogus label.
- The touched files are `gofmt`-clean (the repo has pre-existing comment-indent drift in ~23 other files, unrelated to this PR).

## 3. Overall verdict

**REQUEST_CHANGES** — P0-1 (failing test, red CI) and P1-1 (the display name is never resolved, so the stated feature is not delivered) both block. P2s are non-blocking.

## 4. Suggested path

1. Change the expectation at `bot_summary_create_test.go:101` to `http.StatusCreated`. That alone turns CI green.
2. Add `ResolveUserNameWithDB(uid, imDB)` next to `ResolveSourceNameWithType` in `internal/service/summary.go` — or wire `SetUserNameResolver` once at startup in `cmd/summary-api` and `cmd/summary-worker` — and route the three new call sites through it. Confirm first that bot uids resolve in the IM `user` table.
3. In `ListSummaries`, batch creator + bot uid resolution before the loop (reuse the shape of `batchResolveUserNames`).
4. In `GetSummary`, derive `creator_name` from the already-loaded `participants` slice so detail matches list.
5. Strengthen the test to assert the exact resolved name, plus a non-bot negative case.

## 5. Beyond the diff

- **Other read paths still have no bot marker.** `creator_bot_id` / `creator_bot_name` are exposed only on list, detail, and bot-create. The shared-summary view (`internal/api/handler/share.go`) and the agent summary responses (`internal/api/handler/agent_summary.go`) return summaries without them, so "由 &lt;bot&gt; 创建" will silently disappear on those surfaces. If that is deliberate (e.g. shared links should not reveal the acting bot), it is worth a one-line comment; if not, it is a follow-up.
- **Companion frontend PR.** I reviewed only this repo. Whoever reviews the octo-web side should check how it renders `creator_bot_name` when the value equals the raw uid or is an empty string, so a backend gap does not surface as a uid string in the UI.

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# Code Review — PR #188 (octo-smart-summary)

## Summary

This PR exposes the acting bot's identity on summary API responses so the frontend can mark bot-created summaries as "由 <bot> 创建": `creator_bot_id` + `creator_bot_name` are added to the list items and detail response (plus `creator_name`, which is new to the detail response), and `creator_bot_name` is added to the bot-create response. The two struct-realignment hunks (`apiResponse`, the canonical idempotency-hash payload) are gofmt-only. Overall a small, additive, well-scoped change; the end-to-end data flow holds and I found no blocking issues.

## Verification

Static analysis only at head `a908a22d31c1787836d8da6c8431e5c69be14b26`; build and tests not executed in this environment.

- ✅ **CreatorBotID actually reaches the consumers** — written at create (`internal/api/handler/bot_summary_create.go:463`, pre-existing since #181), column + index exist (`migrations/sql/20260728-01-add-bot-summary-create.sql`), the list query is `SELECT *` scanned into an embedded `model.SummaryTask` (`internal/api/handler/task.go:626`), and detail loads the full row via `authorizeTaskAccess`. Not an empty/short-circuited path.
- ✅ **Name resolution is wired in production** — `service.SetUserNameResolver` is injected against the IM DB in both binaries (`cmd/summary-api/main.go:102`, `cmd/summary-worker/main.go:82`) with parameterized SQL and a uid-echo fallback when `imDB` is nil or the row is missing — the same fallback semantics the existing `creator_name` relies on.
- ✅ **No authorization change** — new fields are serialized after the existing gates: space-scoped load + `canAccessTask` for detail (`internal/api/handler/task.go:888`), visibility `whereSQL` for list. The bot display name is the same exposure class as the already-exposed `creator_name`.
- ✅ **Whitespace hunks are semantics-free** — `git diff --ignore-all-space` reduces the PR to +57/−1 (field additions + the new test). The `canonicalBotCreateRequestHash` payload struct keeps identical JSON tags and values, so the idempotency-hash contract is untouched.

## Findings

No P0/P1 issues; one P2 below.

### P2 — List/detail exposure of the new fields has no regression test (`internal/api/handler/task.go:748`)

The PR's stated goal is exposing the acting bot on list, detail, and create, but only the create response is covered (`TestCreateBotSummaryResponseIncludesBotIdentity`). The list item fields at `internal/api/handler/task.go:748` and the detail fields at `:977` — including `creator_name`, which is brand new to the detail response — would silently regress if the serialization changed. Harnesses already wire both endpoints (`internal/api/handler/bot_citation_test.go:124`, `internal/api/handler/origin_channel_test.go:49`); seeding one TriggerBot task with `CreatorBotID` set and asserting the three fields on `GET /api/v1/summaries` and `GET /api/v1/summaries/:id` would close the gap. Non-blocking.

- Create-path parity: fresh create and both idempotent-replay branches answer through the same `botCreateResponse` (`internal/api/handler/bot_summary_create.go:198`, `:219`, `:229`), so replayed requests return the same shape.
- Surface parity: `/api/v1/bot/summaries*` and `/api/v1/summaries*` route to the same `ListSummaries`/`GetSummary` handlers (`internal/api/router/router.go:63`), so bot and user surfaces get the fields identically.
- Input shape: `CreatorBotID == ""` is guarded before resolution; `ResolveUserName` never returns empty for a non-empty uid, so `creator_bot_name` is non-empty exactly when `creator_bot_id` is.
- Pagination/cursor family in `task.go` is untouched by this diff (proven by the whitespace-only reduction); no notify/emit/counter changes anywhere in the PR.
- One note, not a finding: the list path resolves the bot name live from the IM DB per request while the owner name prefers the frozen participant row — consistent enough for display, and identical to how `creator_name` behaved before.

## Verdict: APPROVED

The feature is wired end-to-end (create → persist → list/detail serialization), authorization is unchanged, and the idempotency contract is intact. The only finding is a test-coverage gap on the list/detail responses, which is worth a follow-up but does not block.

yujiawei (changes requested):
# Code Review — PR #188 (octo-smart-summary) — correction to my earlier review

Still at head `a908a22d31c1787836d8da6c8431e5c69be14b26`. This supersedes my previous review on this same commit.

## ⚠️ Retraction: my P1-1 was wrong

My earlier review claimed that `service.ResolveUserName` has no implementation, that `SetUserNameResolver` "is never called anywhere in the module", and therefore that `creator_bot_name` could only ever echo the raw uid — making the feature inert. **That is false, and I withdraw it.** The resolver is wired against the IM DB in both binaries:

```go
// cmd/summary-api/main.go:101-109  (inside `if imDB != nil`)
service.SetUserNameResolver(func(uid string) string {
    var name string
    imDB.Raw("SELECT name FROM `user` WHERE uid = ? LIMIT 1", uid).Scan(&name)
    if name != "" {
        return name
    }
    return uid
})
```

The same injection exists at `cmd/summary-worker/main.go:82-89`. So in production `creator_bot_name` resolves to the bot's real display name, with a uid echo only when the row is genuinely absent — exactly the semantics the pre-existing `creator_name` already relies on. The feature does what the PR body says it does.

Two mistakes on my side, worth stating plainly because they explain the whole error:

1. **I trusted a negative grep as proof.** My repo-wide search for `SetUserNameResolver` returned only the definition, and I read that absence as "never called" without ever opening `cmd/*/main.go`. The search was silently incomplete: `.gitignore` lists the built binaries as bare names `summary-api` and `summary-worker` (no leading or trailing slash), so an ignore-file-aware searcher treats those as *any* path component and prunes the `cmd/summary-api/` and `cmd/summary-worker/` **directories**. Git itself still tracks those files, so the code is very much there — the search just could not see it. A negative search result is weak evidence for an "it is never called anywhere" claim; I should have opened the entrypoints.
2. **I mistook a test artifact for production behavior.** I cited the failing test's response body (`"creator_bot_name":"bot-1"`) as empirical proof the name never resolves. But unit tests never execute `main()`, so no resolver is injected under test — the uid echo there is fully consistent with a correctly wired production path. That observation could not distinguish the two hypotheses, and I treated it as if it did.

Consequently the **Spec verdict flips to ✅**: nothing is missing, the one arguably-extra field (`creator_name` on detail) is declared in the PR body, and there is no deviation. My earlier "deviation — display-name resolution is inert" line should be disregarded entirely.

## What still blocks: the failing test

**P0-1 stands, unchanged and independently verified.** `internal/api/handler/bot_summary_create_test.go:101` expects `http.StatusOK`, but a fresh create returns `http.StatusCreated` (`internal/api/handler/bot_summary_create.go:229`); `200` is reserved for idempotent replays (`:198`, `:219`). All seven sibling tests in that file expect `201` for a fresh create — this new test is the only one that does not.

Re-confirmed locally just now at this head:

```
--- FAIL: TestCreateBotSummaryResponseIncludesBotIdentity (0.01s)
    bot_summary_create_test.go:102: status=201 body={...,"creator_bot_id":"bot-1",...}
```

And CI agrees — `Test (race, cgo)` is red on this commit. It runs `go test ./...` with cgo (`.github/workflows/ci.yml:142`), which is why the `CGO_ENABLED=0` job passes while this one fails. **Fix:** expect `http.StatusCreated`, or send a second request if the intent was to exercise the replay path.

## Corrected non-blocking findings

### 🟡 P2-1 — per-row IM DB query added inside the `ListSummaries` loop

I previously graded this "latent, zero cost today" on the strength of my mistaken P1-1. With the resolver actually wired, `internal/api/handler/task.go:719-721` issues a **real** `SELECT name FROM user WHERE uid = ?` round-trip per bot-created row, on top of the existing per-row participant query at `:708`. Page size defaults to 20 and is capped at 100 (`task.go:493-498`), so a bot-heavy page costs up to 100 extra IM DB round-trips.

This is the same class of access the pre-existing `creator_name` path already performs, so it is not a new pattern and I do not consider it blocking — but it is a genuine incremental cost, not a hypothetical one. Batching the unique creator and bot uids once before the loop would fix both; `internal/worker/processor.go:860` (`batchResolveUserNames`) already implements `SELECT uid, name FROM user WHERE uid IN ?` and can be reused.

### 🟡 P2-2 — detail and list can now disagree on `creator_name`

The list path prefers the frozen participant row and only falls back to a live lookup (`task.go:706-713`); the new detail path always resolves live (`task.go:963`). With a working resolver these are genuinely different values: if a user is renamed in IM after the task was created, the card shows the frozen name and the detail page shows the new one. `GetSummary` already loads every participant at `task.go:912-913`, so matching the list's precedence costs no extra query. Cosmetic, but easy to align.

### 🟡 P2-3 — the new assertion still cannot fail

`bot_summary_create_test.go:120` asserts only that `creator_bot_name` is non-empty. Since no resolver is injected under test, `ResolveUserName` echoes the uid, so this passes whenever `creator_bot_id` is non-empty regardless of whether resolution works. Install a resolver in the test and assert the exact expected name, and add a non-bot case asserting an empty `creator_bot_name`. (This one is unaffected by the retraction — it was always a test-quality point, not evidence about production.)

### 🟡 P2-4 — no list/detail regression test

The PR's goal spans list, detail, and create, but only create is covered. The list fields at `task.go:747-749` and the detail fields at `:976-979` — including `creator_name`, brand new to detail — would regress silently. Harnesses for both endpoints already exist (`internal/api/handler/bot_citation_test.go`, `internal/api/handler/origin_channel_test.go`); seeding one `TriggerBot` task with `CreatorBotID` set and asserting the three fields on each endpoint would close it.

### 🟡 P2-5 — I/O inside a response-serialization helper

`botCreateResponse` (`bot_summary_create.go:545-549`) is a `model → gin.H` mapper that now performs a DB round-trip. Prefer resolving in the handler and passing the name in.

## Confirmed sound

- No schema change needed — `creator_bot_id` already exists (`internal/model/model.go:133`, from #181), with column and index in `migrations/sql/20260728-01-add-bot-summary-create.sql`. Worker and generation chain untouched.
- `CreatorBotID` genuinely reaches both consumers: written at `bot_summary_create.go:463`, list scans the full row, detail loads it via `authorizeTaskAccess`.
- Authorization unchanged. Detail still passes through `authorizeTaskAccess` (`task.go:111-145`) — space-scoped and fail-closed on an empty `X-Space-Id`; list is filtered by the existing visibility SQL. New fields serialize only after those gates, and the bot name is the same exposure class as the already-public `creator_name`.
- Idempotency contract intact: the realigned `canonicalBotCreateRequestHash` payload struct keeps identical JSON tags and values, and all three create branches answer through the same `botCreateResponse`.
- The two struct-realignment hunks are gofmt-only; touched files are gofmt-clean.
- `creator_bot_id != ""` guards mean non-bot tasks get an empty string rather than a bogus label.

## Verdict

**REQUEST_CHANGES** — on P0-1 alone (failing test, red CI). Everything else here is non-blocking, and my earlier P1 is withdrawn. Once the status-code expectation is corrected, I would expect this to pass; the P2s are reasonable follow-ups rather than gates.

One item I still cannot verify from source, offered as a question rather than a finding: whether personal-bot uids exist as rows in the IM `user` table. If they do not, the resolver's uid echo means `creator_bot_name` degrades to a raw uid in the UI. That risk is identical in kind to the existing `creator_name` behaviour, so it is not a blocker — but it is worth one manual check against a real environment before the frontend ships the label.

wanyiwang06:
Fixes for the remaining blockers, pushed in `1bd5cb8` (test-only, no production code touched).

**P0-1 — failing test, red CI.**
> "Critical: `TestCreateBotSummaryResponseIncludesBotIdentity` expects `http.StatusOK` for a first-time create … but `CreateBotSummary` returns `http.StatusCreated` for new tasks." — @Jerry-Xin #4894280127
> "Fix: expect `http.StatusCreated`, or issue a second (replay) request if the intent was to exercise the idempotent path." — @yujiawei #4894547685

The intent is a fresh create, so the test now expects `http.StatusCreated`, matching the seven sibling tests in the same file.

**P2-3 — the name assertion could never fail.**
> "Install a resolver in the test and assert the exact expected name, and add a non-bot case asserting an empty `creator_bot_name`." — @yujiawei #4894547685
> "Consider installing a temporary resolver in the test and asserting the exact bot display name, then restoring it after the test." — @Jerry-Xin #4894280127

Done both: the identity test installs a temporary resolver via `service.SetUserNameResolver` (restored in `t.Cleanup`) and asserts the exact display name; a new `TestBotCreateResponseNonBotTaskHasEmptyBotName` asserts the non-bot case, with a resolver that poisons an empty uid so removing the `CreatorBotID` guard makes it fail.

On P1-1: independently verified the retraction — `service.SetUserNameResolver` is wired in production at `cmd/summary-api/main.go:102` and `cmd/summary-worker/main.go:82`, so display-name resolution is live.

Verified locally at this head: `CGO_ENABLED=1 go test ./internal/api/handler/` green; full `CGO_ENABLED=1 go test -race ./...` matrix green.

Jerry-Xin (dismissed):
This in-scope PR correctly exposes bot identity, with no blocking correctness or security issues found.

💬 Non-blocking

- 🟡 Warning — `internal/api/handler/task.go:720` adds one IM-database lookup per bot-created list item, creating an additional N+1 query pattern for pages of up to 100 items. Consider batch-resolving unique bot IDs.
- 🔵 Suggestion — Add list/detail response tests covering the fields introduced at `internal/api/handler/task.go:748` and `internal/api/handler/task.go:977`; current coverage only verifies create responses.
- 🔵 Suggestion — The tests at `internal/api/handler/bot_summary_create_test.go:111` restore the resolver to identity rather than its previous value. A scoped resolver API would avoid future global-state interference.

✅ Highlights

- Empty bot IDs are guarded, preventing misleading names on non-bot tasks.
- Create, replay, list, and detail response paths consistently expose bot identity.
- Authorization remains enforced before detail data is returned.
- Exact display-name resolution is asserted rather than merely checking non-emptiness.

Note: full handler tests could not be run locally because a native library dependency was unavailable; correctness was verified by reading the diff and surrounding handler code.

yujiawei (changes requested):
# Code Review — PR #188 (octo-smart-summary)

Reviewed at head `1bd5cb82c03c7532c05ecd32145e2d55581163db`, merge-base `62a2d15`. I checked out the branch, built it, ran the two new tests and the full `internal/api/handler` package under `-race`, and test-merged against `origin/main`.

Two verdicts below: first whether the diff implements exactly what it claims, then implementation quality.

## 1. Spec compliance

**Spec: ✅**

- **Missing: none.** All three surfaces named in the PR description are wired — list (`internal/api/handler/task.go:748-749`), detail (`internal/api/handler/task.go:977-979`), bot-create (`internal/api/handler/bot_summary_create.go:544-549`). The replay branches also benefit, because `findBotIdempotentTaskWithHash` loads the full row with `creator_bot_id = ?` (`bot_summary_create.go:532`), so a 200 replay carries the same identity fields as the 201 create.
- **Extra: none blocking.** The detail-only `creator_name` is declared in the PR description, so it is in scope. The two struct-realignment hunks (`canonicalBotCreateRequestHash`, `apiResponse`) are gofmt-only churn unrelated to the feature — `git diff --ignore-all-space 62a2d15..HEAD` reduces the PR to **+100/−1**, and the idempotency-hash payload keeps identical JSON tags and values, so the hash contract is untouched.
- **Deviation: none.** I re-verified the concern from the previous round that `creator_bot_name` could only ever echo the raw uid: it does not hold. `SetUserNameResolver` is injected against the IM DB in both binaries (`cmd/summary-api/main.go:102-109`, `cmd/summary-worker/main.go:82-89`), with a uid echo only when `imDB` is nil or the row is genuinely missing — the same semantics the pre-existing `creator_name` already relies on.
- **Previous round's blocker is fixed.** The status-code mismatch is resolved (`bot_summary_create_test.go:100` now expects `http.StatusCreated`), and the display-name assertion is now falsifiable — the test installs a resolver and asserts the exact string `"Bot Display Name"` rather than merely non-empty (`bot_summary_create_test.go:105-142`). Both new tests pass locally, and `go test -race ./internal/api/handler/` is green at this head.

## 2. Code quality

The handler code itself is correct and I would approve it as written. The blocker is the state of the branch, not the logic.

### 🔴 P1 — the branch is unmergeable, and the conflict resolution can silently revert a privacy control from #190

`mergeable: CONFLICTING`, and the required `check-sprint` job is red for exactly this reason:

```
##[error]PR has merge conflicts. Resolve conflicts and push to re-trigger this check.
```

`git merge-tree --write-tree HEAD origin/main` conflicts in `internal/api/handler/task.go` — and the conflict lands in **the two `gin.H` literals this PR edits**. Since the merge-base, #190 (`a993977`) rewrote both of them:

```go
// origin/main, task.go:87-99 — privacy mask added by #190
func wireOriginChannelID(t model.SummaryTask) string {
	if t.OriginFromDerived {
		return ""
	}
	return t.OriginChannelID
}
```

`origin/main` now emits `wireOriginChannelID(t)` / `wireOriginChannelType(t)` in **both** the list item and the detail response, plus three new list fields (`referenceable`, `reference_artifact_type`, `reference_unavailable_reason`). This branch still emits the raw `t.OriginChannelID` / `t.OriginChannelType` in both places (`task.go:754-755`, `task.go:983-984`).

**Failure scenario:** the natural way to resolve a conflict inside a map literal you just edited is to keep your own side. Doing that here reverts the `OriginFromDerived` wire mask on both endpoints — re-leaking a derived-inherited origin channel id to viewers who may not be members of the backfilled channel — and drops the three `referenceable*` fields the frontend now expects. Both regressions compile cleanly and neither field is covered by a test that would catch it, so nothing downstream would flag it.

**Requested action:** rebase onto `origin/main` and resolve by taking **main's** projection for `origin_channel_id` / `origin_channel_type` (the `wire*` helpers) and **main's** `referenceable*` entries, then re-add only the new keys from this PR (`creator_bot_id`, `creator_bot_name` on list; `creator_name`, `creator_bot_id`, `creator_bot_name` on detail). Once pushed I'll re-review the merged form — this is the only reason this is not an approve.

### 🟡 P2-1 — detail resolves `creator_name` by a weaker path than list, so the same task can render two different names

List prefers the persisted participant snapshot and only falls back to the IM lookup:

```go
// task.go:705-712
creatorName := ""
var creatorParticipant model.SummaryParticipant
if err := h.db.Where("task_id = ? AND user_id = ?", t.ID, t.CreatorID).First(&creatorParticipant).Error; err == nil {
	creatorName = creatorParticipant.UserName
}
if creatorName == "" {
	creatorName = service.ResolveUserName(t.CreatorID)
}
```

Detail skips the snapshot entirely:

```go
// task.go:963
creatorName := service.ResolveUserName(task.CreatorID)
```

Two distinct ways this diverges:

1. **Creator renamed after the task was created.** The list serves the historical snapshot; detail serves the current IM name. The card and the detail page of the same summary then identify the creator differently.
2. **IM row absent** (deactivated account, cross-tenant uid, IM DB unreachable). `ResolveUserName` echoes the uid, so detail shows a raw uid while the list still shows the snapshot name.

`GetSummary` has already loaded `participants` into memory at `task.go:912-913`, so mirroring the list's snapshot-first logic costs zero extra queries and actually removes one.

### 🟡 P2-2 — N+1 IM-DB lookups in the list loop

`service.ResolveUserName` is a per-uid `SELECT name FROM user WHERE uid = ? LIMIT 1` with no caching (`cmd/summary-api/main.go:103`). It is now called up to twice per list item (`task.go:711` and the new `task.go:720`), inside the loop. With `page_size` capped at 100 (`task.go:497`), a page of bot-created summaries costs up to 100 additional IM round trips on top of the pre-existing per-item participant query.

The repo already has the batch shape for exactly this — `internal/agent/message_enrichment.go:58-74` collects a uid set and resolves it in one statement:

```go
if err := imDB.WithContext(ctx).Raw(
	"SELECT uid, name FROM `user` WHERE uid IN ? AND name != ''",
	uids,
).Scan(&rows).Error; err != nil {
```

Collecting the distinct `creator_id` + `creator_bot_id` set before the loop and resolving once would fix both the new cost and the pre-existing one. Not blocking — bot-created summaries are still rare, and the loop was already N+1 before this PR — but it is the natural moment to do it.

### 🟡 P2-3 — the endpoints this PR is actually about have no test

The PR's stated goal is exposing the bot on **list and detail**; the only new coverage is the create response (`bot_summary_create_test.go:98`, `:145`). `grep -rn "creator_bot_name" --include=*_test.go internal/` returns nothing outside that one file.

Concretely: deleting `creator_bot_id`, `creator_bot_name`, and `creator_name` from both `ListSummaries` and `GetSummary` leaves the entire test suite green. The majority of the advertised contract can regress undetected — including via the conflict resolution in P1. Harnesses for both endpoints already exist (`origin_channel_test.go`, `edit_test.go`), so a case asserting the fields on a `TriggerBot` task and their emptiness on a manual one is cheap.

### 🟡 P2-4 — the non-bot test passes even if the key disappears entirely

```go
// bot_summary_create_test.go:163-168
if got, _ := resp["creator_bot_id"].(string); got != "" {
	t.Fatalf("creator_bot_id=%q, want empty for non-bot task", got)
}
if got, _ := resp["creator_bot_name"].(string); got != "" {
	t.Fatalf("creator_bot_name=%q, want empty for non-bot task", got)
}
```

If `botCreateResponse` stops emitting either key, the map lookup yields `nil`, the comma-ok assertion returns `("", false)`, and `got != ""` is false — so the test still passes. I confirmed this empirically with a standalone case. The always-present-key contract that the frontend relies on (the model's own tag is `json:"creator_bot_id,omitempty"`, so presence here is a deliberate handler-side choice) is therefore unenforced. Assert the key exists and is a string before asserting it is empty:

```go
v, ok := resp["creator_bot_name"]
if !ok {
	t.Fatal("creator_bot_name key missing")
}
if s, isStr := v.(string); !isStr || s != "" {
	t.Fatalf("creator_bot_name=%v, want empty string", v)
}
```

### 🔵 P2-5 (nit) — tests restore a fabricated default rather than the previous resolver

```go
// bot_summary_create_test.go:111-113
t.Cleanup(func() {
	service.SetUserNameResolver(func(uid string) string { return uid })
})
```

This mutates the package-level `resolveUserNameFn` (`internal/service/summary.go:172-179`) and restores an identity closure rather than whatever was installed before.

It is correct today and **not** a live race: no test in this package calls `t.Parallel()`, `main()` never runs under test so the package default *is* identity, and `go test -race ./internal/api/handler/` passes at this head. But it is load-bearing on an invariant nothing enforces — the first test or `TestMain` that installs a real resolver, or the first `t.Parallel()` added to this package, will be silently clobbered or will race on an unsynchronized global. Adding a getter (or having `SetUserNameResolver` return the previous fn) would make save/restore possible; injecting the resolver into `TaskHandler` would remove the global entirely.

## 3. Overall verdict

**CHANGES_REQUESTED**

Spec ✅, and the handler logic is approvable as written — the previous round's blocking finding is genuinely fixed. What blocks is that the branch cannot merge, and the conflict sits on top of a privacy control added by #190 in the very lines this PR touches, where the intuitive resolution is the wrong one. Rebase with the resolution described in P1 and I'll re-review the merged head promptly.

## 4. Suggested changes

1. **Required:** rebase onto `origin/main`; keep main's `wireOriginChannelID(t)` / `wireOriginChannelType(t)` and its three `referenceable*` list keys, and layer only this PR's new keys on top.
2. **Recommended:** in `GetSummary`, derive `creator_name` from the already-loaded `participants` slice with `ResolveUserName` as fallback, matching `ListSummaries` (P2-1).
3. **Recommended:** add a list-and-detail test asserting the new fields for a `TriggerBot` task and their emptiness for a manual one (P2-3).
4. **Recommended:** assert key presence, not just empty value, in the non-bot test (P2-4) — it is two lines and closes a real hole.
5. **Optional:** batch the name resolution for a page instead of resolving per item (P2-2).
6. **Optional:** make the test resolver save/restore the previous value (P2-5).

## 5. Notes outside the diff

- **Authorization is unchanged and remains correct.** The new fields serialize after the existing gates — `authorizeTaskAccess` scopes the detail load by space and fails closed on an empty `X-Space-Id` (`task.go:121-137`), and the list applies its visibility `whereSQL`. `creator_bot_id` is the same exposure class as the already-public `creator_id`.
- **`ResolveUserName` is not space-scoped** — it looks up the IM `user` table by bare uid. That is pre-existing behaviour shared with `creator_name`, not introduced here, but it means a bot's display name is resolvable regardless of which space the caller is in. Worth a separate look if bot display names are ever considered tenant-scoped.
- **Coverage limits of this review.** I did not exercise the endpoints against a real MySQL + IM database (local runs use the SQLite harness, where some IM-side queries fail on collation and are logged and skipped), so the resolver's production behaviour is verified by reading the injection site rather than by observation. I also did not review the companion octo-web PR, so I cannot confirm the frontend consumes `creator_bot_name` under the same key, nor whether it depends on the `referenceable*` fields that the conflict resolution could drop. Whether bot identities are considered sensitive at runtime is not answerable from this repo alone.

Jerry-Xin (approved):
This in-scope PR exposes `creator_bot_id` + `creator_bot_name` on list/detail/create by resolving the acting bot's display name through the same `service.ResolveUserName` path already used for `creator_id`/`creator_name`. No blocking correctness, security, or privacy issue found. Approving.

### ✅ Verified — peer REQUEST_CHANGES claims do not hold on this head (f29a22e73b35)

- **Merge conflict / privacy-control regression** — Not present. `mergeable=MERGEABLE`; the PR's merge-base is a direct recent ancestor of `main` (linear history), so there is no conflict to resolve and no resolution-induced regression.
- **"Uses old raw fields / not synced with privacy-masking"** — The privacy-masking layer in this handler package applies to *citations* (raw chat messages, `callerPlainCitationsVisible`) and to derived-inherited *origin* fields — not to creator identity. `creator_id`/`creator_name` are already unconditionally exposed on these exact responses; the new `creator_bot_id`/`creator_bot_name` follow the identical, pre-existing pattern. There is no creator-identity masking layer to bypass.
- **Resolver wiring** — `service.SetUserNameResolver` IS installed in production at `cmd/summary-api/main.go:102` (IM-DB lookup on the `user` table), the same resolver that already backs `creator_name`. So the advertised display names are delivered in production, not raw IDs.

### 💬 Non-blocking

- 🟡 The create path has strong new tests (`TestCreateBotSummaryResponseIncludesBotIdentity` with a falsifiable resolved-name assertion, plus `TestBotCreateResponseNonBotTaskHasEmptyBotName` for the empty-bot guard). The list (`ListSummaries`) and detail (`GetSummary`) HTTP responses have no dedicated test asserting the new fields. The resolution logic is identical and trivially guarded, so this is low risk, but a list/detail response test would close the contract.
- 🟡 `ListSummaries` resolves the bot name per row via `service.ResolveUserName`, adding one IM-DB lookup per bot-created item for pages up to 100. Consider batch-resolving unique bot IDs if list latency matters.

### ✅ Highlights

- Bot IDs and resolved names are exposed consistently across create, list, and detail.
- Empty `CreatorBotID` is guarded before name resolution, so non-bot tasks keep `creator_bot_name` empty.
- Detail adds `creator_name` (owner) alongside the bot name to support "由 <bot> 代 <owner> 创建".

yujiawei (approved):
# Code Review — PR #188 (octo-smart-summary)

Reviewed at head `f29a22e73b35b9dad17a5f9f99f2be845af27a02`, merge-base `a993977` (current `origin/main`). I checked out the branch, built it, ran both new tests and the full `internal/api/handler` package under `-race`, test-merged against `origin/main`, and ran a mutation check on the new list/detail fields.

Two verdicts below: first whether the diff implements exactly what it claims, then implementation quality.

## 1. Spec compliance

**Spec: ✅**

- **Missing: none.** All three surfaces named in the PR description are wired — list (`internal/api/handler/task.go:867-868`), detail (`internal/api/handler/task.go:1110-1112`), bot-create (`internal/api/handler/bot_summary_create.go:544-549`). The idempotent-replay branches benefit too: `findBotIdempotentTaskWithHash` loads the full row (`bot_summary_create.go:532`), so a `200` replay carries the same identity fields as the `201` create.
- **Surface coverage is complete.** `grep '"creator_id"'` over non-test sources returns exactly three response projections (`task.go:866`, `task.go:1109`, `bot_summary_create.go:549`) — all three now also emit `creator_bot_id`. No surface that exposes the human creator is left without the bot counterpart.
- **Extra: none blocking.** The detail-only `creator_name` is declared in the PR description, so it is in scope. The two struct-realignment hunks are gofmt-only churn: `git diff --ignore-all-space a993977..HEAD` reduces the PR to **+100/−1**, and the `canonicalBotCreateRequestHash` payload keeps identical JSON tags and values, so the idempotency-hash contract is untouched.
- **Deviation: none.**

### The previous round's blocker is fixed, and fixed the right way

The prior review blocked on the branch being unmergeable, with the specific concern that the natural conflict resolution would silently revert the `OriginFromDerived` wire mask and drop the three `referenceable*` fields that #190 had added to the very `gin.H` literals this PR edits.

Verified resolved:

- The branch is now rebased directly onto `a993977` (#190), `mergeable: MERGEABLE`, and `git merge-tree --write-tree HEAD origin/main` completes cleanly with no conflict.
- The resolution kept **main's** projection, not the branch's. Both literals emit `wireOriginChannelID(...)` / `wireOriginChannelType(...)` (`task.go:876-877` for list, `task.go:1121-1122` for detail), and the list retains all three `referenceable*` keys (`task.go:889-891`) plus the `referenceableFromLoaded` computation that feeds them (`task.go:853`).
- Diffing the pre-rebase head against this one confirms the direction of change: `t.OriginChannelID` → `wireOriginChannelID(t)`, with `referenceable`/`reference_artifact_type`/`reference_unavailable_reason` added rather than lost. Nothing from #190 was reverted.
- `bot_summary_create.go` and its test are byte-identical across the rebase.

## 2. Code quality

**Quality: Approved**

The handler logic is correct, the feature does what the description says, and CI agrees: `Build`, `Lint`, `Vet`, `Test`, and `Test (race, cgo)` all pass on this head. Locally `go test -race ./internal/api/handler/` is green and both new tests pass. Everything below is P2 — worth doing, none of it blocks merge.

### 🟡 P2-1 — detail resolves `creator_name` by a weaker path than list, so one task can render two different names

List prefers the persisted participant snapshot and only falls back to the IM lookup:

```go
// task.go:797-805
creatorName := ""
var creatorParticipant model.SummaryParticipant
if err := h.db.Where("task_id = ? AND user_id = ?", t.ID, t.CreatorID).First(&creatorParticipant).Error; err == nil {
	creatorName = creatorParticipant.UserName
}
if creatorName == "" {
	creatorName = service.ResolveUserName(t.CreatorID)
}
```

Detail skips the snapshot entirely:

```go
// task.go:1094
creatorName := service.ResolveUserName(task.CreatorID)
```

Two ways this diverges:

1. **Creator renamed after the task was created.** The card serves the historical snapshot; the detail page serves the current IM name. Same summary, two identities.
2. **IM row absent** (deactivated account, cross-tenant uid, IM DB unreachable). `ResolveUserName` echoes the uid, so detail renders a raw uid while the list still renders the snapshot name.

`GetSummary` already loads `participants` into memory at `task.go:1045-1046`, well before the response is built, so mirroring the list's snapshot-first logic costs zero extra queries and actually removes one.

### 🟡 P2-2 — N+1 IM lookups in the list loop

`service.ResolveUserName` is an uncached per-uid `SELECT name FROM user WHERE uid = ? LIMIT 1` (`cmd/summary-api/main.go:102-109`). The new call at `task.go:811` runs inside the per-item loop, on top of the pre-existing per-item participant query at `task.go:799` and the fallback resolve at `task.go:804`. With `page_size` capped at 100, a page of bot-created summaries costs up to 100 additional IM round trips.

The repo already has the batch shape for exactly this — `internal/agent/message_enrichment.go:58-74` collects a uid set and resolves it in one statement:

```go
if err := imDB.WithContext(ctx).Raw(
	"SELECT uid, name FROM `user` WHERE uid IN ? AND name != ''",
	uids,
).Scan(&rows).Error; err != nil {
```

Collecting the distinct `creator_id` + `creator_bot_id` set before the loop and resolving once would fix both the new cost and the pre-existing one.

Sizing this honestly: this is an *increment* on a loop that was already N+1 before the PR, not a new class of problem, and bot-created summaries are a minority of rows — which is why it is P2 and not a blocker. But the mechanism is real (I confirmed the resolver does uncached I/O), and this is the natural moment to batch it.

### 🟡 P2-3 — the two endpoints this PR is actually about have no test

The stated goal is exposing the bot on **list and detail**; the only new coverage is the create response (`bot_summary_create_test.go:98`, `:147`). `grep -rn "creator_bot_name" --include=*_test.go internal/` returns nothing outside that one file.

Verified empirically rather than by inspection: I deleted `creator_bot_id` + `creator_bot_name` from the list literal and `creator_name` + `creator_bot_id` + `creator_bot_name` from the detail literal, and `go test ./internal/api/handler/` stayed **green** (`ok ... 11.343s`). The majority of the advertised contract can regress with zero test signal.

Harnesses for both endpoints already exist (`origin_channel_test.go`, `referenceable_http_contract_test.go`), so a case that seeds one `TriggerBot` task with `CreatorBotID` set and asserts the three fields on `GET /api/v1/summaries` and `GET /api/v1/summaries/:id` — plus their emptiness on a manual task — is cheap. Given #190 just added a privacy mask to these same two literals, a test that pins their contents also protects the *next* person who has to merge into them.

### 🟡 P2-4 — the non-bot test passes even if the key disappears entirely

```go
// bot_summary_create_test.go:163-168
if got, _ := resp["creator_bot_id"].(string); got != "" {
	t.Fatalf("creator_bot_id=%q, want empty for non-bot task", got)
}
if got, _ := resp["creator_bot_name"].(string); got != "" {
	t.Fatalf("creator_bot_name=%q, want empty for non-bot task", got)
}
```

If `botCreateResponse` stops emitting either key, the map lookup yields `nil`, the comma-ok assertion returns `("", false)`, and `got != ""` is false — the test still passes. The always-present-key contract matters here: the model's own tag is `json:"creator_bot_id,omitempty"` (`internal/model/model.go:133`), so unconditional presence in the response is a deliberate handler-side choice that nothing currently enforces. Two extra lines close it:

```go
v, ok := resp["creator_bot_name"]
if !ok {
	t.Fatal("creator_bot_name key missing")
}
if s, isStr := v.(string); !isStr || s != "" {
	t.Fatalf("creator_bot_name=%v, want empty string", v)
}
```

### 🔵 P2-5 (nit) — tests restore a fabricated default rather than the previous resolver

```go
// bot_summary_create_test.go:111-113 (and :157-159)
t.Cleanup(func() {
	service.SetUserNameResolver(func(uid string) string { return uid })
})
```

This mutates the package-level `resolveUserNameFn` (`internal/service/summary.go:172-179`) and restores an identity closure rather than whatever was installed before.

It is correct today and **not** a live race — no test in this package calls `t.Parallel()` (confirmed: zero occurrences), `main()` never runs under test so the package default *is* identity, the only goroutine `CreateBotSummary` spawns is `triggerWorker`, which never touches the resolver, and `-race` is green. But it is load-bearing on an invariant nothing enforces: the first `TestMain` that installs a real resolver, or the first `t.Parallel()` added here, silently clobbers it or races on an unsynchronized global. Having `SetUserNameResolver` return the previous fn — or injecting the resolver into `TaskHandler` — would remove the hazard.

### 🔵 P2-6 (nit) — detail resolves the creator name unconditionally

`task.go:1094` calls `ResolveUserName(task.CreatorID)` with no guard, while the bot name two lines down is correctly gated on a non-empty id (`task.go:1097-1099`). `CreatorID` is always populated from an authenticated uid at every construction site (`task.go:407`, `agent_summary.go:335`, `schedule.go:575`, `share.go:422`, `bot_summary_create.go:463`, `scheduler.go:155`), so this is not a live bug — but it does add one unconditional IM query to every detail request where there was none before. Folding it into the snapshot-first change from P2-1 makes the guard moot.

### 🔵 Nit — unused assertion field

`bot_summary_create_test.go:123` declares `TriggerType int` in the response struct and never asserts on it. Either assert it is `TriggerBot` (which would be a genuinely useful addition, since it is the field the whole feature keys off) or drop it.

## 3. Overall verdict

**APPROVE**

Spec ✅ and Quality Approved. The previous round's blocking finding is genuinely resolved, and resolved in the specific way that was requested — the rebase kept #190's origin-channel wire mask and its `referenceable*` fields in both literals rather than reverting them, which was the actual risk. The remaining findings are all quality/test-hardening items that do not warrant another round-trip.

### One non-code item that still blocks merge

`check-sprint` is red on this head, and it is **not** a code problem:

```
##[error]No linked issue found. Please add a 'Closes #<issue>' reference to the PR description, then re-run this check.
```

Adding a `Closes #<issue>` line to the PR description (and having a maintainer put that issue on the board with a Sprint set) clears it. Every other check — `Build`, `Lint`, `Vet`, `Test`, `Test (race, cgo)`, `osv-scan`, `gitleaks`, `dependency-review` — passes.

## 4. Suggested changes

1. **Recommended:** in `GetSummary`, derive `creator_name` from the already-loaded `participants` slice with `ResolveUserName` as fallback, matching `ListSummaries` (P2-1, subsumes P2-6).
2. **Recommended:** add a list-and-detail test asserting the new fields for a `TriggerBot` task and their emptiness for a manual one (P2-3).
3. **Recommended:** assert key presence, not just empty value, in the non-bot test (P2-4) — two lines, closes a real hole.
4. **Optional:** batch the name resolution for a page instead of resolving per item (P2-2).
5. **Optional:** make the test resolver save/restore the previous value (P2-5).

None of these need to land in this PR.

## 5. Notes outside the diff

- **Bot display names really do resolve.** Bots are rows in the IM `user` table — `candidates.go:88-96` filters humans with `u.robot = 0 AND u.uid NOT IN (SELECT robot_id FROM robot)`, which only makes sense if bot uids are present in `user`. So `SELECT name FROM user WHERE uid = ?` hits for the ordinary case. A system bot registered only in `robot` and absent from `user` would fall back to the raw uid — the same degradation class the pre-existing `creator_name` already accepts, not something this PR introduces.
- **Authorization is unchanged and remains correct.** The new fields serialize after the existing gates — `authorizeTaskAccess` scopes the detail load by space and fails closed on an empty `X-Space-Id`, and the list applies its visibility `whereSQL`. `creator_bot_id` is the same exposure class as the already-public `creator_id`.
- **`ResolveUserName` is not space-scoped** — it looks up the IM `user` table by bare uid. Pre-existing behaviour shared with `creator_name`, not introduced here, but it means a bot's display name is resolvable regardless of the caller's space. Worth a separate look if bot display names are ever considered tenant-scoped.
- **Coverage limits of this review.** I did not exercise the endpoints against a real MySQL + IM database (local runs use the SQLite harness, where some IM-side queries fail on collation and are logged and skipped), so the resolver's production behaviour is verified by reading the injection site rather than by observation. I also did not review the companion frontend PR, so I cannot confirm it consumes `creator_bot_name` under the same key, nor whether it depends on the `referenceable*` fields that the rebase touched. Whether bot identities are considered sensitive at runtime is not answerable from this repo alone.

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #188 (octo-smart-summary)
Head `f29a22e73b35b9dad17a5f9f99f2be845af27a02`, merge-base `a99397758f4afba9750e392efeee0424bf9120a7` (== current `main` tip). Re-review; previous round verdict was CHANGES_REQUESTED at head `1bd5cb82` (1 P1 + 5 P2).

## Summary

This PR exposes the acting bot's identity on the summary API so the frontend can mark bot-created summaries as "由 <bot> 创建": `creator_bot_id` + `creator_bot_name` are added to the list items, the detail response (plus `creator_name`, new to detail), and the bot-create response (both fresh 201 and idempotent-replay 200 paths). The previous round's only blocker was the branch state: it conflicted with #190's privacy-mask rewrite of exactly the two `gin.H` literals this PR edits, where the intuitive resolution would have silently reverted the `OriginFromDerived` wire mask and dropped the three `referenceable*` keys. The branch is now rebased onto `a993977` (#190 itself, current `main` tip), so that conflict no longer exists and both #190's mask and this PR's new keys are present at head. The handler logic itself was already approvable in the previous round and is unchanged. Remaining findings are carried-over P2/nit items, none blocking.

## Verification

Local checks at head `f29a22e7` (executed in this environment):
- `go build ./...` — clean
- `go vet ./internal/api/handler/ ./internal/service/` — clean
- `go test ./internal/api/handler/ -count=1 -race` — pass (full package, 15.3s)
- `gofmt -l` — the three PR-changed files are clean (the 12 flagged files in the package are pre-existing and untouched by this PR)

Data-flow trace (each consumed value followed to its producer):
- ✅ **`CreatorBotID` really reaches every consumer** — written at create from the authenticated bot identity (`internal/api/handler/bot_summary_create.go:463`, `botUID` taken from `c.Get("bot_id")` set by bot-token auth middleware, 401-guarded at `:125`, never a client field); persisted via column + index (`migrations/sql/20260728-01-add-bot-summary-create.sql`, `internal/model/model.go:133`); list loads it through `SELECT *` (`internal/api/handler/task.go:626`); detail loads the full row via `authorizeTaskAccess` (`internal/api/handler/task.go:132`, fail-closed on empty `X-Space-Id`); idempotent replay loads the full row keyed on `creator_bot_id` (`internal/api/handler/bot_summary_create.go:532`). Not an empty/short-circuited path.
- ✅ **Name resolution is wired in production** — `service.SetUserNameResolver` is installed against the IM `user` table in both binaries (`cmd/summary-api/main.go:102`, `cmd/summary-worker/main.go:82`), falling back to the uid only when the row is missing or `imDB` is nil — the same semantics the pre-existing `creator_name` already relies on. The guard at `task.go:810`/`:1098` skips resolution only for non-bot tasks (`CreatorBotID == ""`), which is the intended semantics.
- ✅ **#190 privacy mask intact after the rebase** — `wireOriginChannelID`/`wireOriginChannelType` (`internal/api/handler/task.go:87-99`) are emitted by both the list projection (`task.go:879-880`) and the detail response (`task.go:1124-1125`), and the three `referenceable*` keys are present (`task.go:892-894`). The PR's hunks do not touch these lines; nothing from #190 was reverted.
- ✅ **Idempotency-hash contract untouched** — the only other hunk in `bot_summary_create.go` (68-78) is gofmt realignment of the canonical-hash payload struct; JSON tags and values are byte-identical, so existing idempotency keys replay cleanly.

## Cross-round blocker re-check

Previous round (`1bd5cb82`) findings, verified against this head:
- **P1 — merge conflict over #190's privacy mask (risk of silently reverting `wireOriginChannel*` and dropping `referenceable*`)** → **fixed**. Rebased onto `a993977`; merge-base == `main` tip, `mergeable=MERGEABLE`; both the mask and the `referenceable*` keys verified present on list and detail at this head, with this PR's new keys layered on top (`task.go:867-868`, `task.go:1110-1112`). Exactly the requested resolution.
- **P2-1** (detail resolves `creator_name` by a weaker path than list) → not addressed; finding 1 below.
- **P2-2** (per-item IM lookup, N+1) → not addressed; finding 2 below (was optional).
- **P2-3** (no list/detail response tests) → not addressed; finding 4 below.
- **P2-4** (non-bot test passes even if the keys disappear) → not addressed; finding 3 below.
- **P2-5** (test cleanup restores a fabricated default resolver) → not addressed; recorded as a Nit below (was optional).

## Findings

No P0/P1 issues; four P2 items, all carried over from the previous round and non-blocking.

### P2 — Detail resolves `creator_name` by a weaker path than list (`internal/api/handler/task.go:1096`)

`GetSummary` resolves `creator_name` directly via `service.ResolveUserName(task.CreatorID)`, while `ListSummaries` prefers the persisted participant snapshot and only falls back to the IM lookup (`task.go:797-804`). The same task can therefore render different creator names on list vs detail, and within the detail response `creator_name` can disagree with `participants[].user_name` (loaded at `task.go:1045-1046`) if the IM-side name changed after the snapshot was written. Diff-scope: introduced by this PR (the field is new to detail). Not R1-blocking: both values are real names of the same user, so this is display inconsistency, not wrong attribution. Fix: derive `creator_name` from the already-loaded `participants` slice with `ResolveUserName` as fallback, matching the list path.

### P2 — Per-item IM lookup adds an N+1 on bot-heavy pages (`internal/api/handler/task.go:811`)

Each bot-created list item now adds one synchronous IM-DB lookup inside the per-item loop, on top of the pre-existing per-item creator-name lookup; a full page of bot-created summaries doubles that N+1 against the IM database. Diff-scope: new with this PR, but follows the pre-existing per-item pattern (which is why it stays P2/optional). Fix direction: batch-resolve the unique creator/bot uids for the page in one query, or cache by uid within the request.

### P2 — Non-bot test passes even if the keys disappear entirely (`internal/api/handler/bot_summary_create_test.go:163`)

`TestBotCreateResponseNonBotTaskHasEmptyBotName` asserts via comma-ok type assertion: if `botCreateResponse` ever stops emitting `creator_bot_id`/`creator_bot_name`, `resp[key]` is nil, `("", false)` passes the `got != ""` check, and the test still goes green. The always-present-key contract the frontend relies on is therefore unenforced. Fix: assert presence and string type before asserting emptiness:

```go
v, ok := resp["creator_bot_name"]
if !ok {
    t.Fatal("creator_bot_name key missing")
}
if s, isStr := v.(string); !isStr || s != "" {
    t.Fatalf("creator_bot_name=%v, want empty string", v)
}
```

### P2 — New list/detail fields have no response-level tests (`internal/api/handler/task.go:867`)

The new tests only exercise the bot-create response; the list (`task.go:867-868`) and detail (`task.go:1110-1112`) projections are verified by inspection only. A regression there (guard removed, key dropped on one surface) would not be caught. Fix: add a handler test that creates a `TriggerBot` task and asserts `creator_bot_id`/`creator_bot_name` through `ListSummaries` and `GetSummary`, plus emptiness for a manual task.

### Nit — Tests restore a fabricated default resolver (`internal/api/handler/bot_summary_create_test.go:112`)

Both new tests restore an identity closure in `t.Cleanup` instead of the previously installed resolver. Safe today (no `t.Parallel` in the package, `main()` never runs under test, and `go test -race` is green), but load-bearing on an invariant nothing enforces. Consider `SetUserNameResolver` returning the previous fn, or injecting the resolver into `TaskHandler`.

## Blind-spot checklist

- **C1 dual-path parity** — checked: all three declared surfaces emit both keys; fresh-create (201) and idempotent replay (200) share `botCreateResponse` (`bot_summary_create.go:198/:219/:229`); a grep for every `gin.H` projection carrying `creator_id` returns exactly these three surfaces — no sibling projection was missed.
- **C2 control-flow ordering / reuse** — checked: `botCreateResponse` is reused by both replay branches without double effect; no new parsing of user-controlled strings (no sanitize/canonicalization surface).
- **C3 authorization boundary** — checked: the new fields serialize after the existing gates (list visibility `whereSQL`; `authorizeTaskAccess` fail-closed on empty space). `creator_bot_id` is the same exposure class as the already-exposed `creator_id`; no capability is newly reachable.
- **C4 container/member cascade** — N/A: no auth-status or membership-state logic changed.
- **C5 build ≠ runtime** — cleared by execution: build + `-race` tests actually run; resolver injection verified by reading both `main()` wiring sites (runtime path), not just by compilation.
- **C6 governance/docs** — N/A: no docs/policy changes.

Security pass (STRIDE): no new authentication surface; no injection paths (no new user-controlled strings reach SQL/HTML/shell); disclosure class unchanged (bot uid/name exposed under the same gates as owner uid/name); no unbounded allocation or loop. `ResolveUserName` is not space-scoped (a bot's display name resolves regardless of caller space) — pre-existing behaviour shared with `creator_name`, not introduced here.

## Verdict: APPROVED

The previous round's sole blocker (the merge-conflict trap sitting on #190's privacy mask) is resolved by the rebase onto `a993977`, and I verified the required resolution landed correctly: the `OriginFromDerived` wire mask and the `referenceable*` keys are intact on both endpoints with the new identity keys layered on top. The feature's data flow holds end-to-end (create → persist → list/detail/replay projections), name resolution is wired in both production binaries, authorization is unchanged and fail-closed, and `go build` / `go vet` / `go test -race ./internal/api/handler/` are all green at this head. The remaining items are carried-over P2/nit quality suggestions, none of which block merging.

[Octo-Q] verdict: APPROVE — 上轮唯一 blocker（与 #190 隐私掩码的合并冲突陷阱）已通过 rebase 到 a993977 解决并逐行验证掩码与 referenceable* 键完好；数据流端到端验证通过，build/vet/race 测试全绿；仅剩 P2/nit 遗留项，无阻塞。

PR #189, opened 2026-08-11
fix(summary): fail-closed empty X-Space-Id on version-history reads

wanyiwang06:
## 背景

渗透测试报告(复测)4.11「逻辑漏洞·垂直越权」的核心手法:空间级鉴权依赖 `X-Space-Id` 请求头,**删掉该头后后端未做默认拒绝**即可绕过。报告的 PoC 打在主站接口,但同一机制适用于 summary——本 PR 修复 summary 侧最后一处 fail-open 读接口并补齐回归。

## 问题

`ListSummaryVersions` / `GetSummaryVersion` 按 `id=? AND space_id=?` 查询,但**唯独缺少** siblings(`authorizeTaskAccess` / `requireTaskInSpace` / `ListSummaries` / `ListSchedules`)都已具备的 `spaceID=="" → 404` 硬门。由于 `SummaryTask.SpaceID` 为 `not null default ''`,请求方只要**不带 X-Space-Id**,查询就变成 `space_id=''`,会命中任何异常的 `space_id=''` 任务行,并把其版本历史返回给该任务的 creator/participant —— 即报告所述的删头越权。

严重度:低(需存在 `space_id=''` 异常行 + 调用者为其成员),但属于「让空 X-Space-Id 在全读面一律 fail-closed」应堵的一致性缺口。

## 改动

- `edit.go`:`ListSummaryVersions` / `GetSummaryVersion` 在查询前短路 `spaceID=="" → 40008/404`,与其它读路径一致。
- 回归测试(此前该读面无空头测试):
  - `edit_version_space_gate_test.go` —— 两个被修接口的 **FAIL-before/PASS-after**(种真实 `space_id=''` 任务)+ 正确 space 的 no-regression。
  - `list_space_gate_test.go` —— 锁定 `ListSummaries` 空/错 space 门。
  - `schedule_space_gate_test.go` —— 锁定 `ListSchedules` / `GetSchedule` 空 space 门。

## 验证

`golang:1.25-bookworm`(CGO + libtokenizers)内:新增用例 + 既有全套空间隔离用例 + **整个 handler 包**全部通过。

## 审计结论(其余读接口)

`GetSummary/GetResult/Regenerate/Delete/Cancel`、`Accept/Decline/Submit/GetPersonal/Members/Personal-versions`、`ListSchedules/GetSchedule`、`StreamSummary` 均已 fail-closed;`GetTemplates`(空 space 跳过查询)、`InferScope`(不碰 DB)、agent `History`(按 uid)、`ShareHandler.Get`(等值 space + 成员校验)无跨空间泄露面。修复后 summary 全读接口对空 X-Space-Id 一致 fail-closed。

### DIFF
--- a/internal/api/handler/edit.go
+++ b/internal/api/handler/edit.go
@@ -564,6 +564,15 @@ func (h *EditHandler) ListSummaryVersions(c *gin.Context) {
 		}
 	}
 	spaceID := middleware.GetSpaceID(c)
+	// fail-closed hard gate: GET requests are NOT caught by StrictSpaceMiddleware,
+	// and SummaryTask.SpaceID is `not null default ''`, so rows with space_id=''
+	// may exist; querying `space_id=''` would MATCH them, leaking a cross-space
+	// version-history read. Reject an empty X-Space-Id before any query (mirrors
+	// authorizeTaskAccess / requireTaskInSpace / ListSummaries).
+	if spaceID == "" {
+		c.JSON(http.StatusNotFound, apiResponse{Code: 40008, Message: "任务不存在"})
+		return
+	}
 	var task model.SummaryTask
 	if err := h.db.Where("id = ? AND space_id = ? AND deleted_at IS NULL", taskID, spaceID).First(&task).Error; err != nil {
 		c.JSON(http.StatusNotFound, apiResponse{Code: 40008, Message: "任务不存在"})
@@ -618,6 +627,15 @@ func (h *EditHandler) GetSummaryVersion(c *gin.Context) {
 		return
 	}
 	spaceID := middleware.GetSpaceID(c)
+	// fail-closed hard gate: GET requests are NOT caught by StrictSpaceMiddleware,
+	// and SummaryTask.SpaceID is `not null default ''`, so rows with space_id=''
+	// may exist; querying `space_id=''` would MATCH them, leaking a cross-space
+	// version read. Reject an empty X-Space-Id before any query (mirrors
+	// authorizeTaskAccess / requireTaskInSpace / ListSummaries).
+	if spaceID == "" {
+		c.JSON(http.StatusNotFound, apiResponse{Code: 40008, Message: "任务不存在"})
+		return
+	}
 	var task model.SummaryTask
 	if err := h.db.Where("id = ? AND space_id = ? AND deleted_at IS NULL", taskID, spaceID).First(&task).Error; err != nil {
 		c.JSON(http.StatusNotFound, apiResponse{Code: 40008, Message: "任务不存在"})
--- a/internal/api/handler/edit_version_space_gate_test.go
+++ b/internal/api/handler/edit_version_space_gate_test.go
@@ -1,0 +1,142 @@
+//go:build cgo
+
+package handler
+
+import (
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"net/http/httptest"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/Mininglamp-OSS/octo-smart-summary/internal/middleware"
+	"github.com/Mininglamp-OSS/octo-smart-summary/internal/model"
+	"github.com/gin-gonic/gin"
+	"gorm.io/gorm"
+)
+
+// ---------------------------------------------------------------------------
+// Pentest 4.11 (逻辑漏洞·垂直越权) — empty-X-Space-Id fail-closed gate on the
+// summary version-history reads.
+//
+// GET /summaries/:id/versions and /summaries/:id/versions/:result_id previously
+// queried `id=? AND space_id=?` with NO spaceID=="" short-circuit — the only
+// summary read paths missing the hard gate every sibling has
+// (authorizeTaskAccess / requireTaskInSpace / ListSummaries / ListSchedules).
+// Because SummaryTask.SpaceID is `not null default ''`, an anomalous space_id=''
+// row could be read version-by-version by a header-less request from its
+// creator/participant (the exact "delete the X-Space-Id header" bypass). The
+// handlers now short-circuit spaceID=="" -> 40008/404 before any query.
+//
+// These tests SEED a real space_id='' task (+ result version) so the gate is
+// genuinely exercised: FAIL-before -> the empty-header query `space_id=''`
+// matched the seeded row and returned 200; PASS-after -> 404/40008.
+// ---------------------------------------------------------------------------
+
+func setupVersionGateRouter(h *EditHandler) *gin.Engine {
+	gin.SetMode(gin.TestMode)
+	r := gin.New()
+	r.Use(middleware.AuthMiddleware(&mockTokenResolver{}), middleware.SpaceMiddleware())
+	r.GET("/api/v1/summaries/:id/versions", h.ListSummaryVersions)
+	r.GET("/api/v1/summaries/:id/versions/:result_id", h.GetSummaryVersion)
+	return r
+}
+
+// seedEmptySpaceVersionTask seeds a space_id='' task owned by creator1 with one
+// SummaryResult version. Returns (taskID, resultID).
+func seedEmptySpaceVersionTask(t *testing.T, db *gorm.DB) (int64, int64) {
+	t.Helper()
+	now := time.Now().UTC()
+	task := model.SummaryTask{
+		TaskNo:      "TST-EDIT-EMPTYSPACE-001",
+		SpaceID:     "", // <-- the anomalous space_id='' row the gate must hide
+		CreatorID:   "creator1",
+		SummaryMode: model.ModeByPerson,
+		Status:      model.StatusCompleted,
+	}
+	db.Create(&task)
+	db.Create(&model.SummaryParticipant{
+		TaskID: task.ID, UserID: "creator1", UserName: "Creator", Status: model.ParticipantCompleted,
+	})
+	result := model.SummaryResult{
+		TaskID:      task.ID,
+		Content:     "empty-space version content",
+		Version:     1,
+		GeneratedAt: now,
+	}
+	db.Create(&result)
+	return task.ID, result.ID
+}
+
+func doGetWithSpace(r *gin.Engine, path, userID, spaceID string) *httptest.ResponseRecorder {
+	w := httptest.NewRecorder()
+	req := httptest.NewRequest(http.MethodGet, path, nil)
+	if userID != "" {
+		req.Header.Set("Token", userID)
+	}
+	if spaceID != "" {
+		req.Header.Set("X-Space-Id", spaceID)
+	}
+	r.ServeHTTP(w, req)
+	return w
+}
+
+// 4.11: ListSummaryVersions with NO X-Space-Id against a real creator of a
+// space_id='' task must 404/40008 and NOT return the version list.
+// FAIL-before: empty header -> `space_id=''` matched -> 200 with versions.
+func TestListSummaryVersions_EmptySpaceHeader_Returns404(t *testing.T) {
+	db := setupEditDB(t)
+	taskID, _ := seedEmptySpaceVersionTask(t, db)
+	h := NewEditHandler(db)
+	r := setupVersionGateRouter(h)
+
+	w := doGetWithSpace(r, fmt.Sprintf("/api/v1/summaries/%d/versions", taskID), "creator1", "")
+	assertCrossSpace404(t, w)
+
+	var resp struct {
+		Data struct {
+			Versions []interface{} `json:"versions"`
+		} `json:"data"`
+	}
+	_ = json.Unmarshal(w.Body.Bytes(), &resp)
+	if len(resp.Data.Versions) != 0 {
+		t.Errorf("empty-space ListSummaryVersions must NOT return versions, got %d; body=%s", len(resp.Data.Versions), w.Body.String())
+	}
+}
+
+// 4.11: GetSummaryVersion with NO X-Space-Id against a real creator of a
+// space_id='' task must 404/40008 and NOT return the version content.
+// FAIL-before: empty header -> `space_id=''` matched -> 200 with the version.
+func TestGetSummaryVersion_EmptySpaceHeader_Returns404(t *testing.T) {
+	db := setupEditDB(t)
+	taskID, resultID := seedEmptySpaceVersionTask(t, db)
+	h := NewEditHandler(db)
+	r := setupVersionGateRouter(h)
+
+	w := doGetWithSpace(r, fmt.Sprintf("/api/v1/summaries/%d/versions/%d", taskID, resultID), "creator1", "")
+	assertCrossSpace404(t, w)
+
+	if strings.Contains(w.Body.String(), "empty-space version content") {
+		t.Errorf("empty-space GetSummaryVersion must NOT return the version content; body=%s", w.Body.String())
+	}
+}
+
+// No-regression inverse: a genuine in-space task read with the CORRECT header
+// still returns the versions (200), proving the new gate did not over-block.
+func TestSummaryVersions_EmptySpaceGate_NoRegressionInSpace(t *testing.T) {
+	db := setupEditDB(t)
+	taskID, resultID, _ := seedEditableTask(t, db) // space1, creator1, one result
+	h := NewEditHandler(db)
+	r := setupVersionGateRouter(h)
+
+	wList := doGetWithSpace(r, fmt.Sprintf("/api/v1/summaries/%d/versions", taskID), "creator1", "space1")
+	if wList.Code != http.StatusOK {
+		t.Fatalf("in-space ListSummaryVersions must be 200, got %d: %s", wList.Code, wList.Body.String())
+	}
+	wGet := doGetWithSpace(r, fmt.Sprintf("/api/v1/summaries/%d/versions/%d", taskID, resultID), "creator1", "space1")
+	if wGet.Code != http.StatusOK {
+		t.Fatalf("in-space GetSummaryVersion must be 200, got %d: %s", wGet.Code, wGet.Body.String())
+	}
+}
--- a/internal/api/handler/list_space_gate_test.go
+++ b/internal/api/handler/list_space_gate_test.go
@@ -1,0 +1,116 @@
+//go:build cgo
+
+package handler
+
+import (
+	"bytes"
+	"net/http"
+	"testing"
+
+	"github.com/Mininglamp-OSS/octo-smart-summary/internal/model"
+	"gorm.io/gorm"
+)
+
+// ---------------------------------------------------------------------------
+// Pentest 4.11 (逻辑漏洞·垂直越权) regression — ListSummaries empty-X-Space-Id
+// fail-closed hard gate.
+//
+// The report's mechanism: space-level authorization hinges on the X-Space-Id
+// header; when a caller simply REMOVES that header the backend must NOT fall
+// open. summary's SpaceMiddleware deliberately allows an empty space_id
+// (space_id="" => "no isolation") and StrictSpaceMiddleware only enforces the
+// header on WRITE methods, so a GET like /api/v1/summaries reaches the handler
+// with space_id="". SummaryTask.SpaceID is `not null default ''`, so a
+// space_id='' row CAN exist; without an explicit guard the list query
+// `space_id=''` would MATCH such a row and return it (fail-open cross-space
+// read == the 4.11 bypass).
+//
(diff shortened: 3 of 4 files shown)


Jerry-Xin (approved):
Project relevance gate passed: this PR fixes summary API authorization behavior in `Mininglamp-OSS/octo-smart-summary`. The change is in scope and ready to merge.

✅ Highlights

🔵 The new empty-space checks in `internal/api/handler/edit.go:566` and `internal/api/handler/edit.go:629` match the existing fail-closed pattern used by sibling task and schedule read paths.

🔵 The tests seed real `space_id = ""` rows and exercise both the blocked path and normal in-space reads, which is the right regression shape for this issue: `internal/api/handler/edit_version_space_gate_test.go:89`, `internal/api/handler/list_space_gate_test.go:58`, and `internal/api/handler/schedule_space_gate_test.go:51`.

💬 Non-blocking

🟡 Test verification caveat: I could not run the handler tests locally because the environment is missing `libtokenizers`; linking fails with `ld: library 'tokenizers' not found`. This appears environmental, not caused by the PR.

🔵 Test hygiene: a few new seed helpers ignore `db.Create` errors, for example `internal/api/handler/edit_version_space_gate_test.go:59` and `internal/api/handler/list_space_gate_test.go:48`. Existing tests use this style in places, so this is not blocking, but checking those errors would make future schema/setup failures easier to diagnose.

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #189 (octo-smart-summary)
Head: `a3a542955209371c724813da2c05011db1e2fc71` · Base/merge-base: `8edbb3a4` (main, PR is a single commit on current main) · Round: first review of this PR

## Summary

This PR closes the last empty-`X-Space-Id` fail-closed gap on the summary version-history reads: `ListSummaryVersions` and `GetSummaryVersion` now short-circuit `spaceID == ""` to 404/40008 before any query, mirroring the established gates in `authorizeTaskAccess` / `requireTaskInSpace` / `ListSummaries`. Three new cgo test files seed genuine `space_id=''` rows and drive the real handlers through the middleware chain (fail-before assertions plus in-space no-regression inverses). The change is correct, minimal, and consistent with the sibling gates; one pre-existing same-class surface (WS subscribe) is noted below as a non-blocking follow-up.

## Verification

Static analysis only at head `a3a54295`; build and tests not executed in this environment.

- ✅ **Gate placement** — both gates sit after param parsing and before the first DB query (`internal/api/handler/edit.go:567-575`, `:630-638`); no query is reachable with `spaceID==""`.
- ✅ **Comment claims true** — `StrictSpaceMiddleware` enforces the header only for POST/PUT/DELETE (`internal/middleware/space.go` `isWriteMethod`), so GETs reach handlers with `space_id=""` set from the header (`router.go:70` mounts it on the `v1` group); `SummaryTask.SpaceID` is `NOT NULL DEFAULT ''` (`migrations/legacy/001_init.sql:7`), so matching rows can exist.
- ✅ **Response shape parity** — 404 + `apiResponse{Code: 40008, Message: "任务不存在"}` is byte-identical to the task-not-found branch and every sibling gate, so task existence is not leaked and clients see one uniform error.
- ✅ **No legitimate regression** — tasks are created via POST paths where the middleware (or bot auth) rejects an empty space, so no healthy task carries `space_id=''`; only anomalous/legacy rows are hidden. Inverse tests lock in-space 200s.
- ✅ **Test package compiles (static)** — `mockTokenResolver` reused from `auth_test.go` (no redeclaration); all referenced helpers (`setupEditDB`, `seedEditableTask`, `assertCrossSpace404`, `setupListTestDBs`, `setupListRouter`, `doRequestWithSpace`, `parseListResponse`, `seedListTask`, `newScheduleTestDB`, `scheduleReq`, `sid`) exist with matching signatures; `//go:build cgo` matches package convention. Tests exercise the production handlers through `AuthMiddleware`+`SpaceMiddleware` (equivalent to production GET behavior), not hand-fed mocks.

## Findings

No P0/P1 issues; one P2 (pre-existing, out of this diff's scope).

### P2 — Same-class gap remains on the WS subscribe surface (pre-existing) (`internal/api/handler/edit.go:567`)

The gate this PR adds for HTTP reads has a twin gap on `/ws` + `/ws/summaries`: those routes use non-strict `SpaceMiddleware` (`internal/api/router/router.go:40-41`), and `subscribe` admits tasks via `taskBelongsToSpace(tid, cw.spaceID)` (`internal/api/ws/hub.go:117-127`). With the header omitted, `spaceID==""` matches exactly the anomalous `space_id=''` tasks this PR hides, so a header-less authenticated client can still subscribe to them (task IDs are sequential int64, trivially enumerable) and receive broadcast events (`internal/api/handler/internal.go:76-78`, `internal/api/handler/personal.go:561`). The comment at `internal/api/ws/hub.go:84` ("WS will receive all events") is also stale relative to the actual filter. Diff-scope three-question answer: pre-existing (hub.go untouched by this PR), not amplified by this PR, base behavior identical — hence P2/follow-up, not a blocker. Recommend a follow-up PR rejecting empty `spaceID` at WS upgrade or fail-closing in `subscribe`.

## Human-verify

1. Confirm whether `space_id=''` rows actually exist in the production `summary_task` table (the exploit precondition for both the fixed HTTP paths and the residual WS path). **Not a merge blocker for this PR.**
2. Confirm the WS surface above should be tracked as the next pentest-4.11 follow-up. **Not a merge blocker for this PR.**

- **Full GET-surface parity audit**: every other space-scoped read is gated or fail-closed — `GetSummary`/`GetResult` via `authorizeTaskAccess` (`internal/api/handler/task.go:111-130`), `ListSummaries` (`:487`), `StreamSummary` (`internal/api/handler/stream.go:81`), `GetPersonal`/`ListPersonalVersions`/`GetPersonalVersion`/`RestorePersonalVersion` via `requireTaskInSpace` (`internal/api/handler/personal.go:155-160`, `internal/api/handler/personal_refine.go:488-600`), `GetMembers` (`internal/api/handler/personal.go:1073-1081`), `ListSchedules`/`GetSchedule` (`internal/api/handler/schedule.go:801-808`, `:850-857`), `GetTemplates` guarded by `spaceID != ""` before the custom-template query (`internal/api/handler/task.go:1497`), `shareH.Get` fail-closed via `canReadGrant`→`activeSpaceMember("")==false` (`internal/api/handler/share.go:95`). Write methods (incl. `RestoreSummaryVersion` POST) are rejected by `StrictSpaceMiddleware`; bot-realm space is server-resolved and required non-empty.
- **No over-blocking**: gate returns the same 404/40008 as a missing task; no existence oracle introduced.

## 数据流回溯 (data-flow traceback)

- `spaceID` ← `middleware.GetSpaceID(c)` ← gin context key `space_id` ← `StrictSpaceMiddleware` reading `X-Space-Id` (fallback `X-Org-Id`). On GET the middleware sets `""` and continues, so runtime data genuinely flows an empty value into both handlers → the new gate is reachable and fires before any consumer (the task query). No upstream early-return/gate short-circuits it.
- `taskID`/`resultID` ← path params, `ParseInt`-validated before use (unchanged).
- In tests, `Token: creator1` → `mockTokenResolver.ResolveUID` returns the token verbatim → `userID="creator1"`, and `SpaceMiddleware` on the test router has identical empty-space semantics to production `StrictSpaceMiddleware` for GET — so the seeded `space_id=''` rows genuinely reach the production gate (no hand-fed bypass).

## 盲点 checklist (C1–C6)

- **C1 双路径 parity** — hit & cleared: both endpoints of the pair (list+get version) gated identically; sibling-path parity proven by the full GET-surface audit above (grep over every `GetSpaceID` consumer).
- **C2 control-flow ordering / reuse** — cleared: gate is inline per handler (no shared-helper reuse hazard), placed after param parsing and before the first query in both; non-canonical probe (whitespace-only `X-Space-Id`) passes the `==""` compare but cannot match any row not created with the identical header value — no exploitable delta, consistent with all sibling gates.
- **C3 授权边界** — N/A beyond covered: no new endpoint/tool/capability exposed; the PR only narrows reachability.
- **C4 授权生命周期/容器级联** — N/A: no membership-status or container-state logic changed.
- **C5 build/tests ≠ runtime** — tests not executed here (static only); runtime path reasoned: middleware chain, route registration (`router.go:89-90`), and response wiring verified statically; test helpers all exist, so the cgo test package has no static compile blocker.
- **C6 治理/文档** — N/A: no governance/policy docs changed.

## 跨轮 blocker 复检 (R6)

N/A — first review round of this PR (no prior Octo findings on this issue).

## Verdict: COMMENT

The two added gates are correct, fail-closed, parity-consistent with every sibling gate, and covered by genuine fail-before/pass-after regression tests; no P0/P1 exists in the diff. The single P2 is a pre-existing same-class exposure on the WS subscribe surface that deserves a tracked follow-up but does not block this PR.

[Octo-Q] verdict: APPROVE — 仅 P2（既有、diff 外）；diff 内无 P0/P1，按 R4 映射 APPROVE。WS subscribe 面同类缺口建议另开 follow-up。

yujiawei (approved):
# Code Review — PR #189 (octo-smart-summary)

Reviewed at head `a3a5429`, merge-base `8edbb3a`. Diff is 4 files, +365/-0: two handler guards in `internal/api/handler/edit.go` and three new test files. I checked out the head commit, re-ran the audit the description claims, and mutation-tested every new test.

---

## 1. Spec compliance

**Spec: ✅**

- **Missing work: none.** The description promises (a) a `spaceID == ""` short-circuit in `ListSummaryVersions` and `GetSummaryVersion`, and (b) regression tests for those two endpoints plus lock-in tests for `ListSummaries` / `ListSchedules` / `GetSchedule`. All present: `edit.go:567-574`, `edit.go:630-637`, and the three `*_space_gate_test.go` files.
- **Extra work: none.** No new flags, fields, or endpoints. Zero deletions. The two lock-in test files cover endpoints not touched by this diff, but they are explicitly declared under 改动 in the description, so this is stated scope, not creep.
- **Deviation: none.** The guard shape (short-circuit before any query, `40008`/404, identical to a genuinely missing task so existence is not leaked) matches `authorizeTaskAccess` (`task.go:122-131`), `requireTaskInSpace` (`personal.go:154-162`), `GetMembers` (`personal.go:1073-1081`), and `ListSummaries` (`task.go:483-490`) exactly.

One caveat on the **description's** audit statement, not on the diff — see P2-1 and P2-2. The claim "修复后 summary 全读接口对空 X-Space-Id 一致 fail-closed" is not quite true; two read surfaces outside this diff remain fail-open. That does not block this change, which is a strict improvement.

### Vulnerability and fix, independently confirmed

I reverted only the two guards and re-ran the new tests. Both fail, and they fail by *leaking*, not by an assertion technicality:

```
--- FAIL: TestListSummaryVersions_EmptySpaceHeader_Returns404
    expected 404 for cross-space access, got 200:
    {"code":0,...,"versions":[{"result_id":1,"version":1,...}]}
--- FAIL: TestGetSummaryVersion_EmptySpaceHeader_Returns404
    expected 404 for cross-space access, got 200:
    {"code":0,...,"content":"empty-space version content",...}
```

`TestSummaryVersions_EmptySpaceGate_NoRegressionInSpace` stays green with the guards removed, so the pair is a real before/after, not a tautology. Restoring the guards turns all three green.

I ran the same mutation on the three lock-in tests by removing the pre-existing gates at `task.go:485-490`, `schedule.go:802-808`, and `schedule.go:851-857`. All three tests fail, each with a 200 leaking the seeded `space_id=''` row. None of the new tests is vacuous.

Full suite with CGO (matching the `test-race` CI job, which is the only job where these `//go:build cgo` files actually run):

```
ok  .../internal/api/handler    11.196s
ok  .../internal/middleware      0.007s
go vet ./internal/api/handler/   (clean)
```

Build-tag choice is consistent with the package — 22 of 35 test files there are already `//go:build cgo`.

### Re-audit of every remaining space-scoped read route

I walked the router rather than trusting the description. Result for GET routes:

| Route | Empty-header behaviour | Where |
|---|---|---|
| `/summaries` | 404 gate | `task.go:485` |
| `/summaries/:id`, `/summaries/:id/result` | 404 gate via `authorizeTaskAccess` | `task.go:122` |
| `/summaries/:id/versions{,/:result_id}` | **404 gate (this PR)** | `edit.go:567,630` |
| `/summaries/:id/personal-versions{,/:version_id}` | 404 gate via `requireTaskInSpace` | `personal.go:154` |
| `/summaries/:id/personal`, `/summaries/:id/members` | 404 gate | `personal.go:154,1073` |
| `/summaries/:id/stream` | 403 on empty space | `stream.go:80-84` |
| `/summary-schedules{,/:id}` | 404 gate | `schedule.go:802,851` |
| `/summary-templates` | skips the query when space is empty | `task.go:1497` |
| `/summary-infer` | no DB access | `task.go:1367` |
| `/summary-member-candidates` | 403 `40300` | `candidates.go:70` |
| `/summary-shares/:share_id` | fail-closed, but *indirectly* — see below | `share.go:370` |
| `/agent/chat/history` | uid-scoped, space-independent | `agent_chat.go:411` |
| `/summary-chat-candidates` | **fail-open** | `candidates.go:170,215,261` |
| `/ws`, `/ws/summaries` | **fail-open** | `ws/hub.go:83-84,121` |
| `/api/v1/bot/*` | space comes from the bot token, non-empty enforced | `middleware/space.go:41` |

`ShareHandler.Get` deserves a note because the description files it under "等值 space". Equality alone would *not* be enough: an empty header plus a `space_id=''` snapshot satisfies `snapshot.SpaceID != spaceID` (`share.go:371`). What actually closes it is `activeSpaceMember`, which returns `false` outright when `spaceID == ""` (`share.go:95`), and `canUseTarget` funnels both the DM and Group branches through it (`share.go:112,122`). So the conclusion holds, but via the membership check rather than the equality check. Worth knowing if anyone later refactors `activeSpaceMember`.

---

## 2. Code quality

**Quality: Approved**

No P0 or P1. The change is minimal, placed before any DB access in both handlers, leaves the downstream creator/participant checks untouched, and returns a response byte-identical to the not-found path. Nothing below blocks merge.

### Considered and dismissed: whitespace bypass of `spaceID == ""`

The most obvious attack on an exact-empty-string comparison is `X-Space-Id: " "` — the application check passes, and under a MySQL `PAD SPACE` collation `WHERE space_id = ' '` still matches a `space_id = ''` row, defeating the patch. `middleware.GetSpaceID` does no trimming (`middleware/space.go:58-66,71-75`), so the guard's correctness rests entirely on the HTTP layer.

It holds. Go's `net/http` strips leading and trailing OWS from header values at parse time, so an all-whitespace value reaches the handler as `""`:

```
X-Space-Id: <empty>     -> "" (len 0)
X-Space-Id: " "         -> "" (len 0)
X-Space-Id: "\t"        -> "" (len 0)
X-Space-Id: "  \t  "    -> "" (len 0)
X-Space-Id: "a "        -> "a" (len 1)
```

(Measured with `http.ReadRequest` on raw wire bytes.) Non-OWS whitespace such as `\v` survives trimming but is not space-padded by MySQL either, so it cannot collide with `''`. No change needed — but adding `strings.TrimSpace` inside `GetSpaceID` would make the guard robust without depending on the transport, and would also normalise the `space_id` used in every `WHERE` clause. Optional.

### P2-1 — `/summary-chat-candidates` drops the space filter entirely on an empty header

`candidates.go:170-172` and `215-217` apply the space predicate only conditionally:

```go
if spaceIDStr != "" {
    q = q.Where("g.space_id = ?", spaceIDStr)
}
```

and the direct-chat branch does the same inline via `(? = '' OR ...)` at `candidates.go:261`. With no `X-Space-Id`, the space scope disappears and the only remaining filter is `gm.uid = <caller>`, so the response merges the caller's groups and threads across every space they belong to. Impact is bounded — it is only the caller's own memberships, not another tenant's data — which is why this is P2 and not a blocker. But it is the same fail-open shape this PR is closing, and its sibling `SearchCandidates` was already hardened to `403 40300` at `candidates.go:70-73`. This is the reason the description's "全读接口一致 fail-closed" claim overreaches.

### P2-2 — `/ws` and `/ws/summaries` are fail-open on empty space, and subscribe has no participant check

`ws/hub.go:83-84` is explicit about it:

```go
spaceID := middleware.GetSpaceID(c)
// Allow empty spaceID: WS will receive all events (no space isolation in that case)
```

Those routes mount the non-strict `SpaceMiddleware` (`router.go:40-41`), and `subscribe` authorises through `taskBelongsToSpace(tid, cw.spaceID)` (`hub.go:121`), which becomes `WHERE space_id = ''` — matching exactly the anomalous rows this PR is defending against.

Two separate gaps, and the second is the larger one: `taskBelongsToSpace` checks *space only*, with no creator/participant condition, so any authenticated member of space X can subscribe to any task ID in space X and receive its event stream. Compare `StreamSummary`, the SSE equivalent, which requires both a non-empty space and `canAccessTask` (`stream.go:80-86`). Out of scope for this diff — I would not hold the PR for it — but it deserves its own issue, and it is worth deciding which of the two is the real priority.

### P2-3 — seed helpers swallow `db.Create` errors, so a schema drift would silently make the tests vacuous

`seedEmptySpaceVersionTask` (`edit_version_space_gate_test.go:59-81`) and `seedEmptySpaceListTask` (`list_space_gate_test.go:39-51`) discard every `db.Create` return value. If a future `AutoMigrate` change stops creating one of these tables, the seed no-ops, the endpoint 404s because the row is absent rather than because the gate fired, and the test still passes — the FAIL-before property is lost silently. `seedEmptySpaceSchedule` (`schedule_space_gate_test.go:39-49`) already does this right with `t.Fatalf` on error; worth matching in the other two. Cheap insurance for tests whose entire value is the negative case.

### P2-4 — `doGetWithSpace` duplicates an existing helper

`doGetWithSpace` (`edit_version_space_gate_test.go:83-94`) is line-for-line identical to `doRequestWithSpace(r, "GET", path, userID, spaceID)` in `auth_test.go:88-99`, which the sibling `list_space_gate_test.go` already uses. Prefer the existing one.

### P2-5 — per-handler enforcement will keep leaking through new endpoints

Five call sites now carry a near-identical copy of this guard, each with its own copy of the explanatory comment. The comment itself names the root cause: `StrictSpaceMiddleware` only enforces the header on writes (`middleware/space.go:167-174`), so every space-scoped GET has to remember the boilerplate. The next GET endpoint that forgets it reintroduces this exact bug, and there is no test or lint that would catch the omission — that is precisely how these two handlers ended up as the last stragglers.

The structural fix is to enforce it once at the route group. That is a larger change than this PR should carry, and the per-handler guard is the right defence-in-depth regardless, so this is not an objection to the approach here. Two directions when someone picks it up: extend `StrictSpaceMiddleware` to reject an empty header on all methods for the `v1`/`p2` groups (the WS and bot groups already differ and would need care), or introduce a small `requireSpace(c) (string, bool)` helper so the five copies collapse to one and new endpoints have something obvious to call.

---

## 3. Overall verdict

**APPROVE**

Spec ✅ and Quality Approved. The vulnerability is real, the fix is correct and minimal, and the tests genuinely fail without it — I verified all three claims by mutation rather than by reading. Every P2 above is either outside this diff or a test-hygiene nit.

## 4. Suggested follow-ups

1. File an issue for the WS subscribe path (P2-2). The missing participant check is the more serious half; the empty-space fall-through is the half that matches this PR's threat model. Mirror `StreamSummary`'s two checks.
2. Decide on `/summary-chat-candidates` (P2-1): either fail-closed like `SearchCandidates`, or document why cross-space mixing of the caller's own chats is intended. Right now it is neither.
3. In this PR if convenient, otherwise trivially later: add `t.Fatalf` on the seed `Create` calls (P2-3) and drop `doGetWithSpace` in favour of `doRequestWithSpace` (P2-4).
4. Separately, consider `strings.TrimSpace` in `GetSpaceID` so the guard does not depend on `net/http` OWS stripping, and so `space_id` is normalised for every query that uses it.

## 5. Additional observations

- `resolveSummaryTaskParam` (`task.go:288-295`) resolves `task_no` → id with `WHERE task_no = ? AND deleted_at IS NULL` and no space predicate. Not exploitable: `GetSummary` immediately re-scopes through `authorizeTaskAccess`, and both failure paths return the same `40008`, so no existence oracle. Noting it because it is the one place in the read path that touches the task table before the space check.
- The test routers use `SpaceMiddleware` rather than the production `StrictSpaceMiddleware`. For GET both behave identically (neither rejects an empty header), so fidelity is preserved, and this matches the convention across the package.
- Only the `test-race` CI job exercises these tests; the default `Test` job runs `CGO_ENABLED=0` and skips every `//go:build cgo` file. Correct per existing convention, just worth remembering when reading a green check on the main test job.

PR #191, opened 2026-08-17
fix(summary): resolve agent-path source names so chips stop showing raw channel ids

### LINKED ISSUE #192: agent 保存路径不解析 source_name：总结详情 chip 显示裸 channel ID（ST20260811jcw59o84）
## 现象

总结详情/列表的 summary-detail-source-chip 显示裸 channel ID 而非可读来源名。

实测案例（2026-08-17，线上）：

- 总结 `ST20260811jcw59o84` 的 chip 显示「来自fef0c26327664232a2fd73ce2fb98ec6」
- 该群在 IM 库 `group` 表名字完好（`一个群聊`，UTF-8 正常）——不是数据损坏，是代码没查

## 根因

`POST /api/v1/summaries/agent`（CreateAgentSummary）写 `summary_source` 行时只存 `source_id`、`source_name` 留空（代码原注释：「a future PR can plumb imDB in and call ResolveSourceNameWithType」）。即时总结（task.go CreateSummary）、定时总结（buildScheduledTaskSources）、bot 总结（bot_summary_create.go）三条路径全都实时解析名字，唯独 agent 路径没接 imDB。

前端 chip 渲染 `{src.source_name || src.source_id}`，空名字 fallback 显示裸 ID。

线上 `octo_summary.summary_source` 54 行中有 10 行 source_name 为空，全部出自 agent 路径。

## 验收标准

- agent 保存路径创建 source 行时解析 source_name（与即时路径对齐；客户端传的名字不信；nil imDB / 查不到落确定性兜底「来源-xxxxxxxx」，不存空串）
- 存量空名行在 list/detail/share 读取时现场 resolve 兜底（已存非空名字是创建时快照，不被覆盖）
- 回归测试 mutation-verified（修复行 revert 即 RED）

wanyiwang06:
Closes #192

fix(summary): resolve agent-path source names so chips stop showing raw channel ids

## 背景

主人实测反馈(2026-08-17):总结详情 ST20260811jcw59o84 的 summary-detail-source-chip
显示「来自fef0c26327664232a2fd73ce2fb98ec6」——一长串裸 channel ID,不是可读的群名。
线上 octo_summary.summary_source 共 54 行:43 行正常、**10 行 source_name 为空**(全部
出自 agent 保存路径)、另有 29 行是 IM 库种子数据双重编码的乱码(另一根因,不在本 PR)。

## 问题

`POST /api/v1/summaries/agent`(CreateAgentSummary)写 source 行时**只存 source_id,
source_name 留空**——代码原注释自己承认:「We do not currently resolve source names via
IM DB ... a future PR can plumb imDB in and call ResolveSourceNameWithType」。对比之下,
即时总结(task.go CreateSummary)、定时总结(buildScheduledTaskSources)、bot 总结
(bot_summary_create.go)三条路径全都实时调 `ResolveSourceNameWithType` 解析名字。

前端 chip 渲染逻辑是 `{src.source_name || src.source_id}`——source_name 为空就 fallback
显示裸 ID。讽刺的是 ST20260811jcw59o84 对应群 `fef0c26...ec6` 在 IM 库里名字完好
(`一个群聊`),纯粹是 agent 路径没接 imDB。

## 改动

两层修复(修未来 + 修存量):

1. **写侧**:AgentSummaryHandler 接上 imDB(router.go 已持有,仅穿线),
   CreateAgentSummary 创建 source 行时调 `ResolveSourceNameWithType`——与即时路径完全
   对齐:客户端传的 source_name 不信(即时路径同样丢弃),imDB 为 nil / 查不到时落到
   确定性兜底「来源-xxxxxxxx」,永不再存空串。
2. **读侧**:新增 `displaySourceName`(task.go)——已存的非空名字是创建时快照,**原样
   返回不覆盖**(后续群改名不改写历史);空名字现场查 IM DB 兜底。接到三处用户可见
   序列化点:ListSummaries、GetSummary、share.Create 快照(ShareHandler 已持有 imDB)。
   这样 10 行存量空名记录无需数据迁移即可显示可读名字。

不碰 reference-context prompt(buildReferencedSummariesContext)——那是 LLM 上下文非
用户可见 chip,且给 AgentChatHandler 通 imDB 会扩大本 PR 面。

## 回归测试

新增 `source_name_display_test.go`,7 个用例,全部 mutation-verified(修复行 revert 即 RED):
- 写侧:IM 库解析带类型后缀「一个群聊(群聊)」;客户端伪造名被丢弃改用真实名;
  nil imDB 落「来源-fef0c263(群聊)」兜底
- 读侧:空名 detail/list 现场解析复现 ST20260811jcw59o84 场景;已存快照不被改名覆盖;
  nil imDB 落确定性兜底而非空串

现有测试 16 处 NewAgentSummaryHandler 调用点签名同步(imDB=nil,均不断言 source 名)。

## 验证

- CGO_ENABLED=0 go test ./... 全绿
- CGO_ENABLED=1 go test -race ./internal/api/handler/ 全绿(16.9s)
- 三个方向的 mutation 验证:写侧 revert→RED、读侧兜底 revert→RED、无条件覆盖快照→RED

### DIFF
--- a/internal/api/handler/agent_summary.go
+++ b/internal/api/handler/agent_summary.go
@@ -36,6 +36,7 @@ import (
 //     boundary the Chat handler already enforces.
 type AgentSummaryHandler struct {
 	db           *gorm.DB
+	imDB         *gorm.DB
 	llmApiURL    string
 	llmApiKey    string
 	llmModel     string
@@ -60,9 +61,10 @@ type refineRunner interface {
 	RunWithHistory(ctx context.Context, system string, history []agent.Message, userInput string) (string, []agent.Message, error)
 }
 
-func NewAgentSummaryHandler(db *gorm.DB, llmApiURL, llmApiKey, llmModel string, llmTimeout, llmMaxTokens int) *AgentSummaryHandler {
+func NewAgentSummaryHandler(db, imDB *gorm.DB, llmApiURL, llmApiKey, llmModel string, llmTimeout, llmMaxTokens int) *AgentSummaryHandler {
 	return &AgentSummaryHandler{
 		db:           db,
+		imDB:         imDB,
 		llmApiURL:    llmApiURL,
 		llmApiKey:    llmApiKey,
 		llmModel:     llmModel,
@@ -357,10 +359,12 @@ func (h *AgentSummaryHandler) CreateAgentSummary(c *gin.Context) {
 
 		// Sources: agent-produced summaries carry their own source list from
 		// the front-end (which knows the origin channel + any additional
-		// referenced channels). We do not currently resolve source names via
-		// IM DB — the deliverable already contains channel names in prose —
-		// so we store source_id only; a future PR can plumb imDB in and call
-		// ResolveSourceNameWithType if the UI wants the resolved display name.
+		// referenced channels). Source names are resolved from the IM DB the
+		// same way the instant path (CreateSummary) does, so the detail chip
+		// shows a human-readable name instead of a raw channel id; the
+		// client-supplied name is deliberately NOT trusted (the instant path
+		// drops it too — see task.go's CreateSummary). ResolveSourceNameWithType
+		// handles a nil imDB with a deterministic "来源-xxxxxxxx" fallback.
 		//
 		// R10: dedup by (source_type, source_id) — uk_summary_source_task_type_id
 		// (migration 20260814-01) would turn duplicate front-end input into a
@@ -379,6 +383,7 @@ func (h *AgentSummaryHandler) CreateAgentSummary(c *gin.Context) {
 				TaskID:     createdTaskID,
 				SourceType: s.SourceType,
 				SourceID:   s.SourceID,
+				SourceName: service.ResolveSourceNameWithType(s.SourceID, s.SourceType, h.imDB),
 			}
 			if err := tx.Create(&src).Error; err != nil {
 				return fmt.Errorf("create summary_source: %w", err)
--- a/internal/api/handler/agent_summary_origin_derive_test.go
+++ b/internal/api/handler/agent_summary_origin_derive_test.go
@@ -92,7 +92,7 @@ func seedPipelineRefTask(t *testing.T, h *AgentSummaryHandler, taskNo string) mo
 // rows → save succeeds with the source channel as origin.
 func TestCreateAgentSummary_DeriveOriginFromSingleSource(t *testing.T) {
 	db := setupAgentSummaryTestDB(t)
-	h := NewAgentSummaryHandler(db, "", "", "", 0, 0)
+	h := NewAgentSummaryHandler(db, nil, "", "", "", 0, 0)
 	r := setupAgentSummaryRouter(h)
 
 	ref := seedPipelineRefTask(t, h, "ST-PIPE-001")
@@ -142,7 +142,7 @@ func TestCreateAgentSummary_DeriveOriginFromSingleSource(t *testing.T) {
 // precedent.
 func TestCreateAgentSummary_DeriveOriginMultiSourceInheritsFirst(t *testing.T) {
 	db := setupAgentSummaryTestDB(t)
-	h := NewAgentSummaryHandler(db, "", "", "", 0, 0)
+	h := NewAgentSummaryHandler(db, nil, "", "", "", 0, 0)
 	r := setupAgentSummaryRouter(h)
 
 	ref := seedPipelineRefTask(t, h, "ST-PIPE-002")
@@ -190,7 +190,7 @@ func TestCreateAgentSummary_DeriveOriginMultiSourceInheritsFirst(t *testing.T) {
 // row wins. If nothing usable remains the save still 40001s.
 func TestCreateAgentSummary_DeriveOriginSkipsUnusableRows(t *testing.T) {
 	db := setupAgentSummaryTestDB(t)
-	h := NewAgentSummaryHandler(db, "", "", "", 0, 0)
+	h := NewAgentSummaryHandler(db, nil, "", "", "", 0, 0)
 	r := setupAgentSummaryRouter(h)
 
 	ref := seedPipelineRefTask(t, h, "ST-PIPE-003")
@@ -240,7 +240,7 @@ func TestCreateAgentSummary_DeriveOriginSkipsUnusableRows(t *testing.T) {
 // the fail-closed contract is unchanged for genuinely origin-less summaries.
 func TestCreateAgentSummary_DeriveOriginNoSourcesStill40001(t *testing.T) {
 	db := setupAgentSummaryTestDB(t)
-	h := NewAgentSummaryHandler(db, "", "", "", 0, 0)
+	h := NewAgentSummaryHandler(db, nil, "", "", "", 0, 0)
 	r := setupAgentSummaryRouter(h)
 
 	ref := seedPipelineRefTask(t, h, "ST-PIPE-004") // no summary_source rows
@@ -285,7 +285,7 @@ func TestCreateAgentSummary_DeriveOriginNoSourcesStill40001(t *testing.T) {
 // source channels — same authz posture as tier-3.
 func TestCreateAgentSummary_DeriveOriginNoAccessRefused(t *testing.T) {
 	db := setupAgentSummaryTestDB(t)
-	h := NewAgentSummaryHandler(db, "", "", "", 0, 0)
+	h := NewAgentSummaryHandler(db, nil, "", "", "", 0, 0)
 	r := setupAgentSummaryRouter(h)
 
 	now := time.Now()
@@ -334,7 +334,7 @@ func TestCreateAgentSummary_DeriveOriginNoAccessRefused(t *testing.T) {
 // consulted (regression guard for the pre-existing path).
 func TestCreateAgentSummary_Tier3BorrowStillWins(t *testing.T) {
 	db := setupAgentSummaryTestDB(t)
-	h := NewAgentSummaryHandler(db, "", "", "", 0, 0)
+	h := NewAgentSummaryHandler(db, nil, "", "", "", 0, 0)
 	r := setupAgentSummaryRouter(h)
 
 	now := time.Now()
@@ -396,7 +396,7 @@ func TestCreateAgentSummary_Tier3BorrowStillWins(t *testing.T) {
 // same as an inaccessible task.
 func TestCreateAgentSummary_OriginBorrowRefusesDeletedReferencedTask(t *testing.T) {
 	db := setupAgentSummaryTestDB(t)
-	h := NewAgentSummaryHandler(db, "", "", "", 0, 0)
+	h := NewAgentSummaryHandler(db, nil, "", "", "", 0, 0)
 	r := setupAgentSummaryRouter(h)
 
 	now := time.Now()
@@ -449,7 +449,7 @@ func TestCreateAgentSummary_OriginBorrowRefusesDeletedReferencedTask(t *testing.
 
 func TestCreateAgentSummary_OriginBorrowRefusesNonCompletedReferencedTask(t *testing.T) {
 	db := setupAgentSummaryTestDB(t)
-	h := NewAgentSummaryHandler(db, "", "", "", 0, 0)
+	h := NewAgentSummaryHandler(db, nil, "", "", "", 0, 0)
 	r := setupAgentSummaryRouter(h)
 
 	now := time.Now()
@@ -505,7 +505,7 @@ func TestCreateAgentSummary_OriginBorrowRefusesNonCompletedReferencedTask(t *tes
 // partial lineage.
 func TestCreateAgentSummary_OverCapReferencedTaskIDsRejected400(t *testing.T) {
 	db := setupAgentSummaryTestDB(t)
-	h := NewAgentSummaryHandler(db, "", "", "", 0, 0)
+	h := NewAgentSummaryHandler(db, nil, "", "", "", 0, 0)
 	r := setupAgentSummaryRouter(h)
 
 	overCap := make([]int64, 0, maxReferencedTaskIDs+3)
@@ -550,7 +550,7 @@ func TestCreateAgentSummary_OverCapReferencedTaskIDsRejected400(t *testing.T) {
 // dedup-then-check posture).
 func TestCreateAgentSummary_DuplicateIDsCollapseBelowCapAccepted(t *testing.T) {
 	db := setupAgentSummaryTestDB(t)
-	h := NewAgentSummaryHandler(db, "", "", "", 0, 0)
+	h := NewAgentSummaryHandler(db, nil, "", "", "", 0, 0)
 	r := setupAgentSummaryRouter(h)
 
 	ref := seedPipelineRefTask(t, h, "ST-PIPE-DUP-01")
@@ -594,7 +594,7 @@ func TestCreateAgentSummary_DuplicateIDsCollapseBelowCapAccepted(t *testing.T) {
 // borrowable citations the marker dangles and must be stripped before save.
 func TestCreateAgentSummary_DanglingMarkersStrippedWhenBorrowRedacted(t *testing.T) {
 	db := newFileBackedAgentSummaryTestDB(t)
-	h := NewAgentSummaryHandler(db, "", "", "", 0, 0)
+	h := NewAgentSummaryHandler(db, nil, "", "", "", 0, 0)
 	r := setupAgentSummaryRouter(h)
 
 	now := time.Now()
@@ -670,7 +670,7 @@ func TestCreateAgentSummary_DanglingMarkersStrippedWhenBorrowRedacted(t *testing
 // happy path.
 func TestCreateAgentSummary_MarkersPreservedWhenBorrowSucceeds(t *testing.T) {
 	db := newFileBackedAgentSummaryTestDB(t)
-	h := NewAgentSummaryHandler(db, "", "", "", 0, 0)
+	h := NewAgentSummaryHandler(db, nil, "", "", "", 0, 0)
 	r := setupAgentSummaryRouter(h)
 
 	now := time.Now()
--- a/internal/api/handler/agent_summary_test.go
+++ b/internal/api/handler/agent_summary_test.go
@@ -61,7 +61,7 @@ func setupAgentSummaryRouter(h *AgentSummaryHandler) *gin.Engine {
 // behavior is unchanged (no resolve, direct validation and pass).
 func TestCreateAgentSummary_ProvidedOriginChannelDirectPass(t *testing.T) {
 	db := setupAgentSummaryTestDB(t)
-	h := NewAgentSummaryHandler(db, "", "", "", 0, 0)
+	h := NewAgentSummaryHandler(db, nil, "", "", "", 0, 0)
 	r := setupAgentSummaryRouter(h)
 
 	// Seed: session with assistant message (deliverable content)
@@ -117,7 +117,7 @@ func TestCreateAgentSummary_ProvidedOriginChannelDirectPass(t *testing.T) {
 // and successfully creates the task with the resolved values.
 func TestCreateAgentSummary_NotProvidedResolveSuccess(t *testing.T) {
 	db := setupAgentSummaryTestDB(t)
-	h := NewAgentSummaryHandler(db, "", "", "", 0, 0)
+	h := NewAgentSummaryHandler(db, nil, "", "", "", 0, 0)
 	r := setupAgentSummaryRouter(h)
 
 	// Seed: session with assistant message (deliverable) and fetch_channel tool call
@@ -183,7 +183,7 @@ func TestCreateAgentSummary_NotProvidedResolveSuccess(t *testing.T) {
 // returns 400 40001 with the new error message.
 func TestCreateAgentSummary_NotProvidedResolveFailure(t *testing.T) {
 	db := setupAgentSummaryTestDB(t)
-	h := NewAgentSummaryHandler(db, "", "", "", 0, 0)
+	h := NewAgentSummaryHandler(db, nil, "", "", "", 0, 0)
 	r := setupAgentSummaryRouter(h)
 
 	// Seed: session with only assistant content, no fetch_channel tool call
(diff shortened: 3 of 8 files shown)


Jerry-Xin (approved):
The PR is relevant and correctly fixes agent-created summary source names without changing historical snapshot semantics.

💬 Non-blocking

- 🟡 Warning — `internal/api/handler/task.go:791` (`ListSummaries`): may perform one additional IM query per legacy unnamed source. Consider batch-resolving names to avoid amplifying the existing N+1 behavior on pages containing many legacy records. (New rows store a snapshot name so `displaySourceName` short-circuits; only legacy empty-name rows trigger the live lookup.)
- 🔵 Suggestion — `internal/api/handler/share.go:411`: add a share-specific regression test confirming that an empty stored name is resolved into the persisted share snapshot.
- 🔵 Suggestion — `internal/api/handler/agent_summary.go:372`: apply the existing `maxSourceCount` limit to agent requests as well, to bound IM lookups and inserts consistently with the standard create endpoint.

✅ Highlights

- Preserves non-empty stored names as historical snapshots (a later group rename must not rewrite history).
- Ignores untrusted client-provided names and resolves canonical names consistently with the instant path (`CreateSummary`).
- Handles missing IM data (nil imDB / lookup miss / malformed thread id) with deterministic, non-empty fallbacks — the chip never regresses to a raw id or blank.
- Read parity: list / detail / share all resolve via the shared `displaySourceName` helper.
- Correctly excludes derived sources from user-visible projections.
- New regression tests are falsifiable (each goes RED if its fix line is reverted): write-side resolution, spoofed-name rejection, nil-imDB fallback, live resolution of legacy empty-name rows, snapshot non-overwrite, and list+detail behavior.

yujiawei (approved):
# Code Review — PR #191 (octo-smart-summary)

Reviewed at head SHA `f4376e757f0b0ae1d9ba1df3a4706310871e090e`. I checked out that exact commit, read the full diff plus the surrounding call graph, ran the test suite, and independently re-ran the mutation checks claimed in the PR description.

---

## 1. Spec compliance

**Spec: ✅**

- **Missing work:** none. All three read-side serialization points named in the description are actually wired: `ListSummaries` (`internal/api/handler/task.go:791`), `GetSummary` (`internal/api/handler/task.go:1046`), and the share snapshot (`internal/api/handler/share.go:411`). The write-side fix is at `internal/api/handler/agent_summary.go:386`.
- **Extra work:** none. The `NewAgentSummaryHandler` signature change (`internal/api/handler/agent_summary.go:64`) and the single router call site (`internal/api/router/router.go:166`) are the minimum plumbing required; the 16 test call-site updates are mechanical. The description's stated exclusion — leaving `buildReferencedSummariesContext` alone — is honored: `internal/api/handler/agent_reference_context.go:338` still reads `s.SourceName` raw, and that is LLM prompt context, not a user-visible chip.
- **Deviation:** none. The claimed trust boundary holds — the client-supplied `sourceReq.SourceName` (`internal/api/handler/task.go:292-296`) is genuinely dropped on the agent path, matching `CreateSummary` (`internal/api/handler/task.go:447`), `bot_summary_create.go:476`, and `scheduled_replace_helpers.go:77`.

Two verification notes that I confirmed rather than took on faith:

1. **The `//go:build cgo` tag does not hide the new tests from CI.** `.github/workflows/ci.yml:142` runs a `test-race` job with `CGO_ENABLED=1 ... go test -race -shuffle=on ./...` on every non-draft PR, so `source_name_display_test.go` executes in CI even though the `CGO_ENABLED=0` job at line 106 skips it.
2. **The mutation claims are real.** Reverting `agent_summary.go:386` to store an empty name turns all three write-side tests RED; reverting `task.go:1046` back to `s.SourceName` turns the detail tests RED. The assertions are behavioral, not vacuous.

---

## 2. Code quality

**Quality: Approved**

No P0 or P1. Everything below is P2 — worth a follow-up, none of it should hold up the merge.

### P2-1 — A write-time IM outage bakes in a placeholder the read-side healing can never repair

`internal/api/handler/agent_summary.go:386`

```go
SourceName: service.ResolveSourceNameWithType(s.SourceID, s.SourceType, h.imDB),
```

`ResolveSourceNameWithType` collapses three different outcomes into one string: success, lookup miss, and query error all exit through `fallbackSourceName` (`internal/service/summary.go:58-60`, `113`, `117-131`), which always returns a **non-empty** `来源-xxxxxxxx(群聊)`. The read-side healing keys strictly on emptiness:

```go
// internal/api/handler/task.go:184-189
func displaySourceName(s model.SummarySource, imDB *gorm.DB) string {
	if strings.TrimSpace(s.SourceName) != "" {
		return s.SourceName
	}
	return service.ResolveSourceNameWithType(s.SourceID, s.SourceType, imDB)
}
```

**Failure scenario:** the IM DB is unreachable for ten minutes; users save 50 agent summaries in that window. Each row persists `来源-fef0c263(群聊)`. When IM recovers, list, detail, and every future share render that placeholder permanently — the self-healing path this PR just built is skipped because the name is non-empty. The exact symptom the PR set out to kill (an unreadable chip) returns under a different trigger, and this time it is unrepairable without a data migration.

Two mitigating facts that keep this at P2 rather than P1: the three pre-existing write paths already behave identically, so this is parity rather than a new class of defect; and the pre-PR agent-path behavior was strictly worse (raw hex id unconditionally, not just during an outage). The PR is a clear net improvement.

**Suggested fix:** distinguish *error* from *miss* in the resolver — e.g. a `ResolveSourceNameWithTypeOK(...) (string, bool)` variant. Persist `""` on a transient error so the read path heals it later, and persist the placeholder on a genuine miss so a permanently-unknown channel does not pay a lookup on every read. Simply teaching `displaySourceName` to treat a `来源-` prefix as unresolved would fix the outage case but re-open the read-path lookup for every genuinely-unresolvable source, which feeds P2-2.

### P2-2 — `ListSummaries` adds a per-source IM query on top of an existing per-task query

`internal/api/handler/task.go:776-793`

```go
var sources []model.SummarySource
h.db.Where("task_id = ?", t.ID).Find(&sources)
...
"source_name": displaySourceName(s, h.imDB),
```

The list handler already issues one source query per task in the page; the helper now adds a synchronous IM lookup per empty-name source on top of that. Neither call carries `c.Request.Context()`, so a slow IM DB is not cancelled when the client disconnects.

I want to be precise about the blast radius rather than quote a scary multiplier: the lookup only fires for rows with an **empty** stored name, derived rows are skipped before the call (`task.go:783-786`), and every write path now persists a non-empty name (`task.go:447`, `agent_summary.go:386`, `bot_summary_create.go:476`, `scheduled_replace_helpers.go:77`/`:241`, `source_backfill.go:113`). So the affected set is the fixed inventory of legacy rows — it shrinks over time and cannot grow. That is what keeps this at P2. It is still worth a batched resolve (collect the empty-name `source_id`s across the page, one `WHERE group_no IN (...)` query) if the list endpoint is latency-sensitive, and a one-off `UPDATE` to backfill the legacy rows would let you delete the read-side fallback entirely.

### P2-3 — The share wire-up is the one integration point with no test

`internal/api/handler/share.go:411`

The new `source_name_display_test.go` has 7 tests: 3 write-side, 3 detail, 1 list. `grep -i share` over that file returns nothing. The description says the fix is wired into three user-visible serialization points; two of the three are pinned. The share path is also the one with the most distinctive semantics (see P2-4), so it is the one most worth a test.

### P2-4 — The `TrimSpace` guard in `share.Create` is now dead code, and the snapshot text changes

`internal/api/handler/share.go:405-412`

```go
if name := displaySourceName(source, h.imDB); strings.TrimSpace(name) != "" {
	names = append(names, name)
}
```

`fallbackSourceName` never returns an empty string (`internal/service/summary.go:117-131` — worst case `来源-(群聊)` for an empty `source_id`), so this condition is now always true.

Two consequences, and I don't think either is wrong — just undocumented:

- **Probably intended, arguably a fix:** `len(names)` now always equals `visibleCount`, so `SourceName` and `SourceCount` on the snapshot (`share.go:427`) can no longer disagree. Previously an empty-name source was counted but unnamed.
- **Probably not intended:** a share card now renders `来自 来源-fef0c263(群聊)` where it previously omitted that source entirely. A share link reaches a broader audience than task participants, so it is worth confirming a hex-fragment placeholder is preferred over omission there. If omission is preferred, the guard needs to test resolution success, not string emptiness.

Either way, replace the now-tautological condition with the actual intent so a future reader doesn't assume it still filters.

### P2-5 — The agent path has no `maxSourceCount` cap, and this PR makes each element more expensive

`internal/api/handler/agent_summary.go:85` declares `Sources []sourceReq` with no bound. `maxSourceCount = 30` (`internal/api/handler/task.go:30`) is enforced on the traditional path (`task.go:378`) and the bot path (`bot_summary_create.go:260`), but never on the agent path.

The unbounded insert loop pre-dates this PR. What changes is the per-element cost: `agent_summary.go:386` now performs a cross-database IM query for every distinct source **inside the open `h.db.Transaction`** (`agent_summary.go:372-391`), and resolver errors degrade to a fallback name rather than aborting, so the loop runs to completion at full cost. A request carrying thousands of unique source ids holds the primary-DB transaction open across thousands of serial IM round-trips.

A one-line `if len(req.Sources) > maxSourceCount { ... 40003 ... }` brings this path to parity with the other two and caps the amplification.

---

## 3. Verdict

**APPROVE**

Spec ✅ and Quality Approved. The diagnosis is precise, the two-layer fix (repair the write path, heal existing rows on read) is the right shape, the snapshot-vs-live-resolution semantics are correct and deliberately documented, and the tests genuinely fail when the fix is reverted — I verified that myself rather than taking the description's word for it. Everything above is a follow-up, not a blocker.

## 4. Suggested follow-ups

1. Split *error* from *miss* in `ResolveSourceNameWithType` so a transient IM outage stores `""` (healable) instead of a permanent placeholder — P2-1 is the finding most likely to bite in production.
2. Add a share-path test, and decide explicitly whether a share card should show a placeholder or omit an unresolvable source (P2-3, P2-4).
3. Add the `maxSourceCount` check to the agent path (P2-5) — one line, brings all three create paths into agreement.
4. Consider a one-off `UPDATE summary_source SET source_name = ... WHERE source_name = ''` for the legacy rows; that would let the read-side fallback be deleted rather than carried forever.

## 5. Additional observations (outside the diff, not blocking)

- **The new tests only exercise `source_type=1` (group).** The thread branch (`internal/service/summary.go:80-87`) uses MySQL-specific syntax — backtick quoting and `COLLATE utf8mb4_unicode_ci` — and the DM branch reads the `user` table (`summary.go:108`). Neither is covered by the sqlite-backed tests here, so `displaySourceName`'s behavior on threads and DMs is unverified by this PR.
- **`Pluck` is unscoped and unlimited.** `internal/service/summary.go:57` and `:108` filter on `group_no` / `uid` alone, with no `space_id` predicate and no `LIMIT`, into a scalar destination — with a duplicate key across spaces the scan lands on the last row read. This pre-dates the PR and affects every write path equally; flagging it for follow-up, not asserting a defect here.
- **`SummaryShareSnapshot.SourceName` is `varchar(500)`** (`internal/model/summary_share.go:22`) with no truncation guard in `share.go`. Since `names` is now monotonically longer than before, a many-source task sits marginally closer to that ceiling. Pre-existing, low risk at the current data volumes.
- **`buildSnapshotV1` still emits `ChannelNames: []string{}`** (`internal/api/handler/agent_summary.go:~600`, marked "P0.2 will populate"). Now that resolution is wired into this handler, that field is cheap to fill — a natural follow-on, explicitly out of scope here.

## 6. Coverage note — what I did not verify

- No live MySQL or IM database: all execution was against the sqlite-backed test suite, so the thread/DM resolver branches and the real query plans behind the N+1 concern were not exercised.
- The front-end claim that the chip renders `source_name || source_id` is taken from the PR description; that code lives in another repository and I did not read it.
- No load or latency measurement for P2-2 — the severity reasoning is based on bounding the affected row set, not on measured timings.
- Migration and production data state (the reported 10 empty-name rows, and the separately-reported double-encoding corruption the description scopes out) were not independently confirmed.

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #191 (octo-smart-summary)
Head: `f4376e757f0b0ae1d9ba1df3a4706310871e090e` · Base: `main` (merge-base `a9939775`)

## Summary

The agent summary path (`POST /summaries/agent`) historically stored `summary_source` rows with an empty `source_name`, so the front-end source chip (`source_name || source_id`) rendered raw hex channel ids. This PR fixes both ends. The write side plumbs the IM DB into `AgentSummaryHandler` (`internal/api/router/router.go:166`) and resolves names with the same `service.ResolveSourceNameWithType` the instant path already uses (`internal/api/handler/agent_summary.go:386`) — client-supplied names remain untrusted, and input is deduped by (type, id) ahead of the unique index. The read side adds a `displaySourceName` helper (`internal/api/handler/task.go:184`): a non-empty stored name is a creation-time snapshot and is returned as-is (a later rename must not rewrite history); an empty name falls back to a live IM lookup. It is wired into list, detail, and share (`task.go:791`, `task.go:1046`, `internal/api/handler/share.go:411`). A new 375-line test file exercises real HTTP routers; three other test files are mechanical constructor updates.

## Verification

Static analysis only at head `f4376e75`; build and tests not executed in this environment. All conclusions come from full-diff reading plus data-flow tracing against base `a9939775`.

- ✅ **Write path resolves names identically to the instant path** — `agent_summary.go:386` calls the same `service.ResolveSourceNameWithType` used by `CreateSummary` (`internal/api/handler/task.go:447`); the client-supplied name is dropped, and (type,id) dedup precedes the insert so duplicate front-end input cannot 500 on `uk_summary_source_task_type_id`.
- ✅ **Snapshot semantics preserved on read** — `displaySourceName` (`task.go:184`) returns a non-empty stored name unchanged; only empty-name rows take the live-lookup branch, so existing named rows are never rewritten.
- ✅ **All user-visible read paths covered** — list (`task.go:791`), detail (`task.go:1046`), and the share snapshot (`share.go:411`) all go through `displaySourceName`; derived rows are still filtered before display and `SourceCount` still reflects visible rows only (`share.go:422-427`).
- ✅ **Failure containment** — `ResolveSourceNameWithType` handles nil imDB, malformed thread ids, and lookup misses with deterministic non-empty `来源-xxxxxxxx` placeholders (`internal/service/summary.go:43-131`); `cmd/summary-api/main.go:63-68` treats an unreachable IM DB as non-fatal (`imDB = nil`), so name resolution can never 500 either endpoint.
- ✅ **Tests are not hand-fed** — `source_name_display_test.go` drives the real gin routers and handler wiring for list/detail/share (including nil-imDB and empty-name fallback cases) instead of calling the helper with fabricated structs; the three other test updates only track the new constructor signature.

## Findings

No P0/P1 issues; two P2 items below (both non-blocking), plus one pre-existing nit recorded under 额外发现.

### P2 — N+1 serial IM lookups on the read path (`internal/api/handler/task.go:184`)

Every legacy agent-path row with an empty `source_name` triggers one live IM query per source, per render. A list render at the 100-pageSize cap (`task.go:553`) with up to 30 sources per task (`maxSourceCount`, `task.go:30`) is up to ~3000 serial queries on a single request in the worst case; detail and share share the cost shape. Nothing backfills the empty stored names, so this cost persists forever rather than decaying. Diff-scope: new behavior introduced by this PR (the read-side fallback), but the path stays correct and pagination-bounded, hence P2 not P1. Fix direction: batch-resolve per request (one query per source type) or a one-time backfill populating empty `source_name` rows so the fallback stops firing.

### P2 — Transient IM errors get baked into permanent placeholders (`internal/api/handler/agent_summary.go:386`)

`ResolveSourceNameWithType` collapses query error and lookup miss into the same non-empty fallback (`internal/service/summary.go:113-131`), and the new write side persists whatever comes back. If the IM DB is briefly unreachable exactly when an agent summary is saved, the placeholder `来源-xxxxxxxx(群聊)` is written to `source_name` — and because `displaySourceName` only heals *empty* stored names (snapshot semantics, correct in itself), that placeholder renders forever and the read path cannot repair it. Diff-scope: new (this PR is the first time the agent path persists a resolved name at save time). Not P1 — the endpoint keeps working and the fallback is deterministic and type-labeled — but a transient failure becomes permanent wrong-but-plausible display data. Fix direction: distinguish error from miss in the resolver and persist an empty name on transient errors so the read-side fallback (or a backfill) can heal it.

- Constructor call sites: every `NewAgentSummaryHandler` caller updated (router + tests); no stale signature remains.
- Worker-side consumers (`internal/worker/specified_sources.go:30`, `:50`) intentionally read the stored name for machine-facing source maps — not a user-visible display path, no change needed there.
- The deprecated `ResolveSourceName` shim (`internal/service/summary.go:149`) is untouched and unused by every path this PR changes.
- Thread-name resolution keeps the #93 guard: malformed/unnamed threads always carry the `short_id` fragment and never collapse to the parent group name (`internal/service/summary.go:68-112`).

## 数据流回溯（consumed data → upstream → does it really flow to the consumer）

- Chip `source_name` (list/detail/share) ← `displaySourceName(s, h.imDB)` ← stored `summary_source.source_name` written at create time by `task.go:447` / `agent_summary.go:386` / `internal/api/handler/bot_summary_create.go:476` / `internal/worker/scheduled_replace_helpers.go:77,241` / `internal/worker/source_backfill.go:113`. All five write paths resolve through the same resolver; all three user-visible read paths now go through `displaySourceName`; empty-name legacy rows genuinely reach the live-lookup branch (the `TrimSpace` guard only short-circuits on a real name).
- `h.imDB` in all three handlers ← `cmd/summary-api/main.go:64-68` (nil-tolerant) ← `router.go:166` and the pre-existing Task/Share wiring. Flows to every `displaySourceName`/resolver call site; nil is handled inside the resolver.
- Group/thread/DM names ← IM DB `group`/`thread`/`user` tables, with the same COLLATE-safe join convention used by `internal/pipeline/fetch.go:141`; error and miss paths both terminate in the deterministic placeholder, never an empty string.

## 盲点 checklist

- **C1 双路径 parity**: hit & cleared. Write side: all five `summary_source` write paths resolve identically (enumerated above). Read side: all three user-visible read paths (list/detail/share) resolve; the two stored-value consumers (LLM reference context, worker source map) are deliberately non-display and were audited as such.
- **C2 control-flow ordering / 嵌套复用**: cleared. `displaySourceName` is invoked per source row inside existing loops with no nested re-entry; the resolver is pure per call (no shared mutable state), so list renders cannot cross-contaminate.
- **C3 授权边界**: cleared — no auth/permission/tool-exposure changes; visibility rules (derived-row filtering, share visibility) are unchanged.
- **C4 授权生命周期/容器级联**: N/A — no container/member status logic touched.
- **C5 build 过 ≠ 运行期正确**: build/tests not executed here (environment discipline); runtime behavior established by data-flow tracing plus the PR's real-router tests covering nil-imDB, empty-name fallback, and snapshot preservation.
- **C6 治理/文档自洽**: N/A — no governance/policy/security docs touched.

## 跨轮 blocker 复检

N/A — first review round for this PR; no prior blockers to re-verify.

## 额外发现

- Nit (pre-existing, non-blocking): the LLM reference-context bullets at `internal/api/handler/agent_reference_context.go:338` still print the raw stored `s.SourceName`, so pre-PR agent rows render an empty name inside the LLM prompt. Not user-visible, file untouched by this PR (R2: pre-existing, base behavior identical); the builder lives on `AgentSummaryHandler`, which now holds `imDB`, so routing it through `displaySourceName` would be a cheap follow-up.

## Verdict: APPROVED

The fix is correct and complete for the bug it targets: agent-path summaries now store resolved source names, legacy empty rows display readable names instead of raw channel ids, snapshot semantics protect historical names, and nil-IM/miss/error paths all degrade to deterministic placeholders rather than 500s or empty chips. The two P2 items are durability/efficiency concerns — permanent placeholders after transient IM errors, and unbounded-but-paginated N+1 lookups on legacy rows — not correctness regressions: no working path becomes unavailable and no user-visible data becomes wrong beyond the deterministic placeholder the resolver already defines. Recommend approval with the two P2s tracked as follow-ups.

[Octo-Q] verdict: APPROVE — no P0/P1; source-name resolution is correct on every write/read path with snapshot semantics preserved; two non-blocking P2 follow-ups (error-vs-miss placeholder persistence at agent_summary.go:386, N+1 legacy lookups at task.go:184).

yujiawei:
Addendum to my review above — sharpening the first P2 after tracing the IM DB lifecycle. No change to the verdict (still approved); this only changes how urgent that follow-up looks.

I originally framed the placeholder-persistence issue as a narrow race: the IM DB has to be unreachable at the exact moment an agent summary is saved. Tracing the wiring, the window is considerably wider than that.

`cmd/summary-api/main.go:64-67`:

```go
imDB, err := db.New(cfg.IMMySQLDSN)
if err != nil {
	log.Printf("[main] connect IM DB (non-fatal): %v", err)
	imDB = nil
}
```

That non-fatal degrade is the right call for availability — it is exactly why name resolution can never 500 either endpoint. But `imDB` is assigned once, at boot, and nothing reassigns it afterwards (`grep` for `imDB =` across `cmd/` and `internal/` finds only this line and the worker's equivalent). So if the IM DB happens to be unreachable when the API process starts, `h.imDB` is nil for the entire process lifetime, not for a few seconds.

Combined with `agent_summary.go:386`, every agent summary saved between that boot and the next restart persists `来源-xxxxxxxx(群聊)` into `source_name`. Because `displaySourceName` heals only *empty* names — correct, snapshot semantics — none of those rows are repairable on read after IM recovers. A single unlucky restart ordering (IM DB slower to come up than summary-api, which is a routine deploy pattern) can therefore mint an unbounded number of permanently-unreadable chips, rather than the handful a transient blip would produce.

This does not change my assessment that it is non-blocking: the endpoint stays up, the placeholder is deterministic and type-labeled, and the pre-existing write paths behave the same way. It does raise the value of the fix I suggested, and adds a second case worth handling explicitly:

- distinguish *error* from *miss* in the resolver, and
- treat `imDB == nil` as "unresolved, retry later" rather than "resolved to a placeholder" on the **write** path specifically — a nil handle is not evidence that the channel has no name.

Storing `""` in both of those cases keeps the read-side fallback able to heal them once IM is reachable again, while a genuine lookup miss can still persist the placeholder so a permanently-unknown channel does not pay a lookup on every render.

Worth noting the tension this creates with the third test in the new file, which currently pins nil-imDB → stored placeholder as the desired behavior. If you take this suggestion, that assertion is the one that flips.

PR #193, opened 2026-08-17
feat: add document agent summary endpoint

wanyiwang06:
## Summary
- add Agent-only document summary endpoint at POST /api/v1/summaries/agent/document
- fetch ACL-checked document summary source from document service and wrap it in prompt fences
- persist document sources, document citations, personal result, and idempotency records
- expose document source metadata on summary detail

## Verification
- git diff --check
- Not run: go test / gofmt, local Go toolchain is unavailable in this workspace

PR #194, opened 2026-08-17
feat: add document agent summary endpoint

wanyiwang06:
## Summary
- add Agent-only document summary endpoint at POST /api/v1/summaries/agent/document
- fetch ACL-checked document summary source from document service and wrap content in prompt fences
- persist document source rows, document citations, personal result, and idempotency records

## Verification
- git diff --check
- Not run: go test / gofmt, local Go toolchain is unavailable in this workspace

### DIFF
--- a/internal/api/handler/agent_summary.go
+++ b/internal/api/handler/agent_summary.go
@@ -42,7 +42,8 @@ type AgentSummaryHandler struct {
 	llmModel     string
 	llmTimeout   int
 	llmMaxTokens int
-	store        agentHistoryStore
+	store          agentHistoryStore
+	documentClient documentSourceClient
 	// runnerFactory is an optional test-only hook for injecting a fake agent
 	// runner without going through the real LLM. When nil (production path),
 	// newRunner falls back to buildRunner with handler's LLM config.
--- a/internal/api/handler/agent_summary_test.go
+++ b/internal/api/handler/agent_summary_test.go
@@ -37,6 +37,7 @@ func setupAgentSummaryTestDB(t *testing.T) *gorm.DB {
 		&model.SummaryParticipant{},
 		&model.SummaryResult{},
 		&model.PersonalResult{},
+		&model.SummaryDocumentAgentIdempotency{},
 	); err != nil {
 		t.Fatalf("failed to migrate: %v", err)
 	}
--- a/internal/api/handler/document_agent_summary.go
+++ b/internal/api/handler/document_agent_summary.go
@@ -1,0 +1,460 @@
+package handler
+
+import (
+	"context"
+	"crypto/sha256"
+	"encoding/hex"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"log"
+	"net/http"
+	"net/url"
+	"os"
+	"sort"
+	"strings"
+	"time"
+	"unicode/utf8"
+
+	"github.com/Mininglamp-OSS/octo-smart-summary/internal/middleware"
+	"github.com/Mininglamp-OSS/octo-smart-summary/internal/model"
+	"github.com/Mininglamp-OSS/octo-smart-summary/internal/service"
+	"github.com/Mininglamp-OSS/octo-smart-summary/internal/timezone"
+	"github.com/gin-gonic/gin"
+	"gorm.io/gorm"
+	"gorm.io/gorm/clause"
+)
+
+const defaultDocumentSummaryRequirement = "总结这份文档"
+
+var errDocumentSourceNotConfigured = errors.New("document summary source API is not configured")
+
+type documentRefReq struct {
+	DocumentID string `json:"document_id"`
+	Version    string `json:"version,omitempty"`
+}
+
+type createDocumentAgentSummaryReq struct {
+	DocumentRefs   []documentRefReq `json:"document_refs"`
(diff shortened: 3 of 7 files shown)


wanyiwang06:
Closing because we will continue with the original combined backend PR #193.

Jerry-Xin (changes requested):
Re-review of the same change previously reviewed as PR #193 (both reviewers requested changes there). Byte-verification against #193's head shows **6 of the 7 changed files are byte-identical** to the previously reviewed commit; the only delta is that `internal/api/handler/task.go` now *removes* the `source_type` / `source_documents` projection from `GetSummary`. None of the previously flagged blocking defects were fixed, and the removal regresses document provenance instead of repairing it.

### 🔴 Blocking (carried over from #193, all independently re-verified on this head)

- 🔴 **Idempotency race → duplicate work + HTTP 500.** The binding lookup at `internal/api/handler/document_agent_summary.go:159` happens before the document fetches and the LLM call, while the binding insert only occurs at `internal/api/handler/document_agent_summary.go:432-438` inside the transaction. Two concurrent same-key requests both miss the lookup, both pay for fetch + generation, and the loser trips `uk_doc_agent_idempotency` and surfaces as `500` at `internal/api/handler/document_agent_summary.go:213` — which also echoes the raw driver error to the caller. This file is byte-identical to #193. Use the established conflict-tolerant pattern from `internal/api/handler/bot_summary_create.go` (`OnConflict` DoNothing + RowsAffected check + re-fetch) and claim the key before doing expensive work; add a concurrent same-key test — the sequential replay test at `internal/api/handler/document_agent_summary_test.go:109` cannot catch this. (Originally flagged in #193 by both reviewers.)

- 🔴 **Multi-version document refs still collapse silently.** `normalizeDocumentRefs` (`internal/api/handler/document_agent_summary.go:224-247`) intentionally keeps `(doc, v1)` and `(doc, v2)` as separate refs, and citations embed the per-chunk version, but persistence writes one row per document under `uk_summary_source_task_type_id (task_id, source_type, source_id)` with `OnConflict{DoNothing}` (`internal/api/handler/document_agent_summary.go:381`) — no version column — so the second version's source row is dropped without error or log. The previous partial exposure of this (`source_documents`) was deleted in this revision rather than fixed, so a summary generated from v1+v2 now reports a single unversioned document and the provenance mismatch is no longer even visible. Decide the contract: persist and expose the resolved version, or reject multi-version refs with 400. Silent collapse is not acceptable. (Originally flagged in #193 by both reviewers.)

- 🔴 **Regenerate on a document task destroys the summary with no recovery path.** `Regenerate` (`internal/api/handler/task.go:1415`) still has no guard for `SourceDocument` tasks: the creator passes every precondition, the result content and citations are wiped, and the worker is re-dispatched — but the document path is not re-fetchable through the worker's source pipeline, so the original deliverable is permanently lost. Reject regenerate/personal-regenerate for tasks with document source rows, or route them back through the document path. (Originally flagged in #193 by yujiawei, P1-2.)

### 🟡 Non-blocking (also unchanged from #193 unless noted)

- No size/length validation before paid work: `idempotency_key` is only checked non-empty (`:139`) against a `VARCHAR(128)` column (and stored untrimmed), `document_id` / title flow unchecked into `VARCHAR(64)` / `VARCHAR(200)` columns after the LLM call has already been paid for. Validate lengths up front; drop the raw `err.Error()` from the `:213` response.
- The document-service response body is decoded with no limit (`:112-113`), and chunk content is copied verbatim into both the prompt and `citations_json` with no per-chunk or aggregate cap — unlike the token-budgeting and 200-rune citation truncation the conversation pipeline already applies.
- Replay response (`:169-170`) omits `task_no` / `created_at` and hardcodes `status`, diverging from the fresh-response shape.
- Cross-host redirects: the bare `http.Client` (`:75`) follows up to 10 redirects; the custom `Token` / `X-User-Id` / `X-Space-Id` headers are not stripped across hosts. Refuse redirects via `CheckRedirect`.
- Fence sanitizer still replaces only two exact literals (`:347-351`) — full-width/zero-width variants and a forged section boundary pass through.
- `gofmt -l` still reports `internal/api/handler/agent_summary.go` and `internal/model/model.go` (re-verified locally on this head).
- Test coverage unchanged: happy path, sequential replay, conflict only — no concurrency, validation, oversized-input, or failure-path tests. Note the SQLite test DB enforces none of the production `VARCHAR` widths.
- `DOCUMENT_SUMMARY_SOURCE_API_URL` / `DOCUMENT_SOURCE_API_URL` (`:68-71`) are still not registered in `CONFIGURATION.md`; until set, the endpoint returns 502 in every deployment with no documented knob.

### ℹ️ Process

- The `check-sprint` CI check fails because the PR description has no `Closes #<issue>` reference — link the issue (and assign it to the current sprint) so the check can pass.

### What flips this to APPROVE

1. Claim the idempotency key before fetch/generation; on conflict re-fetch and return the existing task (bot-summary pattern) — never a 500.
2. Persist and expose document version, or reject multi-version refs explicitly.
3. Guard regenerate/personal-regenerate against document-source tasks.
4. Move all length/size validation ahead of the document fetch and LLM call; cap response bodies, chunks, and the aggregate prompt; stop echoing raw DB errors.

PR #229, opened 2026-08-26
feat(summary): let a pending submission raise attention (red dot + count)

### LINKED ISSUE #230: Multi-person summary: pending submission has no red dot, and the sidebar badge disagrees with the cards
## Problem

Multi-person summaries have three moments that need the user's attention, but only two of them ever surface:

| Scenario | Card red dot | Sidebar badge |
|---|---|---|
| Invited as a participant | ✅ | ✅ (#1359) |
| **My personal summary is generated — my turn to `/submit`** | ❌ | ❌ |
| Summary completed (unread result) | ✅ | ❌ |

The middle row is the worst case. `ListSummaries` has computed `has_pending_submission` since #167:

```sql
CASE WHEN pr.worker_status = ? AND pr.submitted_at IS NULL
     AND (SELECT COUNT(*) FROM summary_participant wait_sp WHERE wait_sp.task_id = sub.id) > 1
     THEN 1 ELSE 0 END AS has_pending_submission
```

…but it never reaches `needs_attention` (task.go) or `attention_count`. The frontend only uses it to relabel a card's status badge as "waiting". So a participant whose personal summary finished generating gets **no indicator anywhere** that the whole team is blocked on their submit. They have to remember to go look.

The third row is a leftover from the badge's history: #852 fed the nav badge from `attention_count` (unread + invitations), #1192/#1213 removed the badge and its 5s poll entirely on a product decision, and #1359/#1366 brought the badge back but only restored the invitation half. So the sidebar number and the card dots have been counting different sets ever since — a user can see "1" on the rail, open Smart Summary, and find three dotted cards.

## Proposal

Make `needs_attention` and `attention_count` the single source of truth for all three scenarios, and have the sidebar badge read the same value the cards do.

Backend (`octo-smart-summary`):

- `needs_attention` ORs `has_pending_submission`
- `attention_count` gains the same term, plus a `pending_submission_count` breakdown
- list ordering: invitation > pending submission > unread
- `MarkSummaryRead` returns `has_pending_submission` and includes it in its recompute

Frontend (`octo-web`):

- sidebar badge reads `attention_count` instead of `pending_invitation_count`
- `handleSummaryRead_` fallback must not drop the submission signal on read
- refresh the badge after submit / invite response from the detail page

No schema change, no new endpoint, and no reinstated polling — #1213 removed the 5s poll for good reasons (~720 req/user/hour) and the existing event-driven refreshes are enough.

## Decided semantics

**Reading is not submitting** (owner, 2026-08-26). Opening the detail page records the read cursor, but the participant still owes the team an explicit `/submit`, so the dot survives `MarkSummaryRead` and clears only when `submitted_at` is set. This is intentionally different from "red dot = unread": what is being nudged is the *action*, not the *content*.

## Out of scope

- No IM/DM notification for invitations. `internal/notify` keeps only its `completed` / `failed` kinds; invitations stay red-dot-only.
- The scheduled pre-run reminder (`summary_task.reminder_sent_at` + `summary_schedule.confirm_lead_minutes`, both added in `20260611-01-team-schedule-confirm.sql` but never read or written by any Go code) is a separate, unimplemented feature and is not addressed here.

wanyiwang06:
## Summary

`ListSummaries` has computed `has_pending_submission` since #167, but the flag never reached `needs_attention` or `attention_count`. The frontend only used it to relabel a card's status badge as "waiting", so a participant whose personal summary had finished generating had **no indicator anywhere** that the team was blocked on their `/submit`.

This wires the existing signal into the two projections that actually drive UI.

## Related Issue

Closes #230

Follow-up to #159 (per-user unread + attention model) and #167 (`has_pending_submission`). Pairs with octo-web `feat/summary-attention-badge`, which switches the sidebar badge to `attention_count`.

## Changes

- `needs_attention` (per-card red dot) now ORs `has_pending_submission`.
- `attention_count` (sidebar badge) gains the same term, plus a new `pending_submission_count` breakdown alongside `unread_count` / `pending_invitation_count`.
- List ordering places pending submissions between invitations and unread content.
- `MarkSummaryRead` returns `has_pending_submission` and includes it in its own `needs_attention` recompute.

### Terminal tasks are excluded (added in review round 2)

`pendingSubmitStatusGuard` excludes `Failed` and `Cancelled` at all four sites that carry the predicate (list projection, attention-count query, `MarkSummaryRead`, and the pre-existing `status=waiting` filter). Without it, the stuck-task reaper or a manual cancel leaves a participant holding a red dot — and a permanently non-zero `attention_count` — for a task nobody can revive, since `Regenerate` and `Delete` are creator-gated. `Completed` stays eligible: a personal-only regenerate intentionally waits for an explicit submit.

### Reading is deliberately not submitting

Owner decision 2026-08-26: opening the detail page records the read cursor, but the participant still owes the team an explicit `/submit`. The dot therefore **survives `MarkSummaryRead`** and clears only when `submitted_at` is set. `MarkSummaryRead` returning the flag is what lets a client recompute `needs_attention` without dropping it.

### Placeholder ordering (the risky part)

`attentionCountSQL` is hand-assembled with positional placeholders. The new `has_submit` term sits between the invitation term and `countInner`, so `model.PersonalStatusCompleted` must be appended to `countArgs` at exactly that position. Both dialect branches (the MySQL-only `JSON_TABLE` roster predicates in `schedulePendingInvitationExpr` / `scheduleVisibilityExpr`) are unaffected — they still contribute their `userID` conditionally, before and after the new argument respectively.

A comment now documents the full placeholder order so a future edit does not shift the arguments silently:

```
//   1. has_invite      -> ParticipantPending [+ userID when schedulePendingExpr is live (MySQL only)]
//   2. has_submit      -> PersonalStatusCompleted
//   3. countInner      -> spaceID, userID, userID [+ userID when scheduleVisibilityExpr is live]
//   4. attentionJoins  -> userID x4
```

The same insertion applies to `MarkSummaryRead`'s `stateArgs`.

## Testing

New `internal/api/handler/attention_pending_submit_test.go`:

- a generated-but-unsubmitted personal result raises the dot and both counts, without leaking into the unread/invitation breakdown
- submitting clears the dot and the counts
- single-person tasks never pend submission (nobody to submit to — the existing `COUNT(*) > 1` gate)
- `MarkSummaryRead` keeps the dot, and the list projection agrees with the read response
- a reaper-failed or creator-cancelled task drops the dot in both producers
- a `Completed` task with a regenerated personal result still holds it
- the `status=waiting` filter agrees with the badge

`attention_pending_submit_mysql_test.go` (opt-in, `SUMMARY_MYSQL_TEST_DSN`) covers the MySQL placeholder ordering for the attention-count query and `MarkSummaryRead` — the two statements the existing MySQL test does not touch. It seeds a live `confirm_policy=1` roster so `schedulePendingInvitationExpr` actually binds its placeholder, and asserts count values rather than booleans (a misaligned argument yields 0 silently instead of erroring).

Commands:

```
CGO_LDFLAGS="-L/usr/local/lib" go test ./internal/api/handler/ -count=1   # full package, 29.9s, pass
CGO_ENABLED=0 go vet ./...                                               # clean
```

- [x] Unit tests added/updated
- [x] Manually verified (SQLite handler tests; the MySQL-only schedule predicate path is unchanged and still covered by the opt-in `SUMMARY_MYSQL_TEST_DSN` test)

## Rollout note

List ordering gains a `has_pending_submission DESC` term between the invitation and unread terms. Any client asserting on list order will see a shift when this deploys, independent of the frontend change.

## Compatibility

- No schema change. `summary_user_read`, `summary_personal_result.submitted_at` and `worker_status` all already exist.
- `pending_submission_count` is additive; existing clients ignore it.
- Deploying this **before** the octo-web change is harmless: the card dot appears immediately, the sidebar badge simply keeps its old narrower meaning until the frontend lands.

## Checklist

- [x] PR description is in English
- [x] Added tests for my changes
- [x] Followed commit message conventions (Conventional Commits)

### DIFF
--- a/internal/api/handler/attention_pending_submit_mysql_test.go
+++ b/internal/api/handler/attention_pending_submit_mysql_test.go
@@ -1,0 +1,149 @@
+//go:build cgo
+
+package handler
+
+import (
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"os"
+	"testing"
+	"time"
+
+	"github.com/Mininglamp-OSS/octo-smart-summary/internal/model"
+	"github.com/Mininglamp-OSS/octo-smart-summary/internal/timezone"
+	"gorm.io/driver/mysql"
+	"gorm.io/gorm"
+)
+
+// MySQL placeholder-ordering coverage for the pending-submission argument.
+//
+// The SQLite harness cannot exercise schedulePendingInvitationExpr: it needs
+// JSON_TABLE, so on SQLite the predicate collapses to the constant "0" and
+// contributes no placeholder. On MySQL it contributes one `?` that sits
+// immediately BEFORE the new PersonalStatusCompleted argument in every
+// hand-assembled statement. A mutation that swapped those two would pass the
+// whole SQLite suite and only break in production.
+//
+// schedule_attention_mysql_test.go already pins the list query's ordering. This
+// covers the two statements it does not touch: the attention-count query (whose
+// new has_submit term is a third placeholder position) and MarkSummaryRead's
+// stateSQL (which had no MySQL coverage at all).
+//
+// Opt-in like its sibling: point SUMMARY_MYSQL_TEST_DSN at an isolated,
+// migrated database.
+func TestPendingSubmission_MySQL_PlaceholderOrderingAndMarkRead(t *testing.T) {
+	dsn := os.Getenv("SUMMARY_MYSQL_TEST_DSN")
+	if dsn == "" {
+		t.Skip("SUMMARY_MYSQL_TEST_DSN is not set")
+	}
+	db, err := gorm.Open(mysql.Open(dsn), &gorm.Config{DisableForeignKeyConstraintWhenMigrating: true})
+	if err != nil {
+		t.Fatalf("open mysql: %v", err)
+	}
+	for _, table := range []string{
+		"summary_task", "summary_source", "summary_participant", "summary_result",
+		"summary_personal_result", "summary_personal_result_version", "summary_user_read", "summary_schedule",
+	} {
+		if !db.Migrator().HasTable(table) {
+			t.Fatalf("mysql integration database is not migrated: missing %s", table)
+		}
+	}
+
+	tx := db.Begin()
+	if tx.Error != nil {
+		t.Fatalf("begin transaction: %v", tx.Error)
+	}
+	defer tx.Rollback()
+
+	const spaceID = "pending-submit-mysql-space"
+	const me = "submit-owner"
+	now := timezone.Now()
+
+	// A schedule whose roster still has an unconfirmed member makes
+	// schedulePendingInvitationExpr live (confirm_policy=1), so its placeholder
+	// really is bound during this test rather than short-circuiting.
+	schedule := model.SummarySchedule{
+		SpaceID: spaceID, CreatorID: me, Title: "scheduled", SummaryMode: model.ModeByPerson,
+		CronExpr: "0 0 * * *", TimeRangeType: 1,
+		ParticipantConfig: model.JSON(`{"participants":[{"user_id":"other-invitee","confirmed":false}],"confirm_gate_passed":false}`),
+		ConfirmPolicy:     model.SchedConfirmRequire, IsActive: 1,
+	}
+	if err := tx.Create(&schedule).Error; err != nil {
+		t.Fatalf("create schedule: %v", err)
+	}
+
+	task := model.SummaryTask{
+		TaskNo: "MYSQL-PENDING-SUBMIT", SpaceID: spaceID, CreatorID: me,
+		SummaryMode: model.ModeByPerson, Status: model.StatusProcessing,
+		TriggerType: model.TriggerScheduled, ScheduleID: &schedule.ID,
+		TimeRangeStart: time.Now().Add(-time.Hour), TimeRangeEnd: time.Now(),
+	}
+	if err := tx.Create(&task).Error; err != nil {
+		t.Fatalf("create task: %v", err)
+	}
+	for _, uid := range []string{me, "teammate"} {
+		if err := tx.Create(&model.SummaryParticipant{
+			TaskID: task.ID, UserID: uid, UserName: uid, Status: model.ParticipantAccepted,
+		}).Error; err != nil {
+			t.Fatalf("create participant %s: %v", uid, err)
+		}
+	}
+	version := model.PersonalResultVersion{
+		TaskID: task.ID, UserID: me, Content: "personal", Version: 1,
+		GeneratedAt: now, CreatedAt: now, UpdatedAt: now,
+	}
+	if err := tx.Create(&version).Error; err != nil {
+		t.Fatalf("create personal version: %v", err)
+	}
+	if err := tx.Create(&model.PersonalResult{
+		TaskID: task.ID, UserID: me, Content: "personal draft",
+		WorkerStatus: model.PersonalStatusCompleted, SubmittedAt: nil,
+		CurrentVersionID: &version.ID, CreatedAt: now, UpdatedAt: now,
+	}).Error; err != nil {
+		t.Fatalf("create personal result: %v", err)
+	}
+
+	// --- list projection + attention-count query ---
+	r := setupListRouter(NewTaskHandler(tx, nil, ""))
+	w := doRequestWithSpace(r, http.MethodGet, "/api/v1/summaries", me, spaceID)
+	if w.Code != http.StatusOK {
+		t.Fatalf("expected 200, got %d: %s", w.Code, w.Body.String())
+	}
+	resp := parseAttentionList(t, w)
+	if len(resp.Data.Items) != 1 || resp.Data.Items[0].TaskID != task.ID {
+		t.Fatalf("expected the seeded task: %+v", resp.Data)
+	}
+	if !resp.Data.Items[0].HasPendingSubmission || !resp.Data.Items[0].NeedsAttention {
+		t.Fatalf("MySQL list projection lost the pending submission: %+v", resp.Data.Items[0])
+	}
+	// A wrong argument position would not necessarily error — it would silently
+	// compare worker_status against a uid, yielding 0. Assert the value.
+	if resp.Data.PendingSubmissionCount != 1 || resp.Data.AttentionCount != 1 {
+		t.Fatalf("MySQL attention-count placeholders are misaligned: attention=%d submission=%d",
+			resp.Data.AttentionCount, resp.Data.PendingSubmissionCount)
+	}
+
+	// --- MarkSummaryRead stateSQL ---
+	rr := setupReadRouter(NewTaskHandler(tx, nil, ""))
+	rw := markReadRequestWithSpace(rr, task.ID, me, spaceID, fmt.Sprintf(`{"personal_version_id":%d}`, version.ID))
+	if rw.Code != http.StatusOK {
+		t.Fatalf("expected 200 from mark-read, got %d: %s", rw.Code, rw.Body.String())
+	}
+	var readResp struct {
+		Data struct {
+			IsUnread             bool `json:"is_unread"`
+			HasPendingSubmission bool `json:"has_pending_submission"`
+			NeedsAttention       bool `json:"needs_attention"`
+		} `json:"data"`
+	}
+	if err := json.Unmarshal(rw.Body.Bytes(), &readResp); err != nil {
+		t.Fatalf("unmarshal mark-read response: %v; body=%s", err, rw.Body.String())
+	}
+	if readResp.Data.IsUnread {
+		t.Fatalf("the rendered version was recorded: %+v", readResp.Data)
+	}
+	if !readResp.Data.HasPendingSubmission || !readResp.Data.NeedsAttention {
+		t.Fatalf("MySQL stateSQL lost the pending submission on read: %+v", readResp.Data)
+	}
+}
--- a/internal/api/handler/attention_pending_submit_test.go
+++ b/internal/api/handler/attention_pending_submit_test.go
@@ -176,6 +176,122 @@ func TestListSummaries_SinglePersonTaskNeverPendsSubmission(t *testing.T) {
 	}
 }
 
+// A dead task must not keep nagging. The stuck-task reaper flips Processing ->
+// Failed once retries are exhausted (worker/scheduler.go) and CancelSummary
+// writes Cancelled; neither touches summary_personal_result, so without the
+// status guard the member who DID generate their summary keeps a red dot and a
+// non-zero space-level attention_count for a task nobody can revive
+// (Regenerate and Delete are creator-gated). A badge that cannot return to zero
+// is worse than no badge.
+func TestListSummaries_TerminalTaskDropsPendingSubmission(t *testing.T) {
+	for _, tc := range []struct {
+		name   string
+		taskNo string
+		status int
+	}{
+		{"reaper marked it failed", "SUBMIT-FAILED", model.StatusFailed},
+		{"creator cancelled it", "SUBMIT-CANCELLED", model.StatusCancelled},
+	} {
+		t.Run(tc.name, func(t *testing.T) {
+			db, imDB := setupListTestDBs(t)
+			taskID := seedPendingSubmitTask(t, db, tc.taskNo, "participant1", []string{"participant2"})
+			if err := db.Model(&model.SummaryTask{}).Where("id = ?", taskID).
+				Update("status", tc.status).Error; err != nil {
+				t.Fatalf("set terminal status: %v", err)
+			}
+
+			r := setupListRouter(NewTaskHandler(db, imDB, ""))
+			w := doRequest(r, http.MethodGet, "/api/v1/summaries", "participant1")
+			resp := parseAttentionList(t, w)
+			if len(resp.Data.Items) != 1 {
+				t.Fatalf("expected the seeded task, got %+v", resp.Data)
+			}
+			if resp.Data.Items[0].HasPendingSubmission || resp.Data.Items[0].NeedsAttention {
+				t.Fatalf("a terminal task must not pend submission: %+v", resp.Data.Items[0])
+			}
+			if resp.Data.AttentionCount != 0 || resp.Data.PendingSubmissionCount != 0 {
+				t.Fatalf("terminal task must not hold the badge above zero: %+v", resp.Data)
+			}
+
+			// MarkSummaryRead must agree — it is the other needs_attention producer.
+			rr := setupReadRouter(NewTaskHandler(db, imDB, ""))
+			now := timezone.Now()
+			version := model.PersonalResultVersion{
+				TaskID: taskID, UserID: "participant1", Content: "personal", Version: 1,
+				GeneratedAt: now, CreatedAt: now, UpdatedAt: now,
+			}
+			if err := db.Create(&version).Error; err != nil {
+				t.Fatalf("create personal version: %v", err)
+			}
+			if err := db.Model(&model.PersonalResult{}).
+				Where("task_id = ? AND user_id = ?", taskID, "participant1").
+				Update("current_version_id", version.ID).Error; err != nil {
+				t.Fatalf("attach current version: %v", err)
+			}
+			rw := markReadRequest(rr, taskID, "participant1", fmt.Sprintf(`{"personal_version_id":%d}`, version.ID))
+			var readResp struct {
+				Data struct {
+					HasPendingSubmission bool `json:"has_pending_submission"`
+					NeedsAttention       bool `json:"needs_attention"`
+				} `json:"data"`
+			}
+			if err := json.Unmarshal(rw.Body.Bytes(), &readResp); err != nil {
+				t.Fatalf("unmarshal mark-read response: %v; body=%s", err, rw.Body.String())
+			}
+			if readResp.Data.HasPendingSubmission || readResp.Data.NeedsAttention {
+				t.Fatalf("mark-read must agree that a terminal task pends nothing: %+v", readResp.Data)
+			}
+		})
+	}
+}
+
+// Completed is deliberately NOT guarded. A personal-only regenerate on a
+// finished task leaves submitted_at NULL on purpose and waits for an explicit
+// submit (worker/personal_processor.go), so the dot is correct there. Pinning
+// this stops a future "just exclude every terminal status" simplification from
+// silently dropping the signal on the one terminal status that still needs it.
+func TestListSummaries_CompletedTaskStillPendsSubmission(t *testing.T) {
+	db, imDB := setupListTestDBs(t)
+	taskID := seedPendingSubmitTask(t, db, "SUBMIT-REGEN", "participant1", []string{"participant2"})
+	if err := db.Model(&model.SummaryTask{}).Where("id = ?", taskID).
+		Update("status", model.StatusCompleted).Error; err != nil {
+		t.Fatalf("set completed status: %v", err)
+	}
+
+	r := setupListRouter(NewTaskHandler(db, imDB, ""))
+	w := doRequest(r, http.MethodGet, "/api/v1/summaries", "participant1")
+	resp := parseAttentionList(t, w)
+	if len(resp.Data.Items) != 1 {
+		t.Fatalf("expected the seeded task, got %+v", resp.Data)
+	}
+	if !resp.Data.Items[0].HasPendingSubmission || !resp.Data.Items[0].NeedsAttention {
+		t.Fatalf("a regenerated personal result on a completed task still owes a submit: %+v", resp.Data.Items[0])
+	}
+	if resp.Data.AttentionCount != 1 || resp.Data.PendingSubmissionCount != 1 {
+		t.Fatalf("completed task must still hold the counts: %+v", resp.Data)
+	}
+}
+
+// The status=waiting list filter carries the same predicate as the badge. If
+// the two disagree, a card can be counted by the badge but hidden by the
+// filter the user clicks to find it (or vice versa).
+func TestListSummaries_WaitingFilterAgreesWithTerminalGuard(t *testing.T) {
+	db, imDB := setupListTestDBs(t)
+	liveID := seedPendingSubmitTask(t, db, "SUBMIT-WAIT-LIVE", "participant1", []string{"participant2"})
+	deadID := seedPendingSubmitTask(t, db, "SUBMIT-WAIT-DEAD", "participant1", []string{"participant2"})
+	if err := db.Model(&model.SummaryTask{}).Where("id = ?", deadID).
+		Update("status", model.StatusFailed).Error; err != nil {
+		t.Fatalf("set terminal status: %v", err)
+	}
+
+	r := setupListRouter(NewTaskHandler(db, imDB, ""))
+	w := doRequest(r, http.MethodGet, fmt.Sprintf("/api/v1/summaries?status=%d", model.StatusPending), "participant1")
+	resp := parseAttentionList(t, w)
+	if len(resp.Data.Items) != 1 || resp.Data.Items[0].TaskID != liveID {
+		t.Fatalf("waiting filter must keep the live task and drop the terminal one: %+v", resp.Data)
+	}
+}
+
 // Owner decision 2026-08-26: reading is not submitting. Opening the detail page
 // records the read cursor, but the participant still owes the team a /submit,
 // so the dot must survive MarkSummaryRead.
--- a/internal/api/handler/attention_pending_submit_test.go
+++ b/internal/api/handler/attention_pending_submit_test.go
@@ -1,0 +1,236 @@
+//go:build cgo
+
+package handler
+
+import (
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	"github.com/Mininglamp-OSS/octo-smart-summary/internal/model"
+	"github.com/Mininglamp-OSS/octo-smart-summary/internal/timezone"
+	"gorm.io/gorm"
+)
+
+// Pending-submission attention (multi-person "your turn to submit" red dot).
+//
+// has_pending_submission was already computed by ListSummaries but never fed
+// needs_attention / attention_count, so the dot never rendered. These tests pin
+// the three rules that make it a usable signal:
+//
+//  1. it only fires for MULTI-person tasks (single-person owners have nobody to
+//     submit to),
+//  2. it clears on submitted_at, not on read — MarkSummaryRead must leave it
+//     standing (owner decision 2026-08-26),
+//  3. it participates in the space-level attention_count that drives the
+//     sidebar badge, with its own pending_submission_count breakdown.
+
+type attentionListResponse struct {
+	Code int `json:"code"`
+	Data struct {
+		Total                  int `json:"total"`
+		AttentionCount         int `json:"attention_count"`
+		UnreadCount            int `json:"unread_count"`
+		PendingInvitationCount int `json:"pending_invitation_count"`
+		PendingSubmissionCount int `json:"pending_submission_count"`
+		Items                  []struct {
+			TaskID               int64 `json:"task_id"`
+			IsUnread             bool  `json:"is_unread"`
+			HasPendingInvitation bool  `json:"has_pending_invitation"`
+			HasPendingSubmission bool  `json:"has_pending_submission"`
+			NeedsAttention       bool  `json:"needs_attention"`
+		} `json:"items"`
+	} `json:"data"`
+}
+
+func parseAttentionList(t *testing.T, w *httptest.ResponseRecorder) attentionListResponse {
+	t.Helper()
+	var resp attentionListResponse
+	if err := json.Unmarshal(w.Body.Bytes(), &resp); err != nil {
+		t.Fatalf("unmarshal list response: %v; body=%s", err, w.Body.String())
+	}
+	return resp
+}
+
+// seedPendingSubmitTask builds a task whose personal result for userID has
+// finished generating but has not been submitted. extraParticipants controls
+// whether the task is multi-person (the gate for the submission signal).
+func seedPendingSubmitTask(t *testing.T, db *gorm.DB, taskNo, userID string, extraParticipants []string) int64 {
+	t.Helper()
+	now := timezone.Now()
+
+	task := model.SummaryTask{
+		TaskNo:      taskNo,
+		SpaceID:     "space1",
+		CreatorID:   "creator1",
+		SummaryMode: model.ModeByPerson,
+		Status:      model.StatusProcessing,
+	}
+	if err := db.Create(&task).Error; err != nil {
+		t.Fatalf("create task: %v", err)
+	}
+
+	if err := db.Create(&model.SummaryParticipant{
+		TaskID: task.ID, UserID: userID, UserName: userID, Status: model.ParticipantAccepted,
+	}).Error; err != nil {
+		t.Fatalf("create participant: %v", err)
+	}
+	for _, uid := range extraParticipants {
+		if err := db.Create(&model.SummaryParticipant{
+			TaskID: task.ID, UserID: uid, UserName: uid, Status: model.ParticipantAccepted,
+		}).Error; err != nil {
+			t.Fatalf("create participant %s: %v", uid, err)
+		}
+	}
+
+	// Personal summary generated (worker_status=Completed) but not submitted.
+	// No current_version_id / current_result_id is set, so is_unread stays false
+	// and the assertions below isolate the submission signal.
+	if err := db.Create(&model.PersonalResult{
+		TaskID: task.ID, UserID: userID, Content: "personal draft",
+		WorkerStatus: model.PersonalStatusCompleted, SubmittedAt: nil,
+		CreatedAt: now, UpdatedAt: now,
+	}).Error; err != nil {
+		t.Fatalf("create personal result: %v", err)
+	}
+
+	return task.ID
+}
+
+func TestListSummaries_PendingSubmissionDrivesAttention(t *testing.T) {
+	db, imDB := setupListTestDBs(t)
+	taskID := seedPendingSubmitTask(t, db, "SUBMIT-MULTI", "participant1", []string{"participant2"})
+
+	r := setupListRouter(NewTaskHandler(db, imDB, ""))
+	w := doRequest(r, http.MethodGet, "/api/v1/summaries", "participant1")
+	if w.Code != http.StatusOK {
+		t.Fatalf("expected 200, got %d: %s", w.Code, w.Body.String())
+	}
+	resp := parseAttentionList(t, w)
+	if len(resp.Data.Items) != 1 || resp.Data.Items[0].TaskID != taskID {
+		t.Fatalf("expected the seeded task in the list, got %+v", resp.Data)
+	}
+	item := resp.Data.Items[0]
+	if !item.HasPendingSubmission {
+		t.Fatalf("generated-but-unsubmitted personal result must flag pending submission: %+v", item)
+	}
+	if item.IsUnread || item.HasPendingInvitation {
+		t.Fatalf("test isolates the submission signal; unread/invitation must be off: %+v", item)
+	}
+	if !item.NeedsAttention {
+		t.Fatalf("pending submission must raise needs_attention (the card red dot): %+v", item)
+	}
+	if resp.Data.AttentionCount != 1 || resp.Data.PendingSubmissionCount != 1 {
+		t.Fatalf("pending submission must feed the sidebar counts: attention=%d submission=%d",
+			resp.Data.AttentionCount, resp.Data.PendingSubmissionCount)
+	}
+	if resp.Data.UnreadCount != 0 || resp.Data.PendingInvitationCount != 0 {
+		t.Fatalf("submission must not leak into the unread/invitation breakdown: %+v", resp.Data)
+	}
+}
+
+func TestListSummaries_SubmittedPersonalResultClearsAttention(t *testing.T) {
+	db, imDB := setupListTestDBs(t)
+	taskID := seedPendingSubmitTask(t, db, "SUBMIT-DONE", "participant1", []string{"participant2"})
+	now := timezone.Now()
+	if err := db.Model(&model.PersonalResult{}).
+		Where("task_id = ? AND user_id = ?", taskID, "participant1").
+		Updates(map[string]interface{}{"submitted_at": now, "submit_source": model.SubmitSourceManual}).Error; err != nil {
+		t.Fatalf("mark submitted: %v", err)
+	}
+
+	r := setupListRouter(NewTaskHandler(db, imDB, ""))
+	w := doRequest(r, http.MethodGet, "/api/v1/summaries", "participant1")
+	resp := parseAttentionList(t, w)
+	if len(resp.Data.Items) != 1 {
+		t.Fatalf("expected the seeded task, got %+v", resp.Data)
+	}
+	if resp.Data.Items[0].HasPendingSubmission || resp.Data.Items[0].NeedsAttention {
+		t.Fatalf("submitting must clear the dot: %+v", resp.Data.Items[0])
+	}
+	if resp.Data.AttentionCount != 0 || resp.Data.PendingSubmissionCount != 0 {
+		t.Fatalf("counts must drop after submit: attention=%d submission=%d",
+			resp.Data.AttentionCount, resp.Data.PendingSubmissionCount)
+	}
+}
+
+func TestListSummaries_SinglePersonTaskNeverPendsSubmission(t *testing.T) {
+	db, imDB := setupListTestDBs(t)
+	// Sole participant: there is no team to submit into, so an unsubmitted
+	// personal result is just the owner's own summary.
+	seedPendingSubmitTask(t, db, "SUBMIT-SOLO", "participant1", nil)
+
+	r := setupListRouter(NewTaskHandler(db, imDB, ""))
+	w := doRequest(r, http.MethodGet, "/api/v1/summaries", "participant1")
+	resp := parseAttentionList(t, w)
+	if len(resp.Data.Items) != 1 {
+		t.Fatalf("expected the seeded task, got %+v", resp.Data)
+	}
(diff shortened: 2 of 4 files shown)


yujiawei (changes requested):
# Code Review — PR #229 (octo-smart-summary)

Reviewed at head `4972954`. I read the full `ListSummaries` / `MarkSummaryRead` bodies (not just the diff), traced every positional placeholder in the three hand-assembled statements against both dialect branches, ran the package tests, and mutation-tested the new test file.

---

## 1. Scope check — ✅

The diff does exactly what the description says, no more:

- `needs_attention` ORs `has_pending_submission` (`internal/api/handler/task.go:918`, `:1056`)
- `attention_count` gains the same term plus a `pending_submission_count` breakdown (`task.go:938`, `:960`)
- Ordering slots the new signal between invitation and unread (`task.go:692`)
- `MarkSummaryRead` returns the flag and keeps it in its own recompute (`task.go:1032`, `:1055`)

No missing items, no unrequested additions, no schema change. The four new tests map 1:1 onto the four claims in the description.

## 2. Code quality — Changes requested

### P1 — the pending-submission signal never clears on a Failed or Cancelled task

`task.go:682` (card), `task.go:938` (space count), `task.go:1032` (mark-read)

The predicate is `worker_status = Completed AND submitted_at IS NULL AND participants > 1`. Nothing in it excludes a task that has already reached a terminal state, and two ordinary paths get a task there while a participant's personal result is still `Completed`/unsubmitted:

1. **The stuck-task reaper.** `internal/worker/scheduler.go:417-433` flips a `Processing` task to `StatusFailed` once `retry_count >= maxRetry-1`. That is precisely the outcome of the situation this feature exists to prevent: `metaCompletionReady` (`internal/worker/meta_processor.go:375-401`) blocks aggregation until *every* accepted participant is terminal, so one member who never submits stalls the task until the reaper fails it.
2. **Manual cancel.** `CancelSummary` (`task.go:2185-2191`) accepts `Pending/WaitingConfirm/Processing` and writes only `status` + `error_message`.

Neither path touches `summary_personal_result`, so `worker_status` stays `Completed` and `submitted_at` stays `NULL`.

Failure scenario: a two-person manual task stalls, the reaper marks it `Failed`. The participant who *did* generate their summary now has `needs_attention = true` on that card and a permanently non-zero `attention_count` / `pending_submission_count` in the sidebar — telling them the team is waiting on a submission for a task that is dead.

There is no recovery for a non-creator: `Regenerate` is creator-gated (`task.go:1462-1465`), and so is delete. `POST /submit` would still clear it — `personal.go:495-500` has no task-status guard — but the UI is unlikely to expose a submit affordance on a terminal task. If the paired `feat/summary-attention-badge` branch does keep submit reachable there, say so and I'll drop this; otherwise it needs a server-side gate, because a global badge that can never return to zero destroys the credibility of the whole signal.

Suggested fix — exclude **only** `Failed` and `Cancelled`, deliberately **not** `Completed`:

```go
// status guard, not a placeholder: see the ordering comment below
pendingSubmitStatusGuard := fmt.Sprintf(" AND sub.status NOT IN (%d,%d)",
    model.StatusFailed, model.StatusCancelled)
```

`Completed` must stay eligible: a personal-only regenerate on a finished task intentionally leaves `submitted_at = NULL` and waits for an explicit submit (`internal/worker/personal_processor.go:413-418`; `personal.go:433`, `:461` reset the column). The dot is correct there.

Please inline the constants with `fmt.Sprintf` rather than adding two `?`s — given the placeholder hazard this PR itself documents, a status guard that adds arguments in three separate statements is exactly the kind of edit the comment is warning about.

Worth applying the same guard to the pre-existing `status=waiting` filter at `task.go:588`, which carries the identical expression, so the filter and the badge stay in agreement.

### P2 — the MySQL placeholder ordering is not covered by CI

The new tests are SQLite-only. On MySQL, `schedulePendingInvitationExpr` (`task.go:69-77`) contributes an extra `?` that sits *before* the new `PersonalStatusCompleted` argument in all three statements. I verified by reading that the ordering is right in every branch, but nothing enforces it:

- the only MySQL test, `schedule_attention_mysql_test.go:20-23`, skips unless `SUMMARY_MYSQL_TEST_DSN` is set, and `ci.yml:106` / `:142` do not set it;
- that test also never touches `MarkSummaryRead`, so `stateSQL`'s new argument has no MySQL coverage at all.

This is not hypothetical. I mutated `countArgs` to append `PersonalStatusCompleted` *after* `spaceID` and two SQLite tests failed immediately — good signal, but a mutation confined to the MySQL branch would sail through. Adding a `MarkSummaryRead` case to the opt-in MySQL test, or running it against a MySQL service container in CI, would close this.

### P2 — `pending_submission_count` adds a correlated subquery over the whole space

`task.go:938`. `attentionCountSQL` evaluates `(SELECT COUNT(*) FROM summary_participant wait_sp WHERE wait_sp.task_id = sub.id)` once per folded task in the space — the count query deliberately ignores pagination, unlike the page query where the same expression only ran over `pageSize` rows.

Largely mitigated: `uk_summary_participant_task_user (task_id, user_id)` (`migrations/sql/20260610-01-schedule-one-to-many.sql:4`) makes each evaluation a unique-index probe. Not blocking, but worth a glance at p99 list latency in a large space after rollout.

### P2 — list ordering changes for every existing client

`task.go:692` inserts `has_pending_submission DESC` between the invitation and unread terms. Intentional and documented, but any client asserting on list order will see a shift the moment this deploys, independent of the frontend change. No action needed — flagging for the rollout note.

---

## 3. Things I checked that are correct

Recording these so a future reviewer doesn't re-litigate them:

- **Placeholder ordering is correct in all three statements, both dialects.** I traced each `?` textually: `pageSQL` (`task.go:693-700`), `attentionCountSQL` (`:940-950`), `stateSQL` (`:1042-1046`). The new argument is inserted at the right position in every branch. The ordering comment at `:933-937` is accurate and worth keeping.
- **The `COUNT(*) > 1` gate is right, despite looking under-specified.** It counts `Pending` and `Declined` rows, which reads like a bug until you check who decides that a submit is required: `personal_processor.go:361-364` picks the multi-person branch with the *identical* plain `COUNT(*)`, and `Submit` uses the same shape at `personal.go:538-543`. So whenever this gate says `> 1`, the worker took the meta path and `metaCompletionReady` really will block on this user's submit — the dot is accurate. Narrowing it to accepted-only would make the badge *disagree* with the code that decides whether a submission is owed. Please leave it as is.
- **No soft-delete hazard.** `summary_participant` has no `deleted_at` column (`migrations/sql/20260101-00-baseline.sql:32-46`, `model.go:249-260`), and `Leave` / `RemoveMember` hard-delete both the participant and the personal-result rows, so a departed member cannot inflate the roster count.
- **The tests are not tautological.** Reverting `needs_attention` at `task.go:918` fails `TestListSummaries_PendingSubmissionDrivesAttention` and `TestMarkSummaryRead_KeepsPendingSubmissionAttention`; shifting the `countArgs` position fails the same two. Both mutations were reverted.
- **Build and tests are green locally.** `CGO_LDFLAGS="-L/usr/local/lib" go test ./internal/api/handler/ -count=1` → ok, 29.8s. `CGO_ENABLED=0 go vet ./internal/api/handler/` → clean.

## 4. Coverage note

What this review did **not** cover, so you can weigh the verdict accordingly:

- The paired `feat/summary-attention-badge` frontend branch. The P1 turns partly on whether the UI still offers submit on a terminal task; that answer lives in the web repo, not here.
- MySQL was read, not executed — the `JSON_TABLE` roster predicates were verified by inspection only.
- WebSocket / notification delivery. As far as the diff shows, this dot is only ever produced by the list and mark-read projections, but I did not audit the push paths for a stale copy of `needs_attention`.

---

## Verdict

**CHANGES_REQUESTED**, on the strength of the P1 alone. Everything else here is careful, well-tested work, and the placeholder handling — the part the description flags as risky — is correct in every branch. The fix is a one-line status guard in three places; happy to re-review quickly.

Jerry-Xin (dismissed):
✅ **APPROVED** — reviewed at head `4972954f`.

Wires the existing `has_pending_submission` signal (#167) into `needs_attention`, `attention_count` (plus a new `pending_submission_count` breakdown), list ordering, and the `MarkSummaryRead` response. Byte-verified and test-verified.

**Verification (independent, at head)**

- *Positional placeholder ordering* (the risky part): traced the textual `?` order against argument assembly in all three hand-assembled queries — `attentionCountSQL`/`countArgs`, the page query's `pageArgs`, and `MarkSummaryRead`'s `stateSQL`/`stateArgs` (`internal/api/handler/task.go`). The new `PersonalStatusCompleted` argument is inserted exactly at the `has_submit` term's position — after the MySQL-conditional schedule-invitation argument, before the `countInner`/join arguments — in all three. Placeholder/argument counts line up on both dialect branches (SQLite 9/9, MySQL 11/11 for the count query).
- *Projection consistency*: the `has_submit` (count), `has_pending_submission` (list) and `pending_submission` (read) predicates are the identical expression (`worker_status=Completed AND submitted_at IS NULL AND COUNT(participants)>1`), matching the pre-existing gate in the `status=pending` filter. Both `needs_attention` producers (list + read) are updated identically.
- *Lifecycle / clear semantics*: the badge is derived per query — no stored counter to drift or leak. It clears exactly when `submitted_at` is set; both writers (`POST /submit` and the scheduled-task system backfill) use conditional `UPDATE ... WHERE submitted_at IS NULL`, so racing writers cannot double-write or clobber each other. Failed generations (`worker_status=Failed`) never raise the dot; `Leave` physically removes the participant + personal-result rows (dot gone); soft-deleted tasks drop out of the counts; `Regenerate` resets `submitted_at`/`worker_status` (dot drops, re-raises on new completion). Manual tasks keep the dot until explicit `/submit` (owner decision, test-covered); scheduled tasks clear via backfill.
- *Tests*: ran the four new tests plus the full handler package locally — all green (`ok internal/api/handler 31.0s`); `CGO_ENABLED=0 go vet ./...` clean. Coverage includes raise, clear-on-submit, solo-task gate, read-does-not-clear, and list/read agreement — not just happy path.

💬 Non-blocking notes

- The `COUNT(*) > 1` gate counts participants regardless of status (declined members still count). That is established repo semantics since #167 and stays consistent with the meta-summary's accepted-member gate in practice; if a declined member should ever drop the dot, that predicate is the single place to change (three copies: list projection, count query, `MarkSummaryRead`).
- The MySQL-only branches (JSON_TABLE schedule predicates) cannot execute under the SQLite test harness; consider extending the opt-in `SUMMARY_MYSQL_TEST_DSN` coverage to `MarkSummaryRead` and the count query's new argument position (the list query already has coverage).

✅ Highlights

- The positional-argument comment block documenting the full placeholder order is exactly the guard this hand-assembled SQL needed.
- Race-safe submit design (conditional UPDATE + idempotent loser path) keeps the badge's clear boundary robust.
- Additive API surface: `pending_submission_count` is a new field; deploying before the octo-web counterpart is harmless.

mochashanyao (changes requested):
[Octo-Q · automated review]

---

# Code Review — PR #1577 (octo-web)

## Summary

本 PR 把侧边栏「智能总结」红点口径从只读 `pending_invitation_count` 改为读 space 级 `attention_count`（未读 ∪ 未处理邀请 ∪ 待提交），让侧边栏角标与卡片 `needs_attention` 红点同源；工具模块相应由 `summaryMenuBadge` 更名重做为 `summaryAttentionBadge`，并在提交 / 邀请应答 / schedule 确认后补了显式重算钩子，标读回退也扩展为覆盖待提交信号。对齐方向正确、测试覆盖充分，但「标读」这条最高频路径上遗留一处 P1：角标与红点会再次分叉。

## Verification

Static analysis only at head `d05b95d8`; build and tests not executed in this environment.

- ✅ **Badge 写入路径穷举** — 仅两处：全局列表 `loadData`（`packages/dmworksummary/src/pages/SummaryListPage.tsx:238`）与 `refreshSummaryAttentionBadge`（`packages/dmworksummary/src/utils/summaryAttentionBadge.ts:48`）；带 channelId 的嵌入式实例不写全局 badge（有测试锁定）。
- ✅ **更名迁移完整** — 全仓无 `summaryMenuBadge` / `*PendingInvitationBadge` 残留引用；module / ConfirmPage / DetailPage 调用点与对应 mock 全部迁移。
- ✅ **竞态守卫** — `refreshSeq` + await 后比对 spaceId，跨 space 迟到响应被丢弃（deferred 测试覆盖）；相同值不重复触发 `menus.refresh()`。
- ✅ **`attention_count` 后端契约先于本 PR 存在** — 该字段与逐条 `needs_attention` / `has_pending_submission` 均在 base 分支已声明（`packages/dmworksummary/src/types/summary.ts:202`、`:512`），卡片红点在 base 即由 `needs_attention` 驱动；响应缺字段时 `?? 0` 为显式且有测试的降级。
- ✅ **标读回退语义** — `??` 区分 `false` 与 `undefined`；现服务端返回 `needs_attention` 时恒取服务端权威值，回退仅对旧版本生效。
- ✅ **变更动作的主动重算均在成功分支** — submit / 邀请接受拒绝 / schedule-confirm 都在 await 成功后调用 `refreshSummaryAttentionBadge()`；列表挂载时另有 `summary-task-regenerated` → `loadData` 全量回写兜底。

## Findings

One P1 blocker described below.

### P1 — 标读后侧边栏 badge 不更新：卡片红点清掉但角标持续过计 (`packages/dmworksummary/src/pages/SummaryListPage.tsx:123`)

`handleSummaryRead_` 消费 `summary-read` 事件时只更新 `items`（`is_unread` / `needs_attention`），从不写 badge。数据流回溯：点开卡片走 `WKApp.routeRight.push`（分栏视图，列表保持挂载）→ `SummaryDetailPage` 调 `markSummaryRead` 清未读并派发 `summary-read` → 卡片红点（`packages/dmworksummary/src/components/SummaryCard.tsx:147`，由 `needs_attention` 驱动）消失；但 badge 的唯一写入点是 `loadData` 与 `refreshSummaryAttentionBadge`（space 切换 / 提交 / 邀请应答 / schedule 确认），标读一个都不触发，且 batch poll 只覆盖 PENDING / WAITING_CONFIRM / PROCESSING，COMPLETED 条目无人再拉。结果：读完一条未读后角标停在 N，列表里只剩 N−1 个红点，直到用户切菜单 / 切 space 才自愈——正是本 PR 要消灭的「角标与红点不一致」的反向形态（diff-scope: new；base 的 `pending_invitation_count` 不受标读影响，不存在该分叉）。

```ts
// 修复方向：handleSummaryRead_ 内，当某条 needs_attention 由 true → false 时
// 主动重算 space 级计数（本地递减在过滤/分页下不正确）
void refreshSummaryAttentionBadge();
```

补一条「标读后角标随之递减」的回归测试即可锁定。

- 双 refresh 竞态与跨 space 迟到响应（utility 测试以 deferred 锁定先后序）。
- `attention_count` 缺失时归零为显式口径，列表与 utility 两处均有测试。
- 新增 `has_pending_submission?` / `pending_submission_count?` 类型字段为纯扩展，无消费端破坏；`page_size=1` 只取计数，参数形态与现有 `listSummaries` 契约一致。

## Verdict: CHANGES_REQUESTED

补齐标读路径的 badge 回写后，本 PR 建立的「侧边栏 = needs_attention 同源」契约才算闭环；其余检查项均非阻塞。

[Octo-Q] verdict: REQUEST_CHANGES — 标读（markSummaryRead → summary-read）只刷卡片红点不写侧边栏角标，读掉未读后角标与红点长期不一致（P1，本 PR 新引入）；修复仅需在 needs_attention true→false 时调 refreshSummaryAttentionBadge()。

wanyiwang06:
Thanks — the P1 is real and I've fixed it. Pushed `129a4a8`.

## P1 — terminal tasks: fixed

I verified both paths you named before changing anything:

- `scheduler.go` flips `Processing -> Failed` on `retry_count >= maxRetry-1` and writes only `status` / `processing_deadline` / `retry_count` / `error_message`.
- `CancelSummary` (`task.go`) writes only `status` + `error_message`.

Neither touches `summary_personal_result`, so `worker_status` stays `Completed` and `submitted_at` stays `NULL`. And your point about recovery holds: `Regenerate` is creator-gated (`task.go`, 40004 "仅创建者可重新生成"), so a non-creator participant has no way out. Confirmed too that `Submit` has no task-status guard — only `pr.WorkerStatus != Completed` — so the dot *is* technically clearable, but only if the UI still renders a submit button on a dead task, which is not something I want the badge's correctness to depend on.

Implemented as you suggested, `fmt.Sprintf` with inlined constants rather than two more placeholders:

```go
func pendingSubmitStatusGuard(taskAlias string) string {
	return fmt.Sprintf(" AND %s.status NOT IN (%d,%d)", taskAlias, model.StatusFailed, model.StatusCancelled)
}
```

Applied at **four** sites, not three — I took your suggestion about `task.go:588` and treated it as part of the fix rather than optional. If the `status=waiting` filter and the badge disagree, a card can be counted by the badge but hidden by the very filter the user clicks to find it. Sites: list projection, attention-count query, `MarkSummaryRead`'s `stateSQL`, and the waiting filter.

`Completed` stays eligible, exactly as you argued. `TestListSummaries_CompletedTaskStillPendsSubmission` pins it so a future "just exclude every terminal status" simplification can't silently drop the one terminal status that still needs the dot.

New tests:

- `TestListSummaries_TerminalTaskDropsPendingSubmission` — sub-tests for reaper-Failed and creator-Cancelled, asserting the card flag, `needs_attention`, both counts, **and** that `MarkSummaryRead` agrees (it's the other `needs_attention` producer, so a guard applied to only one of them would be a split brain).
- `TestListSummaries_CompletedTaskStillPendsSubmission`
- `TestListSummaries_WaitingFilterAgreesWithTerminalGuard`

Mutation-checked: stubbing `pendingSubmitStatusGuard` to `""` fails `TerminalTaskDropsPendingSubmission` and `WaitingFilterAgreesWithTerminalGuard`; reverting restores green.

## P2 (MySQL placeholder coverage) — fixed

Agreed this was the gap that mattered most after the P1, precisely because a MySQL-confined mutation sails through the whole SQLite suite. Added `attention_pending_submit_mysql_test.go`, opt-in on `SUMMARY_MYSQL_TEST_DSN` like its sibling, covering the two statements `schedule_attention_mysql_test.go` doesn't touch: the attention-count query and `MarkSummaryRead`'s `stateSQL`.

Two details worth noting:

- It seeds a schedule with `confirm_policy=1` and an unconfirmed roster member, so `schedulePendingInvitationExpr` is genuinely *live* during the test rather than short-circuiting. Otherwise its placeholder wouldn't actually be bound and the ordering wouldn't be exercised.
- It asserts the count **values**, not just booleans. A misaligned argument wouldn't necessarily error — it would silently compare `worker_status` against a uid and yield 0.

`markReadRequest` gained a Space-explicit variant (`markReadRequestWithSpace`) since the integration tests seed into their own space and can't use the `space1` default.

I did not wire a MySQL service container into CI in this PR — that's an infra change affecting every job and belongs on its own. The test is ready for whoever does it.

## P2 (correlated subquery cost) — noted, not changed

Your read matches mine: `uk_summary_participant_task_user` makes each evaluation a unique-index probe. I'd rather not restructure the count query speculatively; I'll take the p99 look after rollout as you suggest.

## P2 (list ordering shift) — carried into the description

Added to the PR body's rollout note.

## On your coverage caveat

You flagged that the P1 turned partly on whether the UI still offers submit on a terminal task. It's moot now — the guard is server-side, so the badge is correct regardless of what the frontend renders. For the record, the paired `feat/summary-attention-badge` (octo-web#1577) has its own P1 fix landed in `04ae34d2` and doesn't touch submit affordance gating.

One correction to a claim in the other review on that PR: it states this backend PR is "already merged and deployed". It is not — #229 is still open and unmerged as I write this. The frontend's optional-field handling means it degrades quietly either way, but nobody should plan a rollout on that assumption.

yujiawei (approved):
# Code Review — PR #229 (octo-smart-summary), round 2

Re-reviewed at head `129a4a8`. My previous verdict was against `4972954`; per the re-review rule that approval does not carry forward, this is a fresh verdict against the current head, not a ratification of the fix commit.

## 1. Scope check — ✅

`129a4a8` addresses the blocker and the test-coverage gap, and nothing else. No scope creep, no drive-by refactors beyond the one test helper the new MySQL test genuinely needs (`markReadRequestWithSpace`, `auth_test.go:404-412`).

It also picked up the optional consistency item I flagged — the pre-existing `status=waiting` filter (`task.go:620`) now carries the same guard, so the filter and the badge cannot disagree. Good call; that was the failure mode where a card is counted by the badge but hidden by the filter the user clicks to find it.

## 2. Code quality — Approved

### P1 (previous round) — resolved

`pendingSubmitStatusGuard` (`task.go:80-110`) is applied at **all four** sites that carry the predicate — `task.go:620` (waiting filter), `:714` (card), `:970` (space count), `:1064` (mark-read). I grepped for `wait_sp.task_id` and there is no fifth, unguarded occurrence.

Verified in detail:

- **Only `Failed`/`Cancelled` are excluded; `Completed` is deliberately kept.** This is the subtle half and it is right: a personal-only regenerate on a finished task intentionally leaves `submitted_at` NULL and waits for an explicit submit. `TestListSummaries_CompletedTaskStillPendsSubmission` pins it, and the doc comment explains why — that stops a future "just exclude all terminal statuses" cleanup from silently killing the signal.
- **No new placeholders.** The constants are inlined via `fmt.Sprintf`, so the positional-argument ordering in all four hand-assembled statements is byte-for-byte unchanged. That was the right call given the hazard the ordering comment warns about; binding two params in four statements would have been the riskiest possible way to fix this.
- **Alias hygiene.** `sub.status` / `t.status` / `summary_task.status` are all qualified, so there is no collision with the `status` columns on the `me` (participant) and `p` joins in the same statements. `taskAlias` is a compile-time literal at every call site — no injection surface.
- **Mutation-tested.** Making the guard return `""` fails `TerminalTaskDropsPendingSubmission` (both subcases) and `WaitingFilterAgreesWithTerminalGuard`; swapping `Cancelled` for `Completed` fails `CompletedTaskStillPendsSubmission` plus the cancelled subcase. The new tests genuinely constrain the fix rather than restating it. Both mutations reverted.

### P2 (previous round, MySQL coverage) — resolved

`attention_pending_submit_mysql_test.go` covers exactly the two statements that had no MySQL coverage: `attentionCountSQL` and `MarkSummaryRead`'s `stateSQL`. The design detail I want to call out approvingly — the fixture deliberately makes `schedulePendingInvitationExpr` live (`confirm_policy=1` with an unconfirmed roster member) so its placeholder is actually *bound* rather than short-circuited, and the assertions check the *value* rather than just absence of error. A swapped argument would bind `worker_status` against a uid string and silently yield 0; asserting `PendingSubmissionCount == 1` catches that. That is the right shape for a placeholder-ordering test.

Still opt-in behind `SUMMARY_MYSQL_TEST_DSN`, which `ci.yml` does not set — so this remains documentation-plus-a-manual-gate rather than enforcement. Not blocking (it matches the existing sibling test's contract), but wiring a MySQL service container into CI would turn both of these into real guards.

### Carried forward, still non-blocking

- The correlated `COUNT(*)` in `attentionCountSQL` still runs per folded task across the whole space; index-served by `uk_summary_participant_task_user`. Unchanged by this commit.
- The list ordering change (`has_pending_submission DESC` between invitation and unread) still shifts results for existing clients. Intentional and documented.

## 3. One correction to my previous review

I need to walk back a claim I made, because it was load-bearing for the P1 and it was wrong.

I wrote that a non-creator would have *no user-facing recovery*, on the assumption that the UI would not offer a submit affordance on a terminal task. I have since read the shipped frontend: the submit entry point is gated on `worker_status === 2 && !submitted_at && members.length > 1 && summary_mode === BY_PERSON` and on **no task status at all** (`SummaryDetailPage.tsx` `shouldShowMySubmit`, plus `isMultiCollab` / `canRevealPersonalContent`). The `members.length > 1` term matches this backend gate exactly, because `GetMembers` (`personal.go:1099-1100`) returns the roster unfiltered. So the button was in fact reachable, and the badge was clearable.

That makes the original symptom "a badge that nags you to submit into a task that can never aggregate again" rather than "a badge that can never be cleared" — the fix is still correct and still worth making, but I overstated the severity. Flagging it so the record is accurate rather than quietly letting the stronger claim stand.

Usefully, the same reading confirms this fix does not overshoot: the only consumers of `has_pending_submission` on the client are the card's status label and the mark-read `needs_attention` fallback, **not** the submit button. So suppressing the flag on a dead task removes the nag without removing the escape hatch for anyone who still wants to submit. That is the right combination.

## 4. Verification performed

- `CGO_ENABLED=0 go vet ./internal/api/handler/` → clean
- `CGO_LDFLAGS="-L/usr/local/lib" go test ./internal/api/handler/ -count=1` → ok, 30.6s
- Two independent mutations of the new guard, both caught (detail above); working tree restored to `129a4a8`
- Grepped for unguarded occurrences of the predicate — none

Not covered: MySQL was not executed (no DSN available here), so the new integration test is verified by reading only.

## Verdict

**APPROVED.** The blocker is fixed at the right layer, with the one nuance that made it tricky handled correctly and pinned by tests that resist the obvious wrong simplifications.

Jerry-Xin (approved):
# Re-review — PR #229 at head `129a4a8`

Re-reviewed the true delta against the previously approved head `4972954`: one commit (`fix(summary): drop the pending-submission dot on terminally dead tasks`) touching `internal/api/handler/task.go`, `internal/api/handler/auth_test.go`, and two test files. No rebase noise, no collateral changes beyond the claimed fixes (the `auth_test.go` change is a pure helper parameterization, `markReadRequestWithSpace`, needed so the MySQL integration test can seed into its own space).

## Verdict: APPROVE

All three peer findings are resolved on this head. Verified byte-level, not by description.

### 1. Failed/Cancelled tasks no longer hold the dot (yujiwei P1) — fixed

`pendingSubmitStatusGuard` (internal/api/handler/task.go:80-110) appends `AND <alias>.status NOT IN (4,5)` — `StatusFailed`/`StatusCancelled` inlined via `fmt.Sprintf`, deliberately adding **zero placeholders**, so every positional argument list in the hand-assembled statements is untouched. I enumerated every task status against the guarded predicate:

| task status | predicate | bounded? |
|---|---|---|
| Pending(0) / WaitingConfirm(1) / Processing(2) | may fire | live task — correct |
| Failed(4) / Cancelled(5) | forced false | reaper (`internal/worker/scheduler.go`) and `CancelSummary` no longer strand a dot |
| Completed(3) | may fire | correct: a personal-only regenerate resets `submitted_at` and waits for an explicit `/submit` (`internal/worker/personal_processor.go`, `internal/api/handler/personal.go`), and `Submit` has no task-status guard, so the dot is always clearable |

No terminal path remains where the predicate is true with no clear path. The guard is applied consistently at **all four** occurrences of the raw predicate: the `status=waiting` filter (`task.go:620`), the list projection (`task.go:714`), the space-level count (`task.go:970`), and `MarkSummaryRead`'s state query (`task.go:1064`). Aliases check out: `sub` projects `status` via `SELECT *` from `summary_task`, `t` is `summary_task` itself.

### 2. Attention count on cancel/fail (yujiwei P1 companion) — fixed

Rather than writing personal-result rows in the cancel/fail paths, the count query now uses the exact same guarded expression as the card projection (`has_submit` in `attentionCountSQL`), so card red dot, `pending_submission_count`, and `attention_count` share one source of truth and cannot diverge. This is the safer design — no second writer into `summary_personal_result`.

### 3. Read path / badge consistency (mochashanyao) — backend side consistent; the remaining fix lives in the frontend repo

Note for the record: that CR body reviews octo-web #1577 (`SummaryListPage.tsx`'s `handleSummaryRead_` not calling `refreshSummaryAttentionBadge()`) — its fix belongs to the paired frontend branch, not this repo. On this side, `MarkSummaryRead` returns the authoritative post-read attention state (`is_unread`, `has_pending_submission`, `needs_attention`), and on this head its `pending_submission` expression is the identical guarded predicate as list/count — no stored counter exists to drift. The client-side badge refresh on read events is tracked on the octo-web side.

### 4. Tests at head — green

Ran locally at `129a4a8` (`go test ./internal/api/handler/ -count=1`): full package passes (~31s), build and `go vet` clean. The new tests pin exactly the fix paths, not just happy paths:

- `TestListSummaries_TerminalTaskDropsPendingSubmission` — Failed and Cancelled subtests, each also asserting `MarkSummaryRead` agrees;
- `TestListSummaries_CompletedTaskStillPendsSubmission` — pins the deliberate Completed exclusion so a future "exclude all terminal" simplification can't silently drop the signal;
- `TestListSummaries_WaitingFilterAgreesWithTerminalGuard` — waiting filter and badge stay in agreement;
- `TestPendingSubmission_MySQL_PlaceholderOrderingAndMarkRead` (opt-in, `SUMMARY_MYSQL_TEST_DSN`) — covers the two statements the sibling MySQL test never touched: the attention-count query and `MarkSummaryRead`'s stateSQL, with a live `schedulePendingInvitationExpr` (confirm_policy=1, unconfirmed roster member) so the extra placeholder is actually bound. This answers the MySQL-coverage gap raised on the previous head.

### Non-blocking

- The correlated participant-count subquery now appears in four statements; it probes `uk_summary_participant_task_user` each time, but a glance at p99 list latency in large spaces after rollout is still worthwhile (carried over from the previous round, unchanged by this delta).

yujiawei (approved):
# Code Review — PR #229 (octo-smart-summary)

Re-review at head `129a4a8` (round 2). I re-traced every positional placeholder in all four hand-assembled statements against both dialect branches at the current head, read the surrounding `ListSummaries` / `MarkSummaryRead` bodies rather than the diff alone, verified the lifecycle claims against `worker/scheduler.go`, `worker/meta_processor.go` and `worker/personal_processor.go`, and ran the package tests plus `vet`/`gofmt`. Round-2 delta is exactly the status guard plus tests — no drive-by changes.

---

## 1. Spec compliance — ✅

Against issue #230's backend section, four for four:

| Required | Where |
|---|---|
| `needs_attention` ORs `has_pending_submission` | `internal/api/handler/task.go:950` |
| `attention_count` gains the term + `pending_submission_count` breakdown | `task.go:970`, `:992` |
| Ordering: invitation > pending submission > unread | `task.go:724` |
| `MarkSummaryRead` returns the flag and keeps it in its recompute | `task.go:1087-1088` |

Nothing missing, no schema change, and the out-of-scope items in #230 (IM/DM notification for invitations, the unimplemented pre-run reminder) are untouched.

The one addition beyond the issue's literal text is `pendingSubmitStatusGuard` (`task.go:80-110`). I do not count it as scope creep: it is a defect fix inside the exact surface this PR touches, it was raised in round 1, it is confined to the four sites that already carried the predicate, and it is now described in the PR body. Calling it out explicitly so the owner can veto if they disagree.

## 2. Code quality — Approved

### The round-1 P1 is fully fixed

`pendingSubmitStatusGuard(alias)` returns ` AND <alias>.status NOT IN (4,5)` and is applied at **all four** sites carrying the raw predicate:

- list projection — `task.go:714`
- attention-count query — `task.go:970`
- `MarkSummaryRead` `stateSQL` — `task.go:1064`
- pre-existing `status=waiting` filter — `task.go:620`

Including the fourth site was the right call and goes beyond what round 1 asked for. The waiting filter carries the identical expression; leaving it out would let a card be counted by the badge while being hidden by the very filter a user clicks to find it. `TestListSummaries_WaitingFilterAgreesWithTerminalGuard` pins that agreement.

The lifecycle reasoning behind the guard checks out against the code:

- `scanStuckTasks` (`internal/worker/scheduler.go:391-433`) resets `Processing → Pending` while retries remain and writes `StatusFailed` at exhaustion, touching only `status` / `processing_deadline` / `retry_count` / `error_message`.
- `CancelSummary` (`task.go:2217-2222`) accepts `Pending/WaitingConfirm/Processing` and writes only `status` + `error_message`.
- Neither touches `summary_personal_result`, so without the guard `worker_status` stayed `Completed` and `submitted_at` stayed `NULL` forever.
- Excluding only `Failed`/`Cancelled` is correct. `Completed` must stay eligible because the personal-only regenerate branch (`internal/worker/personal_processor.go:415-419`) deliberately skips `backfillSystemSubmittedAt` and waits for an explicit `/submit`. `TestListSummaries_CompletedTaskStillPendsSubmission` pins exactly that, which stops a future "just exclude every terminal status" simplification.

Inlining the constants via `fmt.Sprintf` rather than adding two `?`s in four statements is the right trade here, and the reason is documented at the function.

### Placeholder ordering — verified correct, all four statements, both dialects

I traced the textual `?` order against argument assembly rather than trusting the comment:

- **`pageSQL`** (`task.go:722-731`): `attentionSelect` precedes `innerSQL` in the emitted text, so `ParticipantPending` → `[userID if mysql]` → `PersonalStatusCompleted` → `args...` → `userID ×4` → `pageSize, offset`. Matches `pageArgs` exactly, including the `PersonalStatusCompleted` append at `:729`.
- **`attentionCountSQL`** (`task.go:970-985`): 11 placeholders on MySQL, 9 on SQLite, matching `countArgs` in both branches. The new argument at `:981` sits between the invitation term and `countInner`, which is where the `has_submit` `?` textually appears.
- **`stateSQL`** (`task.go:1059-1075`): 7 on MySQL, 6 on SQLite; the append at `:1074` lands between the invitation `?` and the three join `userID`s.
- **`whereSQL`** waiting branch (`task.go:620-624`): unchanged placeholder count — the guard contributes none.

The guard itself binds nothing, which is what keeps this edit safe.

### Tests

Run locally at head:

```
CGO_LDFLAGS="-L/usr/local/lib" go test ./internal/api/handler/ -count=1   # ok, 30.6s
CGO_ENABLED=0 go vet ./...                                               # clean
gofmt -l <changed files>                                                 # clean
```

The seven SQLite tests are not tautological — they assert count *values*, not just booleans, and the terminal-task case asserts both `needs_attention` producers agree, which is the split-brain risk when a guard is applied to one projection but not the other.

`attention_pending_submit_mysql_test.go` is a genuine improvement over a boolean smoke test: it seeds `confirm_policy=1` with an unconfirmed roster member so `schedulePendingInvitationExpr` actually binds its placeholder instead of short-circuiting, and it asserts counts because a misaligned argument compares `worker_status` against a uid and yields `0` silently rather than erroring.

### Non-blocking (P2)

1. **MySQL coverage is still opt-in.** `.github/workflows/ci.yml:106,142` do not set `SUMMARY_MYSQL_TEST_DSN`, so both MySQL files skip in CI. A mutation confined to the MySQL branch would still ship green. Agreed this belongs in a separate infra change; flagging so it does not get forgotten.
2. **Waiting-filter placeholders have no MySQL coverage.** The new MySQL test does not issue a `?status=0` request, so `task.go:620`'s MySQL argument shape stays uncovered. Pre-existing rather than introduced here (this PR adds no placeholder there), so it is a cheap add whenever (1) is done.
3. **Correlated subquery over the whole space.** `attentionCountSQL` evaluates `(SELECT COUNT(*) FROM summary_participant wait_sp WHERE wait_sp.task_id = sub.id)` once per folded task, and the count query deliberately ignores pagination. `uk_summary_participant_task_user` (`migrations/sql/20260610-01-schedule-one-to-many.sql:4`) makes each evaluation a unique-index probe, so this is a "watch p99 after rollout" item, not a change request.
4. **Positional raw SQL remains fragile by construction.** Four statements assemble fragments and append arguments conditionally on dialect. The ordering comment at `task.go:965-969` plus the opt-in MySQL test are reasonable mitigations for now; if a fifth term ever lands here, named parameters would be worth the refactor. Out of scope for this PR.
5. **Frontend lag window.** Until octo-web#1577 lands, an old client's optimistic `handleSummaryRead_` recompute drops the submission term on read, so the dot briefly disappears until the next list refresh. The server value is authoritative and self-heals; #230 already lists this as a frontend item.

## 3. Things I checked that are correct

Recording these so a later reviewer does not re-litigate them:

- **No missed call site.** Grepping the non-test tree for the raw predicate finds exactly the four guarded locations; no other handler, view, or migration reproduces it. `GetSummary` and `BatchStatus` do not project `needs_attention`, so there is no third producer to keep in sync.
- **No join fan-out.** `uk_summary_participant_task_user (task_id, user_id)` plus `uk_task_participant (task_id, participant_ref_id)` on `summary_personal_result` mean the `pr` LEFT JOIN yields at most one row per task, so `SUM(has_submit)` cannot double-count.
- **No soft-delete hazard on the roster count.** `summary_participant` has no `deleted_at` column (`migrations/sql/20260101-00-baseline.sql`), so a `deleted_at IS NULL` term in the `COUNT(*) > 1` gate would be meaningless.
- **No cross-user leak.** `MarkSummaryRead` authorizes via `authorizeTaskAccess` and the new `pending_submission` term reads `pr` joined on the caller's own `user_id`.
- **`ORDER BY` alias reference** is legal in both dialects, and `has_pending_submission` was already computed per row, so the new sort term adds no per-row cost to the page query.

## 4. Coverage note

What this review did **not** cover:

- MySQL was read, never executed. The `JSON_TABLE` roster predicates and the MySQL argument shapes are verified by inspection only; the opt-in integration test could not be run without a DSN.
- No query plans were measured; the p99 remark on P2-3 is reasoning about index shape, not a benchmark.
- The paired octo-web#1577 branch was not reviewed here.
- WebSocket / notification push paths were not audited for a stale copy of `needs_attention`; the diff shows the flag is only produced by the list and mark-read projections, but I did not prove the negative.

## 5. One merge-gate note

`reviewDecision` on this PR is currently `CHANGES_REQUESTED`, held by a review whose body reviews a different repository's pull request (octo-web #1577) rather than this one. That review will need to be dismissed or re-submitted against this PR before the decision can flip, independent of this approval.

---

## Verdict

**APPROVED.** The round-1 P1 is fixed at all four sites, with the fourth one caught by the author rather than the review. `Completed` is correctly left eligible and pinned by a test. Placeholder ordering — the part this PR itself flags as risky — is correct in every statement and every dialect branch. Remaining items are non-blocking and mostly belong to a CI infra change.

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #229 (octo-smart-summary)

## Summary

This PR activates the previously dormant pending-submission attention signal. `has_pending_submission` was already computed by `ListSummaries` but never fed `needs_attention` or the sidebar counts, so the red dot never rendered. The change wires the signal into `needs_attention` and `attention_count` (with a new `pending_submission_count` breakdown), ranks pending submissions between pending invitations and unread content in the list ordering, and mirrors the signal into `MarkSummaryRead` so opening the detail page does not clear the dot — reading is not submitting. It also adds a server-side status guard excluding terminally dead tasks (Failed via the stuck-task reaper, Cancelled via `CancelSummary`) from the predicate in all four hand-assembled SQL statements, since neither terminal path touches `summary_personal_result` and a badge that can never return to zero would destroy the signal's credibility. The test suite pins the rules end-to-end through the real router: multi-person gate, clear-on-submit, survive-read, terminal suppression, waiting-filter parity, and Completed deliberately unguarded. Well-framed change with careful positional-argument discipline.

## Verification

Static analysis only at head `129a4a89`; build and tests not executed in this environment.

- ✅ **Positional argument ordering** — in all four hand-assembled statements (`internal/api/handler/task.go:620`, `:714`, `:970`, `:1064`) the `args`/`pageArgs`/`countArgs`/`stateArgs` assembly matches the textual `?` order on both dialects: `schedulePendingInvitationExpr` collapses to `0` on SQLite (no placeholder) and every conditional extra `userID` append is gated on the same dialect check; `pendingSubmitStatusGuard` inlines its constants as `%d` and contributes no placeholders.
- ✅ **Guard parity (C1)** — `pendingSubmitStatusGuard` is applied at every site carrying the pending-submission predicate (status=waiting filter, per-card attention select, attention-count query, mark-read state query) with a valid alias in each scope (`summary_task`/`sub`/`sub`/`t`), and both `needs_attention` emitters (`internal/api/handler/task.go:950`, `:1088`) include `HasPendingSubmission`. No fifth consumer of the predicate exists in the repo.
- ✅ **Data-flow traceback** — the worker sets `worker_status=Completed` and leaves `submitted_at` NULL (`internal/worker/personal_processor.go:57-61`); scheduled multi-person tasks are system-backfilled (`internal/worker/personal_processor.go:432-433`) so they never pend; manual tasks wait for `/submit`, which clears the dot via the conditional `submitted_at IS NULL` update (`internal/api/handler/personal.go:513-522`). Reaper `Processing→Failed` (`internal/worker/scheduler.go:387-456`) and `CancelSummary` (`internal/api/handler/task.go:2205-2221`) verified as the two strand paths the guard suppresses; Completed is correctly left unguarded for the personal-regenerate flow.
- ✅ **Cardinality** — `uk_summary_participant_task_user` (`migrations/sql/20260610-01-schedule-one-to-many.sql:4`) plus the personal-result unique key rule out `pr`/`me` join fanout; the `wait_sp COUNT(*) > 1` gate counts distinct users and agrees with the meta completion gate, so the dot always prompts an action that is possible and clears it.
- ✅ **Auth/visibility and contract** — no new endpoint; all changed SQL lives inside the pre-existing space- and user-scoped statements; new response fields (`pending_submission_count`, mark-read `has_pending_submission`) are purely additive.
- ✅ **Tests exercise the production path** through the real router (`internal/api/handler/attention_pending_submit_test.go`), with MySQL placeholder-ordering coverage available via `SUMMARY_MYSQL_TEST_DSN` (`internal/api/handler/attention_pending_submit_mysql_test.go:38`), mirroring the established `schedule_attention_mysql_test.go` convention.

## Findings

No P0/P1/P2 issues.

- Guard alias resolution inside the correlated `EXISTS` (status=waiting filter) and inside the derived-table `CASE` terms; constants inlined as integers, no injection surface (aliases are compile-time literals).
- `ORDER BY` insertion keeps the existing tiebreakers (`has_pending_invitation`, then `is_unread`, `activity_at`) intact.
- `/submit` has no task-status precondition, so even on terminal tasks the dot remains clearable; the guard suppresses it as a product decision, not as the only escape.
- Schedule folding (`rn=1`) semantics unchanged — the signal is computed on the representative row exactly like the pre-existing unread/invitation signals.
- Roster gate counting declined/pending members is consistent with the meta gate (only accepted members block aggregation), so the viewer's submit is always meaningful when the dot fires.
- No schema change required; no docs enumerate these response fields, so no contract drift.

## Verdict: APPROVED

The wiring is complete and symmetric across every statement carrying the predicate, the positional-argument ordering is correct on both dialects at this head, and the terminal-status guard is justified by verified worker behavior and pinned by tests. No blocking or should-fix issues found.

[Octo-Q] verdict: APPROVE — 无 P0/P1：pending-submission 信号接线、终态守卫、四处手写 SQL 的位置参数顺序均在本 head 验证正确，测试覆盖生产路径；零阻塞。

wanyiwang06:
由 PR #233 合并实现，本 PR 关闭。

PR #238, opened 2026-09-05
feat(summary): add unified summary workspace backend

wanyiwang06:
## What this PR delivers

- Adds the unified Summary Workbench backend contract for session-scoped chat, history, preview/revision, save, and direct team workflow creation.
- Keeps selected chats, participants, template, time range, and referenced summaries in one server-authoritative scope.
- Lets the Agent understand free-text source and time-range requirements while trusted template/system actions retain deterministic workflow routing.
- Preserves historical previews as read-only versions and allows only the latest preview to be saved.
- Provides an environment-level rollback through `SUMMARY_WORKBENCH_ENABLED=false`, which fails workspace chat, history, confirmation, and save operations closed.

## Scope and evidence guarantees

- Channel discovery only registers authorised candidates; it does not make those channels readable.
- `set_summary_scope` is the single declaration boundary. It runs after discovery and before fetch, can succeed only once per turn, and freezes further discovery.
- The declaration's complete final scope is used consistently for message access, persisted `scope_json`, coverage, and `preview.effective_scope`.
- Replace and extend both preserve exactly the model-declared channel set; discovery results that were not selected cannot enter the summary.
- Preview saves resolve citations from the exact generating run's evidence session, including rotated identities used by first-turn, replace, extend, and time-only conversations.

## Latest review fixes

- Fixed saved previews losing citations when the generating turn used a rotated internal session identity.
- Fixed cold starts without a valid declaration: they now return a clarification instead of producing an ungrounded preview or retryable 500.
- Enforced discovery → declaration → fetch ordering, including tool calls emitted in the same model turn.
- Enforced the 30-channel limit, 90-day range limit, and 256-character range-label limit before retrieval.
- Removed the dead keyword-based scope/time classifier and its stale tests; user free text is interpreted through the structured Agent scope tool instead.
- Made Workspace save idempotency ignore client scope fields that are replaced by server-authoritative state.

## Behavioural notes

- A cold start must declare its first source with `replace`; if no unique source can be determined, the Agent asks for clarification.
- Existing selected chats remain readable during ordinary revisions. Newly discovered chats become readable only after a successful replace/extend declaration.
- Multi-person summaries still start directly from the trusted `start_team_workflow` action after source, participant, time-range, and reference validation.
- Workflow dispatch remains after the Workspace state transaction commits.

## Verification

- [x] `CGO_ENABLED=0 go test ./...`
[...]

PR #239, opened 2026-09-08
fix(summary): 修复工作台私聊来源显示会话 ID

wanyiwang06:
## 问题与根因

入口合并后，开启总结工作台时，私聊来源使用规范会话 ID `uid_a@uid_b`。
旧的来源名称解析器仍用整个 `source_id` 查询 IM 的单个用户 UID，导致查询未命中，
保存并显示 `来源-xxxxxxxx(私聊)`。关闭工作台的旧入口使用对方 UID，因此没有该问题。

这是新工作台和后端名称解析之间的兼容缺陷，不是字符集乱码。修复位于后端；
本 PR 不包含前端布局、模板交互或入口开关的变更。

## 修改内容与兼容边界

1. 新增以可信的总结创建者为基准的名称解析，支持旧单用户 UID 和双人会话 ID；
   工作台预览保存、个人/多人工作流创建均保存 IM 中的真实私聊对象名称，
   不使用客户端传入的名称。
2. 保留原有规范 `source_id`，不改动消息检索与权限校验。创建者不属于该会话、
   ID 格式错误、用户不存在或 IM 数据库不可用时继续安全降级，不猜测会话对象。
3. 列表、详情及新分享快照兼容已保存的错误名称：只对空名称及精确匹配的
   规范私聊占位名称实时解析。使用任务创建者而非当前查看者，避免名字随查看者反转。
4. 保留真实历史名称快照、旧格式非空名称、群聊/子区名称规则及既有分享快照；
   不改写数据库历史行，不新增迁移。
5. 补充回归测试及 `docs/dm-source-name-regression.md`，记录触发条件与修复边界。

## 回归验证

- [x] 双人 ID 的两种顺序、旧单用户 UID、未知用户、空名称和 nil IM 数据库。
- [x] 非会话成员、缺少创建者、缺少一方、自身重复及多余分隔符安全降级。
- [x] 工作台真实保存路径、个人/多人工作流：保存正确名称，规范会话 ID 不变，客户端名称不可信。
- [x] 创建者与参与者读取列表/详情、新建分享快照；历史真实名称保持不变，读取不更新历史行。
- [x] `go test ./internal/service ./internal/api/handler ./internal/worker -count=1`
  （启用 CGO，并配置 libtokenizers 链接路径）；`Dockerfile.api` 构建通过。

## 本地验收与部署范围

已将本次后端修复部署到本地 28140 对应的 API 镜像，保留现有前端、worker 和测试数据。
浏览器对同一条历史总结验证：`来源-7ecf56dc(私聊)` 已变为 IM 中的
`总结参与者(私聊)`，无需重新生成总结。

未部署 test 或生产环境。回滚可恢复上一 API 镜像，无需回滚数据迁移。


yujiawei (approved):
# Code Review — PR #239 (octo-smart-summary)

Reviewed at head SHA `5bfa0340a38eae9b65cf157cdb848379904a9966` (round 1, no prior reviews on this PR).

Verified locally: `go build ./...` clean; `CGO_ENABLED=1 CGO_LDFLAGS="-L/usr/local/lib" go test ./internal/service ./internal/api/handler -run 'DM|SourceName' -count=1` passes.

---

## 1. Spec compliance

**Spec: ✅**

The PR body scopes itself to: resolve canonical `uid_a@uid_b` DM source names on the backend; keep `source_id`, message retrieval and permission checks untouched; do not rewrite historical rows; no migration; no frontend changes; add regression coverage plus a doc. Checked each claim against the diff:

- **Not missing anything in scope.** Write paths both threaded: `internal/service/summary_workflow.go:461` (personal/team workflow, `in.creatorID` set at `:348-350, :405`) and `internal/api/handler/agent_summary.go:662` (workspace/agent save). Read paths all three: `internal/api/handler/task.go:693` (list), `task.go:967` (detail), `internal/api/handler/share.go:411` (share snapshot).
- **Nothing extra built.** No new endpoints, request fields, model columns, config or feature flags. `ResolveSourceNameWithType` is retained as a thin compat wrapper (`internal/service/summary.go:43-45`) so the six unrelated callers keep compiling with unchanged semantics.
- **No divergence from the stated approach.**
  - `source_id` genuinely untouched — `ResolveSourceNameForActor` changes only the *lookup key* (`internal/service/summary.go:113-121`); the persisted `SourceID` is written from the input in every path.
  - Creator perspective, not viewer perspective, at all three read sites — this is the right call, otherwise the chip name would flip depending on who opens the summary.
  - No migration files in the diff; `TestSummarySourceName_CanonicalDMHistoricalListAndDetail` (`internal/api/handler/source_name_dm_test.go:98-104`) asserts the stored row is byte-identical after a read, and it passes.
  - Client-supplied names are genuinely untrusted on the workspace path: `applyWorkspaceScopeToSaveRequest` (`internal/api/handler/agent_summary_save.go:229-232`) rebuilds `req.Sources` from the server-locked scope, so both the id and the name come from the server.
  - All 9 changed files are Go or docs — no frontend.

### Test effectiveness (mutation-checked)

The PR claims each test goes RED if its fix line is reverted. I verified this rather than taking it on trust — three mutations, reverted after each:

| Mutation | Result |
|---|---|
| Drop the peer derivation in `ResolveSourceNameForActor` (`summary.go:114-120`) | 6 tests RED across both packages |
| Drop the canonical-DM placeholder exception in `ResolveStoredSourceName` (`summary.go:149-152`) | 3 tests RED |
| Pass `""` instead of the creator at the 3 threaded call sites | 2 tests RED |

No tautological assertions found. This is well above the usual bar for a regression-fix PR.

---

## 2. Code quality

**Quality: Approved** — no P0/P1. Seven P2 items below; none blocks merge.

### P2-1 — Three implementations of the same canonical-DM peer derivation, two of them semantically different

The new `sourceNameDMPeerID` is the third parser of the `uid_a@uid_b` format in this binary:

- `internal/service/summary.go:130-142` (new) — `strings.Split`, `len != 2` rejects, rejects `parts[0] == parts[1]`
- `internal/api/handler/agent_summary_workspace.go:1964-1976` `summaryWorkspaceDMPeerID` — logically identical to the above
- `internal/api/handler/bot_summary_create.go:405-419` `deriveDMPeer` — `strings.SplitN(canonicalID, "@", 2)`, and **returns the owner for the self-DM `owner@owner`**, which its call site explicitly depends on (`bot_summary_create.go:359-361`: "Self-DM ('note to self', peer == owner) is by construction inside the owner's own space membership set")

So one string format now has two different self-DM behaviours and two different multi-`@` behaviours, in three places, with nothing tying them together in tests.

Failure scenario: a later change makes `deriveDMPeer` reject self-DM "for consistency" and silently breaks bot self-DM authorization; or makes `sourceNameDMPeerID` accept self-DM while the workspace scope validator at `agent_summary_workspace.go:1924` still rejects it. Suggest one exported helper next to `pipeline.NormalizeDMChannelID` — the function that *creates* the format — with all three delegating to it.

### P2-2 — `strings.Split` diverges from the `SplitN(..., 2)` convention used everywhere else

`internal/service/summary.go:131`:

```go
parts := strings.Split(sourceID, "@")
if actorID == "" || len(parts) != 2 || ... {
```

`pipeline.NormalizeDMChannelID` (`internal/pipeline/fetch.go:328-330`) splits on the **first** `@` via `IndexByte`, and `deriveDMPeer` matches that with `SplitN(..., 2)`. The new parser instead requires exactly two components overall.

Consequence: any UID containing `@` — e.g. an email-style UID from a directory integration — produces a canonical id like `creator@user@example.com` that the existing convention resolves and this one rejects, degrading to `来源-creator@(私聊)`. `ResolveStoredSourceName` won't repair it either, since the exception requires `sourceNameDMPeerID != ""`.

There is also a **behavioural regression** for legacy (non-canonical) rows: before this PR a bare peer UID containing `@` was looked up directly and could resolve; now it is unconditionally routed through the pair parser and returns a placeholder when the actor is neither half.

The `extra separator` case at `internal/service/summary_dm_name_test.go:41` codifies this behaviour as expected rather than flagging it. The rest of the system does assume `@`-free UIDs, so this may be unreachable in production — but the PR makes that assumption load-bearing in a new place without asserting it. Cheap hardening that cannot regress: on peer-derivation failure, retry the lookup with the full `sourceID` before falling back to the placeholder.

### P2-3 — Self-DM ("note to self") still shows the raw placeholder

`sourceNameDMPeerID` rejects `parts[0] == parts[1]`, so a self-DM canonical id resolves to `来源-xxxxxxxx(私聊)` — exactly the symptom this PR sets out to remove. `internal/service/summary_dm_name_test.go:42` (`{"self pair", "creator@creator", "creator", "来源-creator@(私聊)"}`) encodes that as expected with no explanatory comment, yet self-DM is a first-class concept elsewhere in this codebase (`bot_summary_create.go:359-361`).

Not blocking — it degrades safely and the case is narrow — but the correct display for a self-DM is a product decision, and right now it falls out of an unexplained equality guard rather than being decided. Either resolve it to the user's own name, or add a comment at `summary.go:131` stating that self-DM is deliberately excluded and why.

### P2-4 — Same bug class left unfixed at `bot_summary_create.go:474`

```go
row := model.SummarySource{TaskID: task.ID, SourceType: source.SourceType, SourceID: source.SourceID,
    SourceName: service.ResolveSourceNameWithType(source.SourceID, source.SourceType, h.imDB)}
```

`source.SourceID` here is already canonical — `resolveAuthorizedBotSources` sets it via `canonicalBotSourceID` → `pipeline.NormalizeDMChannelID` (`bot_summary_create.go:305`, `:393`). So bot-created DM summaries persist precisely the placeholder this PR eliminates elsewhere, and the correct actor `ownerUID` is already in scope on the same line (it is used for `CreatorID` at `:462`).

Mitigating: the read paths repair it, because the creator is by construction a party to the DM there, so list/detail/share render correctly. That is why this is P2 and not P1. What is **not** repaired is the persisted value feeding server-side consumers — see P2-6.

One-line fix: `service.ResolveSourceNameForActor(source.SourceID, source.SourceType, ownerUID, h.imDB)`.

### P2-5 — Scheduled-task write paths likewise actor-less

`internal/worker/scheduled_replace_helpers.go:77` and `:241` both call the actor-less resolver. `summary_schedule.source_config` stores `req.Sources` verbatim and un-canonicalized (`internal/api/handler/schedule.go:555-558`), so a schedule created from the unified workspace carries canonical DM ids and every scheduled run writes the placeholder. `task.CreatorID` is already in scope in the same file (`:133`, `:154`, `:271`) and at `scheduler.go:196`, so threading it through is mechanical.

Same mitigation as P2-4 (read paths repair the display), hence P2.

For completeness: `internal/worker/source_backfill.go:113` is also actor-less but has no impact — it writes `Derived: true` rows, which are filtered out of list (`task.go:686`), share (`share.go:404`) and `explicitSpecifiedSources` (`internal/worker/specified_sources.go:24`).

### P2-6 — Read-time repair does not reach server-side consumers

`ResolveStoredSourceName` is wired into the HTTP display paths only. Two consumers read `SummarySource.SourceName` straight from the DB:

- `internal/worker/specified_sources.go:30` and `:50` → `sourceNameForGeneration` → `internal/worker/personal_processor.go:957` → `p.llm.CallMapStreamWithModel(..., sourceName, ...)`. The placeholder becomes the source label inside the LLM prompt.
- `internal/api/handler/agent_reference_context.go:335` — the placeholder enters the agent's referenced-summary context text.

Net effect for rows written by the paths in P2-4/P2-5: the UI shows `momo_test(私聊)` while generation and reference context still see `来源-creator@(私聊)`. Fixing P2-4 and P2-5 closes this for new rows; historical rows would need the backfill the PR deliberately declines.

### P2-7 — Read-path repair is permanent and un-batched

Because nothing is written back, every affected row pays a single-row `imDB` query on **every** list/detail/share render, indefinitely. `ListSummaries` already issues one `summary_source` query per task (`task.go:684`); this adds up to one IM query per placeholder DM source on top, bounded by `page_size` (default 20, max 100 — `task.go:448-453`).

For a user whose history is mostly workspace DM summaries — exactly the population this bug created — that is 20–100 extra sequential round-trips to a second database per list page. It is bounded and does not affect correctness, so P2, but it is a permanent cost rather than a transitional one and the PR body does not mention it. Two options, neither needing a migration:

1. Memoize `uid → name` for the duration of one list request. DM peers repeat heavily across a page, so this collapses most of the cost for a few lines.
2. Or collect the page's peer UIDs and issue one `WHERE uid IN (...)`.

### P2-8 (informational, no change requested) — placeholder detection is string equality

`internal/service/summary.go:149-152` identifies a bug-placeholder by `source.SourceName == fallbackSourceName(source.SourceID, source.SourceType)`. There is no provenance flag, so in principle a real display name could collide and lose its snapshot semantics.

Adjudicating this rather than just listing it: the colliding name would have to be exactly `来源-` + the first 8 characters of the canonical DM id it appears in — i.e. self-referential to the row it lives in. Reaching that requires deliberate construction, and the only robust alternative (a provenance column) needs the migration this PR intentionally avoids. The narrow exact-match design is the right tradeoff here; `internal/service/summary_dm_name_test.go:70` (`similar name`) already pins that near-misses are preserved. Noted for the record only.

### Informational — creator-perspective resolution widens the audience for the peer's name

Using the creator rather than the viewer is correct. The consequence is that team participants, and anyone holding a share link (`share.go:401-403` notes shares are a broader audience than participants), now see the creator's DM counterparty's real name where they previously saw `来源-xxxxxxxx(私聊)`.

This restores pre-workspace behaviour — the legacy path stored the peer UID and resolved it actor-lessly — and sharing a DM summary already discloses the DM's content, so it is not a new leak. Flagged only so it is a conscious decision.

---

## 3. Overall verdict

**APPROVE**

Spec ✅ and Quality Approved. The root-cause analysis is correct and specific, the blast radius is properly contained (lookup key only, no id rewrite, no migration, no retrieval or authorization change), the degradation paths all fail safe rather than guessing a participant, and the tests are mutation-proven rather than merely present.

The residual items are all P2. The two most worth a follow-up are the unfixed sibling write paths (P2-4, P2-5 — both one-line, actor already in scope) and the parser triplication (P2-1, P2-2). Neither blocks this merge; the read-path repair means users see correct names either way.

## 4. Suggested follow-ups

1. Thread the actor into `bot_summary_create.go:474` and `scheduled_replace_helpers.go:77,241` — one line each, `ownerUID` / `task.CreatorID` already in scope. Closes the generation-prompt inconsistency in P2-6 for new rows.
2. Collapse `sourceNameDMPeerID`, `summaryWorkspaceDMPeerID` and `deriveDMPeer` into one exported helper beside `pipeline.NormalizeDMChannelID`, and pick one self-DM semantic deliberately. Add a test that pins the parse of a canonical id produced by `NormalizeDMChannelID` round-trip.
3. Add a per-request `uid → name` memo in `ListSummaries` (P2-7). Small, self-contained, no schema change.
4. Decide and document the intended self-DM display (P2-3), even if the decision is "keep the placeholder".

## 5. Additional observations outside the task scope

- **Legacy agent-save path trusts client source ids.** `applyWorkspaceScopeToSaveRequest` only runs when `workspaceSave` is true; the legacy `POST /summaries/agent` path persists `req.Sources` as sent (`agent_summary.go:649`) with no channel-access check. Combined with DM name resolution this lets an authenticated caller resolve an arbitrary UID to a display name by submitting `<self>@<target>`. **This predates the PR** — the old code resolved a client-supplied bare UID the same way, so exposure is unchanged and this is explicitly not a finding against this diff. Worth a separate look at whether the legacy path should validate sources against `pipeline.GetUserChannels` the way the bot path does at `bot_summary_create.go:313-395`.
- `docs/dm-source-name-regression.md:33-36` records local-only verification on port 28140. Reads accurate and appropriately scoped — worth confirming the same before/after check on test once deployed, since the read-path repair is the part that carries the historical rows.

## 6. Review coverage

What this review did **not** cover, so the gaps are explicit:

- No live IM database, so production UID format is unconfirmed. Whether `@` can appear in a real UID (P2-2) and how many historical placeholder rows exist (P2-7 cost) are both unverified.
- Only the `DM|SourceName` test subsets were executed, plus `go build ./...`. The full suite and the `Dockerfile.api` build were not re-run — both are claimed green in the PR body.
- Query counts were not instrumented; the P2-7 figures are derived from reading the loops and `page_size` bounds, not measured.
- Frontend rendering of `source_name` lives outside this repository and was not checked.
- The `internal/worker` package tests were not run.

Jerry-Xin (approved):
## Verdict rationale

Independent byte-level verification on the checked-out head, plus executable probes. The claimed defect is real, the fix is correct and narrowly scoped, and every fix line is pinned by content-level tests.

### 1. Defect reproduction at merge-base (confirmed)

Probe against the merge-base tree (in-memory IM fixture with both participants present):

- `ResolveSourceNameWithType("creator@peer-user", SourceDirect, db)` → `来源-creator@(私聊)` — the canonical `uid_a@uid_b` ID is looked up verbatim as a single `user.uid`, misses, and degrades to the truncated placeholder. This is exactly the reported `来源-xxxxxxxx(私聊)` symptom.
- Legacy single-UID form (`"peer-user"`) resolves to the real name at merge-base — consistent with the report that the old (non-workspace) entry was unaffected.
- The workspace arc introduced in #238 canonicalises DM chat IDs (`canonicalizeSummaryWorkspaceContextForActor` → `pipeline.NormalizeDMChannelID`, CRC32-ordered), so the save path feeds canonical IDs into the old resolver. Root-cause narrative in the PR body and `docs/dm-source-name-regression.md` matches the code.

### 2. Fix correctness

- `ResolveSourceNameForActor` keeps the lookup key change only; persisted/retrieval `source_id` stays canonical (asserted byte-level in tests). Both CRC32 orderings resolve; the legacy UID form keeps working.
- Degradation matrix is fail-safe and pinned: malformed IDs (missing part, extra `@`, self-pair), actor-less calls, creators not in the conversation, unknown users, whitespace-only IM names, and nil IM DB all fall back to the deterministic placeholder — no guessing of a participant.
- `ResolveStoredSourceName` read-side repair is correctly narrow: only empty names and byte-exact canonical-DM placeholder matches (type=DM + creator is a party + stored name == `fallbackSourceName(...)`) are re-resolved. Real snapshots, similar-but-different placeholders, legacy-form placeholders and group/thread rows are preserved. Reads do not rewrite persisted rows (asserted).
- Creator perspective (not viewer) is used everywhere, so the displayed name is stable across viewers; the participant-viewer case is explicitly pinned. Client-supplied display names remain untrusted on the save path (test seeds an untrusted name and asserts the resolved row).

### 3. Consumer census (whole tree at head)

- All `displaySourceName` call sites updated with the creator ID: list (task.go:693), detail (task.go:967), share snapshot creation (share.go:411 — uses the task creator, not the sharer; snapshot semantics preserved).
- Write paths routed through the actor-aware resolver: workspace save (agent_summary.go:662) and workflow persist (summary_workflow.go:461), the latter also covering the legacy instant path (`CreateFromLegacyHTTP` → `persist`, creatorID falls back to the authenticated actor).
- Worker backfill cannot hit the gap: DM channels are never backfilled (R9 P1 guard in source_backfill.go).

### 4. Mutation probes (each fix line individually reverted → corresponding tests go RED)

1. Disable canonical `@` handling in the resolver → `TestResolveSourceNameForActor_DM`, `TestResolveStoredSourceName_DMSnapshots`, `TestSummaryWorkflowCanonicalDMSourceName` all red.
2. Remove the stored-placeholder exception → snapshots test red + historical list/detail red (both viewers) + share snapshot red.
3. Revert workspace save to the actor-less resolver → `TestCreateAgentSummary_WorkspaceSaveCanonicalDMSourceName` red at the HTTP layer, stored row shows the placeholder.
4. Revert workflow persist → personal and team subtests red.
5. Drop creator plumbing on list/detail reads → historical test red for both viewers.

Full suites green at head: `internal/service`, `internal/api/handler`, `internal/worker` (sqlite in-memory fixtures; the only touched IM query is a single-column parameterised `uid` lookup — no dialect-sensitive SQL, no collation mixing, unchanged query shape vs merge-base).

### Non-blocking findings

- **P2 — write-side unification is incomplete (pre-existing, outside the claimed scope).** `bot_summary_create.go:474` and `scheduled_replace_helpers.go:77,241` still resolve names via the actor-less `ResolveSourceNameWithType`. Bot-created tasks normalise DM sources to canonical IDs against the owner (who is also `CreatorID`), and schedule configs persist client-supplied source IDs un-normalised — so those rows will keep storing the exact placeholder. The read-side repair covers all user-visible chips (list/detail/new share snapshots) because the creator is by construction a party, but consumers that read the stored name directly bypass the repair: `sourceNameForGeneration` (personal_processor.go:957, generation label), `explicitSpecifiedSources` (specified_sources.go:30, fetch-constraint payload) and the reference-context builder (agent_reference_context.go:335). This is not a regression (merge-base behaviour is identical) and the PR/docs claims are scoped accurately, but a follow-up routing these three write sites through `ResolveSourceNameForActor` with the task creator would make the stored data consistent with the new save paths.
- **Observation — read-path lookup volume.** The repair adds one single-row IM lookup per matching row on every list/detail/share-create request. It is bounded to empty-name and exact-placeholder rows (same envelope as the pre-existing empty-name fallback) and hits an indexed `uid`, so this is acceptable; just worth remembering if list payloads ever grow large numbers of placeholder rows.
- **Nit — stale log prefix.** summary.go:76 and :98 still log `ResolveSourceNameWithType:` after the rename to `ResolveSourceNameForActor`.

APPROVE. The reported defect reproduces at merge-base, the fix is correct, conservative about historical data, and every changed line is pinned by content-level tests that go red on reversion.

mochashanyao (approved):
[Octo-Q · automated review]

---

PR: https://github.com/Mininglamp-OSS/octo-smart-summary/pull/239
Head SHA: 5bfa0340a38eae9b65cf157cdb848379904a9966

# Code Review — PR #239 (octo-smart-summary)

## Summary

The unified summary workspace canonicalises DM sources to `uid_a@uid_b`, but the name resolver kept querying the IM `user` table with the whole conversation id, so every workbench DM summary persisted and displayed the placeholder `来源-xxxxxxxx(私聊)`. This PR threads a trusted actor (the task creator) through resolution: `ResolveSourceNameForActor` derives the counterparty from the canonical pair, `ResolveStoredSourceName` repairs already-persisted rows at read time only when the stored name is empty or byte-identical to that placeholder, and list/detail/share now resolve from the creator's perspective rather than the viewer's. Source ids, message retrieval and permission checks are untouched; no migration, no history rewrite.

The fix is correct on the paths it covers, and the tests drive real HTTP handlers rather than only the helper. My concerns are all about reach — the paths it does not cover.

## Verification

Static analysis only at head `5bfa0340a38eae9b65cf157cdb848379904a9966`; build and tests not executed in this environment.

- ✅ **Peer derivation fails closed** — `sourceNameDMPeerID` (`internal/service/summary.go:131-143`) rejects an empty actor, a part count other than two, either part blank, a self-pair, and an actor matching neither half. Each returns `""`, which the DM branch turns into the deterministic placeholder instead of a guess.
- ✅ **Actor-less API is output-identical** — `ResolveSourceNameWithType` delegates with `actorID=""`, and for a canonical id that yields exactly the same `fallbackSourceName` the old `WHERE uid = 'a@b'` miss produced, so untouched callers neither improve nor regress.
- ✅ **Repair cannot rewrite a real snapshot** — the exception in `ResolveStoredSourceName` (`internal/service/summary.go:148-156`) needs `SourceType == model.SourceDirect`, a resolvable creator, and an exact match against `fallbackSourceName`. Real names, legacy single-UID rows and group/thread placeholders all return as-is, and every read path only projects.
- ✅ **Creator is populated at each repaired site** — `ListSummaries` selects `sub.*` over `SELECT * FROM summary_task` (`internal/api/handler/task.go:540`) into an embedded `model.SummaryTask` (`:588-596`); `GetSummary` loads the full row via `authorizeTaskAccess` (`:197-201`); `share.go` via `resolveTask`; `agent_summary.go` sets `task.CreatorID = userID` before the source loop. No column-limited projection silently yields an empty creator.
- ✅ **Write-time and read-time actors agree** — `createFromAgent` sets `CreatorID: in.ActorID` (`internal/service/summary_workflow.go:283`) and both routes funnel into the same `persist` that resolves with `in.creatorID` (`:461`), the value written to `task.CreatorID` (`:437`).
- ✅ **Tests hit the production path** — `internal/api/handler/source_name_dm_test.go` drives the detail, list, share and workspace-save endpoints, asserts the creator perspective for both a creator and a participant viewer, and asserts the stored row is not rewritten. Its new `"user"` table does not collide with `setupSourceNameImDB` / `setupListTestDBs` / `setupShareTest`.

## Findings

No P0/P1 issues; three P2 items and one nit below.

### P2 — Two source-name write paths still persist the bug's placeholder (`internal/api/handler/agent_summary.go:662`)

Only two of the four non-derived `summary_source` writers were migrated. `internal/api/handler/bot_summary_create.go:474` still calls `service.ResolveSourceNameWithType`, and its DM sources are *always* canonical — `canonicalBotSourceID` (`:305`) runs `pipeline.NormalizeDMChannelID(sourceID, ownerUID, ChannelTypeDM)`, which always returns the `uid_a@uid_b` form (`internal/pipeline/fetch.go:323-338`). With `actorID==""` the resolver falls through to `fallbackSourceName`, so bot-created DM summaries keep writing exactly the placeholder this PR removes; the scheduled writers at `internal/worker/scheduled_replace_helpers.go:77` and `:241` have the same shape. Read-time repair hides this in list/detail/share, but two consumers read `SourceName` raw — the reference-context bullet at `internal/api/handler/agent_reference_context.go:335`, which its own comment says is visible to any task reader, and `sourceNameForGeneration` (`internal/worker/specified_sources.go:50`), feeding the generation prompt at `internal/worker/personal_processor.go:957`. Both writers already hold the right identity (`ownerUID`; the schedule task's `CreatorID`).

### P2 — Legacy HTTP route feeds a client-supplied identity into the "trusted" actor (`internal/service/summary_workflow.go:461`)

The new doc comment (`internal/service/summary.go:47-50`) says the actor must be the trusted source owner, never a client-supplied value. That holds for the agent route but not for `CreateFromLegacyHTTP`: `CreateSummary` sets `CreatorID: req.UID` from the request body (`internal/api/handler/task.go:359`) and `normalize` falls back to `ActorID` only when it is empty (`internal/service/summary_workflow.go:348-350`). On `POST /api/v1/summaries` a caller therefore picks which half of the pair counts as "self" — `uid=V` with `sources=[{source_type:3, source_id:"V@W"}]` resolves and persists W's real IM display name. The override is pre-existing and a plain `source_id=W` already resolved W's name before this PR, so this amplifies an existing hole rather than opening a new class of disclosure — not a change I'd block on. Deriving the actor from `middleware.GetUserID(c)` on this route, or validating `req.UID` against it, would make the comment true; `docs/dm-source-name-regression.md`'s "以可信的总结创建者身份解析" should say which routes guarantee it.

### P2 — Repair puts an uncached per-row IM query on the polled list path (`internal/api/handler/task.go:693`)

`ResolveStoredSourceName` re-resolves any row whose stored name equals the canonical placeholder, so the pre-PR short-circuit on non-empty names no longer covers those rows. In `ListSummaries` that runs inside the per-task × per-source loop: each affected row issues its own `SELECT name FROM user WHERE uid = ?` against the separate IM database — up to `page_size` (default 20, max 100, `internal/api/handler/task.go:448-453`) times sources per request, uncached, with no statement timeout. The affected row set is the bug's whole blast radius (every workbench DM summary persisted since the canonical-id change) on the endpoint the sidebar polls. Empty-name rows already paid this cost before the PR, so the pattern is pre-existing and only widened, which is why I would not block. Batching a page's peer UIDs into one `WHERE uid IN (?)`, or memoizing per request, keeps the repair off the hot path.

### Nit — DM branch trims for the guard but not for the return (`internal/service/summary.go:125`)

The guard was tightened to `strings.TrimSpace(name) != ""` (`:124`) but the value returned is still the untrimmed `name + "(私聊)"`, so an IM row stored as `" Alice "` renders with its padding. Returning `strings.TrimSpace(name) + "(私聊)"` matches the intent of the changed guard.

## Human-verify

1. The repair newly surfaces the counterparty's real display name to an audience wider than the DM: every task participant reads `source_name` on list/detail, and `internal/api/handler/share.go:411` bakes it into a snapshot the surrounding comment describes as readable by anyone with the link. Before this PR those rows always rendered the opaque placeholder. The raw `source_id` (`uid_a@uid_b`) was already emitted beside it and the DM's summarized content is already in the snapshot, so the incremental disclosure looks small and the creator-perspective choice is clearly deliberate — worth a product confirmation that naming a private-chat counterparty to invited participants is intended. **Not a merge blocker for this PR.**
2. The PR body and `docs/dm-source-name-regression.md` report `go test ./internal/service ./internal/api/handler ./internal/worker` and a `Dockerfile.api` build as passing under CGO with `libtokenizers`. Those cannot be reproduced here and are taken on the author's word. **Not a merge blocker for this PR.**

- `pipeline.ApplySourceConstraints` matches on `source_id`/`source_type` only, so the raw `source_name` passed by `internal/worker/specified_sources.go:30` cannot influence channel selection.
- `internal/worker/source_backfill.go:75-77` skips DM channels outright, so its untouched resolver call at `:113` cannot produce a canonical-DM placeholder; those rows are `Derived: true` and excluded from user-visible projections anyway.
- Placeholder matching is exact Go string comparison (no collation or trailing-space widening) and `model.SourceDirect == 3` matches the resolver's `case 3`.
- Existing share snapshots keep their stored `SourceName`; only new snapshots resolve — matching the stated "既有分享快照不改写" boundary. The `source_name_display_test.go` edit only relaxes a comment from "NEVER overwritten" to the new scoped exception; no assertion was weakened.

## Verdict: COMMENT

The core fix is sound: peer derivation fails closed, the read-time repair cannot rewrite a real snapshot or a legacy row, the creator identity is genuinely populated at every repaired call site, and the tests cover the production handlers. Nothing breaks a working path or produces wrong user-visible data, so there is no blocker. The three P2s are about reach rather than correctness — two write paths still persist the placeholder and leak it into the reference-context and generation prompts, the legacy route's "trusted" actor is client-supplied, and the repair adds uncached cross-database lookups to a polled endpoint. All are reasonable as follow-ups, though the bot and scheduled writers are cheap enough that folding them in here would make the fix complete.





---

# Octo 附加段（automated review内部证据，不进入对外 GitHub review 正文）

## 5. 数据流回溯（每个被消费数据 → 上游来源 → 是否真流到消费点）

1. **`actorID` @ `internal/service/summary.go:117`（DM 分支）** ← `ResolveSourceNameForActor` 形参 ← 4 个生产调用点：
   - `internal/service/summary_workflow.go:461` = `in.creatorID` ← `normalize`（`:348-350`）← `in.CreatorID` 否则 `in.ActorID`。两条入口：`createFromAgent`（`:283` `CreatorID: in.ActorID`，ActorID 来自已鉴权身份）→ **真流到**；`CreateSummary`（`internal/api/handler/task.go:359` `CreatorID: req.UID`，来自请求体）→ **流到但客户端可控**（见 F2）。
   - `internal/api/handler/agent_summary.go:662` = `task.CreatorID`，在 `:573` 由 `userID`（middleware）赋值，且在 tx 内 `tx.Create(&task)` 之后使用 → **真流到**。
   - `internal/service/summary.go:44` = 常量 `""`（兼容包装）→ 有意为之，输出与旧实现逐字节一致。
2. **`creatorID` @ `internal/service/summary.go:150`（占位名修复闸门）** ← `displaySourceName` ← 三个读点全部核实**非列裁剪**：
   - `internal/api/handler/task.go:693` `t.CreatorID` ← `listTaskRow` 内嵌 `model.SummaryTask`（`:588-596`）← `pageSQL = SELECT sub.* ...`（`:583`）← `innerSQL = SELECT * FROM summary_task`（`:540`）→ `creator_id` 在投影内 ✅（这是本 PR 最容易静默失效的点：若列表用了列裁剪，闸门恒 false、修复空转）。
   - `internal/api/handler/task.go:967` `task.CreatorID` ← `authorizeTaskAccess` 的 `First(&task)` 整行加载（`:197-201`）✅。
   - `internal/api/handler/share.go:411` `task.CreatorID` ← `resolveTask`，且同函数 `:363` 的 `canAccessTaskDB(..., task.CreatorID)` 已先行消费该字段 ✅。
3. **`source.SourceID` / `source.SourceType` @ `internal/service/summary.go:149-151`** ← 持久化 `summary_source` 行 ← 全仓 6 个写入点。规范 `uid_a@uid_b` 形态由 `pipeline.NormalizeDMChannelID` 保证：工作台路由 `internal/api/handler/agent_summary_workspace.go:1290/1636/1917`（selfUID=`actorID`）、bot 路由 `internal/api/handler/bot_summary_create.go:305`（selfUID=`ownerUID`）。→ 这正是"精确匹配占位名"能命中的前提 ✅；同时说明 **bot 路由必然产出规范 ID，却仍走无 actor 的解析器**（F1）。
4. **`name` @ `internal/service/summary.go:124`** ← `imDB.Table("user").Where("uid = ?", peerID).Pluck("name", &name)`。核实查询键是**对端 UID**而非整串会话 ID（即原 bug 本身）✅；`err != nil` / 命中空 / 纯空白 → 统一落 `fallbackSourceName`，错误未被吞成一个"看起来正确的错名字" ✅。
5. **`snapshot.SourceName` @ `internal/api/handler/share.go:411`** ← `strings.Join(names, "、")` ← 每条非 derived 源经 `displaySourceName`。写入后为只读快照，`Get` 路径不再二次解析 ✅（与 PR 声称的"既有分享快照不改写"一致）。
6. **`s.SourceName`（裸读，未经修复闸门）** @ `internal/api/handler/agent_reference_context.go:335` 与 `internal/worker/specified_sources.go:50` → `internal/worker/personal_processor.go:957` 的生成 prompt。对 bot / scheduled 写入的规范 DM 行，**该占位名仍会流到 prompt 与参考语境 bullet**（F1 的实际后果链）。
7. **`m.SourceName` @ `internal/worker/source_backfill.go:88/113`** ← pipeline `ChannelInfo.ChannelName`；`:75-77` 明确 `sourceType == model.SourceDirect` 直接 `continue` → 未迁移的解析器调用**不可能**看到规范 DM ID ✅（此路径判 clear，不算 parity 缺口）。

## 6. 盲点 checklist（C1–C6）

- **C1 双路径 parity — HIT（→ F1, P2）**。写路径 4 条（非 derived）：`summary_workflow.go:461` ✅迁移、`agent_summary.go:662` ✅迁移、`bot_summary_create.go:474` ❌未迁移、`scheduled_replace_helpers.go:77`/`:241` ❌未迁移；第 5 条 `source_backfill.go:113` 因 DM 短路判 N/A。读路径 3 条（list/detail/share）全部迁移 ✅，`rg` 确认无遗漏的 `displaySourceName` 旧签名调用点。兼容包装 `ResolveSourceNameWithType → ResolveSourceNameForActor(.., "", ..)` 逐分支比对，输出与 base 完全一致 ✅。守卫自身等价写法覆盖：修复闸门用**精确字符串相等**匹配占位名，而占位名由同一个 `fallbackSourceName` 生成 → 生成端与匹配端同源，不存在"守卫漏扫等价写法"的假绿。
- **C2 control-flow ordering / 嵌套复用 — clear**。`ResolveStoredSourceName → ResolveSourceNameForActor → sourceNameDMPeerID` 三层嵌套：同一行内 `sourceNameDMPeerID` 被调用两次（闸门 + 解析），入参完全相同、纯函数无副作用 → 不产生双重作用或顺序错位。非 canonical 形式试穿闸门：前导/尾随空格（`"a @b"` → `parts[0] == actorID` 精确比较失败 → 安全兜底，不误判）、多个 `@`（`len(parts) != 2` → `""`）、自配对（`parts[0] == parts[1]` → `""`）、空 actor（→ `""`）、大小写/连字符变体（精确比较 → 兜底）。全部落确定性占位，不会解析出错误的人名。
- **C3 授权边界 ≠ 能力边界 — HIT（→ F2, P2）**。解析器新增能力：把"与 actor 成对的任意 UID"翻译成 IM 真实显示名。追踪"谁能到达这个能力"：agent/工作台路由 = 已鉴权 `ActorID`（不可控）✅；legacy HTTP 路由 = 请求体 `req.UID`（**可控**）→ 调用方自选配对中哪一半算"自己"，从而指定要落库/展示哪个 UID 的真名。未新增 endpoint/tool；`canAccessTask` / `authorizeTaskAccess` / `ApplySourceConstraints` 均未被本 PR 触碰（消息检索与权限校验未变，已核实）。按 R2 判 amplified 而非 new：`task.CreatorID = req.UID` 覆盖本身是既有行为，且 base 分支下 `source_id=W`（裸 UID）已能解析 W 的名字 → 同类能力既有，本 PR 只扩展到 `a@b` 形态 → P2，不升 P1。
- **C4 授权生命周期 / 容器-成员状态级联 — N/A（附理由）**。本 PR 不改任何鉴权/成员/状态门控，不新增查询过滤，也不涉及父容器（space/schedule）禁用时的授权级联。IM `user` 表查询无 status/deleted 过滤，但这一点在 base 分支的裸 UID 分支上完全相同（`internal/service/summary.go:123` 与旧实现同形），属未触碰的既有语义，掩码/授权语义无变化。
- **C5 build/note 通过 ≠ 运行期路径正确 — clear（含诚实边界）**。无打包/相对 URL/heredoc/扩展/环境差异改动。本次**未执行** `go build` / `go test`，因此不以"作者说测试通过"作为缓解；改为静态核实运行期可编译性与真实路径：`displaySourceName` 全部 3 个调用点均已改为 3 参、全仓 `rg` 无残留 2 参调用；`internal/api/handler/task.go` 仍使用 `strings` 12 处、`gorm`/`service`/`model` 导入仍被需要；`//go:build cgo` 标签与同目录既有测试一致；新测试依赖的 helper（`setupSourceNameImDB`/`setupListTestDBs`/`setupShareTest`/`seedWorkspaceSaveFixture`/`newSummaryWorkflowTestService`）均存在，且它们都不建 `"user"` 表 → `seedSourceNameDMUsers` 的 `CREATE TABLE "user"` 不会撞表。作者的 `go test` / `Dockerfile.api` 结论记入 Human-verify #2，不作为定级依据。
- **C6 治理/策略/安全文档自洽性 — HIT（轻微，并入 F2）**。`docs/dm-source-name-regression.md` 与代码逐条比对：「只对空名称及精确匹配的规范私聊占位名称实时解析」= `ResolveStoredSourceName` 实际语义 ✅；「不改消息检索与权限校验」= diff 未触碰相关文件 ✅；「不改写数据库历史行、不新增迁移」= 无 migration 文件、读路径只投影 ✅；回归覆盖 1–5 条与两个新测试文件的用例逐条对得上 ✅。**一处过度声明**：「以可信的总结创建者身份解析」对 legacy HTTP 路由不成立（`req.UID` 客户端可控）→ 已在 F2 内要求同步修正文档措辞。未触碰 SECURITY.md / label 策略 / 披露流程，无治理冲突或泄露口。

### R3 分布式/多实例副作用 checklist
- **多客户端 fan-out**：N/A — 无通知/推送/事件 emit/计数/缓存/广播改动；分享快照创建仍由既有 `requestHash` + `Idempotency-Key` 幂等保护（未被本 PR 触碰）。
- **observer ≠ actor**：**HIT，且正是本 PR 修对的轴** — 名称解析基准从"当前查看者"改为"任务创建者"，三个读点（`task.go:693`、`task.go:967`、`share.go:411`）已逐一核实传入的是 `CreatorID` 而非 `middleware.GetUserID(c)`；新测试用 creator 与 participant 两种 viewer 断言结果一致 ✅。
- **调度/无界面触发**：**HIT（→ F1）** — 后台 worker 路径 `scheduled_replace_helpers.go` 不依赖任何客户端观察者，仍走无 actor 解析器，规范 DM 行在无人界面触发时继续写入占位名。

## 2/3/4. 定级、建议与额外发现

### diff-scope 三问 + R1/R4 定级
| # | finding | new / amplified / pre-existing | R1 检验（能否让可用路径不可用 / 产出用户可见错误数据 / 比修复前更糟） | 定级 |
|---|---|---|---|---|
| F1 | bot + scheduled 写路径未迁移，占位名仍落库并流入 prompt / 参考语境 bullet | **pre-existing**（两文件未被触碰；输出与 base 逐字节相同） | 否 — 没有任何路径比修复前更糟；读路径修复反而让 list/detail/share 变好。属"修复不完整"而非"引入回归" | **P2** |
| F2 | legacy HTTP 路由把客户端 `req.UID` 当作"可信 actor" | **amplified**（`:461` 新消费了一个客户端可控值；覆盖机制与裸 UID 名字预言机均既有） | 否 — 泄露类别既有（`source_id=W` 已可解析 W 名字），本 PR 仅扩展到 `a@b` 形态；未让任何路径不可用 | **P2** |
| F3 | 修复闸门把未缓存的跨库 IM 查询放上被轮询的列表热路径 | **amplified**（空名行既有同型 N+1；本 PR 扩大了命中行集合） | 否 — 常规运行下仅增加延迟，不会让列表不可用；仅当 IM 库劣化时才放大，属既有风险面 | **P2** |
| F4 | DM 分支 guard 已 TrimSpace、返回值未 TrimSpace | **new**（改动行本身） | 否 — 仅多保留 IM 库中的首尾空格，纯外观 | **Nit** |
| F5 | 修复后私聊对端真名对全体参与者与分享链接可见 | **new**（占位名 → 真名） | 否 — 同对象中 `source_id`（含双方 UID）与该私聊的总结正文**早已**对同一受众可见，增量披露极小；且创建者视角是 PR 明确的设计选择 | **Human-verify（非阻塞）** |

**R4 映射结论：无 P0/P1 → automated review建议 APPROVE。**（按 output-format 规则，存在 P2 时 skill 侧 `review-verdict` 记为 `COMMENT`；两个 marker 语义不同，不冲突：`automated review-verdict` 按 R4「仅 P2/nit → APPROVED」映射，`review-verdict` 按 skill「有非阻塞 finding → COMMENT」。）

### 建议（具体到行）
1. `internal/api/handler/bot_summary_create.go:474` → `service.ResolveSourceNameForActor(source.SourceID, source.SourceType, ownerUID, h.imDB)`（`ownerUID` 已在同函数签名内）；`internal/worker/scheduled_replace_helpers.go:77` 与 `:241` → 把所属 task 的 `CreatorID` 透传进 `buildScheduledTaskSources` / `syncScheduledTaskSources` 并作为 actor 传入。补一条回归：bot 创建规范 DM 源后断言落库名为 `momo_test(私聊)` 而非占位名。
2. `internal/api/handler/task.go:359` → 解析 actor 改用 `userID`（middleware），或在入口处校验/忽略 `req.UID != userID`；随后把 `docs/dm-source-name-regression.md` 的「以可信的总结创建者身份解析」改成明确区分 agent 路由与 legacy 路由。
3. `internal/api/handler/task.go:678-696` → 在投影循环前收集本页所有 DM 对端 UID，用一次 `WHERE uid IN (?)` 批量取名（或 per-request memoize `uid → name`），把修复从热路径上摘下来。
4. `internal/service/summary.go:125` → `return strings.TrimSpace(name) + "(私聊)"`。
5. 可选：若 F5 经产品确认需要收敛，最小改动是在 `ResolveStoredSourceName` 的占位名修复分支上再加一个"查看者 == 创建者 或 查看者是该 DM 成员"的条件；但请先确认，不要默认改（当前行为是 PR 明确设计）。

### 额外发现（不计入定级）
- `internal/service/summary.go:65` 群聊分支仍用 `name != ""`，而 DM 分支已改为 `TrimSpace`；纯空白的群名会渲染成 `" (群聊)"`。既有、未触碰、纯外观。
- `fallbackSourceName` 按**字节**切 `sourceID[:8]`，规范形态下占位名会内嵌创建者 UID 前缀（如 `来源-creator@(私聊)`），在所有未修复投影中可见。既有行为，仅提示。
- 文档记录的本地 28140 验证（`来源-7ecf56dc(私聊)` → `总结参与者(私聊)`）与修复后的代码路径完全对得上，是有力的单点证据；但它是单行验证，覆盖不到 F1 的 bot / scheduled 写路径。

## 7. 跨轮 blocker 复检（R6）
**N/A** — 本 issue 评论历史为空（`Octo-Q issue comment list` 返回 `[]`），本次是 head `5bfa0340a38eae9b65cf157cdb848379904a9966` 上的第一轮审查，不存在上一轮未解决 blocker 需要复检。本报告即针对该 head SHA，merge-base 取 `origin/main` = `391134cc8c25226287e58a26204616d42a200e6d`（`git merge-base` 实测，非推测）。若后续 re-push 产生新 head，需重新给 formal verdict。

## 产物与环境说明
- 结构化 finding：`/tmp/rv-239-5bfa034/review.final.json`（5 条，`finalize` 校验 `validation_valid=true`、`invalid_comments=[]`、`missing_high_risk_families=[]`）。
- 对外 review 正文：`/tmp/rv-239-5bfa034/review.body.md`。
- ⚠️ 未写入 skill 默认的 `/tmp/review.final.json` / `/tmp/review.body.md`：本机 `/tmp` 被多个并发 review run 共用，本次运行中这两个默认路径已被另一 run（内容为 `octo-admin` / `deliveryrouting.go` 的审查）覆盖。为避免污染在飞的其它 PR 审查产物，本次改用 PR 专属目录。**请终审从上面两个专属路径或本 comment 正文取材，不要读默认路径。**
- 审查环境：`pr-base` ref = merge-base `391134c`，`HEAD` = `5bfa034`，`git diff pr-base...HEAD` 覆盖全部 9 个变更文件；coverage ledger 报 `requires_inspect: 0`、`Needs expanded review: false`，9/9 单元已在 brief 内 inspect 完毕。

[Octo-Q] verdict: APPROVE — 按 R4：全部 finding 为 P2/Nit/Human-verify，无 P0/P1。核心修复正确（对端推导 fail-closed、读时修复闸门精确到「空名 或 与 fallbackSourceName 逐字节相等的规范私聊占位名」、三个读点传入的都是 CreatorID 而非 viewer、列表投影经 SELECT sub.* 确认 creator_id 真流到消费点），未让任何既有可用路径变得不可用、也未产出比修复前更糟的用户可见数据。三个 P2 均为「修复覆盖面」而非「正确性」问题：bot/scheduled 两条写路径未迁移（占位名继续落库并流入参考语境 bullet 与生成 prompt）、legacy HTTP 路由把客户端 req.UID 当作可信 actor（既有预言机的放大，非新泄露类别）、读时修复把未缓存跨库查询放上被轮询的列表热路径（既有 N+1 的扩大）。建议合并，P2 作为 follow-up；其中 bot/scheduled 两行改动成本极低，若愿意在本 PR 内补齐可让修复真正闭环。

yujiawei:
## Addendum — three items from a further pass (verdict unchanged: APPROVE)

All three are non-blocking. Two are actionable one-liners; the third is a correction to my own reasoning above.

### P2 — the legacy route feeds a client-supplied value into the new "trusted actor"

The doc comment introduced by this PR at `internal/service/summary.go:47-50` states the actor "must be the trusted source owner (the task creator), **never** the current viewer or a client-supplied display name". That holds for the workspace/agent route, but not for `POST /api/v1/summaries`:

```go
// internal/api/handler/task.go:358-360
workflowInput := service.LegacyCreateSummaryWorkflowInput{
    ActorID:   userID,    // authenticated
    CreatorID: req.UID,   // request body
```

`req.UID` appears exactly once in that file and is never validated against `userID`, and `normalize` prefers it whenever it is non-empty (`internal/service/summary_workflow.go:348-350`), so it is the value that reaches `ResolveSourceNameForActor` at `:461`. A caller can therefore choose which half of a canonical pair counts as "self": `uid=V` with `sources=[{source_type:3, source_id:"V@W"}]` resolves and persists W's real IM display name.

Worth being explicit that this is **not a new disclosure class and not a reason to hold the PR**: before this change, `source_id=W` (a bare UID) already resolved W's name through the same DM branch regardless of `req.UID`. This widens the shape of an existing hole rather than opening one. `LegacyCreateSummaryWorkflowInput`'s own comment ("CreatorID preserves its historical uid override") shows the override is deliberate and predates this work.

The reason to mention it at all is that the PR adds a comment asserting a trust property that one of its four call sites does not satisfy. Either derive the actor from the authenticated identity on that route, or soften the comment to say which routes guarantee it — a comment that is false on one path is worse than no comment. `docs/dm-source-name-regression.md`'s "以可信的总结创建者身份解析" has the same issue.

### Nit — the DM branch trims for the guard but not for the returned value

`internal/service/summary.go:124-125`:

```go
if err == nil && strings.TrimSpace(name) != "" {
    return name + "(私聊)"
}
```

The guard was tightened to `TrimSpace` **in this diff**, but the returned value is still the untrimmed `name`, so an IM row stored as `" Alice "` renders as `" Alice (私聊)"`. `return strings.TrimSpace(name) + "(私聊)"` matches the intent of the changed line.

(Pre-existing, untouched, mentioned only for context: the group branch at `:65` still uses the un-trimmed `name != ""` guard, so a whitespace-only group name renders as `" (群聊)"`.)

### Correction to my own review above

Under P2-5 I wrote that `internal/worker/source_backfill.go:113` has no impact because it writes `Derived: true` rows that are filtered from user-visible projections. That is true but it is not the operative reason, and I should have given the stronger one: `:75-77` skips DM sources outright —

```go
// R9 P1: DM channels are never backfilled (see doc comment).
if sourceType == model.SourceDirect {
    continue
}
```

so that call site can never observe a canonical DM id in the first place. The conclusion (no impact, not a parity gap) stands; the reasoning I published was weaker than the code warrants.

PR #232, opened 2026-08-27
feat(obs): run + attempt latency histograms, configurable REFINE_TIMEOUT

wanyiwang06:
## Problem

`llm_call_duration_seconds_total` (from #219) is a cumulative sum. Divided by `llm_calls_total` it yields a mean, and a mean over this workload describes no real request: refine calls are sub-second while long-context agent turns are documented at 60–100s in `internal/agent/llm.go`.

#220 defers the final `LLM_TIMEOUT` value to per-scenario P95/P99 ("不直接规定 `LLM_TIMEOUT=60`，最终值应结合各场景 P95/P99 决定"). Those percentiles require bucket counts, which no metric in the repo currently produces.

## Relationship to #231

This PR is a **partial delivery of #231 P1 (metrics), scoped to the LLM latency slice**. It does not close #231 and must not be merged as if it did.

What #231 asks for, and where this PR lands against it:

| #231 | This PR |
|---|---|
| §5 metric inventory — LLM latency distributions | ✅ `llm_run_duration_seconds`, `llm_attempt_duration_seconds` |
| §6 cardinality red line (labels from a finite vocabulary) | ✅ labels are `path` / `outcome` only; label-set pins in the tests |
| §7 exposure on the internal port, not the public router | ✅ both histograms land on the existing `/internal/metrics` |
| §9 a test asserting the emitted label key set | ✅ label-set pins on both histograms |
| §1 `internal/obs` OTEL facade, `OTEL_*` config keys | ❌ not here |
| §2 `RunTrace` → span tree | ❌ not here |
| §3 OTEL inside `internal/timing` (~39 call sites) | ❌ not here — called out under *Deliberately out of scope* as S3 |
| §4 cross-process trace propagation (`otelhttp`, W3C `traceparent`) | ❌ not here |
| §8 collector deployment + tail sampling | ❌ not here |

**Implementation approach differs from #231's primary plan, deliberately.** #231 proposes OpenTelemetry and notes the risk that `go.mod` gains a dozen-plus `go.opentelemetry.io/*` modules. This PR adds **no new Go modules** — the histograms are hand-rolled into the existing `internal/llmobs` registry. That is #231's own *Alternative considered*, which it accepts as "a valid fallback for P1" precisely because the facade lets the exporter be swapped later without touching call sites. Choosing it here keeps a security-adjacent dependency surface unchanged while still unblocking the percentile question.

[...]

Issue #231, opened 2026-08-27
feat(obs): adopt OpenTelemetry for agent + worker — metrics always on, traces sampled


wanyiwang06:
## Problem

Both execution paths now have timing instrumentation, but neither produces anything you can put on a dashboard or alert on.

- **Agent path** (#216 / PR #215): `internal/agent/trace.go` records per-step planner latency, prompt size, tool spans and sub-phases, then emits a bounded text block via `log.Printf`. It is gated behind `AGENT_TRACE`, **off by default**.
- **Worker path**: `internal/timing` has ~39 call sites (26 in `personal_processor.go`). Every `Record()` writes to stdout **and** appends to `/var/log/smart-summary/timing.log`, plus a per-task `summary-report.log`.

Both are single-request forensics. Neither answers:

- What is the p95 of an agent request this week vs. last week?
- Is the planner prompt growing over releases?
- Which tool / pipeline stage is the current straggler across all traffic?
- Did the error rate move after a deploy?

And neither can page anyone. `AGENT_TRACE` being off by default makes it useless for intermittent production issues, because by the time you turn it on the incident is over.

There is currently **zero** metrics/tracing dependency in `go.mod`, and no `/metrics` endpoint.

## Proposal

Adopt OpenTelemetry for both paths: **metrics always on** (cheap, no log output, gives trends + alerts), **traces sampled** (gives cross-process drill-down).

### Core principle: do not touch the call sites

`timing.Record()` / `timing.Stage()` / `RunTrace.AddStep()` keep their signatures. The packages gain OTEL emission internally alongside the existing logging. This keeps the diff reviewable and the change revertible.

**Metrics are not logs.** A histogram observation is an in-memory bucket increment with no IO; the data is scraped from `/metrics`. Wiring all ~39 worker call sites adds **zero** log lines. If anything this is the opportunity to demote the `timing` stdout half to debug once Grafana carries the trend.

### 1. New `internal/obs` package (single OTEL dependency boundary)

- `obs.Init(ctx, cfg) (shutdown func(context.Context) error, err error)` — builds TracerProvider + MeterProvider with OTLP export; returns official noop providers when disabled (zero-allocation, same discipline as the nil-safe `*RunTrace`).
- Called once from each of `cmd/summary-api/main.go` and `cmd/summary-worker/main.go`, with `defer shutdown(ctx)`. The worker must flush before exit or the last batch of spans is lost.
[...]

PR #240, opened 2026-09-08
feat(timing): bound LLM-call label cardinality with PurposeClass (#231 S3)

wanyiwang06:
## What

Adds `timing.PurposeClass(purpose string) string`, a pure function that collapses the free-text `LLMCall.Purpose` into a small, closed set of stable identifiers:

`intent · map_chunk · map_single · reduce · team_reduce · post_retrieval_narrow · other`

## Why (#231 S3)

`LLMCall.Purpose` is human-readable, and one value is built as `fmt.Sprintf("Map: 分块总结 chunk#%d", idx)` — the chunk index is **unbounded**. That is fine in a log line, but the moment `Purpose` is used as a metric label it becomes a cardinality bomb: one time series per chunk index, per task, forever.

This is the one live cardinality risk flagged in #231's plan. `PurposeClass` is the seam the upcoming worker/agent metrics (#231 S3 emission, e.g. `summary_stage_duration_seconds{stage=...}`) must label on — the raw purpose stays log-only.

## Scope / dependencies

- Touches **only** `internal/timing/` — deliberately independent of PR #232 (histograms), which never touches this package. No merge conflict either way.
- Pure function + tests. No behavior change to existing logging/reporting.

## Tests

- `TestPurposeClass_ClosedSet` — every real call-site purpose maps to its class.
- `TestPurposeClass_ChunkIndexIsBounded` — 1000 distinct chunk indices collapse to exactly one class.

Closed set by design: unknown purposes fall through to `other` rather than silently widening the label space; add a case (and a test) when a new call site is introduced.

### DIFF
--- a/internal/timing/timing.go
+++ b/internal/timing/timing.go
@@ -178,26 +178,48 @@ func RecordLLMSince(taskNo, purpose string, start time.Time, tokens int) {
 // on PurposeClass, never on the raw Purpose. The chunk index stays in the log
 // only.
 //
-// The returned set is closed: intent, map_chunk, map_single, reduce,
-// team_reduce, post_retrieval_narrow, other. New purposes fall through to
-// "other" rather than silently widening the label space — add a case here (and
-// a test) when a new call site is introduced.
+// The cases below are the census of every RecordLLM/RecordLLMSince call site at
+// head. The retrieval-prep family is split per forced function rather than
+// merged into one class: the forced-function set is closed (recognize_intent,
+// extract_time_range, resolve_channel_scope, resolve_topic_target — see
+// internal/pipeline/*), so per-function classes stay bounded, and #220 needs
+// recognize_intent's latency percentile as a distinct input — merging would
+// bury it.
+//
+// The returned set is closed: intent, extract_time_range, resolve_channel_scope,
+// resolve_topic_target, retrieval_prep, post_retrieval_narrow, map_single,
+// map_chunk, reduce, team_reduce, other. New purposes fall through to "other"
+// rather than silently widening the label space — add a case here (and a test)
+// when a new call site is introduced.
 func PurposeClass(purpose string) string {
 	switch {
-	case purpose == "意图识别":
+	// Retrieval-prep tool calls: purpose is "检索预处理: " + forceFn, or
+	// "检索预处理(tool-call)" when no function is forced (see
+	// worker/{processor,personal_processor}.go). Match the closed forceFn set
+	// exactly so an unforeseen value falls through to "other" rather than
+	// silently joining a class.
+	case purpose == "检索预处理: recognize_intent":
 		return "intent"
+	case purpose == "检索预处理: extract_time_range":
+		return "extract_time_range"
+	case purpose == "检索预处理: resolve_channel_scope":
+		return "resolve_channel_scope"
+	case purpose == "检索预处理: resolve_topic_target":
+		return "resolve_topic_target"
+	case purpose == "检索预处理(tool-call)":
+		return "retrieval_prep"
+	case strings.Contains(purpose, "PostRetrievalNarrow"):
+		return "post_retrieval_narrow"
 	// Order matters: the chunk-map prefix is the unbounded one and must be
 	// matched before any looser "Map:" rule.
 	case strings.HasPrefix(purpose, "Map: 分块总结 chunk#"):
 		return "map_chunk"
-	case strings.HasPrefix(purpose, "Map: 单次总结") || purpose == "单次总结":
+	case strings.HasPrefix(purpose, "Map: 单次总结"):
 		return "map_single"
 	case strings.HasPrefix(purpose, "团队汇总"):
 		return "team_reduce"
-	case strings.HasPrefix(purpose, "Reduce"):
+	case purpose == "Reduce" || strings.HasPrefix(purpose, "Reduce:"):
 		return "reduce"
-	case strings.Contains(purpose, "PostRetrievalNarrow"):
-		return "post_retrieval_narrow"
 	default:
 		return "other"
 	}
--- a/internal/timing/timing.go
+++ b/internal/timing/timing.go
@@ -167,6 +167,42 @@ func RecordLLMSince(taskNo, purpose string, start time.Time, tokens int) {
 	RecordLLM(taskNo, purpose, time.Since(start), tokens)
 }
 
+// PurposeClass collapses a free-text LLMCall.Purpose into one of a small,
+// closed set of stable identifiers.
+//
+// Why this exists (#231 S3): Purpose is a human-readable string, and one of its
+// values is built as fmt.Sprintf("Map: 分块总结 chunk#%d", idx) — the chunk index
+// is unbounded. Purpose is fine as a log line, but the moment it is used as a
+// metric label it becomes a cardinality bomb: one time series per chunk index,
+// per task, forever. Any metric that wants to break down by call kind MUST label
+// on PurposeClass, never on the raw Purpose. The chunk index stays in the log
+// only.
+//
+// The returned set is closed: intent, map_chunk, map_single, reduce,
+// team_reduce, post_retrieval_narrow, other. New purposes fall through to
+// "other" rather than silently widening the label space — add a case here (and
+// a test) when a new call site is introduced.
+func PurposeClass(purpose string) string {
+	switch {
+	case purpose == "意图识别":
+		return "intent"
+	// Order matters: the chunk-map prefix is the unbounded one and must be
+	// matched before any looser "Map:" rule.
+	case strings.HasPrefix(purpose, "Map: 分块总结 chunk#"):
+		return "map_chunk"
+	case strings.HasPrefix(purpose, "Map: 单次总结") || purpose == "单次总结":
+		return "map_single"
+	case strings.HasPrefix(purpose, "团队汇总"):
+		return "team_reduce"
+	case strings.HasPrefix(purpose, "Reduce"):
+		return "reduce"
+	case strings.Contains(purpose, "PostRetrievalNarrow"):
+		return "post_retrieval_narrow"
+	default:
+		return "other"
+	}
+}
+
 func ensureReportFile() *os.File {
 	reportOnce.Do(func() {
 		acctMu.Lock()
--- a/internal/timing/timing_test.go
+++ b/internal/timing/timing_test.go
@@ -324,32 +324,63 @@ func TestRecordLLM_Empty(t *testing.T) {
 	}
 }
 
-// TestPurposeClass_ClosedSet asserts every real call-site purpose maps to its
-// stable class, and — the point of #231 S3 — that the unbounded chunk index in
-// "Map: 分块总结 chunk#%d" collapses to a single class rather than one label
-// value per chunk.
-func TestPurposeClass_ClosedSet(t *testing.T) {
-	cases := []struct {
-		purpose string
-		want    string
-	}{
-		{"意图识别", "intent"},
-		{"Map: 分块总结 chunk#1", "map_chunk"},
-		{"Map: 分块总结 chunk#2", "map_chunk"},
-		{"Map: 分块总结 chunk#9999", "map_chunk"},
-		{"Map: 单次总结(跳过Map-Reduce)", "map_single"},
-		{"单次总结", "map_single"},
-		{"Reduce: 最终总结", "reduce"},
-		{"Reduce: 合并分块总结", "reduce"},
-		{"Reduce", "reduce"},
-		{"团队汇总: 合并各成员总结", "team_reduce"},
-		{"检索后裁剪 PostRetrievalNarrow", "post_retrieval_narrow"},
-		{"something new nobody classified", "other"},
-		{"", "other"},
-	}
-	for _, c := range cases {
+// productionPurposes is the census of every literal (or literal-shaped) value
+// passed to RecordLLM/RecordLLMSince in non-test code at head, paired with the
+// class it MUST map to. Keep this in sync with the call-site census when a new
+// RecordLLM* site is added — the retrieval-prep entries are "检索预处理: " + the
+// closed forceFn set plus the no-force "(tool-call)" case.
+var productionPurposes = []struct {
+	purpose string
+	want    string
+}{
+	{"检索预处理(tool-call)", "retrieval_prep"},
+	{"检索预处理: recognize_intent", "intent"},
+	{"检索预处理: extract_time_range", "extract_time_range"},
+	{"检索预处理: resolve_channel_scope", "resolve_channel_scope"},
+	{"检索预处理: resolve_topic_target", "resolve_topic_target"},
+	{"检索后裁剪 PostRetrievalNarrow", "post_retrieval_narrow"},
+	{"Map: 单次总结(跳过Map-Reduce)", "map_single"},
+	{"Map: 分块总结 chunk#0", "map_chunk"},
+	{"Map: 分块总结 chunk#7", "map_chunk"},
+	{"Reduce: 合并分块总结", "reduce"},
+	{"团队汇总: 合并各成员总结", "team_reduce"},
+}
+
+// TestPurposeClass_RealCallSites pins every production purpose value to its
+// class. Built from the RecordLLM* census rather than the report-rendering
+// strings, so misclassifying any real family fails here.
+func TestPurposeClass_RealCallSites(t *testing.T) {
+	for _, c := range productionPurposes {
 		if got := PurposeClass(c.purpose); got != c.want {
 			t.Errorf("PurposeClass(%q) = %q, want %q", c.purpose, got, c.want)
+		}
+	}
+}
+
+// TestPurposeClass_NoProductionPurposeIsOther is the anti-vacuity guard the
+// review asked for: no real call-site value may land in the "other" junk drawer.
+// Misclassifying any current production family (e.g. dropping the retrieval-prep
+// cases) turns this red, where a table of speculative literals would not.
+func TestPurposeClass_NoProductionPurposeIsOther(t *testing.T) {
+	for _, c := range productionPurposes {
+		if got := PurposeClass(c.purpose); got == "other" {
+			t.Errorf("production purpose %q classified as %q — real latency would be buried in the unclassified bucket", c.purpose, got)
+		}
+	}
+}
+
+// TestPurposeClass_UnknownFallsThrough documents the closed-set contract: a
+// value no call site emits, including a would-be-new forced function, lands in
+// "other" rather than silently widening the label space.
+func TestPurposeClass_UnknownFallsThrough(t *testing.T) {
+	for _, p := range []string{
+		"",
+		"something new nobody classified",
+		"检索预处理: brand_new_forced_fn",
+		"ReducerDebug", // must NOT be caught by the reduce rule
+	} {
+		if got := PurposeClass(p); got != "other" {
+			t.Errorf("PurposeClass(%q) = %q, want other", p, got)
 		}
 	}
 }
--- a/internal/timing/timing_test.go
+++ b/internal/timing/timing_test.go
@@ -1,6 +1,7 @@
 package timing
 
 import (
+	"fmt"
 	"os"
 	"path/filepath"
 	"strings"
@@ -320,5 +321,48 @@ func TestRecordLLM_Empty(t *testing.T) {
 
 	if count != 0 {
 		t.Error("RecordLLM with empty taskNo should not record")
+	}
+}
+
+// TestPurposeClass_ClosedSet asserts every real call-site purpose maps to its
+// stable class, and — the point of #231 S3 — that the unbounded chunk index in
+// "Map: 分块总结 chunk#%d" collapses to a single class rather than one label
+// value per chunk.
+func TestPurposeClass_ClosedSet(t *testing.T) {
+	cases := []struct {
+		purpose string
+		want    string
+	}{
+		{"意图识别", "intent"},
+		{"Map: 分块总结 chunk#1", "map_chunk"},
+		{"Map: 分块总结 chunk#2", "map_chunk"},
+		{"Map: 分块总结 chunk#9999", "map_chunk"},
+		{"Map: 单次总结(跳过Map-Reduce)", "map_single"},
+		{"单次总结", "map_single"},
+		{"Reduce: 最终总结", "reduce"},
+		{"Reduce: 合并分块总结", "reduce"},
+		{"Reduce", "reduce"},
+		{"团队汇总: 合并各成员总结", "team_reduce"},
+		{"检索后裁剪 PostRetrievalNarrow", "post_retrieval_narrow"},
+		{"something new nobody classified", "other"},
+		{"", "other"},
+	}
+	for _, c := range cases {
+		if got := PurposeClass(c.purpose); got != c.want {
+			t.Errorf("PurposeClass(%q) = %q, want %q", c.purpose, got, c.want)
+		}
+	}
+}
+
+// TestPurposeClass_ChunkIndexIsBounded is the cardinality guard: a thousand
+// distinct chunk indices must produce exactly one class, or the label space is
+// unbounded and this function has failed its only job.
+func TestPurposeClass_ChunkIndexIsBounded(t *testing.T) {
+	seen := map[string]struct{}{}
+	for i := 0; i < 1000; i++ {
+		seen[PurposeClass(fmt.Sprintf("Map: 分块总结 chunk#%d", i))] = struct{}{}
+	}
+	if len(seen) != 1 {
+		t.Errorf("chunk-map purposes produced %d classes, want 1 — cardinality is not bounded: %v", len(seen), seen)
 	}
 }


Jerry-Xin (changes requested):
**REQUEST_CHANGES** — the cardinality bound itself is verified and correct, but the classifier's coverage of the real purpose value domain at head is wrong, and the closed-set test does not enforce the census it claims. Both should be fixed before any emission PR consumes this contract.

### Verified (byte + executable evidence at head 0218e127)

- **Boundedness holds for every value the tree can emit.** Full census of `timing.RecordLLMSince` call sites (8): `internal/worker/meta_processor.go:237` (团队汇总→team_reduce), `internal/worker/processor.go:760` and `internal/worker/personal_processor.go:711` (purpose variable), `processor.go:766` / `personal_processor.go:719` (检索后裁剪 PostRetrievalNarrow→post_retrieval_narrow), `personal_processor.go:997` (Map: 单次总结→map_single), `personal_processor.go:1070` (`Map: 分块总结 chunk#%d`→map_chunk), `personal_processor.go:1133` (Reduce: 合并分块总结→reduce). The unbounded chunk-index literal collapses to a single class (the 1000-index test probe is green). The purpose-variable sites resolve to `检索预处理(tool-call)` and `检索预处理: <forceFn>` with forceFn from the closed set {recognize_intent, extract_time_range, resolve_channel_scope, resolve_topic_target} (`internal/pipeline/intent.go:265`, `narrow.go:115`, `resolve_channel.go:712`, `resolve_topic.go:141`) — bounded, so there is no cardinality defect. All `timing.Observe` stage names are literals.
- **Pre-consumption seam, explicitly scoped.** Nothing outside `internal/timing` references `PurposeClass` at head, and no metric emission consumes `LLMCall.Purpose` anywhere (the existing `internal/llmobs` registry labels path/model/position/outcome only). The PR description explicitly defers emission and declares independence from #232 — acceptable scoping, not a vacuous fix.
- **No collision with #238 or #239.** #238's purpose enum (workspace/agent_reference/personal/team) is the conversation-scope domain and never flows into `LLMCall.Purpose`; #239 is not in this PR's base (base = main tip 391134cc, the #238 merge).
- **CI green at head** (Build / Test / Lint / Vet / race). Local `go vet ./internal/timing/` clean; package suite green. (Local `go build ./...` fails only at the `cmd/summary-api` cgo link step — a native tokenizer library missing on the review machine; environmental, corroborated by CI Build being green.)
- **Mutation probes:** identity classifier (`return purpose`) → both new tests red ✓; remove the `map_chunk` case → red ✓; remove the `意图识别` case → red ✓.

### 🔴 P1 — the classifier census is wrong against the real value domain, and the closed-set test does not enforce the coverage it claims

The test comment and PR description state "every real call-site purpose maps to its class." The bytes at head falsify that:

1. **The `intent` class is dead in production.** Intent recognition is recorded through the shared tool-call closure — `internal/pipeline/intent.go:265` forces `recognize_intent` — so its real purpose is `检索预处理: recognize_intent`, which classifies as `other`. The literal `意图识别` appears nowhere in the tree as a `RecordLLM`/`RecordLLMSince` argument; it exists only inside `PurposeClass` itself, its test, and unrelated report-rendering strings (`internal/timing/timing.go:270-272`). The advertised `intent` series will never emit, and real intent-recognition latency is indistinguishable from unclassified junk in `other`.
2. **Five real production purpose values are absent from the table**: `检索预处理(tool-call)`, `检索预处理: recognize_intent`, `检索预处理: extract_time_range`, `检索预处理: resolve_channel_scope`, `检索预处理: resolve_topic_target` — all → `other`. Meanwhile the table pins literals no call site emits (`Reduce: 最终总结`, bare `单次总结`, bare `Reduce`). The enforcement is inverted.
3. **Mutation-proven vacuity:** inserting `case strings.HasPrefix(purpose, "检索预处理"): return "map_chunk"` — misclassifying the entire retrieval-prep family — leaves the whole suite green. No test pins any real production purpose, while a production-dead literal is pinned (removing it goes red).

Why this blocks now: this function is the contract the #231 S3 emission will label on (`summary.llm.duration{purpose_class}` and friends). Intent-recognition and retrieval-prep latency percentiles are exactly the decision inputs #220 says it needs; consuming this class set as shipped buries them in `other`, and relabeling after emission ships means time-series churn across dashboards and alerts. The fix is small and belongs in this PR:

- Enumerate the census (the 8 `RecordLLMSince` sites above) and add class coverage for the retrieval-prep family — either one `retrieval_prep` class or per-forced-function classes (bounded: the forced-function registry is closed) — and decide explicitly whether intent recognition (`recognize_intent`) deserves the already-advertised `intent` identity rather than falling through.
- Rebuild the test table from the real literals emitted at head; keep speculative literals only if clearly marked as forward-compatibility; and make at least one assertion fail if any current production family is misclassified (today none would).

### 🟡 P2 (non-blocking)

- `strings.HasPrefix(purpose, "Reduce")` (`internal/timing/timing.go:197`) is looser than the documented contract "unknown purposes fall through to `other`": any future value starting with `Reduce` (e.g. `ReducerDebug`) would silently join the `reduce` series. `purpose == "Reduce" || strings.HasPrefix(purpose, "Reduce:")` would match the documented rule.
- Title/scope: #231 S3 is "wire agent and worker" (emission). This PR delivers the purpose-normalization precondition — the one live cardinality bomb #231 names — and the description is honest about that. Fine as an S3 installment, but the emission PR must consume the corrected census.

With P1 addressed (classifier + tests aligned to the real value domain), this is a quick re-review.

mochashanyao (changes requested):
[Octo-Q · automated review]

---

# Code Review — PR #240 (octo-smart-summary)

## Summary

The PR adds a single exported helper, `PurposeClass`, to `internal/timing` plus two tests. It maps the free-text `LLMCall.Purpose` recorded by `RecordLLM` onto a closed set (`intent`, `map_chunk`, `map_single`, `reduce`, `team_reduce`, `post_retrieval_narrow`, `other`) so a call-kind metric cannot inherit the unbounded `chunk#%d` suffix as a label value. Nothing else changes: no production caller, no metric wiring, no config or docs. The cardinality bound itself is sound — unknown purposes fall through to `other`, so the label space cannot widen. The mapping table, however, does not match the purpose strings the code actually produces.

## Verification

Static analysis only at head `0218e127612177343057dbce8b517172db02c4e8`; build and tests not executed in this environment.

- ✅ **Cardinality bound holds** — `default: return "other"` (`internal/timing/timing.go:201`) means an arbitrary future purpose cannot add a series, and the new loop test asserts 1000 chunk indices collapse to one class.
- ✅ **Rule ordering survives the tricky literals** — `"Map: 单次总结(跳过Map-Reduce)"` contains the substring `Reduce` but is caught by the `map_single` prefix rule at `internal/timing/timing.go:193` before the `Reduce` rule at `:197`; `"Reduce: 合并分块总结"` cannot hit `map_chunk`, which requires the full `Map: 分块总结 chunk#` prefix.
- ✅ **Imports compile** — `strings` was already imported at `internal/timing/timing.go:21` on the base revision; the test's added `fmt` import is used by the chunk-index loop.
- ✅ **No lint regression from the unused export** — `.golangci.yml` enables only `govet`, so an exported-but-unreferenced helper does not fail CI.
- ✅ **All 13 expectations in `TestPurposeClass_ClosedSet` match the switch** — walked case by case, no mismatch.

## Findings

One P1 blocker and three P2 items.

### P1 — `intent` is unreachable in production and four live purposes fall into `other` (`internal/timing/timing.go:187`)

No production code records purpose `"意图识别"`. The intent-recognition call goes through the unified tool-call callback with `forceFn="recognize_intent"` (`internal/pipeline/intent.go:265`), and the worker composes `purpose = "检索预处理: " + forceFn` before recording it (`internal/worker/personal_processor.go:708`, `internal/worker/processor.go:757`). That string matches no case here, so it returns `other` — as do `"检索预处理: extract_time_range"` (`internal/pipeline/narrow.go:115`), `"检索预处理: resolve_channel_scope"` (`internal/pipeline/resolve_channel.go:712`), `"检索预处理: resolve_topic_target"` (`internal/pipeline/resolve_topic.go:141`) and the `forceFn == ""` fallback `"检索预处理(tool-call)"`. The only places `"意图识别"` appears as a Purpose value are hand-fed test inputs (`internal/timing/timing_test.go:85`, `:181`); elsewhere it is report header text (`internal/timing/timing.go:270`). Since the doc comment at `:177` makes this helper the mandatory label source for any call-kind metric, wiring it as-is reports every intent-recognition and pre-retrieval call as `other` while the `intent` series stays permanently zero — silently wrong operator data for the most frequent non-map/reduce LLM calls, and the per-run report keeps contradicting itself (header `意图识别: 短路=否 LLM调用=1次` next to `用途=检索预处理: recognize_intent`).

```go
case strings.HasPrefix(purpose, "检索预处理: "):
    switch strings.TrimPrefix(purpose, "检索预处理: ") {
    case "recognize_intent":
        return "intent"
    case "extract_time_range", "resolve_channel_scope", "resolve_topic_target":
        return "pre_retrieval_tool"
    }
```

Classify the tool-call family explicitly (per-function classes or one bounded `pre_retrieval_tool` class plus the `forceFn == ""` fallback), then assert each of the nine real purpose strings in the test.

### P2 — No production consumer, so nothing is bounded yet (`internal/timing/timing.go:178`)

`PurposeClass` is referenced only from `internal/timing/timing_test.go`. `Purpose` reaches exactly two sinks today: the in-memory account (`internal/timing/timing.go:161`) and the human-readable report line (`:300`). The metrics registry never labels on it — `internal/llmobs/metrics.go:26` documents that label values come from config or a closed constant set, never request data, and `llm_calls_total` / `llm_call_duration_seconds_total` are labelled by `path`, `position`, `result` only (`internal/llmobs/metrics.go:222`). If this is deliberate staging for a later step of #231, say so in the PR description and in the doc comment; as written, the comment reads as though the cardinality bomb is already defused while the guard is comment-plus-test only.

### P2 — Classification keys off display prose, creating a second taxonomy (`internal/timing/timing.go:191`)

Every rule string-matches the report text rendered at `internal/timing/timing.go:300`, while the repo already has a bounded machine-level call-kind enum feeding the real metrics: `llmfallback.Path` with `worker_map` / `worker_reduce` / `tool_call` / `api_refine` / `agent_chat` / `agent_tool` (`internal/llmfallback/observer.go:11`). The drift is not hypothetical — it already happened in this PR (see the P1). The doc at `:173` also understates the input space: besides the chunk index, `"检索预处理: " + forceFn` is composed at the call site, bounded today by four literals but one prefix rule away from re-exposing a suffix. Prefer passing the class (or the existing `llmfallback.Path`) into `RecordLLM`/`RecordLLMSince` and keeping `Purpose` purely for the log line, so a metric label never depends on parsing prose.

### P2 — Test claims full call-site coverage but omits the tool-call family (`internal/timing/timing_test.go:327`)

The comment asserts "every real call-site purpose maps to its stable class", yet the table omits all five tool-call purposes and includes `"意图识别"`, bare `"单次总结"`, `"Reduce"` and `"Reduce: 最终总结"`, none of which any production call site emits. The suite is green against an inventory that does not exist, which is exactly how the P1 mis-bucketing stays invisible. Derive the cases from the real literals — one per `RecordLLMSince` call site, including the `forceFn`-composed ones — so a new call site without a class fails the test instead of silently landing in `other`.

## Verdict: CHANGES_REQUESTED

The helper is well-intentioned and its cardinality bound is correct, but the delivered mapping is calibrated against test strings rather than the nine purposes production actually records: `intent` can never fire and the whole retrieval-preprocessing family, intent recognition included, collapses to `other`. Fixing the table and the coverage test is enough to clear this; the two design P2s are advisory.





---

# 附录（内部材料 · 供终审腿复核，非对外 review 正文）

## 数据流回溯（每个被消费数据 → 上游来源 → 运行时是否真流到消费点）

**1. `PurposeClass(purpose)` 的输入 `purpose` — 上游 9 个真实产生点（全仓 rg `RecordLLM|RecordLLMSince`，排除 vendor/test）**

| 真实 purpose | 产生点 | 分类结果 |
|---|---|---|
| `"检索预处理(tool-call)"`（forceFn=="") | `internal/worker/personal_processor.go:707`、`internal/worker/processor.go:756` | ❌ `other` |
| `"检索预处理: recognize_intent"` | 组装于 `internal/worker/personal_processor.go:708-711` / `internal/worker/processor.go:757-760`，forceFn 来自 `internal/pipeline/intent.go:265` | ❌ `other`（应为 `intent`） |
| `"检索预处理: extract_time_range"` | forceFn 来自 `internal/pipeline/narrow.go:115` | ❌ `other` |
| `"检索预处理: resolve_channel_scope"` | forceFn 来自 `internal/pipeline/resolve_channel.go:712` | ❌ `other` |
| `"检索预处理: resolve_topic_target"` | forceFn 来自 `internal/pipeline/resolve_topic.go:141` | ❌ `other` |
| `"检索后裁剪 PostRetrievalNarrow"` | `internal/worker/personal_processor.go:719`、`internal/worker/processor.go:766` | ✅ `post_retrieval_narrow` |
| `"Map: 单次总结(跳过Map-Reduce)"` | `internal/worker/personal_processor.go:997` | ✅ `map_single`（未被 `Reduce` 前缀抢走，顺序正确） |
| `fmt.Sprintf("Map: 分块总结 chunk#%d", idx)` | `internal/worker/personal_processor.go:1070` | ✅ `map_chunk`（unbounded idx 被折叠，bound 生效） |
| `"Reduce: 合并分块总结"` | `internal/worker/personal_processor.go:1133` | ✅ `reduce` |
| `"团队汇总: 合并各成员总结"` | `internal/worker/meta_processor.go:237` | ✅ `team_reduce` |

**2. `"意图识别"` 这个 Purpose 值 — 生产端不存在（断流）**
`rg "意图识别"` 全仓命中：`internal/pipeline/intent_rules.yaml:1`、`internal/pipeline/intent_shortcut.go:16,184`（均为注释/文案）、`internal/timing/timing.go:270,272`（报告头文案，非 Purpose）、`internal/timing/timing_test.go:85,99,100,181,203,207,265,336`（hand-feed）。意图识别的真实计量走 `TaskContext.IntentLLMCalls`（`internal/worker/personal_processor.go:762`）与 `timing.RecordSkip(taskNo,"intent_recognition",...)`（`:784`），**不经 Purpose 通道** → `case purpose == "意图识别"`（`internal/timing/timing.go:187`）生产不可达。

**3. `PurposeClass` 的返回值 — 消费点为空（本 PR 的 bound 在生产端不生效）**
全仓 `rg PurposeClass` 10 处命中全部落在定义/注释（`internal/timing/timing.go:170,178,185`）与测试（`internal/timing/timing_test.go:327,331,351,352,357,360,363`）。真实 metric 路径 `internal/llmobs/metrics.go` 的 `labels()` 只吃 `path`/`model`/`position`/`outcome`/`reason`/`from`/`to`/`result`（`:186-224`），并在 `:26-29` 明确写死"every label value … never from request data"。`Purpose` 当前仅两个下游：内存账本 `internal/timing/timing.go:161` 与人读报告行 `:300`。→ 结论：**无生产消费点**。

## 盲点 checklist（C1–C6）

- **C1 双路径 parity — HIT（P1）**：类集合 ↔ 真实 purpose 集合两端不对称，5 个真实 purpose 无对应 class、1 个 class（`intent`）无对应 purpose，证据见上表。守卫等价写法覆盖：新增 test 不是守卫（hand-feed 字面量，不从 call-site 派生），无法保证"新 call site 必须加 case"这条 doc 规则被执行 → 假绿。
- **C2 control-flow ordering / 非 canonical 形式 — CLEAR（附证据）**：逐条走查 switch 顺序（`map_chunk` 先于 `map_single`；`map_single` 先于 `Reduce`，故 `"(跳过Map-Reduce)"` 不被抢；`团队汇总` 先于 `Reduce`）。非 canonical 试穿：全角冒号 `Map：分块总结`、前导空格/tab、`chunk#` 改词、大小写变体 → 均静默落 `other`；当前 4 个 Map/Reduce/团队汇总 字面量精确匹配生产写法，故本轮不成立为缺陷，脆弱性并入 P2（以展示文案做分类键）。
- **C3 授权边界 ≠ 能力边界 — N/A**：diff 不含权限/jail/tool 暴露/凭证改动（2 文件、+80/-0，纯 `internal/timing`）。
- **C4 授权生命周期 / 容器-成员级联 — N/A**：无鉴权、无父容器 status 查询改动。
- **C5 build/note 通过 ≠ 运行期路径正确 — HIT**：doc comment 的"Any metric … MUST label on PurposeClass"是 note 不是修复；test 绿也不等于运行期分类正确（hand-feed 绕过真实 purpose 组装路径）。已按运行期推演定级，未以"build 过 / note 已警告 / tests pass"缓解。补充：本环境未执行 `go build` / `go test`（skill 明令禁止跑构建与测试），故编译/测试结论为静态推演——`strings` 在 base 已导入（`internal/timing/timing.go:21`，base 版本第 217 行已用 `strings.Builder`），test 新增 `fmt` 被 `Sprintf` 使用，`.golangci.yml` 仅启用 `govet`（未启 `unused`），未引用导出函数不会挂 CI。
- **C6 治理/策略/安全文档自洽性 — N/A（已比对）**：diff 不含 SECURITY.md / CONTRIBUTING / CODEOWNERS / label 策略 / 披露流程；`rg` 该仓 `docs/`、`README.md`、`CONFIGURATION.md` 未发现列举 timing purpose 或 metric label 集合的文档需同步。`internal/llmobs/metrics.go:26-29` 的"标签值不来自 request data"不变量与 PurposeClass 的闭集输出相容，无策略冲突或绕过。

## 跨轮 blocker 复检（R6）

N/A — 首轮审查。本 issue comment 历史为空（`Octo-Q issue comment list` → `[]`），且按 skill 约束未查询 GitHub 旧评论/timeline/benchmark。审查对象为当前 head `0218e127612177343057dbce8b517172db02c4e8`；merge-base = `391134cc8c25226287e58a26204616d42a200e6d`（等于 `origin/main` tip），diff 范围 `pr-base...HEAD` = 2 files / +80 / -0（`internal/timing/timing.go` 170-205、`internal/timing/timing_test.go` 4 + 326-368）。

## 定级依据（R1–R4）

- **diff-scope 三问（P1）**：① new — 分类表与声称覆盖全量 call-site 的 test 均为本 PR 新增；② 非 pre-existing — base 无 PurposeClass、无任何 purpose→class 映射；③ 本 PR diff 本身就是该行为契约，不存在"base 行为一致且未触碰"的降级条件。
- **R1 判定**：分类表对真实生产 purpose 错分（`intent` 恒为 0；意图识别 + 三类检索预处理 tool-call 全落 `other`），而本 PR 的 doc 把该函数定为 call-kind metric 的**唯一合法 label 来源** → 按其自身规定接线即产生运维可见错误数据，且比"不提供该函数"更糟（错表带着权威 doc + 假绿 test 进主干，后续作者不会再自行盘点 call-site）。触发条件不是边缘路径：意图识别与检索预处理是每次总结运行的常规 LLM 调用。→ **P1**。
- **诚实边界（供终审裁量）**：本 PR 未接线，故当前生产无实时回归；P1 的成立依赖 doc 自己规定的强制接线路径与"#231 S3 分步"的上下文（仓内 `git log` 未见 S1/S2 接线提交）。若终审判定"纯 staging、下一步才接线"应降级，最低也应是 P2-blocking-for-series；但修表成本 <10 行，带病合入的代价更高，故automated review建议维持 CHANGES_REQUESTED。
- **R3 分布式/多实例副作用 — N/A**：无通知/推送/event emit/计数广播/缓存状态改动，纯内存纯函数，无 fan-out、无 observer≠actor、无调度路径丢失。
- **R4 映射**：存在 1 个 P1 → `CHANGES_REQUESTED`；另有 3 个 P2（无生产消费点/未接线、以展示文案为分类键且与既有 `llmfallback.Path` 闭集重复、test 覆盖声明不实）。无 P0。

## 双模型 fan-out 状态（降级说明）

本机安装的 `requesting-code-review` skill 目录仅含 `scripts/pr_review_tool.py`，**不存在** `glm_review.py` / `qwen_normalize.py` / `merge_reviews.py`，SKILL.md 内亦无 Dual-Model Fan-out / GATING 段。故本轮为单腿执行，无 `/tmp/merged.findings.json`，渲染源为 `/tmp/review.final.json`（`finalize` 校验：`validation_valid=true`、`invalid_comments=[]`、`missing_high_risk_families=[]`；2 条 `weak-failure-path` warning 分别落在两条 P2 上，已在其正文补齐后果描述）。GH_TOKEN 在本环境为 Bad credentials（HTTP 401），故未使用 `gh`，PR 内容通过公开仓匿名 clone + `refs/pull/240/head` 获取，head SHA 与任务单一致；全程零写操作、未提交任何 GitHub review/status。

[Octo-Q] verdict: REQUEST_CHANGES — 理由：新增 `PurposeClass` 的闭集与生产真实 purpose 清单不一致（`intent` 生产不可达；`检索预处理: recognize_intent` / `extract_time_range` / `resolve_channel_scope` / `resolve_topic_target` / `检索预处理(tool-call)` 五项全落 `other`），而其 doc comment 把该函数定为 call-kind metric 的强制 label 来源，接线即产生运维可见错误数据；配套 test 声称覆盖 "every real call-site purpose" 却用 hand-feed 字面量绕过真实组装路径，构成假绿。其余 3 项 P2 为建议级（未接线/无消费点、以展示文案做分类键并与既有 `llmfallback.Path` 闭集重复、测试覆盖声明不实）。修好分类表 + 按真实 call-site 重写测试表即可清除阻塞。

yujiawei (changes requested):
# Code Review — PR #240 (octo-smart-summary)

Reviewed at head `0218e127612177343057dbce8b517172db02c4e8`, merge-base `391134cc` (current `main` tip). Independent pass; a review already exists on this same head and converges on the same P1. I confirm it, and correct one part of its census that makes the problem slightly larger than reported.

## 1. Spec compliance

**Spec: ❌**

The governing spec is #231 S3: *"`timing.RecordLLMSince()` needs `purpose` normalized before it can be a label: `fmt.Sprintf("Map: 分块总结 chunk#%d", idx)` must collapse to a `purpose_class` of `map_chunk`, with the index staying in the log line only."*

**Delivered correctly:** the literal ask is met. `map_chunk` collapses the unbounded index, `TestPurposeClass_ChunkIndexIsBounded` proves it over 1000 indices, and the raw purpose stays log-only (`timing.go:300` is the only consumer of `LLMCall.Purpose`; `internal/timing` has no metric emission at all). Scoping to `internal/timing` with no emission wired is honest and matches the PR description. Build / Test / Test(race) / Vet / Lint are green at head; `go vet ./internal/timing/` and the package suite are green locally too.

**Missing:** the PR description claims *"`TestPurposeClass_ClosedSet` — every real call-site purpose maps to its class."* That is false at head. Full census of the 8 `RecordLLM`/`RecordLLMSince` sites — `meta_processor.go:237`, `processor.go:760`, `processor.go:766`, `personal_processor.go:711`, `:719`, `:997`, `:1070`, `:1133` — yields these purposes actually emitted at runtime:

| Purpose emitted | `PurposeClass` result | Correct? |
|---|---|---|
| `团队汇总: 合并各成员总结` | `team_reduce` | ✅ |
| `Map: 单次总结(跳过Map-Reduce)` | `map_single` | ✅ |
| `Map: 分块总结 chunk#N` | `map_chunk` | ✅ |
| `Reduce: 合并分块总结` | `reduce` | ✅ |
| `检索预处理: recognize_intent` | **`other`** | ❌ |
| `检索预处理: extract_time_range` | **`other`** | ❌ |
| `检索预处理: resolve_channel_scope` | **`other`** | ❌ |
| `检索预处理: resolve_topic_target` | **`other`** | ❌ |

None of those four appears in the test table. Four of the eight call-site purposes are unclassified, and no test would fail if the classifier mishandled any of them.

**Over-built:** two of the seven advertised classes have no producer anywhere in the tree — see P1 and P1-b below.

## 2. Code quality

### 🔴 P1 — the retrieval-preprocessing family falls to `other`, and `intent` is dead

`internal/timing/timing.go:187`

```go
case purpose == "意图识别":
    return "intent"
```

The literal `意图识别` is never passed to `RecordLLM`/`RecordLLMSince`. It exists only inside `PurposeClass`, its test, and unrelated report-rendering text at `timing.go:270-272`. Intent recognition is recorded through the shared tool-call closure:

```go
// internal/worker/personal_processor.go:707-711 (identical at processor.go:756-760)
purpose := "检索预处理(tool-call)"
if forceFn != "" {
    purpose = "检索预处理: " + forceFn
}
timing.RecordLLMSince(taskNo, purpose, callStart, tokens)
```

`forceFn` is a closed set of four literals — `recognize_intent` (`internal/pipeline/intent.go:265`), `extract_time_range` (`narrow.go:115`), `resolve_channel_scope` (`resolve_channel.go:712`), `resolve_topic_target` (`resolve_topic.go:141`). So the real intent purpose is `检索预处理: recognize_intent` → `other`. The `intent` series will never emit a sample, and intent-recognition latency is indistinguishable from genuinely unclassified traffic.

This is bounded, so there is no cardinality defect — but it defeats the function's purpose. `#231` wants to answer *"which pipeline stage is the current straggler"* and `#220` needs per-stage P95/P99 to set `LLM_TIMEOUT`. Retrieval preprocessing is four distinct LLM round-trips per task and is exactly the kind of stage those questions are about; burying it in `other` removes it from the answer. It also breaks the discipline the doc comment sets up: `other` is documented as the sink for *new, unclassified* purposes, so an operator watching `other` grow cannot tell a new unlabeled call site from the four known ones that live there permanently.

Relabeling after `summary.llm.duration{purpose_class}` ships means time-series churn across dashboards and alerts, so it is materially cheaper to fix now, in this PR, while nothing consumes the contract.

### 🔴 P1-b — `post_retrieval_narrow` is dead too (correction to the existing review)

`internal/timing/timing.go:199`

The existing review's census lists `processor.go:766` / `personal_processor.go:719` as a live `post_retrieval_narrow` mapping. It is not. The closure that records that purpose is never invoked:

```go
// internal/pipeline/narrow.go:161-163
// PostRetrievalNarrow remains unchanged — uses LLMCallFn (not Function Call).
func PostRetrievalNarrow(ctx context.Context, messages []Message, topic string, llmFn LLMCallFn) []Message {
	return messages
}
```

`fetch.go:1163` is the sole consumer of the worker's `llmFn` closure, and it hands it to that stub, which returns without calling it. (`NarrowByTopicReport` also takes an `LLMCallFn`, but its only caller is `internal/agent/tool_narrow_channels.go:77` with its own closure, and the agent path does not use `internal/timing`.)

So **two of seven classes are unreachable**, not one, and the reachable set is `{map_chunk, map_single, reduce, team_reduce, other}` — with `other` permanently holding the retrieval-prep family. Please re-derive the census rather than patching the one class named in the earlier review.

The same check kills one more advertised value: `检索预处理(tool-call)` (the `forceFn == ""` branch) is also unreachable, since all four pipeline call sites pass a non-empty `forceFn`.

### 🟡 P2 — free-text matching is the root cause; this repo already has the drift-proof idiom

`internal/timing/timing.go:185-203`

`internal/llmfallback/observer.go:7-18` solves this exact problem the other way round:

```go
// Path identifies which logical call site drove a Run. It is a low-cardinality
// label: keep it to the fixed set below so metrics series stay bounded. Never
// derive it from request data (task ids, chunk indexes, user ids).
type Path string

const (
	PathUnknown Path = "unknown"
	PathAgentChat Path = "agent_chat"
	...
)
```

Typed constants, assigned explicitly at the call site via `WithPath`, already carrying the same cardinality red line in its doc comment. A producer cannot drift from that vocabulary, because the producer *is* the vocabulary.

`PurposeClass` inverts this: it reverse-engineers a taxonomy from free text written elsewhere, and P1 / P1-b are the first two symptoms — the author had no way to see `检索预处理: recognize_intent` or the `PostRetrievalNarrow` stub from inside `internal/timing`. Suggested shape: export the purpose literals as consts in `timing` (or take a typed `Class` argument alongside the human-readable purpose at the `RecordLLM*` sites), so adding a call site without a class becomes a compile-time or grep-able event rather than a silent `other`.

### 🟡 P2 — matcher looseness contradicts the documented contract

- `timing.go:197`: `strings.HasPrefix(purpose, "Reduce")` would silently absorb a future `ReducerDebug` into the `reduce` series. `purpose == "Reduce" || strings.HasPrefix(purpose, "Reduce:")` matches the doc.
- `timing.go:199`: `strings.Contains(purpose, "PostRetrievalNarrow")` is the only substring match in the tree — looser than every sibling case, and the producer is a fixed literal, so `==` is sufficient.

### 🟡 P2 — the test table pins literals nothing emits

`internal/timing/timing_test.go:327-341`: `意图识别`, bare `单次总结`, `Reduce: 最终总结`, and bare `Reduce` have no producer at head, while all four live retrieval-prep purposes are absent. The table's docstring claims it is a call-site census; it is currently the inverse. Rebuilding it from the eight real sites (with a comment mapping each row to `file:line`) makes the next drift visible.

## 3. Overall verdict

**CHANGES_REQUESTED** — Spec ❌ AND Quality Changes-Requested.

To be explicit about proportionality: nothing here is a runtime risk today. The function has no callers, no behavior changes, and the headline cardinality bomb is genuinely defused. The block is that this function's entire value is being the *correct* label vocabulary for the emission PR that follows, and it ships with 4 of 8 real purposes unclassified and 2 of 7 classes dead. The fix is roughly 15 lines plus test rows, and it is much cheaper before a dashboard depends on the class names.

## 4. Suggested changes

1. Add coverage for the retrieval-prep family — either one `retrieval_prep` class or per-forced-function classes (`forceFn` is a closed set of four, so both are bounded). If per-function, `recognize_intent` can claim the already-advertised `intent` identity.
2. Delete `post_retrieval_narrow` and the `检索预处理(tool-call)` handling, or leave them with a comment saying they are forward-compatibility for a stub that does not yet call back. Do not leave them looking like live mappings.
3. Rebuild the test table from the eight `RecordLLM*` sites, annotating each row with `file:line`. A useful invariant: assert that no *currently emitted* purpose returns `other`.
4. Tighten the `Reduce` prefix and the `PostRetrievalNarrow` substring match per the doc contract.
5. Consider the typed-const shape from `llmfallback.Path` so this class of drift becomes a compile-time concern.

## 5. Additional findings (outside this diff)

- **`pipeline.PostRetrievalNarrow` is a silent no-op** (`internal/pipeline/narrow.go:162`). Layer 5 of the personal fetch pipeline does nothing, while `processor.go:714-720` and `personal_processor.go:715-721` each build an LLM closure and thread it through `ResolveAndFetchMessagesForPersonal`'s signature to reach it. Pre-existing, not introduced here, but worth its own issue: either implement it or remove the parameter and the two dead closures.
- **Nothing will enforce that the emission PR uses `PurposeClass`.** #231 §9 asks for *"a test asserting the emitted label key set equals an explicit whitelist."* That belongs in the emission PR, but this guard is inert until then — a future PR can still label on the raw `Purpose` with no test failing. Worth carrying forward as an acceptance item.
- **Independence from #232 checks out.** #232 does not touch `internal/timing`, and this diff touches nothing else, so there is no merge or semantic conflict either way.
- **`check-sprint` is failing** ("Validate Sprint field on Project Board"). Project-board metadata, not code, but it is a merge gate.

## Coverage notes

Checked: all 8 `RecordLLM*` sites and their purpose provenance; all 4 `forceFn` producers; reachability of every class; the `llmFn` / `toolCallFn` call graph into `internal/pipeline`; build, vet, gofmt and the package suite at head; base = current `main` tip; interaction with #232.

Not checked: runtime behavior under a real worker task (no emission exists to observe); Grafana dashboard or alert definitions (not in this repo); the agent path's own timing, which `PurposeClass` does not cover and which will matter when #231 S3 wires agent metrics.

wanyiwang06:
Thanks — the census was exactly right, `PurposeClass` had been built from the test file's placeholder strings rather than the real call sites. Fixed in 638646c.

**P1 — classifier realigned to the RecordLLM* census:**
- Retrieval-prep family now classified per forced function against the closed forceFn set: `检索预处理: recognize_intent`→`intent` (the advertised class is now real, not dead), `extract_time_range` / `resolve_channel_scope` / `resolve_topic_target` → same-named classes, and `检索预处理(tool-call)` (no force) → `retrieval_prep`. Bounded — the forceFn registry is closed. This keeps intent/retrieval percentiles distinct, which is what #220 needs, instead of burying them in `other`.
- Dropped the placeholder-only cases (`意图识别`, bare `单次总结`, bare `Reduce`, `Reduce: 最终总结`) — no call site emits them.

**P2 — reduce rule tightened** to `purpose == "Reduce" || strings.HasPrefix(purpose, "Reduce:")`, so `ReducerDebug` can't silently join the reduce series (there's now a test asserting it falls to `other`).

**Tests rebuilt from the real value domain:**
- `productionPurposes` table pins all 10 real call-site values (incl. the 4 forceFn variants and `chunk#N`).
- `TestPurposeClass_NoProductionPurposeIsOther` — the anti-vacuity guard you asked for: misclassifying any production family (e.g. dropping the retrieval-prep cases) turns this red.
- `TestPurposeClass_UnknownFallsThrough` — pins the closed-set contract for unknowns, including a would-be-new forced function.

`go build/vet/test ./internal/timing/` green. Ready for re-review.

Jerry-Xin (approved):
Round-2 re-review at 638646c5771f05acb76d8612cb3979d051ecec94 (single fix commit 638646c on top of the round-1 head 0218e12; merge-base unchanged at 391134c, main has not moved). Delta touches only internal/timing/timing.go (PurposeClass doc + switch) and internal/timing/timing_test.go (table + tests). Everything the delta touched was re-verified from scratch; unchanged mechanisms carry their round-1 findings.

## Verdict: APPROVE — the round-1 blocker (R1-P1, census/contract defect) is closed with mutation-resistant enforcement.

### R1-P1 closure evidence

1. Production-dead `intent` class — FIXED by aligning to real literals. The classifier now keys on the exact strings the forced-tool closure emits: "检索预处理: recognize_intent" → intent (internal/pipeline/intent.go:265), "检索预处理: extract_time_range" → extract_time_range (narrow.go:115), "检索预处理: resolve_channel_scope" (resolve_channel.go:712), "检索预处理: resolve_topic_target" (resolve_topic.go:141), plus the defensive no-force branch "检索预处理(tool-call)" → retrieval_prep (worker/processor.go:756, worker/personal_processor.go:707). I re-enumerated the full census at this head: all 8 RecordLLMSince call sites (meta_processor.go:237; processor.go:760,766; personal_processor.go:711,719,997,1070,1133) and all 4 pipeline toolCallFn invocations pass one of exactly the literals the table pins. Every production purpose now maps to a meaningful class; none falls to `other`. The old dead literal 意图识别 survives only in report-rendering (timing.go:292-294), which is a separate pre-existing display mechanism and does not flow through PurposeClass.

2. Test table rebuilt from the census — FIXED. `productionPurposes` is keyed to the real code literals; the round-1 fictional entries (意图识别, bare 单次总结, "Reduce: 最终总结", bare "Reduce") are no longer pinned as production values anywhere in the PurposeClass tests. Two new guards: TestPurposeClass_NoProductionPurposeIsOther (anti-vacuity: no real call-site value may land in the junk drawer) and TestPurposeClass_UnknownFallsThrough (closed-set contract, including a "ReducerDebug" pin and a would-be-new forced function "检索预处理: brand_new_forced_fn" → other).

3. Mutation battery re-run — 11 mutants, 10 killed, all value-level red with precise assertion messages:
   - M3 (round-1 survivor: whole retrieval-prep family → map_chunk) is now RED via TestPurposeClass_RealCallSites. The exact mutation that kept the suite green in round 1 now fails on all five family entries.
   - M3b (drop the five retrieval cases → other) RED via both RealCallSites and the anti-vacuity guard.
   - M4 (drop only the recognize_intent case) RED — the intent class is now pinned by a real production literal, unlike round 1.
   - M1 (neutralize classifier to constant other) RED; M2 (remove map_chunk case) RED including the 1000-index boundedness path; M6 (remove tool-call case) RED; M7 (class collision extract_time_range→intent) RED; M8 (remove team_reduce) RED; M10 (remove post_retrieval_narrow) RED.
   - M5 (loosen the reduce rule back to HasPrefix("Reduce")) RED via the ReducerDebug pin — the round-1 P2 on loose prefix matching is not just fixed but test-enforced.

### Round-1 P2 advisories status

- Loose HasPrefix("Reduce"): CLOSED (tightened to == "Reduce" || HasPrefix("Reduce:"), pinned by mutation-red test). Keeping bare "Reduce" as a defensive case is harmless; no current call site emits it and the closed-set comment remains accurate for code literals.
- PR title/scope wording: unchanged, acceptable — scope declaration stays honest (normalization seam only).
- S3 emission: still owed. PurposeClass has no production consumer at this head (verified by repo-wide grep); that matches the PR's explicit deferral. The follow-up emission must consume this census-pinned mapping.

### Boundedness (carried + re-verified)

The chunk-index bomb remains collapsed: TestPurposeClass_ChunkIndexIsBounded (1000 distinct indices → exactly 1 class) passes at this head, and the map_chunk-before-map_single ordering is preserved with an explicit comment.

### Audit of the fix itself (new bytes, full rigor)

- One surviving mutant, Advisory (non-blocking): a hypothetical loosening that replaces the two Map rules with map_single-first plus a catch-all HasPrefix("Map") keeps the suite green, because every production Map literal still lands in the same class; only hypothetical unknown "Map*" values would deviate from the documented unknown→other contract. The shipped code is precise (exact prefixes), so no production value is affected. Suggestion: add one unknown "Map"-family literal to TestPurposeClass_UnknownFallsThrough for symmetry with the ReducerDebug pin, so the ordering property documented in the comment is itself test-enforced.
- Advisory (non-blocking): the PR description still lists the old seven-class set while the implemented closed set is now eleven classes (timing.go doc comment and switch agree with each other; verified all 11 returned labels are enumerated). Updating the description would keep the contract documentation aligned for the S3 emission follow-up.
- No SQL touched; package tests run without any database. go vet ./internal/... clean; go build ./internal/... clean; full internal/timing suite green (-count=1). CI at this head: Lint/Vet/Test/Test(race,cgo)/Build/Generate/osv/dependency-review/secret-scan/pr-title-lint all green; check-sprint fails but it also failed at the round-1 head and at main — pre-existing repo-wide board gate, not introduced by this change.

The round-1 blocker is closed in both fix shapes it could take: real-literal alignment AND a mutation-resistant, census-keyed test enforcement. Approving.

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #240 (octo-smart-summary)
Head: `638646c5771f05acb76d8612cb3979d051ecec94` · Base: `391134cc` (merge-base with `main`) · Range: `pr-base...HEAD` · +133/-0 across 2 files

## Summary

This PR adds one pure classifier, `PurposeClass`, to `internal/timing`, plus four tests. It collapses the free-text `LLMCall.Purpose` values produced by the worker pipeline into a closed 11-value vocabulary so that a future metric can break LLM-call latency down by call kind without labelling on the unbounded `Map: 分块总结 chunk#%d` string. The change is strictly additive: no existing call site, log line, report field or metric series is modified.

The classification itself is correct — I re-derived the call-site census independently rather than trusting the test table, and every production purpose maps to a non-`other` class. The two findings below are about what is *not* there: the classifier has no consumer, and the census that keeps it honest is a hand-maintained copy with no enforcement. Neither is blocking.

## Verification

Static analysis only at head `638646c5`; build and tests not executed in this environment.

- ✅ **Census is accurate at head** — all 8 non-test `RecordLLM`/`RecordLLMSince` sites classify to a real class: `internal/worker/meta_processor.go:237` (团队汇总 → `team_reduce`), `internal/worker/personal_processor.go:711` and `internal/worker/processor.go:760` (closure-derived → retrieval-prep family), `internal/worker/personal_processor.go:719` and `internal/worker/processor.go:766` (→ `post_retrieval_narrow`), `internal/worker/personal_processor.go:997` (→ `map_single`), `internal/worker/personal_processor.go:1070` (→ `map_chunk`), `internal/worker/personal_processor.go:1133` (→ `reduce`).
- ✅ **The "closed forceFn set" claim is true** — exactly four `toolCallFn(...)` invocations pass a forced name (`internal/pipeline/resolve_topic.go:141`, `internal/pipeline/intent.go:265`, `internal/pipeline/resolve_channel.go:712`, `internal/pipeline/narrow.go:115`) and they match the four tool-name literals at `internal/pipeline/resolve_topic.go:26`, `internal/pipeline/intent.go:87`, `internal/pipeline/resolve_channel.go:64`, `internal/pipeline/narrow.go:27`. No fifth forced function exists, so per-function classes stay bounded.
- ✅ **Switch ordering holds** — the unbounded chunk-map prefix at `internal/timing/timing.go:215` precedes `map_single` at `:217`; the tightened reduce rule at `:221` lets `ReducerDebug` fall through to `other`, matching `internal/timing/timing_test.go:380`.
- ✅ **All four new tests hand-traced green** — 11 census rows, 4 fall-through rows, and 1000 chunk indices collapsing to exactly one class all match the switch as written. `strings` was already imported at `internal/timing/timing.go:21`; the new `fmt` import is used at `internal/timing/timing_test.go:394`.
- ✅ **No CI gate breaks** — `.golangci.yml` sets `default: none` and enables only `govet`, so an unused exported symbol in an `internal/` package is not flagged; no generated file, dependency or workflow is touched.

## Findings

No P0/P1 issues; two P2 items below. Both are diff-scope **new** (introduced by this PR, not pre-existing behaviour), and neither makes a working path unusable, produces user-visible wrong data, or leaves the system worse than before the change — so neither reaches the P1 bar.

### P2 — `PurposeClass` ships with no production consumer (`internal/timing/timing.go:194`)

The only references to this symbol in the whole module are the definition itself and `internal/timing/timing_test.go`. No metric is labelled with it: `internal/llmobs/metrics.go` builds every series from the closed `llmfallback.Path` set plus model/position/outcome/reason (`internal/llmobs/metrics.go:186`, `:196`, `:214`), and the only production reader of `LLMCall.Purpose` is the human-readable report line at `internal/timing/timing.go:322` — which this doc comment itself says is fine. So at head the cardinality bomb described at `internal/timing/timing.go:175` is not reachable, the PR changes no runtime behaviour, and the "MUST label on PurposeClass" rule is enforced only by prose: the #231 S4 author gets no compile-time or test-time signal, and the vocabulary can drift unnoticed while it stays unwired. Land the consumer in this stage — the natural seam is `FlushReport` at `internal/timing/timing.go:254`, which already walks `acct[taskNo]` and could emit a `class=PurposeClass(c.Purpose)` series — or state explicitly in the PR description that wiring is S4 and add the drift guard below so the classifier cannot rot in the meantime.

### P2 — the census is hand-copied, and the anti-vacuity guard polices the copy (`internal/timing/timing_test.go:332`)

`TestPurposeClass_NoProductionPurposeIsOther` at `internal/timing/timing_test.go:364` iterates this same `productionPurposes` table, so it goes red only when `PurposeClass` changes — never when a call site does. That is precisely the root cause of the defect the follow-up commit fixes (classes derived from placeholder strings, the whole retrieval-prep family buried in `other`): the table was re-derived by hand, but the derivation is still unenforced, so the next `RecordLLM*` site or the next forced function added under `internal/pipeline` lands in `other` with a green build. The contract is also load-bearing on a format string duplicated byte-for-byte in two closures — `internal/worker/processor.go:756` and `internal/worker/personal_processor.go:707` — so editing one copy's `"检索预处理: "` prefix silently unclassifies that whole path while every test still passes. Give the purpose format one source of truth: extract the duplicated closure body into a helper such as `timing.RetrievalPrepPurpose(forceFn)`, export the closed forceFn set next to the pipeline tool definitions, and assert from a worker-level test that every name in that set classifies to something other than `other`.

## 数据流回溯 (data-flow traceback)

- `purpose` consumed by `PurposeClass` → produced by the two worker closures (`internal/worker/processor.go:756`, `internal/worker/personal_processor.go:707`) and six literal sites → stored into `LLMCall.Purpose` by `RecordLLM` at `internal/timing/timing.go:161` → accumulated in `acct[taskNo]` → drained by `FlushReport` at `internal/timing/timing.go:259` and rendered raw at `:322`. **The consumer end is missing**: the stored value never flows through `PurposeClass`, so the classification product reaches nothing. → P2 #1.
- `forceFn` → four inline literals in `internal/pipeline`, each behind a `topic == "" || toolCallFn == nil` guard. When a guard trips the callback is never invoked and no record is produced, so that path cannot leak an `other` value.
- `forceFn == ""` branch → `"检索预处理(tool-call)"` → `retrieval_prep`. No caller passes an empty forceFn at head, so this class is currently unreachable — defensive, not a defect.
- `idx` in `fmt.Sprintf("Map: 分块总结 chunk#%d", idx)` → loop index over `chunks` at `internal/worker/personal_processor.go:1033`, genuinely unbounded → correctly collapsed by the prefix rule at `internal/timing/timing.go:215`.
- `productionPurposes` → upstream is the author's manual transcription, not the call sites. No runtime data flows from production into the guard. → P2 #2.

## 盲点 checklist (C1–C6)

- **C1 双路径 parity — HIT.** The purpose-format string exists in two byte-identical closures (`internal/worker/processor.go:756`, `internal/worker/personal_processor.go:707`) while `PurposeClass` exact-matches one spelling. Verified identical at head, so classification is correct today; the exposure is future one-sided drift → folded into P2 #2. No add/remove, subscribe/unsubscribe or acquire/release pair is touched by the diff.
- **C2 control-flow ordering / 嵌套复用 — clear.** Pure function with no call sites, so no nested or re-entrant ordering. All 11 arms hand-traced in evaluation order; the `Contains` arm at `internal/timing/timing.go:211` precedes the prefix arms but no production literal satisfies two arms. Non-canonical forms tried: full-width colon `检索预处理：`, trailing space, `ReducerDebug`, bare `Reduce`, lowercase `reduce:`, uppercase `MAP:` — all fall to `other`, which is the documented fail-closed behaviour.
- **C3 授权边界 ≠ 能力边界 — N/A.** No permission, jail, tool-exposure or credential change; the classifier is not reachable from any request path.
- **C4 授权生命周期 / 容器-成员级联 — N/A.** No auth, container/member status, query filter or masking logic touched.
- **C5 build/note 通过 ≠ 运行期路径正确 — HIT.** The new tests are a hand-fed table, so a green suite would not prove production coverage; I re-derived the census from the call sites instead of trusting it. No build or test was executed here, so nothing is claimed from a green pipeline.
- **C6 治理/策略/安全文档自洽性 — N/A.** No `SECURITY.md`, `CONTRIBUTING`, `CODEOWNERS`, label-policy or workflow change. The only prose is a Go doc comment, and its two factual claims (complete census, closed forceFn set) were both verified true.

**Security pass (STRIDE/OWASP):** Spoofing N/A (no identity). Tampering/injection N/A — output is a closed set of 11 ASCII literals never derived from request data, and if it later becomes a label value `escapeLabelValue` at `internal/llmobs/metrics.go:64` already handles exposition-format escaping. Repudiation N/A (no actor attribution). Information disclosure N/A, mildly positive — keeping `chunk#%d` out of labels reduces what a scraper sees. DoS N/A (O(1) string comparisons on a short string, no loop, no allocation). Elevation of privilege N/A. Not a security-sensitive PR.

## 跨轮 blocker 复检 (R6)

Head carries a follow-up commit `638646c` ("fix(timing): align PurposeClass with the real RecordLLM* census (#240 review)") whose message and diff record the prior round, so this re-check uses checkout-local evidence only.

1. **Dead `intent` class / retrieval-prep family in `other` — 已修.** `internal/timing/timing.go:201` now classifies all four forced functions plus the no-force `(tool-call)` case; independently verified against the four real `toolCallFn` call sites listed above.
2. **`HasPrefix(purpose, "Reduce")` too loose (prior P2) — 已修.** `internal/timing/timing.go:221` is now `purpose == "Reduce" || strings.HasPrefix(purpose, "Reduce:")`, with a regression row at `internal/timing/timing_test.go:380`.
3. **Anti-vacuity guard — 部分修.** The guard was added but iterates the hand-written table, so it cannot detect call-site drift → P2 #2 above.
4. **Placeholder-only cases removed — 已修.** `意图识别`, bare `单次总结` and `Reduce: 最终总结` are gone from the switch and are referenced by no non-test code. `internal/timing/timing_test.go:183` and `:252` still feed `"Reduce"` / `"单次总结"` to `RecordLLM` for report-rendering tests; those never reach `PurposeClass`, so the removal breaks nothing.

No unresolved P0/P1 from the prior round survives at head. This is round 2, so no "stop whack-a-mole, redesign" escalation is warranted.

- No behaviour change for the existing report: `FlushReport` still renders raw `Purpose`, so operators keep the chunk index in the log exactly as before.
- The agent runtime is untouched: `internal/agent` makes no `timing.RecordLLM*` calls, so its LLM observability still comes from `llmfallback` path labels. The census claim is correctly scoped to the worker pipeline.
- Concurrency: `PurposeClass` is pure and reads no shared state, so the parallel Map goroutines at `internal/worker/personal_processor.go:1035` are unaffected.
- No dependency, migration, generated-file, workflow or deployment-surface change.

## Verdict: APPROVED

The classifier is correct against the real call-site census, the closed forceFn set claim checks out, the switch ordering is sound, and the change cannot regress any production path because it changes none. The two P2s are about durability rather than correctness: the helper is unwired, and its census guard is self-referential. Both are worth fixing before or alongside the S4 metric that consumes this, but neither blocks this PR.

[Octo-Q] verdict: APPROVE — 零 P0/P1；两个 P2 均为「未接线 + 手工 census 无强制」的耐久性问题，不影响任何现有生产路径（本 PR 为纯新增，`PurposeClass` 在生产中无调用方），按 R1/R4 映射不构成阻塞。

yujiawei (approved):
# Code Review — PR #240 (octo-smart-summary)

Re-review at head `638646c5771f05acb76d8612cb3979d051ecec94`, merge-base `391134cc` (current `main` tip). Round 2. I previously requested changes on head `0218e127`; this is an independent pass on the new head — the earlier verdict does not carry over.

## 1. Spec compliance

**Spec: ✅**

Governing requirement (#231 S3): `purpose` must be normalized before it can be a metric label — `fmt.Sprintf("Map: 分块总结 chunk#%d", idx)` must collapse to a single `map_chunk` class, with the index staying in the log line only.

**Delivered:** met, and the round-1 blockers are genuinely fixed, not papered over.

I re-derived the call-site census independently rather than trusting the doc comment. There are exactly 8 `RecordLLM`/`RecordLLMSince` sites in non-test code at head:

| Site | Purpose emitted | `PurposeClass` | Correct? |
|---|---|---|---|
| `internal/worker/meta_processor.go:237` | `团队汇总: 合并各成员总结` | `team_reduce` | ✅ |
| `internal/worker/processor.go:760` | `检索预处理(tool-call)` / `检索预处理: <forceFn>` | `retrieval_prep` / per-fn | ✅ |
| `internal/worker/processor.go:766` | `检索后裁剪 PostRetrievalNarrow` | `post_retrieval_narrow` | ✅ |
| `internal/worker/personal_processor.go:711` | same as `processor.go:760` | `retrieval_prep` / per-fn | ✅ |
| `internal/worker/personal_processor.go:719` | `检索后裁剪 PostRetrievalNarrow` | `post_retrieval_narrow` | ✅ |
| `internal/worker/personal_processor.go:997` | `Map: 单次总结(跳过Map-Reduce)` | `map_single` | ✅ |
| `internal/worker/personal_processor.go:1070` | `Map: 分块总结 chunk#N` | `map_chunk` | ✅ |
| `internal/worker/personal_processor.go:1133` | `Reduce: 合并分块总结` | `reduce` | ✅ |

`forceFn` is still the closed 4-literal set — `recognize_intent` (`internal/pipeline/intent.go:265`), `extract_time_range` (`narrow.go:115`), `resolve_channel_scope` (`resolve_channel.go:712`), `resolve_topic_target` (`resolve_topic.go:141`) — so the per-function split at `internal/timing/timing.go:200-207` is bounded. Every one of the 10 distinct production purposes now maps to a real class; nothing lands in `other` any more. That was the round-1 P1 and it is closed.

- **Nothing missing.** Round-1 P1 (census wrong, `意图识别` dead, retrieval-prep family falling to `other`) and round-1 P2 (`strings.HasPrefix(purpose, "Reduce")` too loose) are both addressed at `timing.go:200-207` and `timing.go:220`.
- **Nothing over-built.** The class set widened from 7 to 11 because the retrieval-prep family is now split per forced function. That option was explicitly offered in round 1 ("either one `retrieval_prep` class or per-forced-function classes"), the doc comment at `timing.go:186-191` justifies it against #220's need for `recognize_intent`'s percentile, and 11 is still a small closed set. Sanctioned, not scope creep.
- **No divergence.** Still confined to `internal/timing/`, still no emission wired, still no conflict with #232 (which does not touch this package). `PurposeClass` has no consumer outside its own test at head — deliberate staging, declared in the PR description.

## 2. Code quality

**Quality: Approved**

### Verified with executable evidence at head

- **Build / vet / test green locally.** `go build ./internal/...` exit 0, `go vet ./internal/timing/` clean, `go test ./internal/timing/ -count=1` → `ok`. CI is green on Build / Test / Test (race, cgo) / Vet / Lint / Generate (no drift).
- **`strings` was already imported** at `internal/timing/timing.go:21` on the base revision — no new import needed, no build risk.
- **Rule ordering holds for every tricky literal.** `"Map: 单次总结(跳过Map-Reduce)"` contains the substring `Reduce` but is caught by the `map_single` prefix at `:216` before the reduce rule at `:220`; `"Reduce: 合并分块总结"` cannot hit `map_chunk`, which requires the full `Map: 分块总结 chunk#` prefix; the chunk-map case is correctly ordered ahead of the looser `Map:` rule.
- **The test suite is now mutation-resistant.** This is the round-1 vacuity finding, so I probed it rather than reading it. Six mutations, each applied to `timing.go`, `go test ./internal/timing/ -run PurposeClass`, then reverted:

  | Mutation | Result |
  |---|---|
  | `return purpose` (identity classifier) | 🔴 RealCallSites, UnknownFallsThrough, ChunkIndexIsBounded |
  | delete all four retrieval-prep cases | 🔴 RealCallSites, NoProductionPurposeIsOther |
  | insert loose `HasPrefix("检索预处理") → "map_chunk"` first | 🔴 RealCallSites, UnknownFallsThrough |
  | delete the `map_chunk` case | 🔴 RealCallSites, NoProductionPurposeIsOther |
  | revert reduce rule to `HasPrefix("Reduce")` | 🔴 UnknownFallsThrough |
  | `post_retrieval_narrow` → `other` | 🔴 RealCallSites, NoProductionPurposeIsOther |

  All six go red, and the tree is byte-identical after restore. In round 1 the third mutation left the whole suite green; it no longer does. `productionPurposes` (`timing_test.go:332-347`) is the real 10-value domain plus a second chunk index, and it is the census, not a table of speculative literals.
- **Cardinality is bounded for every reachable value.** 1000 chunk indices → exactly one class (`ChunkIndexIsBounded`), unknowns → `other` (`UnknownFallsThrough`, including a would-be-new forced function), so the label space cannot widen without a code change plus a test change.

### 🟡 P2 — non-blocking

1. **The PR description no longer matches the code.** The body advertises the closed set as `intent · map_chunk · map_single · reduce · team_reduce · post_retrieval_narrow · other` (7 values) and names `TestPurposeClass_ClosedSet`. The code returns 11 values (`timing.go:192-194`) and that test no longer exists — it was replaced by `TestPurposeClass_RealCallSites` / `_NoProductionPurposeIsOther` / `_UnknownFallsThrough`. The in-code doc comment is accurate; only the PR body is stale. Worth fixing before merge because this is a durable record of the contract that the emission PR will read, and a stale doc claim is exactly what round 1 tripped over — but it is a description edit, not a code change, so it does not block.

2. **`strings.Contains` is the one remaining loose rule.** `internal/timing/timing.go:211`:

   ```go
   case strings.Contains(purpose, "PostRetrievalNarrow"):
       return "post_retrieval_narrow"
   ```

   Every other case is exact-equality or a deliberate prefix; this one matches anywhere in the string, so a future `"SkipPostRetrievalNarrow"` or `"PostRetrievalNarrow (disabled)"` would silently join the class instead of falling through to `other`. `purpose == "检索后裁剪 PostRetrievalNarrow"` would match the discipline you just applied to the reduce rule. Same note applies more weakly to `strings.HasPrefix(purpose, "团队汇总")` at `:218` (no colon, so `团队汇总X` joins).

3. **Two classes are reachable in source but not at runtime.** Not a defect in this PR, but the emission PR should know before it builds dashboards on them:
   - `post_retrieval_narrow` — the `llmFn` closure that records it (`processor.go:766`, `personal_processor.go:719`) is consumed only by `PostRetrievalNarrow`, which is a stub: `internal/pipeline/narrow.go:162-164` returns `messages` without ever calling `llmFn`. `fetch.go:1163` is its only caller. (`NarrowByTopicReport` also takes an `LLMCallFn`, but its sole caller is `internal/agent/tool_narrow_channels.go:77` with its own closure, and the agent path does not use `internal/timing`.)
   - `retrieval_prep` — the `forceFn == ""` branch. All four `toolCallFn(...)` invocations in `internal/pipeline/` pass a non-empty literal, so the no-force purpose is not produced today.

   Mapping them anyway is the right call — better than letting a source-level purpose fall into `other` — so no change is needed in the classifier. One clause in the doc comment noting that these two are defensive would stop a dashboard author from waiting on a series that will never emit.

4. **Census drift is still guarded only by convention.** Adding a new `RecordLLM*` site with a new literal fails nothing; the guard is the "keep this in sync" comment at `timing_test.go:329-331`. Round 1 happened because a hand-maintained table drifted. A test that greps the tree for `RecordLLM*` string arguments and asserts each is in `productionPurposes` would make the next drift loud. Reasonable to defer to the emission PR.

5. **The silent `other` bucket has no diagnostic.** When the emission PR lands, an unclassified purpose disappears into `other` with no warning log and no separate counter, so a forgotten classifier update is invisible. Consider emitting `other` as its own series (or logging once per unknown value) so the junk drawer is observable.

6. **Class-name asymmetry.** `recognize_intent → "intent"` is abbreviated while the other three forced functions map to their verbatim names (`extract_time_range`, `resolve_channel_scope`, `resolve_topic_target`). These become operator-facing label values; picking one convention now is cheaper than renaming after dashboards exist.

## 3. Overall verdict

**APPROVE**

No P0 or P1. The round-1 blocker was a real defect and the fix is real — I confirmed it by mutation rather than by reading the diff. The cardinality bound holds for every value the tree can emit, the census matches the 8 actual call sites, and the test suite now fails on every misclassification I could construct. The remaining items are all P2: one stale PR description, one loose match rule, and four notes for the emission PR.

## 4. Suggestions

- Before merge: update the PR body's class list to the 11 actual values and drop the reference to `TestPurposeClass_ClosedSet`.
- Tighten `timing.go:211` to `purpose == "检索后裁剪 PostRetrievalNarrow"`.
- Add a sentence to the doc comment marking `post_retrieval_narrow` and `retrieval_prep` as defensive / not currently emitted.
- For the emission PR: make `other` an observable series, and consider automating the census check.

## 5. Additional findings outside the diff

- **`check-sprint / check-sprint` is failing** on this PR (`Validate Sprint field on Project Board`). Every code check passes; this is a project-board metadata gate, not a code defect, but it will block merge until the Sprint field is set on the board entry.
- **`PostRetrievalNarrow` is a no-op stub** (`internal/pipeline/narrow.go:162-164`) that still costs two `RecordLLMSince` closures and a parameter threaded through `ResolveAndFetchMessagesForPersonal`. Either implement it or delete the dead plumbing — unrelated to this PR, worth its own issue.
- **The classification-from-prose pattern remains the underlying fragility.** `internal/llmfallback/observer.go:11` already models the drift-proof shape: a bounded `Path` enum (`worker_map` / `worker_reduce` / `tool_call` / `api_refine` / `agent_chat` / `agent_tool`) carried as a value rather than parsed out of display text. Passing a class into `RecordLLM`/`RecordLLMSince` alongside the human-readable purpose would remove the parsing step entirely. That is a larger refactor than #231 S3 should absorb, and the closed-set discipline here is an acceptable interim, but it is the right long-term direction.

yujiawei:
Follow-up to my approving review — no verdict change, one P2 sharpened with executable evidence.

My review's P2 #4 said census drift is guarded only by convention, but I framed it on the *consumer* side (a new `RecordLLM*` site landing in `other`). There is a second, sharper vector on the *producer* side that I under-weighted: the purpose format string is duplicated byte-for-byte in two closures — `internal/worker/processor.go:753-762` and `internal/worker/personal_processor.go:704-713`, identical except for the `taskNo` expression:

```go
purpose := "检索预处理(tool-call)"
if forceFn != "" {
    purpose = "检索预处理: " + forceFn
}
```

`PurposeClass` exact-matches one spelling of that prefix, so editing one copy silently unclassifies that entire path. I probed it rather than reasoning about it — changed `"检索预处理: "` to `"检索预处理 - "` in `processor.go` only, leaving `personal_processor.go` untouched:

```
go build ./internal/...                              exit 0
go vet ./internal/timing/ ./internal/worker/         exit 0
go test ./internal/timing/ ./internal/worker/        ok    ok
```

Fully green, while every workspace-path retrieval-prep call would classify to `other`. That is the same failure mode as the round-1 defect — a class silently emptied into the junk drawer with a passing suite — just re-armed on the producer side instead of the table side. Nothing is wrong at head; both copies are correct today. It stays P2 because it needs a future edit to trigger. But it is the highest-value P2 in my list, and extracting the format into one source of truth (e.g. `timing.RetrievalPrepPurpose(forceFn)`, with the closed `forceFn` set exported next to the pipeline tool definitions) would collapse it and my P2 #4 into a single fix.

One caution on a remedy worth flagging before the emission stage: `FlushReport` (`internal/timing/timing.go:254`) looks like a convenient seam to wire this into, since it already walks `acct[taskNo]`. It is not a good one. It is the human-report builder, it fires once per run rather than per call — so it would yield per-run aggregates, not the per-call observations a latency histogram needs — and emitting from it would both couple `internal/timing` to the metrics registry and change the report path this PR deliberately leaves untouched. The emission belongs in the metrics registry alongside the existing `llmobs` series, not here.

Verdict stands: **APPROVE**. Every code check is green; the outstanding merge blocker is `check-sprint` (Sprint field on the project board), not the code.

Issue #242, opened 2026-09-09
chore(obs): extract internal/metrics registry (S2) and wire agent/worker app-level metrics (S3-b)


wanyiwang06:
Tracking issue for the remaining observability work after #231 was closed on the histogram (S1, PR #232) and the label-cardinality normalization (S3-a, PR #240) landing. These two pieces were deliberately deferred and now have no open issue.

## S2 — extract a neutral `internal/metrics` package
The registry primitives (`labelSet`/`Labels`, `escapeLabelValue`, `counterVec`, `gaugeVec`, `histogramVec`, `withLE`/`formatBucket`, exposition writing) are currently welded into `internal/llmobs`, which is LLM-fallback-specific. Lift them into a general-purpose `internal/metrics` package with `llmobs` as the first consumer.

- Pure refactor, **zero behavior change**: `llmobs`'s `WritePrometheus` output must stay byte-identical (existing `internal/llmobs` tests are the regression net).
- Motivation: the worker and agent paths (S3-b below) need to emit metrics, and they must not import `llmobs` (wrong layering / import cycle risk). A neutral registry is the prerequisite.

## S3-b — wire agent and worker app-level metrics
S1 only measures the LLM sub-call. The user-facing latency is the whole pipeline (fetch → intent → map → reduce) and the whole agent turn (planning + tools). Emit, using the neutral registry:

- **Worker** (single seam: `internal/timing`): `summary_stage_duration_seconds{stage}`, and LLM-call latency labelled by `timing.PurposeClass(...)` (the closed set from #240 — currently a loaded gun with no trigger).
- **Agent** (single seam: `RunTrace.Report`): `agent_request_duration_seconds{outcome}`, `agent_phase_duration_seconds{phase}`, `agent_tool_duration_seconds{tool}`, `agent_prompt_chars`, `agent_steps`. Metrics always-on; `AGENT_TRACE` keeps gating only the log block.

## Why this matters
- #220 must size `AGENT_STEP_TIMEOUT` (whole agent turn, incl. tools/planning) — that distribution only exists in S3-b, not in the LLM-call histogram from #232.
- #240's `PurposeClass` produces no series until S3-b consumes it.

## Order
S2 → S3-b (S3-b depends on the extracted registry). Both are unblocked now that #232 has merged and `internal/llmobs/metrics.go` has its final shape.

wanyiwang06:
[...]

PR #243, opened 2026-09-09
refactor(obs): extract internal/metrics registry (#242 S2)

wanyiwang06:
## What
Lifts the hand-rolled Prometheus registry primitives out of `internal/llmobs` into a neutral `internal/metrics` package.

- New `internal/metrics`: `LabelSet`/`Labels` (+ value escaping), `CounterVec`, `GaugeVec`, `HistogramVec`, and a `Registry` that renders registered collectors **in registration order**.
- `llmobs.Metrics` now composes `metrics.*` vectors and delegates `WritePrometheus` to a `Registry`, registered in the same order as the previous hand-written write sequence.

## Why (#242 S2)
The registry was welded into `llmobs`, which is LLM-fallback-specific. The upcoming worker/agent app-level metrics (#242 S3-b) must emit without importing `llmobs` (wrong layering / import-cycle risk). A neutral package is the prerequisite.

## Zero behavior change
Pure refactor — no metric names, help text, bucket layouts, or label sets change. Scrape output is **byte-identical**: the existing `internal/llmobs` exposition tests (which parse the text format and assert structure, sorting, escaping, histogram cumulative/`+Inf`/`le` semantics) pass unchanged and are the regression net.

## Tests
- `internal/llmobs`: unchanged, green (byte-identical output guard).
- `internal/metrics`: new standalone tests — label escaping + odd-arg panic, counter/gauge exposition, histogram cumulative buckets / inclusive `le` / `+Inf == _count` / `_sum` / negative-clamp, and `Registry` registration-order determinism. `-race` green.

Tracked in #242. Follow-up: S3-b wires `internal/timing` and `RunTrace.Report` to emit through this registry (consuming `timing.PurposeClass` from #240).

### DIFF
--- a/internal/llmobs/metrics.go
+++ b/internal/llmobs/metrics.go
@@ -12,137 +12,12 @@
 package llmobs
 
 import (
-	"fmt"
 	"io"
-	"sort"
-	"strconv"
-	"strings"
-	"sync"
 	"time"
 
 	"github.com/Mininglamp-OSS/octo-smart-summary/internal/llmfallback"
+	"github.com/Mininglamp-OSS/octo-smart-summary/internal/metrics"
 )
-
-// labelSet is a bounded, ordered key/value list used as a map key. Every label
-// value fed into it comes from configuration (model ids) or a closed constant
-// set (path, position, outcome, reason) — never from request data — so the
-// series count stays bounded by models * paths.
-type labelSet string
-
-func labels(pairs ...string) labelSet {
-	if len(pairs)%2 != 0 {
-		panic("llmobs: labels requires key/value pairs")
-	}
-	var b strings.Builder
-	for i := 0; i < len(pairs); i += 2 {
-		if i > 0 {
-			b.WriteByte(',')
-		}
-		b.WriteString(pairs[i])
-		b.WriteString("=\"")
-		b.WriteString(escapeLabelValue(pairs[i+1]))
-		b.WriteString("\"")
-	}
-	return labelSet(b.String())
-}
-
-// labelEscaper is built once at package scope, NOT per call. strings.NewReplacer
-// compiles a trie on construction, so building it inside escapeLabelValue cost
-// ~3.5µs and 6.7KB of garbage per label value (measured) against ~26ns and zero
-// allocations when hoisted — and escapeLabelValue runs on every label of every
-// event on the LLM hot path.
-//
-// \r is escaped alongside the three the text format requires: it is not a
-// separator so it cannot forge a line, but leaving a bare CR in a value makes
-// log/terminal output overwrite itself, and SafeTextForLog already strips it
-// elsewhere in this codebase.
-// CR is replaced with a space, NOT escaped as \r. The Prometheus text format
-// defines only \\, \" and \n inside a label value; \r is an invalid escape and
-// the reference parser aborts the ENTIRE scrape on it, not just that line. So
-// "hardening" a bare CR into \r traded one ugly character for the loss of every
-// metric in the response. Flattening matches what SafeTextForLog already does.
-var labelEscaper = strings.NewReplacer(`\`, `\\`, `"`, `\"`, "\n", `\n`, "\r", " ")
-
-// escapeLabelValue applies the Prometheus text-format escaping rules.
-func escapeLabelValue(v string) string {
-	return labelEscaper.Replace(v)
-}
-
-// counterVec is a labelled monotonic counter.
-type counterVec struct {
-	name string
-	help string
-	mu   sync.Mutex
-	vals map[labelSet]float64
-}
-
-func newCounterVec(name, help string) *counterVec {
-	return &counterVec{name: name, help: help, vals: map[labelSet]float64{}}
-}
-
-func (c *counterVec) inc(l labelSet) { c.add(l, 1) }
-
-func (c *counterVec) add(l labelSet, delta float64) {
-	c.mu.Lock()
-	c.vals[l] += delta
-	c.mu.Unlock()
-}
-
-func (c *counterVec) write(w io.Writer) {
-	c.mu.Lock()
-	keys := make([]labelSet, 0, len(c.vals))
-	for k := range c.vals {
-		keys = append(keys, k)
-	}
-	snapshot := make(map[labelSet]float64, len(c.vals))
-	for k, v := range c.vals {
-		snapshot[k] = v
-	}
-	c.mu.Unlock()
-
-	sort.Slice(keys, func(i, j int) bool { return keys[i] < keys[j] })
-	fmt.Fprintf(w, "# HELP %s %s\n# TYPE %s counter\n", c.name, c.help, c.name)
-	for _, k := range keys {
-		fmt.Fprintf(w, "%s{%s} %g\n", c.name, k, snapshot[k])
-	}
-}
-
-// gaugeVec is a labelled last-write-wins gauge.
-type gaugeVec struct {
-	name string
-	help string
-	mu   sync.Mutex
-	vals map[labelSet]float64
-}
-
-func newGaugeVec(name, help string) *gaugeVec {
-	return &gaugeVec{name: name, help: help, vals: map[labelSet]float64{}}
-}
-
-func (g *gaugeVec) set(l labelSet, v float64) {
-	g.mu.Lock()
-	g.vals[l] = v
-	g.mu.Unlock()
-}
-
-func (g *gaugeVec) write(w io.Writer) {
-	g.mu.Lock()
-	keys := make([]labelSet, 0, len(g.vals))
-	for k := range g.vals {
-		keys = append(keys, k)
-	}
-	snapshot := make(map[labelSet]float64, len(g.vals))
-	for k, v := range g.vals {
-		snapshot[k] = v
-	}
-	g.mu.Unlock()
-
-	sort.Slice(keys, func(i, j int) bool { return keys[i] < keys[j] })
-	fmt.Fprintf(w, "# HELP %s %s\n# TYPE %s gauge\n", g.name, g.help, g.name)
-	for _, k := range keys {
-		fmt.Fprintf(w, "%s{%s} %g\n", g.name, k, snapshot[k])
-	}
-}
 
 // durationBuckets are the shared second-boundaries for every latency
 // histogram in this package.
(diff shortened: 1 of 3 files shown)


Jerry-Xin (approved):
## Verdict: APPROVE

Pure extraction refactor (#242 S2): lifts the hand-rolled Prometheus primitives out of `internal/llmobs` into a neutral `internal/metrics` package. Byte-level equivalence and the zero-behavior-change claim were independently verified.

### Verification

- **Byte-equivalence of the extraction**: compared the new `internal/metrics` primitives against the merge-base `internal/llmobs/metrics.go` line by line — `Labels` construction, the label escaper (including the CR→space flattening rationale), counter/gauge snapshot+sort+exposition, histogram cumulative rendering (inclusive `le`, `+Inf == _count`, `_sum`, negative-value clamp), `withLE`, `formatBucket`: semantics identical. The only textual delta is the odd-argument panic prefix (`llmobs:` → `metrics:`), a programming-error path, not scrape output.
- **Exposition order preserved**: the previous hand-written `WritePrometheus` sequence (attempts, switches, calls, callSecs, runDur, attemptDur, lastOK) matches the new `Registry.MustRegister` order exactly; the Registry renders in registration order and snapshots under lock before writing to a potentially slow client.
- **Metric identity preserved**: all seven metric names, help texts, label sets, and the shared `durationBuckets` layout are unchanged from the merge base (string literals compared side by side).
- **Call-site census**: the sole consumer of `metrics.NewCounterVec`/`NewGaugeVec`/`NewHistogramVec`/`MustRegister` is `llmobs.NewMetrics`; no leftover references to the removed private types, no other direct registration anywhere in the tree; `WritePrometheus` delegation intact; the `/internal/metrics` route is untouched.
- **Duplicate registration**: `MustRegister` is invoked exactly once per `NewMetrics` on a fresh Registry — no global state, so no double-render or panic path exists.
- **Tests**: `go test -race` green for `internal/metrics` (new: escaping + odd-arg panic, counter/gauge exposition, cumulative buckets / `+Inf` / `_sum` / negative clamp, registration-order determinism) and `internal/llmobs` (existing byte-identical exposition net, unchanged). `go vet` clean. Environment disclosure: `internal/api/router` tests could not link on the review machine due to a missing external `tokenizers` cgo library (the same link failure affects every `cmd/*` binary here); this diff is pure Go across 3 files with no cgo surface, so the failure is environmental and not attributable to the PR.

### 💬 Non-blocking (fast-follow candidates for #242 S3-b)

- 🔵 `NewHistogramVec` copies the bucket slice but does not validate that boundaries are strictly increasing and finite. This is inherited from the merge-base constructor and unreachable at this head (the sole caller passes a fixed valid literal), so it is not a defect in this diff. However, now that the constructor is exported and S3-b will bring worker/agent callers, a construction-time guard (panic on non-increasing, duplicate, NaN, or ±Inf boundaries) plus tests for those cases would turn a silent misclassification into an immediate programming error.
- 🔵 `LabelSet` as an exported string type lets a caller bypass `Labels` and inject unescaped exposition text. A struct with an unexported encoded field — and rejecting the histogram-reserved `le` label in `Labels` — would harden the reusable API surface.

The extraction is faithful, the regression net is intact, and the layering goal (metrics emission without importing `llmobs`) is achieved with no semantic drift.

yujiawei (approved):
# Code Review — PR #243 (octo-smart-summary)

**Verdict: APPROVE.** This is a clean, faithful extraction of the hand-rolled Prometheus primitives out of `internal/llmobs` into a neutral `internal/metrics` package. The "zero behavior change / byte-identical scrape output" claim holds up under line-level scrutiny and under the retained regression net.

## Spec compliance

- **No missing work:** All primitives are moved — `LabelSet`/`Labels`, `CounterVec`, `GaugeVec`, `HistogramVec`, and the new `Registry`. `llmobs.Metrics` now composes `metrics.*` vectors and delegates `WritePrometheus` to the registry.
- **No over-build:** No new metric names, help text, bucket layouts, or label sets. The only additive surface is `Registry` (needed to preserve exposition ordering deterministically) and the exporting of the previously-private types — both squarely within the stated goal of letting non-`llmobs` paths emit without an import cycle (#242 S3-b prerequisite).
- **No deviation:** Registration order in `NewMetrics` is exactly `attempts, switches, calls, callSecs, runDur, attemptDur, lastOK`, matching the previous hand-written `WritePrometheus` sequence 1:1, so family ordering in the scrape output is unchanged.

## Behavior-preservation checks (all pass)

- Escaping (`\\`, `\"`, `\n`, and CR→space flattening), series sorting, `%g` value formatting, inclusive `le` bucket lookup via `SearchFloat64s`, cumulative bucket accumulation, `+Inf == _count`, `_sum`, negative-clamp, and the mutex snapshot-under-lock pattern are all direct equivalents of the prior code.
- `Registry.MustRegister` / `WritePrometheus` lock correctly and snapshot the collector slice under the lock, so concurrent scrape + register is race-free and preserves registration order.
- Nice incidental cleanup: the base file carried two contradictory comment blocks about `\r` handling (one advocating `\r` escaping, one advocating CR→space); the extraction keeps only the correct CR→space rationale.

## Verification performed

- `go build ./...`, `go vet ./internal/metrics/... ./internal/llmobs/...`: clean.
- `go test -race ./internal/metrics/... ./internal/llmobs/...`: green. The **unchanged** `internal/llmobs` exposition tests (which parse the text format and assert sorting, escaping, histogram cumulative/`+Inf`/`le` semantics) are the real regression net and still pass through the new delegation path.
- `gofmt -l`: clean.
- Confirmed the only Go consumer of the new package is `internal/llmobs`; the `/internal/metrics` reference in `internal/api/router` is an HTTP route string, not an import.

## Non-blocking suggestions (nits — for a future hardening pass, NOT this refactor)

These describe latent robustness gaps that exist **identically in the pre-extraction code**. Because this PR's contract is "zero behavior change," preserving them here is correct — they should not be silently fixed inside a refactor. Flagging for a follow-up:

1. `HistogramVec.Observe` does not guard against `NaN` (`h.sums[l] += NaN` would permanently poison `_sum` for that series). Not reachable on the current call path (`v` is always `Duration.Seconds()`), but now that the constructor is exported, a future caller could feed arbitrary input.
2. `NewHistogramVec` copies the bucket slice (good — prevents caller mutation) but does not sort/validate it. `SearchFloat64s` requires sorted, finite, distinct boundaries; an unsorted/NaN/dup boundary from a new external caller would misassign observations. Today only the fixed, valid `durationBuckets` is passed.
3. Exported `LabelSet` is a raw string rendered verbatim inside `{}`. A sibling package can now construct one directly and bypass `Labels`' escaping, enabling malformed/injected exposition lines. The package doc already warns callers to build only from `Labels`; a stricter type or a validating render would close the footgun.
4. `WriteProm` / `WritePrometheus` swallow `io.Writer` errors, so a broken-pipe mid-scrape yields a silently truncated payload the handler can't surface. Pre-existing.

Since none of these are regressions and all are out of scope for a pure refactor, they do not block. Consider tracking them alongside the #242 S3-b follow-up when the package gains its second consumer.

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #243 (octo-smart-summary)

## Summary

This PR lifts the hand-rolled Prometheus text-exposition registry out of `internal/llmobs` into a new domain-free package `internal/metrics`, exporting `LabelSet`/`Labels`, `CounterVec`, `GaugeVec`, `HistogramVec`, and a new `Collector`/`Registry` pair, and rewires `llmobs.Metrics` to build its seven vectors from the new package and render through `m.reg.WritePrometheus`. `internal/llmobs/metrics.go` shrinks by 262 lines to just the LLM-specific names, help text, bucket layout and label choices; `durationBuckets` correctly stays behind. A new 119-line test file covers the extracted primitives. The extraction is faithful: I diffed the moved code with identifiers normalised and the bodies are character-identical apart from export casing, the panic-message prefix (`llmobs:` → `metrics:`) and comment text.

Nothing here is broken. The findings below are all latent hardening gaps that this PR *newly exposes* by turning a package-private writer into a shared library whose stated purpose is worker and agent consumers — no current caller triggers any of them.

## Verification

Local checks at head `4174824`:
- `go build ./...` — clean
- `go vet ./internal/metrics/... ./internal/llmobs/...` — clean
- `gofmt -l` on the three changed files — clean
- `go test ./internal/metrics/... ./internal/llmobs/... ./internal/api/router/...` — pass
- `go test -race -shuffle=on -count=1 ./internal/metrics/... ./internal/llmobs/...` — pass

Methodology: `requesting-code-review` workflow (brief → coverage ledger → per-unit inspect → finalize). `finalize` reported 0 blocking errors, 0 invalid comments, 0 missing high-risk families.

## Findings

No P0/P1 issues. Three P2 items and two nits, all non-blocking.

### P2 — HELP text and metric names reach the exposition stream unescaped (`internal/metrics/metrics.go:140`)

All three writers (here, `:177`, `:258`) do `fmt.Fprintf(w, "# HELP %s %s\n# TYPE %s ...\n", name, help, name)` with the raw strings handed to the constructors. The text format requires HELP text to escape backslash and newline; a help string containing either forges or truncates exposition lines — and the package's own comment at `:60-64` argues precisely this failure mode for a bare CR in a label value ("the reference parser aborts the ENTIRE scrape"). The sanitiser covers only `escapeLabelValue` on values, so the input-shape contract is asymmetric. Nothing breaks today: I checked every help string in `internal/llmobs/metrics.go` and none contains a backslash or newline, so the fix is byte-identical for all current callers. But llmobs's help strings are already multi-sentence paragraphs that clear the constraint only by luck, and the next consumer adding a line break silently kills every metric in the scrape, not just its own family. Run `name`/`help` through a package-scope replacer in the three `WriteProm` methods, or validate at construction and panic the way `Labels` already does for an odd argument count.

### P2 — `Labels` validates arity but writes keys verbatim and allows duplicates (`internal/metrics/metrics.go:46`)

`b.WriteString(pairs[i])` emits the key with no escaping and no validation, while `pairs[i+1]` goes through `escapeLabelValue`. A key containing a newline, quote, space or `}` is emitted raw and can forge a whole sample line; `Labels("path","a","path","b")` renders `path="a",path="b"`, which the text parser rejects as a duplicate label. `Labels` already treats an odd argument count as a panic-worthy programming error, so the guard exists and is simply incomplete. No current caller is affected — all eight call sites in `internal/llmobs/metrics.go:104-163` pass literal keys. Validate keys against `[a-zA-Z_][a-zA-Z0-9_]*` and reject duplicates under the same panic contract, or state in the `Labels` doc that keys must be static identifiers.

### P2 — `MustRegister` performs no validation, so nil/duplicate fails at scrape time (`internal/metrics/metrics.go:90`)

`MustRegister(nil)` appends a nil `Collector` and the nil-interface call at `:103` panics later, inside `Registry.WritePrometheus` during an HTTP scrape — on the internal router that is recovered by `gin.Recovery()` into a 500 with a truncated body, which is the worst place to discover a composition-root mistake. The `Must` prefix also promises a panic it never performs: both the Go convention and `prometheus/client_golang`'s `MustRegister` — the library this package explicitly positions itself against at `:5-10` — panic on failure, and client_golang additionally panics on duplicate registration. A duplicate here silently emits two HELP/TYPE headers for one family, which this repo's own `parseExposition` helper already treats as an error (`internal/llmobs/metrics_test.go:59`). Panic on nil at register time and reject duplicates, or rename to `Register` if no panic semantics are intended.

### Nit — `CounterVec.Add` accepts a negative delta on a `# TYPE counter` family (`internal/metrics/metrics.go:121`)

The exported `Add` does `c.vals[l] += delta` with no sign check, so a consumer can make a monotonic counter go backwards; `rate()`/`increase()` read a decrease as a counter reset and discard the interval. `HistogramVec.Observe` already defends its accumulator against exactly this ("a bad caller must not corrupt `_sum` for everyone else on this series", `:221`); the same reasoning applies here but the guard is missing. `internal/llmobs` only calls `Inc` and `Add(..., e.Duration.Seconds())`, so nothing regresses today.

### Nit — The byte-identical ordering invariant is comment-only (`internal/llmobs/metrics.go:94`)

The comment asserts registration order "must match the previous hand-written `WritePrometheus` sequence to keep scrape output byte-identical", but no test pins llmobs's family order. `TestRegistry_RendersInRegistrationOrder` proves the primitive preserves order generically and `TestWritePrometheus_SeriesAreSorted` pins intra-family sorting only; reordering the `MustRegister` argument list at `:98` would pass the entire suite. Runtime consequence is limited (Prometheus ingestion is family-order-insensitive), so this is a golden-output/diffability concern. Either assert the seven family names appear in order in `TestWritePrometheus_ExpositionFormat`, or drop the byte-identical claim if ordering is not actually a contract.

## Data-flow traceback

Every consumed datum in the diff, traced to its producer and confirmed to actually arrive:

- `m.reg` → assigned in `NewMetrics` (`internal/llmobs/metrics.go:97-98`) → consumed by `Metrics.WritePrometheus` (`:169`) → `internal/api/router/router.go:209`. Cannot be nil on the production path: `&Metrics{...}` is constructed only inside `NewMetrics` (single literal, `:60`; grep repo-wide with `--no-ignore`), and `reg` is set before the constructor returns.
- Registration order → `MustRegister(m.attempts, m.switches, m.calls, m.callSecs, m.runDur, m.attemptDur, m.lastOK)` (`:98`) → `Registry.WritePrometheus` iterates a copy in slice order (`internal/metrics/metrics.go:99-104`). Byte-for-byte the same sequence as the base's hand-written `WritePrometheus` (base `internal/llmobs/metrics.go:391-397`). Confirmed identical, not assumed.
- `durationBuckets` → stays in llmobs (`:36`) → passed to `NewHistogramVec`, which copies the slice (`internal/metrics/metrics.go:203-204`), so `runDur` and `attemptDur` sharing one package-level slice cannot alias or corrupt each other via `sort.SearchFloat64s`.
- Label values → `string(e.Path)`, `e.Model`, `e.Position`, `outcomeLabel(e.Outcome)`, `e.From`, `e.To`, `string(e.Reason)`, `position` → all argument lists unchanged from base; each still passes through `escapeLabelValue`, whose `strings.NewReplacer` is byte-identical to the base version.
- End-to-end producer chain: `cmd/summary-api/main.go:35` and `cmd/summary-worker/main.go:32` call `llmobs.Install(...)` before router setup → `internal/llmobs/default.go:21-27` calls `NewMetrics(nil)`, stores it in `defaultMetrics`, and registers the Observer via `llmfallback.SetDefaultObserver` → `llmfallback.Run` events reach `Observer.Observe*` (`internal/llmobs/observer.go:34`) → `Metrics.Observe*` → the new `metrics.*Vec`. No gate short-circuits and no error is swallowed on this path.

## Blind-spot checklist

- **C1 dual-path parity — HIT, cleared.** Normalised-identifier diff of base `internal/llmobs/metrics.go:26-267` against `internal/metrics/metrics.go:29-287` shows identical bodies. Pairs verified: old private impl ↔ new exported impl; old `WritePrometheus` call sequence ↔ new `MustRegister` order; all eight `Observe*` label call sites ↔ unchanged; `durationBuckets` deliberately not moved. No dangling reference to any removed unexported helper remains in package `llmobs` (the `labels` hits in `internal/llmobs/metrics_test.go:454` and `:478` are a local variable and a comment, not the deleted function). No guard/lint tool changed, so the equivalent-form-coverage sub-rule is N/A.
- **C2 control-flow ordering / nested reuse — HIT, cleared with two P2s.** `Registry.WritePrometheus` copies the collector slice under `r.mu` and renders *outside* the lock, so there is no lock-ordering interaction with each collector's own mutex. `HistogramVec.WriteProm` reads `h.buckets` unlocked, which is safe because buckets are immutable after the constructor's copy. Non-canonical-form probing of the escaping control (`\`, `"`, `\n`, `\r`, plus unicode/whitespace variants) shows label *values* are handled identically to base, but the equivalent control is absent for HELP text, metric names and label keys → the first two P2 findings.
- **C3 authorization boundary ≠ capability boundary — HIT, no consequence.** This PR genuinely widens capability: previously only package `llmobs` could create metric series, now every package in the module can. Traced who can reach the *output*: the only renderer is `internal/api/router/router.go:206-210`, mounted on `SetupInternal` (internal listener, with an explicit comment that it must not be reachable from the public listener), and it renders `llmobs.Default()` only. `internal/metrics` has exactly one importer (`internal/llmobs/metrics.go:19`). So no new external exposure and no new data on the wire. Residual: the closed-set cardinality guard `TestMetrics_LabelCardinalityIsBounded` (`internal/llmobs/metrics_test.go:243`) is llmobs-scoped and did not travel with the extraction, so when S3 lands the first non-llmobs consumer there is nothing to catch unbounded cardinality except the doc comment at `internal/metrics/metrics.go:15-17`. Worth a shared guard (a `Registry`-level series-count assertion, or a repo test that walks every `metrics.Labels(` call site and fails on a non-literal argument) at that point — not a blocker for this PR.
- **C4 authorization lifecycle / container-member cascade — N/A.** No auth, permission, tenant or membership code touched; no SQL, migration or status field in the diff.
- **C5 build/note pass ≠ runtime path correct — HIT, cleared by tracing, not by build status.** I did not treat the green build as mitigation; the full production path is traced in the section above, from `Install` in both composition roots through the Observer to the scrape handler. Verification gotcha worth recording: plain `rg` silently skips `cmd/` in this repo, because `.gitignore` lists `summary-api` and `summary-worker` as root-level binaries and ripgrep applies a slash-free pattern at any depth. My first wiring grep therefore appeared to show `llmobs.Install` had no production caller; re-running with `--no-ignore` found both call sites. Anyone re-verifying this PR should use `--no-ignore` or `grep -r`.
- **C6 governance/policy/security-doc self-consistency — mostly N/A, one doc item.** No `SECURITY.md`, `CONTRIBUTING`, workflow, label-policy or release-process change. `.github/CODEOWNERS` is a single `*` wildcard, so the new directory is already covered; there is no labeler config; the `go generate` drift job only touches `internal/api/handler/fence_guard.go`. Two doc observations: the "why hand-rolled instead of `prometheus/client_golang`" rationale (dependency-review / osv-scanner re-clearing) is now duplicated near-verbatim in `internal/metrics/metrics.go:5-10` and `internal/llmobs/metrics.go:4-11`, and two copies of one justification drift — llmobs could keep only the Observer-seam sentence and point at the new package. On the positive side, the move dropped the stale self-contradictory half of the old CR comment (base `internal/llmobs/metrics.go:55-58` claimed "`\r` is escaped alongside the three the text format requires" while `:59-63` said the opposite); the extracted copy keeps only the correct version.

## Cross-round blocker recheck

N/A — this is the first review round for head `4174824`. The issue carries no prior review comments, and per the review workflow I did not read GitHub PR comments or review history. I confirmed the PR's current `headRefOid` equals the SHA under review, and the merge-base against `main` is `31233927`, so the diff range is correct and this is not an orphan-history branch.

- Output parity is real, not assumed: identical rendering code, identical family order, identical escaping replacer, identical `%g` / `formatBucket` / `withLE` formatting.
- The unchanged llmobs suite still pins the contract end-to-end, including `TestInstall_EndToEndValuesAreExact` (exact series label sets and values driven through a real `llmfallback.Run`), `TestWritePrometheus_EscapesLabelValues`, `TestWritePrometheus_CarriageReturnIsFlattenedNotEscaped`, `TestHistogram_QuantileInputsAreExact` and `TestHistogram_ConcurrentObserveIsRaceFree`. The extracted package is therefore race-covered indirectly, and `-race -shuffle=on` passes.
- `NewHistogramVec`'s defensive bucket copy means `runDur` and `attemptDur` sharing `durationBuckets` cannot corrupt each other.
- Empty-`LabelSet` histogram rendering (`withLE("", le)`) is reachable now that the package is public; I checked the renderer and it produces `{le="..."}` buckets and `name{}` sum/count lines, which the text parser accepts. Benign, but untested.
- No import cycle: `internal/metrics` imports only stdlib; `internal/llmobs` imports `internal/metrics`.
- Dropped imports (`fmt`, `sort`, `strconv`, `strings`, `sync`) are genuinely unused in llmobs now; `io` and `time` are still needed.
- CI gates unaffected: `golangci-lint` config enables only `govet` (clean), `gofmt` clean, no golden files or repo-walking meta-tests exist, and the History Check workflow's common-ancestor requirement is satisfied.

## Verdict: APPROVED

This is a careful, mechanical extraction: the moved code is character-identical modulo export casing, the family ordering that the byte-identity claim depends on is preserved exactly, the full production path from `llmobs.Install` in both composition roots to the `/internal/metrics` handler still carries data, and build, vet, gofmt and the whole test suite (including `-race -shuffle=on`) pass at head `4174824`. The three P2 items and two nits are latent hardening gaps in a package that this PR deliberately turns into a shared library — none has a reachable production consequence today, so none is a merge blocker. I would land this and address the HELP/name escaping and `Labels` key validation before or alongside S3, when the first non-llmobs consumer makes them reachable.

[Octo-Q] verdict: APPROVE — no P0/P1. The extraction is verbatim-faithful (normalised diff proves identical bodies), exposition order and escaping are preserved so scrape output stays byte-identical, the Install→Observer→registry→scrape data flow was traced end-to-end and confirmed live in both `cmd/summary-api` and `cmd/summary-worker`, and build/vet/gofmt/test/-race all pass at head 4174824. Remaining findings are three P2 library-hardening gaps (unescaped HELP/name, unvalidated label keys, non-validating `MustRegister`) plus two nits; all are latent with no current triggering caller, so per the severity rubric they do not block.

PR #245, opened 2026-09-10
feat(summary): unified smart summary — durable run-based generation, one-shot override, streaming refine

wanyiwang06:
## What this does

Backend half of the unified smart summary. One summary now carries its own version series, and every generation is a durable, leased, idempotent **run** rather than an in-request side effect.

### Generation runs
`summary_generation_run` records `operation_type` (`generate` / `regenerate` / `scheduled_generate` / `refine` / `regenerate_override`), an `executor` (`workflow` / `refine`), a lease + `execution_token`, `idempotency_hash` + `request_hash`, `config_revision`, `base_version_id`, and the frozen `input_json`. HTTP admits the run and returns 202; the worker poll loop executes it. A restart recovers an expired lease, and an old token can never commit after takeover.

### Capability gating is engine-neutral
- `can_refine` is **configuration-independent** — agent-created summaries accept 按意见调整 immediately.
- `can_regenerate_direct` requires a complete configuration, so agent summaries are pointed at 定时更新 to fill it in first.
- Both clear while a run is active.

### One-shot topic override
`regenerate_override` freezes an ad-hoc requirement into a single run without touching `summary_task.topic`, `generation_spec_json`, or `config_revision`. It is also **excluded from the scheduled series' incremental watermark**, so an ad-hoc topic cannot consume the scheduled series' start point. An empty override is a 422, not an admitted run.

### Refinement now streams
`ExecuteRefine` takes a `RefineStreamSink` factory that is invoked **only after a successful claim** — a run another worker won never emits a preview into the winning run's channel. Streaming uses the optional `RefineStreamModel` capability; a model without it keeps the buffered path unchanged, and still finalizes the sink. Provider errors are sanitized before they reach the preview.

The sink is built in `worker` and consumed through an interface owned by `service`, so the service layer keeps no transport dependency. Refine reuses the same worker→ingest→SSE channel as full generation.

## Test plan
- `internal/service` and `internal/worker` suites pass, including new coverage for streaming refine (capable model streams, stream failure never emits `done` and leaks no provider text, non-capable model still finalizes) and for the override contract (frozen ad-hoc topic, saved config untouched, applied yet watermark-excluded, while a plain regenerate *does* move the watermark).
- `go build ./...`, `go vet`, and `gofmt` clean.
- Mirror fixtures under `tests/` are hermetic and synthetic-identity only.

Paired with the octo-web PR that splits the 重新生成 / 定时更新 entries.


yujiawei (approved):
# Code Review — PR #245 (octo-smart-summary)

Reviewed at head `fda3aae046bc878c2de77f72f45b8f65ec1ccbf1` against merge-base `391134cc`. Scope: durable run-based generation (`summary_generation_run` with lease + `execution_token` fencing), one-shot topic override (`regenerate_override`), streaming refine, and the three `20260908-*` migrations. Build, `go vet`, and the `internal/service` + `internal/worker` suites are green locally. This PR is security-sensitive; areas flagged for a human sign-off are called out below.

## Spec compliance

The diff implements what the description promises and does not appear to over-build:

- **Durable run admission/commit.** `QueueRefine` / `QueueRegenerate` freeze input and admit a 202 run; the worker poll loop (`ContentGenerationWorker.poll`) executes it. Lease + `execution_token` fencing is correct: `ClaimGeneration` increments the token under the task-row lock, and `checkGenerationLease` / `CompleteRefine` / `CompleteFullGeneration` reject any stale token — an old token can never commit after takeover. `RecoverableGenerations` reclaims expired leases.
- **Capability gating is engine-neutral, and neither branch is accidentally constant.** `singleContentCapabilities` sets `CanRefine = !busy` (config-independent) while `CanRegenerateDirect = config.State == "complete" && !busy`. Both correctly clear while a run is active.
- **One-shot override is correctly isolated.** `regenerate_override` only rewrites the frozen run spec's `Requirement`; it never writes `summary_task.topic`, `generation_spec_json`, or `config_revision`. `successfulGenerationEnd` excludes it from the watermark via `operation_type <> 'regenerate_override'`, so an ad-hoc topic cannot consume the scheduled series' incremental start point. Empty override → 422. Verified against the override tests.
- **Streaming refine** builds the sink only after a successful claim (`ExecuteRefine`), falls back to the buffered path for a non-streaming model, and sanitizes provider errors to a fixed `"summary refinement failed"` string before they reach the preview.

**Verdict — Spec: ✅** (no missing, no extra, no divergent behavior found).

## Code quality

One non-blocking correctness nit; the rest is solid.

### P2 — Full-generation streaming emits `done/completed` before the durable commit

`internal/worker/confirmed_generation.go:39`

```go
content, citations, count, tokens, usedModel, err := p.executePersonalPipelineInput(...)
if err != nil {
    stream.Error("summary generation failed")
    return service.FullGenerationOutput{}, err
}
stream.Done(model.StatusCompleted)   // <-- emitted here, inside GenerateConfirmed
return service.FullGenerationOutput{...}, err
```

The authoritative commit only happens afterward, back in the caller:

`internal/service/content_full_generation.go:203`

```go
result, err := s.CompleteFullGeneration(ctx, id, run.ExecutionToken, output)
```

Failure scenario: the provider returns a body that later fails validation in `CompleteFullGeneration` (e.g. an invalid citation marker → `invalid_generation_output`), or the lease was taken over just before commit (`generation_lease_lost`). The SSE preview channel has already received `done: completed`, but the durable run ends up `failed`/`conflict` or owned by another token. A subscribed client sees a false terminal success on the live preview until its next poll/refresh corrects it.

This is a preview-only inconsistency — the authoritative version is token-fenced and correct, and the client's poll self-heals — so it does not block merge. Note the asymmetry with the refine path, which does this correctly: `ExecuteRefine` (`internal/service/content_refine_executor.go:132-144`) only calls `sink.Done(...)` *after* `CompleteRefine` succeeds, and calls `sink.Error(...)` otherwise. Suggest moving the full-generation `Done`/`Error` emission out of `GenerateConfirmed` and gating it on the `CompleteFullGeneration` result the same way, so the preview's terminal event matches the durable outcome.

### Notes (no change required)

- **Lock ordering is internally consistent.** Legacy/config writers take schedule→task (`lockConfiguration`, `LockLegacyContentTask`); coordinated content writes take task→membership→content and never acquire a schedule lock (`lockContent`). No circular wait between these families.
- **Trust boundaries** hold: `AuthorizeGenerationSources` is re-checked at both admission and commit; `requireSingleGeneration` restricts the pilot to a single-person creator-owned task; a `permission_hidden` citation payload is stripped (markers included) before it can enter a refine prompt (`QueueRefine`). Command routes only mount when `SUMMARY_CONTENT_EXECUTION_SPACES` is non-empty.
- **Migrations** `20260908-01..03` are additive-only with forward-only `Down`, matching their stated contract; the `uk_result_generation` / `uk_personal_generation_user` unique keys are on the new nullable `generation_id` column, so they will not collide with historical rows.

**Verdict — Quality: Approved** (the P2 above is a suggested follow-up, not a blocker).

## Items worth a human eye (security-sensitive)

- The full-generation preview `done`-before-commit gap above: confirm product is comfortable that a live preview can briefly show success for a run that ultimately lands in `conflict`/`failed` before the client poll corrects it.
- `execution_token` fencing and the `active_slot` / `idempotency_hash` unique constraints are the core safety mechanism here — worth a manual concurrency read before enabling the pilot in a new space.

## Overall verdict

**APPROVE.** Spec is fully and faithfully implemented with no scope creep, concurrency fencing is correct, and the only correctness finding is a preview-only cosmetic inconsistency that self-heals on poll. Recommend addressing the P2 streaming-order asymmetry in a follow-up.

Jerry-Xin (changes requested):
Reviewed head `fda3aae046bc878c2de77f72f45b8f65ec1ccbf1` against merge-base `391134cc`, subsystem by subsystem across all 116 files: migrations (`20260908-01..03`), the durable run state machine and service core, the HTTP command/read boundary, the worker poll loop and streaming, the confirmed-generation pipeline, the legacy-write fencing layer, tests, docs and mirror fixtures.

Verification performed on this head: full Go suite green (all packages); `-race` green across every touched package; the entire MySQL-tagged suite executed against a real, disposable MySQL 8.0 database — migration chain on both the fresh-install and legacy-upgrade paths plus replay idempotency, the end-to-end confirmed pipeline through the production poller, the window-function list projection, configuration CAS under concurrent retries, admission races, and deletion run-cancellation fencing — all pass. Mutation probes on the new guard tests: 3/3 killed with precise failure messages (watermark exclusion for one-shot overrides, `execution_token` fencing on lease takeover, idempotency request-hash conflict), controls green before and after. The branch base predates two observability commits now on `main`; a local merge reproduction shows a conflict-free auto-merge whose combined tree builds and passes the service/worker suites, preserving both sides.

The durable-run core is well constructed: token fencing on every claim and commit, a consistent task→run lock order, terminal-state-first rejection commits so revoked runs cannot occupy their slot forever, bounded error codes (no provider text in run rows, logs, or client responses), run DTOs that structurally exclude private input (`json:"-"` plus a serialization contract test), idempotent admission with request-hash conflict detection, and DB-enforced active-slot uniqueness. The legacy-write fencing census is exhaustive across handlers, worker claims/CAS transitions, scheduler scans, retention pruning, and deletion paths, with under-lock rechecks covering stale-scan races. Two issues nevertheless block.

### 🔴 Blocking

**🔴 1. Full-generation preview emits terminal success before the durable commit — a false success on the user-visible stream, never corrected on the channel**

`internal/worker/confirmed_generation.go:39` emits `stream.Done(model.StatusCompleted)` at the end of `GenerateConfirmed`, but the authoritative commit runs afterwards in `ExecuteFullGeneration` (`internal/service/content_full_generation.go:203`, `CompleteFullGeneration`). Every commit-side failure — output validation (`invalid_citation_reference`, `invalid_generation_output`, `invalid_content`), a user cancellation while the model call is in flight, a lease takeover, permission revocation re-checked at commit, or a database failure — leaves the client with an already-delivered `done: completed` while the durable run ends `failed`/`cancelled`/`conflict` with nothing applied. No corrective stream event is emitted on any of those paths: `ExecuteFullGeneration` holds no sink, and the worker poll loop treats `generation_lease_lost` as expected and returns quietly (`internal/worker/content_generation.go:100-110`).

This is more than the documented "view only" caveat, for three reasons:

1. **The repository's own established convention on the very same channel does the opposite.** The legacy personal pipeline (`internal/worker/personal_processor.go:307-412`) emits `finishStreamError("summary generation failed" / "summary generation cancelled" / "summary persistence failed" / "summary completion failed")` on every failure path and only calls `finishStreamDone(StatusCompleted)` after persistence and completion succeeded (lines 400/406/412). This PR's own refine executor follows that convention (`internal/service/content_refine_executor.go`: `sink.Done` only after `CompleteRefine` returns, `sink.Error` otherwise, pinned by `TestContentRefineExecutorSurfacesStreamErrorAndNeverFinishes`). The new full-generation path regresses it.
2. **It is reliably reachable through ordinary user action**: cancelling a run mid-flight. Cancellation does not interrupt the executing model call; the stream completes, `done(completed)` is delivered, and only then does the commit fail the lease check against the cancelled run. The user who just cancelled watches the generation "complete successfully".
3. The hub deliberately guarantees terminal-event delivery (`internal/streaming/hub.go` discards a stale buffered event to make room for `done`/`error`), so the false `done` is delivered reliably and — per point 1 — is never followed by an `error` on the channel.

The durable state itself stays truthful throughout (token fencing and the terminal-first rejection commit do their job), which bounds the blast radius — but the live channel's terminal signal must match the durable outcome. Same family, lower intensity: `ExecuteRefine` emits `sink.Done(model.StatusCompleted)` when `CompleteRefine` returns a run in `conflict` status (candidate saved, not applied); the legacy integer status enum has no conflict value, so the preview again reports completion for a non-applied outcome.

Fix direction: lift the terminal stream emission for full generation to the commit boundary — mirror the refine sink pattern (sink factory invoked after a successful claim inside `ExecuteFullGeneration`, `Done`/`Error` driven by the actual `CompleteFullGeneration` result), and let `Done` reflect the true terminal outcome (applied vs conflict vs failed). Add regression coverage mirroring the existing refine stream-failure test: post-stream commit failures (invalid citation marker, cancellation mid-flight, lease loss) must not surface a completed `done` and must surface a sanitized error.

**🔴 2. Two new test files break the repository's primary CI gates at this head (`Test` and `Vet` are both red)**

`internal/service/content_refine_stream_test.go` and `internal/service/content_regenerate_override_test.go` use `contentTestDB`, which is defined behind a `//go:build cgo` constraint (`internal/service/content_read_test.go:1`), but neither file carries the constraint. Under `CGO_ENABLED=0` — exactly how the CI `Test` job (`CGO_ENABLED=0 go test -shuffle=on -count=1 -timeout 5m ./...`) and `Vet` job (`CGO_ENABLED=0 go vet ./...`) run — the package fails to compile:

```
internal/service/content_refine_stream_test.go:68:8: undefined: contentTestDB
internal/service/content_refine_stream_test.go:103:8: undefined: contentTestDB
internal/service/content_refine_stream_test.go:134:8: undefined: contentTestDB
internal/service/content_regenerate_override_test.go:16:8: undefined: contentTestDB
```

Both CI jobs are red at this head with exactly these four errors, while `Test (race, cgo)` passes because CGO is enabled there — masking the breakage for cgo-only local runs. Reproduced locally byte-for-byte; a fresh merge-base probe tree is green under `CGO_ENABLED=0` vet + test, so this is PR-introduced. The pattern is known inside this PR — `internal/service/content_write_test.go` carries `//go:build cgo` — these two files simply missed it. The guard tests themselves are valuable and discriminating (see mutation results); they must also compile in both CI configurations.

Fix: add `//go:build cgo` to both files (or relocate `contentTestDB` to an unconstrained shared helper), then confirm the `Test` and `Vet` gates go green at the new head.

### 🟡 Non-blocking

1. **`DeleteSchedule` group deletion does not invalidate durable runs.** `internal/api/handler/schedule.go` (`DELETE /summary-schedules/:id`) locks the bound tasks and soft-deletes all of them, but never calls `service.CancelContentGenerationsForDeletion` — unlike the identical group deletion in `DeleteSummary` (`internal/api/handler/task.go:2030-2046`), which locks every affected task and cancels runs in the same transaction, and unlike `CancelContentGenerationsForDeletion`'s own doc comment ("Call before soft-deleting that task, in the same transaction"). Deleting the schedule of an enrolled task that has a pending/running run leaves the run holding its `active_slot` until a worker claim self-heals it (`lockContent` → 404 → run cancelled `permission_revoked`): bounded to roughly one poll interval in a consistently configured deployment, unbounded if no worker scans that space. Late commits remain fenced, so there is no false success — this is a consistency/hygiene gap. Fix: the same per-task cancellation loop the `DeleteSummary` group path uses.

2. **Managed-scheduler admission filter disagrees with its own gate → perpetual no-op requeue.** `QueueDueGenerations` (`internal/service/generation_scheduler.go:27`) selects schedules where *any* bound task is enrolled (`EXISTS ... content_protocol_version > 0 AND generation_spec_json IS NOT NULL`), but `QueueScheduledGeneration` (:55-59) processes only the *newest* bound task (`ORDER BY id DESC LIMIT 1`) and returns `nil` — without advancing `next_run_at` — when that one is unenrolled. A pilot-space schedule whose history contains an older enrolled round plus newer unenrolled rounds (reachable: `SaveGenerationConfiguration` can enroll an older round while legacy fires keep producing newer ones) stays due forever and is re-picked on every poll tick, each time opening a locking transaction (schedule + task `FOR UPDATE`). Correctness is preserved (the legacy engine still advances the schedule; the managed engine no-ops), but it is continuous lock/DB noise. Fix: align the `EXISTS` filter with the newest-bound-task condition the gate applies.

3. **`CONFIGURATION.md` omits the two rollout-critical allowlists.** The new section documents `SUMMARY_CONTENT_READ_SPACES`, but `SUMMARY_CONTENT_WRITE_SPACES` (legacy-write fencing and worker recovery scans) and `SUMMARY_CONTENT_EXECUTION_SPACES` (command-route mounting, pilot execution, managed scheduling) appear only in `docs/unified-summary-versioning-verification.md`. These gate the entire rollout; add them to the canonical environment table with their fail-closed semantics (empty default, `*` not honored).

### Notes (no change required)

- The refine lease (2 minutes) versus the 90-second model budget leaves a ~30-second commit window; a slow commit can trigger a takeover that re-executes the model call. Token fencing keeps this honest (a single winner commits), merely wasteful — acceptable at current scale.
- Idempotent `QueueRefine`/`QueueRegenerate` replays return the original run regardless of its terminal status (failed/cancelled included) — correct replay semantics; the status is visible to the caller.
- An approval already on this head independently identified the 🔴 1 mechanism and rated it non-blocking. The same-channel legacy parity evidence, the reliable reachability via mid-flight cancellation, and the guaranteed terminal-event delivery are why it is rated blocking here; credit for the independent convergence and for the accurate lock-ordering, trust-boundary, and nullable-`generation_id` uniqueness notes there.
- The `check-sprint` CI gate is red repository-wide (known unrelated board gate) and is not counted against this PR.

### Verification summary

| Check | Result |
|---|---|
| Full suite at head (cgo enabled) | Green, all packages |
| `CGO_ENABLED=0` test + vet at head | **Red** — 4× `undefined: contentTestDB` (🔴 2); merge-base probe green → PR-introduced |
| `-race`, all touched packages | Green |
| Real MySQL 8.0: migrations (fresh + legacy upgrade + replay), E2E confirmed pipeline, list window projection, CAS/admission races, deletion fencing | All pass |
| Mutation probes: watermark exclusion / token fencing / idempotency conflict | 3/3 killed; controls green |
| Prior-fix sweep (draft safe-save idempotency, finish gate, structural truncation disclosure, citation-evidence caps, soft-delete self-heal, case/collation identity discipline) | Anchors untouched and live; router changes are additive-only |
| Auto-merge against current `main` (branch base predates two merged observability commits) | Conflict-free; combined tree builds, vets, and passes service/worker suites |

Verdict: REQUEST_CHANGES — 🔴 1 (false terminal success on the live preview channel before the durable commit) and 🔴 2 (primary CI gates red at head) both have small, well-scoped fixes; the remainder of the PR is strong.

wanyiwang06:
Superseded by #246, which delivers the same user-facing feature by reusing the existing edit / version-history / Workflow-regenerate / scheduled-update machinery (+1 nullable column) instead of a new durable-run + versioning layer. Closing in favor of the leaner approach.

PR #246, opened 2026-09-11
feat(summary): 统一个人总结能力并修复 Agent 工具提交

wanyiwang06:
## 改动范围

- 为 Agent 保存的个人总结复用现有编辑、版本历史、Workflow 重新生成和定时更新能力；独立保存实际生成需求，不把显示标题当作原提示词。
- 增加 generation-config 更新接口，校验来源权限并默认持久化补齐的配置；重新生成不改变定时周期。编辑保留版本，重新生成失败恢复已有正文，并重置旧流状态。
- 修复继续优化的工具提交：正文由独立文本生成阶段完成，通过请求内 `content_handle` 提交，后端序列化为原有 `preview.content`，不改变对外响应/SSE。
- 非法工具参数最多原地修复两次；草稿就绪后只允许最终提交。保留授权、Map/Reduce、引用和超时约束，确定性模型/协议错误不重放整轮任务。
- 增加单测、默认关闭且仅记录结构元数据的诊断能力，以及实现/上线说明。数据库备份、私有现场回放、日志和本机镜像部署文件不纳入提交。

## 范围边界

保留“继续优化 → Agent 对话 → 新总结”的原有流程。本 PR **不实现局部修改路由优化**，也不保证继续优化不重新检索。配套前端：Mininglamp-OSS/octo-web#1662（来自 `wanyiwang06:codex/personal-summary-main-parity`）。

## 验证

- [x] `go test ./internal/agent ./internal/llmfallback ./internal/service ./internal/api/handler ./internal/worker ./internal/streaming`（本机 native 库路径已配置）
- [x] `go test -race ./internal/agent -run 'TestPreparedDraft|TestDraftWriter|TestToolArgument|TestInvalidArgument'`
- [x] API/Worker 已完成构建；既有隔离镜像中的最终版本通过 3 次引用场景及 1 次完整 fetch/Map/Reduce/draft/submit 真实模型验收，不等同于 CI 或模型成功率承诺。
- [x] `git diff --check`、提交内容敏感文件检查及与已拉取上游 main 的无冲突合并检查。

## 上线顺序

先备份数据，通过既有迁移机制应用 `migrations/sql/20260911-01-summary-generation-requirement.sql`，再部署后端，最后部署配套前端。新增 nullable TEXT 字段；Down 会删除该字段及保存的生成需求，回滚前须备份。

PR #247, opened 2026-09-14
fix(ci): allowlist documented summary API URL

an9xyz:
## Summary
- allowlist the known `generic-api-key` false positive in `docs/auth-optimization-frontend.md`
- keep the exception scoped to the exact rule and file path
- preserve all default gitleaks detections elsewhere

## Validation
- `git diff --check`
- merge validation requires the post-merge full-history Secret Scan to pass

This unblocks the approved v1.13.0 release.


yujiawei (approved):
# Code Review — PR #247 (octo-smart-summary)

**Head SHA:** `c707945221f4468ad19a087d2066da52cf716e73`
Adds `docs/auth-optimization-frontend.md` to the existing `generic-api-key` path allowlist in `.gitleaks.toml`, alongside the pre-existing `internal/service/llm_test.go` entry.

## 1. Spec compliance

The PR does exactly what it claims and nothing more:

- **Missing?** No. It allowlists the exact file (`docs/auth-optimization-frontend.md`) under the exact rule (`generic-api-key`), and keeps the comment block accurate.
- **Extra?** No. Only the one path is added; the pre-existing `llm_test.go` exemption is preserved; `useDefault = true` is untouched.
- **Deviation?** No. Uses a path-based allowlist (not a commit fingerprint), matching the stated rebase-stability rationale and the existing pattern.

**Spec: ✅**

## 2. Verification of the premise (this is a real false positive)

The failing gate is genuine and correctly diagnosed. The full-history `push`→`main` Secret Scan (`gitleaks git`, not the PR-range scan) is red, and the authoritative CI log flags:

```
RuleID:      generic-api-key
File:        docs/auth-optimization-frontend.md
Line:        128
Commit:      1d0afc57ca36290442560901f92089d8f4f01fb5
```

Line 128 is `… → 转发到 summary-api:8080/api/v1/summaries。` inside a frontend auth design doc. The `generic-api-key` heuristic reads the token before the colon as a key; the value is a documented cluster-local service endpoint and carries no credential. This is a true false positive, so an allowlist (rather than a code change) is the right remedy.

Note on reachability: the flagged file is **not** present on `main` or on this PR's tree — it lives only on unrelated historical refs. `gitleaks git` scans all reachable commits across refs, which is why the full-history push scan trips on it while the PR-range scan is green. A path allowlist matches by path regardless of commit reachability, so this fix is the correct lever for that job.

## 3. Security review (this PR is security-sensitive)

I reproduced the config against gitleaks locally to confirm the exemption does not weaken detection:

- **Default rules are preserved, not replaced.** Redefining `[[rules]] id = "generic-api-key"` under `[extend] useDefault = true` *merges* the allowlist into the default rule rather than overriding it. A `generic-api-key` secret placed in any non-allowlisted file is still detected. ✅
- **The exemption is rule-scoped, not global.** A `private-key` block (or other rule types) placed inside the allowlisted doc is still flagged — only `generic-api-key` is suppressed there. ✅
- **Full-history (`gitleaks git`) suppression works** for the target finding even when the file has since been deleted from the tip. ✅

**Quality: Approved.** No P0/P1 issues.

### Advisory notes (P2 / non-blocking)

- **File-wide suppression blast radius.** The allowlist exempts the *entire* `docs/auth-optimization-frontend.md` from `generic-api-key`, not just line 128. If a real generic-api-key-shaped secret were ever added to that file it would be silently missed. Risk is low here (it's a design doc, it's already absent from `main`, and every other rule still applies), but a tighter allowlist keyed on the specific value/regex (e.g. a `regexes`/`stopwords` entry for the `summary-api:8080` string) would shrink the exempted surface. Optional.
- **Unanchored path regex.** `docs/auth-optimization-frontend\.md` is an unanchored substring match, so a hypothetical adjacent path like `sub/docs-auth-optimization-frontend.md.txt` would also be exempted. This is standard gitleaks behavior and the string is specific enough that collision is unlikely; flagging only for awareness.

## 4. Verdict

**APPROVE.** Spec ✅ AND Quality Approved. The premise is a verified false positive, the fix is correctly scoped to one rule + one file, and it provably does not weaken secret detection anywhere else (default rules preserved, rule-scoped, other rule types still fire inside the doc). The advisory notes are optional hardening, not merge blockers.

## 5. For the human verifier (security-sensitive)

- The suppressed line contains no credential — it is `summary-api:8080/api/v1/summaries`, a cluster-local endpoint.
- Worth a manual sanity glance: this allowlist widens the exempted surface to a whole doc file. If that doc could later carry real secrets, prefer a value-scoped allowlist. As-is it's acceptable.

Jerry-Xin (approved):
# Review — PR #247 (octo-smart-summary) @ `c707945221f4468ad19a087d2066da52cf716e73`

**Verdict: APPROVE** — independently verified end-to-end. The allowlist entry is a precisely scoped fix for a verified false positive that has been keeping the full-history Secret Scan on `main` red; it does not weaken any other detection.

## What the change does

Adds `docs/auth-optimization-frontend\.md` to the `paths` allowlist of the `generic-api-key` rule in `.gitleaks.toml` (the single rule block under `[extend] useDefault = true`), and updates the comment block/description to cover both exemptions. +10/−4, one file.

## Motivation trace (verified against CI history, not taken on faith)

- The failing gate is real: **every recent `push`→`main` full-history Secret Scan is red**, including the run at current main tip `f3cf986e` (2026-09-10). The CI log shows exactly one finding: `RuleID: generic-api-key`, `File: docs/auth-optimization-frontend.md`, `Line: 128`, `Commit: 1d0afc57…`, fingerprint `1d0afc57…:docs/auth-optimization-frontend.md:generic-api-key:128`.
- **Local reproduction (gitleaks 8.30.1, the same pinned version as CI, checksum-verified binary):** full-history scan with the base config → exactly 1 leak, fingerprint byte-identical to the CI log; full-history scan at this PR head with the new config → **0 leaks, exit 0**. The fix provably turns the post-merge full-history scan green and suppresses exactly the one documented finding.
- Ancestry (correcting one detail in the earlier peer review): the commit that added the file (`1d0afc5`, 2026-04-29) and the commit that deleted it (`bf2f860`, 2026-05-09, "remove hardcoded LLM gateway defaults") are **both direct ancestors of `main`** — verified via `git merge-base --is-ancestor` on a fully unshallow-ed history (180 commits). The blob lives in main's own history; it is absent only from the current tree. (Shallow clones misreport these commits as non-ancestors — worth knowing when re-verifying.) The scan therefore trips on `push`→`main` even though nothing in the working tree matches.
- This is a symptom-free fix: the file was already deliberately deleted from the tree in May; the only remaining artifact is the historical blob. History rewriting to silence one documented false positive would be disproportionate.

## Legitimacy of the allowlisted content

Line 128 is `… → 转发到 \`summary-api:8080/api/v1/summaries\`。` in a (now-deleted) frontend auth design doc. `summary-api` is this repo's own service (`cmd/summary-api`, `Dockerfile.api`), `8080` is the documented default `API_PORT` (`CONFIGURATION.md`), and the path is the public API route. It is a cluster-local DNS endpoint with **no credential** — a textbook `generic-api-key` false positive (the heuristic reads the token before the colon as a key; entropy 3.59). Comparable cluster-local URLs are already documented publicly in the current tree (e.g. `octo-search-batch:8080` in README). Nothing here masks a real leak.

## Scope / blast-radius verification (boundary probes, gitleaks 8.30.1)

- **Path-precise:** the exact same trigger string placed in a *non*-allowlisted path is still detected → the rule remains active and the exemption is not global.
- **Rule-scoped:** a valid-shaped `github-pat` token placed *inside* a file at the allowlisted path is still detected → the exemption does not exempt the file from other rules; redefining `[[rules]] id = "generic-api-key"` under `[extend] useDefault` merges the allowlist into the default rule rather than replacing it (consistent with CI behavior: the base config still flagged the docs file while `llm_test.go` stayed exempt).
- **Exactly one finding suppressed:** the base-config full-history scan shows 1 leak total across all 180 commits, so the new path entry silences precisely that finding and nothing else.
- The path regex escapes the dot correctly and matches the real path verbatim.

## Consumer census

Single consumer: the org reusable `Mininglamp-OSS/.github reusable-secret-scan.yml@v1`, auto-detecting `.gitleaks.toml`, invoked by `.github/workflows/secret-scan.yml` on `pull_request` (PR-range scan — unaffected, green on head) and `push`→`main` (full-history — the gate being fixed). No local hooks, Makefile targets, or other scripts in this repo consume the file.

## Convention consistency & merge coordination

- Matches the sibling `internal/service/llm_test.go` entry exactly (path-based not fingerprint-based, escaped-dot regex, comment block extended with rationale, description generalized appropriately).
- Zero file overlap with open PR #246; `git merge-tree` against its head (`0dcefce0`) is clean. No merge-order constraint.

## CI status on head

All checks green (Build, Lint, Test, Test race/cgo, Vet, secret-scan PR-range, osv, dependency-review, code-review) except:
- `check-sprint` ❌ — "No linked issue found": the PR description has no `Closes #<issue>` reference. Note this check is **not** in main's ruleset required list (only `code-review` is), but it is a visible red check; consider adding a linked issue or letting a maintainer handle the board entry before merge. The ruleset additionally requires 2 approvals + CODEOWNERS + last-push approval.

## 💬 Non-blocking advisories

- 🔵 **Anchor the path regex** (`'''^docs/auth-optimization-frontend\.md$'''`) to make the stated exact-file scope literal; the unanchored form is a substring match (same applies to the pre-existing `llm_test.go` entry). Independently converged with the second-opinion pass and the earlier peer review.
- 🔵 **Whole-file, rule-scoped blast radius:** if this path were ever re-added to the tree, the entire file would be exempt from `generic-api-key` (other rules still fire — probe-verified). Acceptable given the file is absent and the trade-off is documented in the config comment; a value/regex-scoped allowlist entry would shrink the surface further. Also converged with the earlier peer review.

Peer activity: one prior APPROVE (yujiawei) on this same head; verdict and both advisories converge with this review, and the ancestry detail above is corrected here for the record.

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #247 (octo-smart-summary)

## Summary

One file, +10/-4: `.gitleaks.toml` gains a second `generic-api-key` allowlist path (`docs/auth-optimization-frontend\.md`) and a rewritten rationale block. The false positive it targets is genuine at the scanner version CI pins — gitleaks 8.30.1 reports `docs/auth-optimization-frontend.md:128` (match `summary-api:8080/api/v1/summaries`, secret `8080/api/v1/summaries`, introduced by `1d0afc57`) — and the new entry does suppress it. My concern is the shape of the exemption, not its intent: a `paths` allowlist switches the rule off for the whole file rather than for the documented value, in a repo where this workflow is the only gitleaks gate. Nothing here blocks the merge; two P2 items below.

## Verification

Commands actually executed at head `c7079452`. gitleaks 8.30.1 was downloaded from the pinned release URL and SHA256-verified against `551f6fc8…`, the checksum the org reusable workflow enforces; 8.21.2 was also run for comparison.

- `gitleaks git --config <pre-PR toml> --log-opts=f3cf986..c707945 .` — 0 findings; identical with the PR config. This PR's own CI range is clean either way.
- `gitleaks git --config <pre-PR toml> --log-opts=origin/main .` — 0 findings. Main's lineage has no gitleaks findings today.
- `gitleaks git --config <pre-PR toml> .` across all local refs — 3 findings (the doc, plus 2 at `internal/api/handler/agent_summary_be2_test.go:301`); with the PR config — 2. The exemption demonstrably works.
- `gitleaks dir` over both historical docs extracted from `1d0afc57` — exactly 1 finding, frontend only.
- Blind-spot probes on a synthetic corpus — see Findings.
- No Go build or test was run; this PR touches no Go code.

Data-flow trace for the only consumed value in this diff: producer `.gitleaks.toml:37` → consumer is the org reusable secret-scan step, which auto-detects the file (`if [ -f .gitleaks.toml ]; then config_args=(--config .gitleaks.toml); fi`) and runs `gitleaks git --log-opts="${BASE_SHA}..${HEAD_SHA}"` on `pull_request` (fetch-depth 2) or full history on `push`/`schedule` (fetch-depth 0). `.github/workflows/secret-scan.yml:13` is the sole caller and passes no `gitleaks-version` override, so 8.30.1 is what consumes the entry. I confirmed the value reaches the consumer for the intended finding, and also reaches paths it was never meant to cover.

## Findings

No P0/P1 issues; two P2 items below.

### P2 — Path allowlist disables `generic-api-key` for the whole file, and the regex is unanchored (`.gitleaks.toml:37`)

A `paths` entry is a file-level kill switch, not a value-level one. Verified with 8.30.1: a genuine-looking `apiKey = "aK9xQ2mZ7pL4vR8nT1wY6uI3oP0sD5fG7hJ2kL9"` placed at `docs/auth-optimization-frontend.md:6` is reported under the default config and under the pre-PR config, but is silently skipped under this PR's config. The regex is also unanchored, so the same key at `internal/docs/auth-optimization-frontend.md:2` is swallowed too. This repo has no L0 gitleaks hook (`scripts/` holds only `download_tokenizer_models.sh`; no husky/lefthook/pre-commit), and the reusable workflow's own header calls L1 "the only secret defense for fork contributors" — so this narrows the only gate, permanently and invisibly. Scope by value instead of by file; this form is tested and produces an identical all-refs result to the PR config while keeping the rest of the file scanned:

```toml
[rules.allowlist]
description = "Documented generic-api-key false positives"
regexTarget = "match"
paths = [
  '''internal/service/llm_test\.go''',
]
regexes = [
  '''summary-api:8080/api/v1/summaries''',
]
```

### P2 — Rationale describes a file that is not in the shipped tree (`.gitleaks.toml:22`)

"`docs/auth-optimization-frontend.md` also contains the internal service URL" is written in the present tense, but the file does not exist at head `c7079452` and is unreachable from `origin/main`: it was added by `1d0afc57` and deleted by `bf2f8600`, and `git merge-base --is-ancestor` is false for both against `origin/main`. The entry only bites because gitleaks scans commit diffs — a future merge PR carrying `1d0afc57` inside `BASE_SHA..HEAD_SHA` trips on the historical add. A maintainer reading this file on main will grep for a path that isn't there, fail to reproduce the finding, and may delete the entry as dead config and re-break that merge. Record the provenance: name the introducing commit, state that the path is absent from the main tree, and note the version sensitivity (8.30.1 flags `:128`; 8.21.2 reports nothing on byte-identical content).

## Human-verify

1. Main's `push`/`schedule` full-history scan cannot reach this doc today, so the entry is inert on main until the lineage containing `1d0afc57` merges — at which point it becomes load-bearing for that merge's green gate. Worth confirming the intended merge path for that lineage. **Not a merge blocker for this PR.**
2. The two remaining all-refs findings at `internal/api/handler/agent_summary_be2_test.go:301` (`"Idempotency-Key": "req-mismatch-…"`, commits `412e565e` / `09b1228f`, on `fork/feat/SUM-BE2-agent-draft-save` and `pr/202`) belong to an unrelated lineage; `origin/main`'s copy uses `idem-mismatch-example`, which does not trigger. That lineage will need its own exemption when it is PR'd. **Not a blocker here.**
3. If the org ever moves the gitleaks pin off 8.30.1, this entry may become dead weight rather than a fix. **Not a blocker.**

- C1, guard replacement — the stub `[[rules]] id = "generic-api-key"` carrying only an allowlist does *not* displace the default rule under `[extend] useDefault = true`: `internal/service/other.go:3` is still detected under both pre-PR and PR configs, and the all-refs scan under the PR config still reports the 2 unrelated findings. The header claim at `.gitleaks.toml:5` holds.
- C1, dual-path parity — sibling `docs/auth-optimization-backend.md`, added by the same commit, needs no exemption: scanning both historical docs yields exactly one finding, frontend only.
- C1, completeness — the all-refs full-history scan with the pre-PR config produces only 3 findings repo-wide, and the doc is the sole one in this lineage, so no second allowlist entry is missing.
- C5 — I did not accept "CI is green" as proof; the pinned scanner was re-run against the PR range, main's lineage, all refs, and the extracted historical file.
- C6 — no conflict with `SECURITY.md` (private coordinated disclosure to security@mininglamp.com); `.github/CODEOWNERS` is a blanket `*` rule, so no ownership path is bypassed and no public disclosure channel is opened.
- C2 / C3 / C4 — N/A: no control-flow ordering, authorization boundary, or container/member status cascade is touched; the diff is a scanner allowlist plus comments.
- Cross-round recheck — no earlier review round exists for head `c7079452`, so there are no prior blockers to re-verify.

## Verdict: APPROVED

The exemption is real, correctly targeted at a reproducible false positive, mechanically effective at the pinned scanner version, and harmless to every other detection path in the repo — I could not construct a P0/P1 from it. Both P2 items are quality-of-the-guard issues with small, tested fixes: narrow the doc exemption from a file-level `paths` kill switch to a value-level `regexes` entry, and make the comment state that the path lives only in history. Neither needs to gate this merge.

PR #248, opened 2026-09-14
feat(summary): 统一个人总结能力并修复 Agent 工具提交

wanyiwang06:
> 替代 #246：基于同一分支当前 head 重新发起 review；旧 PR 保留完整评审历史。

## 改动范围

- 为 Agent 保存的个人总结复用现有编辑、版本历史、Workflow 重新生成和定时更新能力；独立保存实际生成需求，不把显示标题当作原提示词。
- 增加 generation-config 更新接口，校验来源权限并默认持久化补齐的配置；重新生成不改变定时周期。编辑保留版本，重新生成失败恢复已有正文，并重置旧流状态。
- 修复继续优化的工具提交：正文由独立文本生成阶段完成，通过请求内 `content_handle` 提交，后端序列化为原有 `preview.content`，不改变对外响应/SSE。
- 非法工具参数最多原地修复两次；草稿就绪后只允许最终提交。保留授权、Map/Reduce、引用和超时约束，确定性模型/协议错误不重放整轮任务。
- 增加单测、默认关闭且仅记录结构元数据的诊断能力，以及实现/上线说明。数据库备份、私有现场回放、日志和本机镜像部署文件不纳入提交。

## 范围边界

保留“继续优化 → Agent 对话 → 新总结”的原有流程。本 PR **不实现局部修改路由优化**，也不保证继续优化不重新检索。配套前端：Mininglamp-OSS/octo-web#1662（来自 `wanyiwang06:codex/personal-summary-main-parity`）。

## 验证

- [x] `go test ./internal/agent ./internal/llmfallback ./internal/service ./internal/api/handler ./internal/worker ./internal/streaming`（本机 native 库路径已配置）
- [x] `go test -race ./internal/agent -run 'TestPreparedDraft|TestDraftWriter|TestToolArgument|TestInvalidArgument'`
- [x] API/Worker 已完成构建；既有隔离镜像中的最终版本通过 3 次引用场景及 1 次完整 fetch/Map/Reduce/draft/submit 真实模型验收，不等同于 CI 或模型成功率承诺。
- [x] `git diff --check`、提交内容敏感文件检查及与已拉取上游 main 的无冲突合并检查。

## 上线顺序

先备份数据，通过既有迁移机制应用 `migrations/sql/20260911-01-summary-generation-requirement.sql`，再部署后端，最后部署配套前端。新增 nullable TEXT 字段；Down 会删除该字段及保存的生成需求，回滚前须备份。

PR #249, opened 2026-09-15
fix(worker): renew personal-summary leases via heartbeat (#220 §2)

wanyiwang06:
## Problem (data-independent half of #220 §2)
A personal-summary run holds two leases and roots at `context.Background()`. `scanStuckPersonalTasks` re-dispatches any participant whose `worker_started_at` is older than **10 min**; `scanStuckTasks` re-dispatches any task past its `processing_deadline` (**WORKER_TASK_LEASE_MINUTES**, 20m). A degraded run's retry+fallback sequence (~727s at `LLM_TIMEOUT=180s`) — and the fetch stage's ~20m budget — legitimately outlive those windows, so a **slow-but-alive** run looks identical to a **dead** worker and gets a second concurrent dispatch: the same summary computed twice.

## Fix — lease renewal (heartbeat), not a run deadline
A stuck scanner exists to reclaim *dead* workers. Distinguish alive-vs-dead by renewing the leases while the run lives, instead of bounding the run (a bound would starve the configured fallback and truncate the SSE stream — see the discussion history).

`startLeaseHeartbeat(taskID, participantRefID)`, started once both leases are established and stopped via `defer` at the run's exit, every `personalStuckLease/3` (~200s):
- renews the **participant** lease (`worker_started_at`), guarded by `status = Processing`;
- renews the **task** lease (`processing_deadline`), guarded by task `status = Processing`;
- **ceiling** (`personalRunCeiling`, 40m — comfortably above the ~32m legitimate worst case of fetch ~20m + LLM ~12m): past it, renewal stops so the stuck scanner regains its recovery path for a genuinely wedged run rather than pinning the participant forever;
- **fence**: the renewal checks `.Error` (logged) and treats `RowsAffected == 0` as *lease lost* (the row is no longer Processing — a scanner reset or a re-dispatch owns it) and stops renewing.

`personalStuckLease` (10m) is now a shared const — the single source of truth for both the scanner window and the heartbeat — so they cannot drift.

## Behavior
- **Common fast path unaffected** — summaries finishing in seconds/minutes never rely on renewal.
- **Slow-but-alive runs** keep both leases fresh → no double dispatch, fallback still reachable, long pipelines and the SSE stream untouched.
- **Truly wedged runs** stop renewing at the ceiling → reclaimed by the scanner. A lease-lost run stops burning tokens once it notices `RowsAffected == 0` / aborts at the runnable check.

## Tests
[...]

PR #250, opened 2026-09-15
feat(summary): 统一个人总结能力并修复 Agent 工具提交

wanyiwang06:
替代 #248：从最新 main 重新建立分支并重新触发三方 review；旧 PR 保留完整评审历史。

## 改动范围

- 为 Agent 保存的个人总结复用现有编辑、版本历史、Workflow 重新生成和定时更新能力；独立保存实际生成需求，不把显示标题当作原提示词。
- 增加 generation-config 更新接口，校验来源权限并持久化补齐配置；重新生成不改变定时周期。编辑保留版本，重新生成失败恢复已有正文。
- 修复继续优化的工具提交：正文通过请求内 content_handle 提交，后端序列化为原有 preview.content，不改变对外响应/SSE。
- 非法工具参数最多原地修复两次；草稿就绪后只允许最终提交。保留授权、Map/Reduce、引用和超时约束。
- 包含 #248 各轮 review 已完成的 derived source、引用终态和失败恢复修复。

## 范围边界

保留“继续优化 → Agent 对话 → 新总结”的原有流程；不实现局部修改路由优化，也不保证继续优化不重新检索。配套前端：Mininglamp-OSS/octo-web#1674。

## 验证

- go test ./internal/agent ./internal/llmfallback ./internal/service ./internal/api/handler ./internal/worker ./internal/streaming
- git diff --check
- 从最新 upstream/main 移植 7 个提交，无冲突；功能差异仍为 55 files, +4433/-259。

## 上线顺序

先备份数据，通过既有迁移机制应用 migrations/sql/20260911-01-summary-generation-requirement.sql，再部署后端，最后部署配套前端。


wanyiwang06:
@yujiawei @Jerry-Xin @mochashanyao 请基于当前 head 重新 review。此 PR 从最新 main 重建并完整承接 #248 的功能与各轮已修复项；配套前端为 Mininglamp-OSS/octo-web#1674。

Jerry-Xin (dismissed):
# Review — PR #250 @ f4657b95f13b2071c8ea72e1583cc5c4df7c6a38

## Verdict: APPROVE

This is a successor round for #248 (closed unmerged when this PR was created). Verification was successor-style: nothing was blind-ported from the predecessor record; every claim below was re-established at this head by bytes or by a fresh run, and the carried items are explicitly identified as carried with the soundness argument stated.

## 1. Successor verification

### 1a. True merge-base
- Mirror is non-shallow (`git rev-parse --is-shallow-repository` → false).
- `git merge-base f4657b95 origin/main` → **635cb6ad4bd117f1eb01c1032399c19782fdd604**, which is exactly the current tip of main. The branch was rebuilt on latest main as the description claims; no stale-base gap.

### 1b. Patch-equivalence vs predecessor head 21562ef — IDENTICAL
- Predecessor effective change set: `git diff f3cf986e..21562ef` (f3cf986e = merge-base of 21562ef and main, re-confirmed by `git merge-base`).
- Successor effective change set: `git diff 635cb6ad..f4657b95`.
- Both patches: **252,224 bytes**, `cmp` → **byte-identical**. Both ranges: 55 files, +4433/−259 (GitHub API for #248 at close and #250 now report identical stats).
- Cross-check: `git diff 21562ef f4657b95 --shortstat` → 1 file, +10/−4 = `.gitleaks.toml` only, i.e. the sole difference between the two head trees is the main-side commit that landed between the two merge-bases. The PR-side change set is unchanged at the byte level.
- **Classification: IDENTICAL.** Not merely rebase-adapted — the effective diffs are the same bytes.

### 1c. The +4433/−259 vs ≈+4463/−289 stats question — resolved exactly
The ≈+4463/−289 figure is the arithmetic sum of the two #248 review-round deltas: range diff at RC-head eedb399b was +4194/−259, and the fix commit added +269/−30 on top. Naively summing (4194+269 / 259+30 = 4463/289) double-counts the 30 lines the fix commit deleted **from eedb399b's own additions** — those lines cancel out of any base..head diff. True effective delta: 4194 + (269−30) = **+4433**, deletions vs base unchanged at **−259** (the fix's 30 deletions removed PR-added lines, not base lines). Verified by `git diff --shortstat` on f3cf986e..eedb399b, eedb399b..21562ef, and f3cf986e..21562ef directly. No discrepancy remains.

### 1d. Main-delta survey (f3cf986e → 635cb6ad)
- Exactly **one** commit entered main: 635cb6a "fix(ci): allowlist documented summary API URL (#247)".
- It touches only `.gitleaks.toml` (+10/−4). It does not touch any of the 55 PR files (grep of PR file list → zero hits), does not touch citationtext/worker/schedule/agent code, and cannot interact semantically with this change set. No rebase-conflict-resolution risk spots exist: file-level overlap between the main-delta and the PR is empty.

### 1e. Five-fix spot-walk at this head (byte grep, not memory)
All five #248 blockers were re-confirmed present at f4657b95 by direct inspection of the checked-out bytes:
1. **P1-1** (regenerate-restore gate): unconditional restore block present in `markPersonalFailed` (personal_processor.go, function at :489; restore block ~:580-612 with the explanatory comment referencing PR#248 P1-1), executing before the `participantCount <= 1` branch — no longer gated into dead code.
2. **P1-2/P1-3** (compound brackets hard-fail; in-window prose rewrite): `CanonicalizeAdjacent` (citation.go:128) and `adjacentToMarker` (citation.go:169) present; worker finalization call-site personal_processor.go:1214 calls `CanonicalizeAdjacent`; `citation.go` diff vs base contains **zero deletion lines** (pure insertion — old `Canonicalize` byte-untouched); old `Canonicalize` non-test call-sites unchanged (agent/summary_draft.go:259,278; handler/agent_summary_citations.go:289; handler/agent_summary.go:759; service/edit.go:30).
3. **P1-4** (CreateSchedule source semantics): create path `inherited, hadStored, err := scheduleTaskSources(...)` with `if hadStored { sched.SourceConfig = inherited }` (schedule.go:645-657, comment referencing PR#248 P1-4) — zero-row Agent task keeps caller-supplied config instead of deterministic 40005; update path guarded by `if req.Sources != nil` (schedule.go:1285-1299) — nil-body task-scope update no longer rewrites source_config; `[]`-on-stored still 409 via the mismatch checks in scheduleTaskSources.
4. **P1-5** (derived-row leakage): `Where("task_id = ? AND derived = 0", ...)` at schedule_sources.go:27 with the full rationale comment (one-cycle promotion + 409 lockout prevented).
5. **markPersonalFailed restructure**: present as recorded (:489+, restore-then-branch shape).

Because §1b proved the entire effective diff byte-identical to 21562ef, this spot-walk is a belt-and-braces confirmation, not the sole evidence.

### 1f. Carried verification record (#248 fix-confirmation at 21562ef) — what was re-run vs carried
**Re-run at this head (fresh):** true merge-base; full patch-equivalence; main-delta survey; five-fix byte spot-walk; `gofmt -l` on all delta .go files (clean); `go build ./...` (exit 0); `go vet` on all changed package trees (exit 0); full `go test -race -count=1` on all changed packages (14/14 ok, 0 FAIL — counts in §4); CI check-runs; merge-tree vs #249 both orders; pre-write peer rescan.
**Carried from 21562ef (not re-run):** the probe battery (compound brackets [50011-2010], [7;9], [2-1], [9 73] no-hard-fail; in-window [3-5], [1,2], [12-14] byte-identical prose; real clusters still expand), the 8-mutant battery (M1/M2/M3/M5 assertion-killed; M4/M6/M7/M8 survivals recorded as advisories A-1/A-3), and the MySQL 8.0.46 lane (derived=0 predicate + full scheduleTaskSources flow against a real container).
**Why carrying is sound:** the carried results are properties of exact bytes; those bytes are proven identical at this head (§1b), and the only head-tree difference (`.gitleaks.toml` from main) is outside every carried verification surface. The MySQL lane is carried specifically because the schedule/store SQL bytes are patch-identical and the main-delta touched no SQL. Fresh suites were still re-run at this head, per precedent for verified-identical successor bytes.

## 2. #249 interaction / merge-order guidance
- `git merge-tree --write-tree` f4657b95 × 1b728aa8 (current #249 head, 3/3 APPROVE, unmerged): **clean in both orders**, producing the same combined tree (3a67808030df) — mechanical co-merge is conflict-free.
- `personal_processor.go` regions remain disjoint (#249: lease machinery ~:25/:314; #250: restore/finalization ~:580+/:1200+).
- Recommendation: merge **#249 first** (smaller, fully approved), then this PR; or co-merge — either is mechanically safe per merge-tree. Whichever PR merges second should re-run `internal/worker` tests once on the combined tree as cheap insurance.

## 3. Open advisories carried from #248 (all non-blocking, unchanged)
- **A-1**: P1-4 create/update source semantics unpinned by shipped tests (M6/M7/M8 probe-killed only); 3-test prescription recorded in #248 history.
- **A-2**: client-supplied source_name persists via the zero-row fallback path (display-only).
- **A-3**: resolved-member check unpinned (M4 survivor).
- **A-4**: isolated compound-bracket staleness (cosmetic).
- **A-5**: residual cluster ambiguity — recorded deliberate trade-off.
- **A-6**: TestNumericProseDoesNotBecomeCitationCorruption name-vs-coverage; in-window risk pinned by TestCanonicalizeAdjacent against the function the worker actually calls.
- **Spec 1b/1c**: #250's description is materially unchanged from #248's; the "不改变对外响应/SSE" claim (scoped to the content_handle submission path) coexists with observable contract changes elsewhere (generation_requirement field, EffectiveTopic→DisplayTopic, new prepare_summary_draft progress step). Still open as a documentation-accuracy advisory.
- **yujiawei's 7 P2s** (hub Prepare-in-tx commit-failure window; single-slot retiredRunID; maxIndex divergence; draftSource.Scope unsanitized; runner aggregate wall-clock non-stream paths; consecutive user roles vs strict gateways; GetSummary generation_requirement disclosure) and **mochashanyao's** zero-citation save gate (deliberate fail-closed per code comment) + empty-args terminal; **N-series** (N-1 regenerate lost-update fast-follow, N-2, N-3). None re-verified as blockers at this head; all remain advisory/fast-follow candidates.

## 4. Test record at this head
- gofmt on all delta .go files: clean. `go build ./...`: exit 0. `go vet` (agent/api/citationtext/llmfallback/model/service/streaming/worker trees): exit 0.
- `go test -race -count=1` — 14/14 packages ok, 0 FAIL: agent, agent/artifact, agent/eval, agent/finishgate, agent/summaryrun, agent/summaryspec, api/handler, api/router, citationtext, llmfallback, model, service, streaming, worker. Per-package test counts (top-level, -race -count=1 -v): agent 455, agent/artifact 19, agent/eval 4, agent/finishgate 45, agent/summaryrun 12, agent/summaryspec 10, api/handler 1253, api/router 8, citationtext 42, llmfallback 45, model 14, service 221, streaming 11, worker 202 — **2341 total, 0 FAIL**. citationtext (42) and worker (202) match the predecessor fix-confirmation record exactly, as byte-identity predicts.
- CI at f4657b95: Build / Lint / Vet / Generate (no drift) / Test / Test (race, cgo) / gitleaks / osv / dependency-review / history / add-to-project / pr-title-lint / label — all green; `check-sprint` failure is pre-existing on main and non-required (same baseline as predecessor rounds).

## 5. Credit
The five confirmed blocking findings fixed in this change set originated from the #248 review history: yujiawei (RC 5205386424, later APPROVE 5208910737 at 21562ef) and mochashanyao (RC 5205478359). The fix-confirmation record (probes + mutants) from that round applies to these identical bytes.

## 6. Peer state at write time
No other reviews on #250 at the time of writing (author re-review request 15:14:01Z only). Pre-write rescan performed; this review targets head f4657b95f13b2071c8ea72e1583cc5c4df7c6a38 exactly.



---

## ⚠️ CORRECTION (appended post-publication; original body above preserved verbatim)

This APPROVE is withdrawn by its author (2026-09-15) after independent byte-level adjudication of yujiawei's REQUEST_CHANGES (review 5212225941), posted ~11 minutes after this review on the same head. The verdict above no longer represents my position; this review is being dismissed. The authoritative vote at this head is yujiawei's REQUEST_CHANGES. Per duplicate-vote discipline, no replacement review is posted.

**P1-A (wrong-version restore after terminal per-person regeneration failure) — CONFIRMED TRUE, blocking.** Mechanism independently verified byte-for-byte:
- The restore path pins the user's selection: `current_version_id = source.ID` (personal_refine.go:680).
- The per-person regenerate reset nulls it: `"current_version_id": nil` (personal_refine.go:809), committed before the worker is dispatched (personal_refine.go:842-846).
- The worker reloads the row after commit, so in `markPersonalFailed` `pr.CurrentVersionID == nil`, the pinned branch (`query.Where("id = ?", *pr.CurrentVersionID)`) is skipped, and `Order("version DESC").First` restores the **highest** version — silently discarding a prior user restore of an older version.
- The whole-task regenerate reset (task.go:1455-1477) does **not** touch `current_version_id`, so the same restore block behaves correctly on the team path — an internal inconsistency confirming the dead branch on the per-person path, the exact path this fix was written for (per its own code comment).
- The `ensurePersonalVersionBaseline` backfill (personal_refine.go:929-936) only re-pins NULL to the highest version; no compensating path restores the user's selection.

Impact on the original body: §1e item 1 and the "ALL FIVE FIXED" carry in §1f were imprecise. The restore does run unconditionally (that part of the P1-1 fix holds and was correctly verified), but its version-pinning branch cannot fire on the per-person regenerate path — so the P1-1 remediation is incomplete for the restore-then-regenerate-failure scenario: the code's own declared contract ("a restored older version is the actual pre-regeneration body") is not fulfilled, producing a deterministic, user-reachable wrong-data state. This gates merge.

**P1-B (dangling single citation marker survives refinement) — mechanism TRUE, severity assessed as ADVISORY rather than blocking, on base-parity byte evidence.** The base refine path already persisted LLM content with zero citation validation and already called `CleanUnreferencedCitations` (base personal_refine.go:131/:366; base service/edit.go contains the same function), so the dangling-single `[n]` outcome is identical at base. This PR strictly narrows the hole by adding fail-closed compound validation (`NormalizeGeneratedCitations` at head personal_refine.go:131/:381); singles remain unvalidated exactly as at base and fall outside the declared scope ("normalize/validate compound citations"). The asymmetry vs the agent surfaces (`citationtext.Valid` validates singles at summary_draft.go:237 / tool_emit_summary_response.go:287) is real and worth a follow-up, but this is a residual pre-existing gap, not a PR-introduced regression.

**P2 (regenerate lacks lockPersonalWriteTask) — CONFIRMED TRUE, non-blocking**, matching its author's own rating: `task.Status` is checked outside the transaction (personal_refine.go:749) and only the personal-result row is locked inside it (:790-793), while edit/refine/restore all take `lockPersonalWriteTask` (:149/:399/:648, personal.go:658).

Everything else in the original body — the successor verification (byte-identical effective diff vs 21562ef), the +4463/−289 stats resolution, the main-delta survey, the #249 merge-tree guidance, and the test/CI record — remains accurate and is unaffected by this correction. Note that P1-A/P1-B/P2 all live in bytes that are identical to predecessor head 21562ef; they are new findings missed by all reviewers in the #248 rounds, not successor-round regressions.

yujiawei (changes requested):
# Code Review — PR #250 (octo-smart-summary)

Reviewed at head `f4657b95f13b2071c8ea72e1583cc5c4df7c6a38` against merge-base `635cb6a`. Independent line-level review plus a cross-family adversarial pass. 55 files, +4433/−259.

## 1. Spec compliance

The stated scope (docs/personal-summary-main-parity.md + PR body) is: unify Agent-saved and Workflow personal summaries over the existing edit/version/regenerate/schedule services; add a `generation-config` endpoint with source/time validation that does not change schedule recurrence; route the Agent terminal submission through a request-local `content_handle`; bound argument repair to two in-place retries; normalize/validate compound citations fail-closed; restore the last committed body on single-person regeneration failure.

- **No under-build found** — every scoped item is implemented: nullable `generation_requirement` + migration, `PUT /summaries/:id/generation-config` wired in the router, `EffectiveTopic`/`DisplayTopic` split, `prepare_summary_draft` tool + handle serialization, bounded repair (`maxToolArgumentRepairs=2`), `citationtext.Canonicalize`/`CanonicalizeAdjacent`, and the `markPersonalFailed` restore block.
- **No over-build found** — the change stays within the "continue-optimize still opens a new conversation; no local-edit routing; no schedule recurrence change" boundary. Source edits on schedule-bound tasks are correctly rejected (409). No stray flags/fields beyond what the frontend contract needs.

**Spec: ✅**

## 2. Code quality

**Quality: Changes-Requested** — two correctness defects on the regeneration/refine data path.

### P1 — Terminal regeneration failure silently reverts a user's restored version
`internal/worker/personal_processor.go` (~L585–595, `markPersonalFailed` restore block).

The new restore reads the pre-regeneration body as:
```go
query := tx.Where("task_id = ? AND user_id = ?", pr.TaskID, pr.UserID)
if pr.CurrentVersionID != nil {
    query = query.Where("id = ?", *pr.CurrentVersionID)
}
err := query.Order("version DESC").First(&latest).Error
```
But `RegeneratePersonalSummary` (`internal/api/handler/personal_refine.go` ~L809) resets the personal result with `"current_version_id": nil` **before** the worker runs. So when the worker later reloads `pr` and this block executes, `pr.CurrentVersionID` is nil and the query falls through to `Order("version DESC").First` = the **highest** version number. If the user had previously *restored* an older version (which set `current_version_id` to that older row and is the true "last committed body"), a regeneration that then fails terminally restores the wrong version — the highest one — silently discarding the user's restore. Note the whole-task `Regenerate` path in `task.go` was deliberately changed to *stop* nulling `current_version_id` for exactly this reason; the per-person path was not given the same treatment, so the `CurrentVersionID != nil` branch here is effectively dead on the one path it was written for. Fix: capture the active version id before the reset (e.g. snapshot into a field the worker can read, or don't null `current_version_id` on the per-person reset the way the team path already avoids).

### P1 — `NormalizeGeneratedCitations` lets an unauthorized *single* marker survive; metadata is then stripped, producing a dangling citation
`internal/service/edit.go` L21–30 → `internal/citationtext/citation.go` (`Canonicalize`).

`Canonicalize` intentionally only processes **compound** markers (`if !m.Compound { continue }`); single markers are left untouched and never validated. So refinement output containing a single `[2]` when only citation `[1]` exists passes `NormalizeGeneratedCitations` without error. `CleanUnreferencedCitations` then keeps only citations whose index is *referenced*, but there is no citation metadata for index 2 to keep — the result is a persisted `[2]` marker with no backing citation row (a broken/misleading reference in refine + restore + edit paths). The new tests cover invalid *groups* (`[9,74]`, `[9–73]`) but not invalid *singles*, so this slips through. Either validate single in-window markers too (the writer-side `citationtext.Valid` already does this), or document why single markers are exempt at these write surfaces. Given `emit_summary_response` validates singles but the refine/edit surfaces do not, the asymmetry looks unintended.

### P2 — Personal regeneration lacks the task-lock/revalidation the sibling write paths just gained
`internal/api/handler/personal_refine.go` — `RegeneratePersonalSummary` (~L741–815).

Edit / both refine transports / restore were all hardened to `lockPersonalWriteTask` (lock the task first, recheck status) inside their persistence transaction. `RegeneratePersonalSummary` checks `task.Status == Completed` *outside* the transaction and then, inside, locks only the personal result — it never re-locks/re-reads the task. A concurrent whole-task regenerate or a status/config transition landing after the stale check still lets this handler clear the result and dispatch a `personal_regenerate` worker against changed task configuration. Lower severity because the worker re-reads task state, but it's an inconsistency with the serialization model the rest of this PR establishes. Recommend applying `lockPersonalWriteTask` here too.

### Nits (non-blocking)
- `internal/api/handler/generation_config.go` `SaveGenerationConfig`: `saveGenerationScope` re-reads the task to re-check the schedule binding, but the enclosing tx already holds a `FOR UPDATE` lock on the same row from the top of the transaction — the extra read is harmless but redundant.
- `internal/streaming/hub.go` `Prepare`: the `retiredRunID` collapse (`activeRunID` else previous `retiredRunID`) tracks only one retired generation; two rapid requeues before any `EventStart` would let the oldest run's late `EventStart` through. The idle-TTL and the comment acknowledge this is best-effort; acceptable, but worth a follow-up if requeue bursts are real.

## 3. Coverage / blind spots
- **`prepare_summary_draft` + `content_handle` serialization**: verified the handle never enters durable history or the public payload (tests + code path confirm). Good.
- **Advisory legs run this review**: one cross-family reviewer (line-level adversarial) completed and its findings are folded in above (the two P1s and the P2 were independently confirmed against source). A second long-context advisory leg **timed out twice and is therefore absent** — it was NOT treated as agreement. Areas I could not cross-check against that missing leg and verified only directly: the MySQL row-lock concurrency test (`personal_write_mysql_test.go`, opt-in only, not run in CI), and runtime schedule-binding races under real concurrency.
- **Not exercised here**: the migration was not applied against a real MySQL instance; `down` drops the column and discards saved requirements (documented in the rollout section — ensure the ops runbook backs up first).

## 4. Overall verdict
**REQUEST_CHANGES** — Spec ✅ but two P1 correctness defects on the regeneration/refine data path (wrong-version restore on terminal failure; dangling single-citation markers) must be fixed before merge. The P2 lock asymmetry and nits are follow-ups.

mochashanyao (changes requested):
[Octo-Q · automated review]

---

# Code Review — PR #250 (octo-smart-summary)

## Summary

This PR unifies Agent-saved personal summaries with the existing Workflow capability set and reworks how the Agent planner submits its final draft. It adds a nullable `summary_task.generation_requirement` column so the generation instruction is stored separately from the display title, exposes it through `PUT /api/v1/summaries/:id/generation-config`, and lets `regenerate` optionally accept a time range and source set. On the Agent side it introduces `prepare_summary_draft`, a text-only writer that runs after Map/Reduce and returns a request-local `content_handle`; `emit_summary_response` now takes that handle instead of an inline `preview.content`, with the server re-serializing the original response shape. Supporting changes add bounded in-place repair of malformed tool arguments, a new `internal/citationtext` package shared by every citation surface, opt-in run-scoped tool diagnostics, `llmfallback.HTTPError` for status-aware error classification, a stream lifecycle reset on regenerate, and a personal-version restore when a regeneration permanently fails.

The design is coherent and the concurrency work (task-level `FOR UPDATE` on personal writes, baseline revalidation, in-transaction participant reads, generation-scoped stream reset) is a real improvement. The blocking problem is that the new shared citation package is wired with two different strictness policies, and the strict one is applied to surfaces that consume text the lenient one deliberately preserved.

## Verification

Static analysis only at head `f4657b95`; build and tests not executed in this environment (the review workflow forbids running builds or package managers).

- ✅ **Draft handle is request-local and immutable** — `summaryDraftState` is created per `RunWithHistoryOutcome` invocation (`internal/agent/runner.go:155`) and never inherited from a parent run; `resolveSummaryDraftArguments` (`internal/agent/summary_draft.go:319`) rejects a handle that is empty, mismatched, or paired with an inline `content`, so cross-request and inline-body submissions fail closed.
- ✅ **Post-preparation gating cannot spin** — once `state.handle != ""` the schema list collapses to the terminal tool (`internal/agent/runner.go:196`), a non-terminal call returns `nonterminal_after_preparation`, and a tool-less turn is bounded by `draftSubmissionNudges >= 2`, so the `step--` refund at `internal/agent/runner.go:496` cannot produce an unbounded loop.
- ✅ **Argument repair is bounded** — `maxToolArgumentRepairs = 2` (`internal/agent/tool_arguments.go:10`), validation happens before the turn is recorded or sibling tools dispatch, and `repairFinalOnly` prevents a post-budget repair from reopening evidence gathering.
- ✅ **Regenerate roster parity** — `acceptedParticipantIDs` (`internal/api/handler/task.go:1446`) and `resetParticipants` (`internal/api/handler/task.go:1502`) use the identical `status NOT IN (Pending, Declined)` predicate, so every personal stream that is reset corresponds to a personal result that was cleared; no stale-snapshot/empty-content split.
- ✅ **Failure restore is wired end to end** — `Regenerate` keeps `current_version_id` while clearing `content` (`internal/api/handler/task.go:1458`), `ensurePersonalVersionBaseline` runs before the clear, and `markPersonalFailed` (`internal/worker/personal_processor.go:588`) re-exposes that version; the read path returns `pr.Content` unconditionally (`internal/api/handler/personal.go:459`), so the restored body is actually visible.
- ✅ **Hub generation identity** — `Prepare` installs a fresh `*streamState`, and `deleteKeyIfIdle` compares pointer identity (`internal/streaming/hub.go:238`), so a stale timer from the retired generation cannot delete the new one.
- ✅ **Diagnostics are opt-in and redacted** — `toolDiagnosticsEnabled` requires an exact `AGENT_TOOL_DIAGNOSTIC_RUN_ID` match; only sha256 digests, byte counts, JSON root kind and syntax offsets are logged, model-authored tool names are collapsed to `unregistered`, and no prompt or argument value is emitted. The `[0]` index in `calls` is guarded by `err == nil`, which cannot hold for whitespace-only input.
- ✅ **Draft writer input is data-only** — `SummaryDraftSource.Scope` serializes `summaryWorkspaceContext` (`internal/api/handler/summary_workspace_contract.go:83`), which carries channels, participants, template, time range and referenced task IDs only; no planner system prompt or tool protocol reaches the writer.
- ✅ **Error surface tightened** — `Regenerate` no longer echoes `err.Error()` to the client (`internal/api/handler/task.go:1519`), and `llmfallback.HTTPError` preserves the upstream status through wrapping so `classifySummaryWorkspaceServiceError` can separate deterministic rejection from transport failure.

## Findings

One P1 contract break spanning four write surfaces, described first; three P2 items and one nit.

### P1 — Refinement and Agent-save surfaces use the strict citation normalizer on text that generation deliberately preserved (`internal/service/edit.go:30`)

`NormalizeGeneratedCitations` delegates to `citationtext.Canonicalize`, but every generation path in this PR uses `citationtext.CanonicalizeAdjacent` instead (`internal/worker/personal_processor.go:1214`) for two documented reasons: strict `Canonicalize` rewrites isolated in-window numeric prose into false citations, and it hard-fails on any group it cannot resolve. The PR's own tests fix both halves of that contract — `internal/citationtext/citation_test.go:117` asserts generation must persist `预算区间 [3-5] 万元。`, `涉及 [1,2] 两类问题。`, `参见附录 [12-14]。`, `依据 GB/T [50011-2010] 抗震规范执行。`, `编号 [7;9] 待确认。` and `排期 [2026-09] 确认。` byte-identically, while `internal/citationtext/citation_test.go:83` asserts `Canonicalize` errors on `[1-3]` and `[1;2]`.

Those preserved bodies are exactly what the refinement surfaces now feed back into the strict normalizer, and the two policies disagree:

```text
stored by generation          CanonicalizeAdjacent   NormalizeGeneratedCitations
依据 GB/T [50011-2010]        preserved (Indices=nil) ErrInvalid  -> HTTP 500
排期 [2026-09]                preserved (last<first)  ErrInvalid  -> HTTP 500
编号 [7;9]                    preserved (no split)    ErrInvalid  -> HTTP 500
预算区间 [3-5] 万元            preserved               "[3][4][5]" -> false citations
涉及 [1,2] 两类问题            preserved               "[1][2]"    -> false citations
```

`outsideEvidenceWindow` does not rescue the first three: it returns `false` when `len(m.Indices) == 0` (`internal/citationtext/citation.go:213`), so an unparseable group fails closed regardless of how far above the window its numbers are. `排期 [2026-09]` is not exempt either — `dateShape` requires a 4-2-2 shape, and the parsed `last=9 < first=2026` nulls the index list.

The last two rows are worse than a failure because they are silent: the expanded markers are persisted to `PersonalResult.Content` / `SummaryResult.Content` and to a new version row, so prose about a budget range renders as three citation popovers.

Diff scope: **new**. None of these call sites normalized before this PR — `internal/api/handler/personal_refine.go:131` and `:381`, `internal/api/handler/edit.go:302` and `:487`, and `internal/api/handler/agent_summary.go:763` are all added hunks. Base behavior was "save the model output", so both the 500 and the corruption are strictly worse than before.

Reachability of the 500 is not marginal. `CleanUnreferencedCitations` keeps only referenced indices, so a hole-y citation set is the steady state, and `maxIndex` is the largest referenced index rather than a count — any prose range that starts at or below it and touches a hole returns `ErrInvalid`. The refine prompt instructs the model to preserve unmodified content (`internal/api/handler/edit.go:795`), so the failure is deterministic across retries: the affected summary can never be refined again, and the LLM result is discarded each time. `internal/api/handler/refine_numeric_prose_test.go:22` does not catch this because its prose samples (`[2024-2025]`, `[2026-09-14]`, `[100-120]`) are all wholly above the tiny `{1,3}` window it seeds.

Fix direction: route the refinement and save surfaces through the same prose-safe normalizer generation uses — either call `CanonicalizeAdjacent`, or add a shared helper that expands only resolvable adjacent clusters and leaves everything else byte-identical — and reserve strict `Canonicalize` for the Agent draft writer, which is the only surface with a bounded repair loop (`internal/agent/summary_draft.go:186`).

### P1 — Agent save rejects citation-free drafts that contain ordinary bracketed prose (`internal/api/handler/agent_summary.go:759`)

Same contract, distinct consequence, so calling it out separately. The `len(cits) == 0` branch calls `citationtext.Canonicalize(content, nil, -1)`: with `valid == nil` every compound group errors, and `maxIndex = -1` disables the out-of-window prose exemption. A legitimately citation-free draft containing `规划[2024-2025]` or `页码[100-120]` therefore fails with `errWorkspacePreviewCitationUnresolved`, even though `internal/citationtext/citation_test.go:104` asserts that exact shape is prose under a known window and `:98` asserts it is prose under a citation-free known window. The comment's premise — no built metadata is not proof of an empty window — is sound, but `-1` also removes the only mechanism that distinguishes prose from a citation. The writer already gated citations at `internal/agent/summary_draft.go:236` with two repair attempts, so this second strict gate converts a savable run into a user-visible failure without adding a repair path. Fix: keep the strict check for markers that resolve against a known window, and treat unresolvable compound groups as preserved prose here.

### P2 — PersonalEdit is now locked out while a multi-person task is still generating (`internal/api/handler/personal.go:658`)

`lockPersonalWriteTask(tx, taskID, true)` admits only Completed/Failed/Cancelled, so `StatusProcessing` returns 409 / 40005. In BY_PERSON multi-person tasks each member's `PersonalResult` reaches Completed independently while `summary_task` stays Processing until meta aggregation, and the base handler had no task-status or `worker_status` check at all — an early finisher could edit their own body and now cannot until the slowest member finishes. The gate also earns nothing on that path: `lockedTask` is consumed only by `syncSinglePersonalDisplay` at `:716`, which is unreachable when `participantCount > 1` because `reviveCompletedForRecompute` returns first. The parts that do fix the race are the `FOR UPDATE` serialization, `checkPersonalWriteBaseline`, and the new Pending/Processing `worker_status` guard. Fix: keep those, but admit `StatusProcessing` (or gate on `pr.WorkerStatus` alone).

### P2 — Stream reset happens inside the transaction with no rollback compensation (`internal/api/handler/task.go:1505`)

`streamHub.Prepare()` replaces the in-memory `streamState` (`awaitingStart = true`, snapshot cleared, `retiredRunID` pinned) before commit, and only COMMIT can still fail at that point. If it does — deadlock, connection loss — the persisted rows still hold the previous completed content while subscribers sit in `awaitingStart`, where `Publish` drops every non-start frame (`internal/streaming/hub.go:130`), so the client shows an endless generating state for a run that will never start; `deleteKeyIfIdle` cannot reclaim the key while clients are attached. Fix: move the `Prepare` calls to just after a successful commit (the worker poller only sees Pending at commit anyway), or record the prepared keys and undo them when the transaction returns an error.

### P2 — Source guard is stricter on regenerate than on schedule create/update (`internal/api/handler/generation_config.go:62`)

`validateRegenerationConfigDB` rejects any non-nil `req.Sources` on a schedule-bound task, including a byte-identical re-send of the inherited set, while `scheduleTaskSources` (`internal/api/handler/schedule_sources.go:38`) explicitly permits resending the same identifiers and 409s only on add/drop/replace. `docs/personal-summary-main-parity.md` states the intended rule as "may resend the same set, but cannot add, drop, replace, or change source types". If the paired frontend echoes current sources on regenerate, every regeneration of a schedule-bound summary fails with 409 / 40005. Fix: reuse the same equality check so an unchanged set is accepted.

### Nit — Malformed-tool history drop also discards the assistant's prose (`internal/agent/history.go:275`)

`stripInvalidToolHistory` skips the entire assistant message when any call in the batch has malformed arguments, so a legacy turn carrying both a user-visible partial answer and one bad tool call loses that prose from planner context. The stated intent is to keep "any later user-visible answer". Correctness is unaffected because the paired tool messages go too; preserving `Content` as a plain assistant turn with `ToolCalls` cleared would match the comment.

## Human-verify

1. The paired frontend (octo-web#1674) is outside this checkout. Whether the P2 regenerate/source asymmetry and the P2 PersonalEdit gate are user-visible depends on whether it echoes current sources on regenerate and whether it enables personal edit while the task is Processing. Neither is a reason to change the server-side contract analysis above, but both need a client-side confirmation. **Not** independently a merge blocker.
2. This PR replaces #248 and states it carries forward that PR's resolved review items. #248's own review timeline is out of scope for this pass; the previously raised blockers there should be re-confirmed against this head by whoever holds that history. **Not** a merge blocker for this diff.

- Migration `migrations/sql/20260911-01-summary-generation-requirement.sql` is additive and nullable with a symmetric Down; `model.SummaryTask.GenerationRequirement` is `*string` with `type:text` and `omitempty`, so legacy rows and existing JSON consumers are unaffected. `EffectiveTopic` only prefers the new column for `TriggerAgent` with a non-blank value, and the API projections switched to `DisplayTopic` so the wire `topic` stays bounded.
- `SaveGenerationConfig` authorizes twice (space-scoped `authorizeTaskAccess` then creator equality), resolves the legacy requirement exactly once from the `FOR UPDATE`-locked row, re-checks the schedule binding inside `saveGenerationScope` against a freshly read row, and fails closed with 400 when requirement, time range or sources are all missing.
- `scheduleTaskSources` inherits only `derived = 0` rows, matching every other `summary_source` reader, and `hadStored` correctly leaves a zero-source Agent task's caller-supplied config untouched instead of overwriting it with an empty set.
- `saveGenerationScope` delete-then-recreate of `SummarySource` runs inside the transaction, dedupes on `type:id`, and re-resolves `SourceName` server-side, so a mid-loop failure cannot leave a half-replaced source set.
- `extractReferencedIndices` moving to `citationtext.Scan` widens protection (escaped brackets, images, link labels and destinations, reference definitions, indented code, fence delimiter/length) rather than narrowing it; the only index shapes it stops counting are `[0]` and 6+-digit brackets, neither of which `buildCitations` can produce.
- `attemptChat`'s new outgoing-argument guard is defense in depth only: the runner validates `turn.ToolCalls` before recording, and `stripInvalidToolHistory` cleans legacy batches, so the guard cannot fire on the live path and terminate a run.
- `classifySummaryWorkspaceServiceError` maps `InvalidToolArgumentsError`, `SummaryDraftError` and non-retryable `HTTPError` to 502 / 50003 with `transient=false` and a generic message, so upstream bodies are not leaked and a completed discovery/Map/Reduce run is not replayed. The deliberate tradeoff — a genuinely transient upstream 5xx inside the draft writer is also reported non-transient — is documented at `internal/api/handler/agent_summary_workspace.go:150`.
- `prepare_summary_draft` cannot race other tools: `runTools` refuses the whole batch when it is not the sole call (`internal/agent/runner.go:594`), and it is registered only in the `summary_workspace` profile.
- Terminal-submit conversion preserves the public contract: `resolveSummaryDraftArguments` deletes `content_handle` and re-injects `preview.content` before `parseSummaryResponsePayload`, so the persisted payload and SSE shape are unchanged, and `preview: null` still passes through for legacy callers.
- `retry_count` is now reset on both the task and each personal result during regenerate, closing the case where a re-queued run inherited an exhausted retry budget and failed permanently on its first error.

## Verdict: CHANGES_REQUESTED

The P1 is the blocker: one shared citation package is wired with two incompatible strictness policies, and the strict one now sits on four surfaces that consume text the lenient one guarantees will be stored verbatim. The consequence is either a permanent HTTP 500 on refine/team-refine for summaries containing shapes as ordinary as `GB/T [50011-2010]` or `[2026-09]`, or silent expansion of prose ranges into persisted false citation markers. Aligning the refinement and save surfaces with the generation-side normalizer resolves it; the P2 items are worth fixing in the same pass but are not blockers on their own.

[Octo-Q] verdict: REQUEST_CHANGES — P1 阻塞：citation 归一化在「生成路径（CanonicalizeAdjacent，保留 prose）」与「精修/保存路径（Canonicalize，严格）」之间策略不对称，导致本 PR 自己的测试语料（GB/T [50011-2010]、[2026-09]、[7;9]、预算区间 [3-5] 万元）在被生成路径合法落库后，一旦用户继续优化/团队调整就必然 HTTP 500 永久失败，或被静默改写成假引用标记并写入版本历史；按 R1「让本来能工作的路径不可用 + 产生用户可见错误数据 + 比修复前更糟」判 P1，按 R4 映射为 REQUEST_CHANGES。另有 Agent 保存边界同类 P1 一项、P2 三项、Nit 一项。

## 数据流回溯（每个被消费数据 → 上游来源 → 是否真流到消费点）

1. **`content_handle`** — 上游：`internal/agent/summary_draft.go:187`，仅在质量门（`summaryDraftQualityFailure`）与引用门（`canonicalizeDraftCitations`）双双通过后才赋值。消费点：`internal/agent/summary_draft.go:319` `resolveSummaryDraftArguments` ← 由生产 terminal handler `internal/agent/tool_emit_summary_response.go:271` 调用。状态注入：`internal/agent/runner.go:155`（每次 `RunWithHistoryOutcome` 新建，不复用父 run）；writer 输入注入：`internal/agent/runner.go:433` 写入 `toolsCtx`，并在 `:456` 以 `toolsCtx` 调 `runTools`。→ **真的会流到**，链路完整。
2. **`SummaryDraftSource`** — 上游：`internal/api/handler/agent_summary_workspace.go:707` `WithSummaryDraftSource(ctx, …)`。消费点：`summary_draft.go:141`（喂给 writer 的 `source`）与 `:271`（无取数改写时的引用索引集）。`analyses` 取自 `input.messages` 中 `Role=="tool" && Name=="merge_summaries"` 的结果；`toolsCtx` 在追加 prepare 调用**之前**快照 `msgs`，因此 Reduce 结果确实在内。→ **真的会流到**；未执行 Map 时 `analyses` 为空属预期，此时 `hasEvidence` 亦为 false，不会误触发「先合并再撰写」硬失败。
3. **`generation_requirement`** — 写入点三处：`agent_summary.go:634`（Agent 首存，仅 `requirement != ""` 时赋值）、`task.go:1416`（regenerate）、`generation_config.go:168`（配置保存）。读取点：`model.go:221`（`EffectiveTopic`，供生成）、`generation_config.go:23`（`generationRequirement`）、`task.go:1037`（详情 API）。→ 重点排查「写入空非 nil 字符串后永久关闭 `DisplayTopic` / AgentMessage 回退」这一 #296 类死 getter 陷阱：三条写入路径均在持久化前校验非空（regenerate 与 config-save 经 `validateRegenerationConfigDB` 以 400 拦住 `requirement == ""`，首存仅在非空时赋值），**空非 nil 不可达**，回退链未被永久短路。
4. **失败恢复的正文** — 上游：`task.go:1431` 在清空 `content` **之前** 调 `ensurePersonalVersionBaseline`；`task.go:1458` 的重置列刻意**移除** `current_version_id`（保留指针）。消费点：`personal_processor.go:592` 读 `pr.CurrentVersionID` → 回填 content/citations/generated_at 等 → `personal.go:459` 的读取路径**无条件**返回 `pr.Content`。→ **真的会流到用户**；恢复正文不会因 `worker_status=Failed` 或 `submitted_at=NULL` 被上层 getter 吞掉（已逐点确认 GET 路径不按 worker_status 过滤）。
5. **重置名单 vs 清空名单** — `task.go:1446` `acceptedParticipantIDs` 与 `task.go:1502` `resetParticipants` 使用完全相同的 `status NOT IN (Pending, Declined)` 谓词，且后者在前者把成员置为 Accepted 之后求值，集合不变。→ 不存在「结果被清空但流未重置」的 #343 类角标/列表分裂。
6. **`citationIndices`** — 上游：`tool_emit_summary_response.go:158` `setSummaryCitationWindow`，仅收 `CitationIndex > 0`。消费点：`summaryCitationIndexAllowed` 只在 `citationWindowKnown` 为真时查表，否则退回 `n>0 && n<=len(evidenceKeys)`。→ 双分支回退均已验证；窗口已知时**按集合而非区间**判定，in-window 空洞 fail closed。
7. **⚠️ 断裂的一条链（即 P1）** — worker `finalContent` →`personal_processor.go:1214` `CanonicalizeAdjacent` **刻意原样保留** prose 组 → 落库 `summary_personal_result.content` / `summary_result.content` → 精修接口读出 `pr.Content` 拼进 prompt → 模型按 `edit.go:795`「尽量保留用户没有要求修改的内容」原样吐回 → `personal_refine.go:131` / `edit.go:302` `NormalizeGeneratedCitations` → 严格 `Canonicalize` → `ErrInvalid`（HTTP 500，确定性重放）或静默展开成假引用并写入正文与版本历史。数据**确实流到了消费点**，问题是**生产端与消费端的策略互相矛盾**：生产端保证「这些形状必须原样落库」，消费端保证「这些形状必须报错或改写」。

## 盲点强制 checklist（C1–C6）

**C1 双路径 parity — HIT（P1 + P2 各一）**
- 🔴 命中：citation 归一化的生成端（`CanonicalizeAdjacent`，永不报错、永不改写孤立 prose）与精修/保存端（`Canonicalize`，严格、会改写、会硬失败）不对称，涉及 `personal_refine.go:131`/`:381`、`edit.go:302`/`:487`、`agent_summary.go:759`/`:763`、`agent_summary_citations.go:289` 共 6 个消费点。
- 🔴 命中（P2）：`generation_config.go:62` 对 schedule-bound 任务的 regenerate **拒绝同集合重发**，而 `schedule_sources.go:38` 对 schedule create/update **允许同集合重发**——同一份「不得更换来源」契约的两个入口策略不一致。
- ✅ 已逐对 grep 确认一致的成对路径：sources 的 delete↔recreate（`saveGenerationScope` 同一 tx、按 `type:id` 去重、失败整体回滚）；hub 的 `Prepare`↔`Publish` 代际（`hub.go:238` 指针身份校验，旧 timer 不会删新代）；stream 重置名单↔结果清空名单（同谓词）；`CreateSchedule`↔`UpdateSchedule` 的来源继承（两处都走 `scheduleTaskSources`，且 `hadStored=false` 时都保留调用方集合）；`Up`↔`Down` 迁移对称。
- ✅ 守卫/lint 自身等价写法覆盖：`invalidToolArguments`（`tool_arguments.go:33`）覆盖空串、非法 JSON、非 `{` 开头三类；多值拼接 `{"a":1}{"b":2}` 会被 `json.Unmarshal` 判错并归入 `invalid_json`，未漏。`terminalSchemas` 只按 name 过滤，模型幻觉出的未注册工具名由 `runner.go:214` 的 `nonterminal_after_preparation` 兜住，非守卫盲区。`runTools`（`runner.go:594`）对「prepare 未单独调用」整批拒执行，等价写法（与别的工具混批）已覆盖。

**C2 control-flow ordering / 嵌套复用 — clear**
- `runner.go:496` 的 `step--` 是唯一可疑的循环变量回退点：仅在 `len(ToolCalls)==1 && name==prepare && state.handle!=""` 时触发一次；下一轮 schema 收窄为 terminal-only，非终止调用直接 `nonterminal_after_preparation` 返回，无工具轮次受 `draftSubmissionNudges >= 2` 上限约束 → **不存在无限循环**。已 grep 全部 `step` 写点确认无第二处回退。
- `Canonicalize` 被 5 处复用、`CanonicalizeAdjacent` 被 1 处复用，语义各自内部一致；问题是**策略选择**（C1）而非 ordering 或双重作用。
- 安全控件非 canonical 形式试穿（`citationtext.Scan` / `protectedSpans`）：前导空格/tab `[ 1 - 3 ]` → 匹配且与 `[1-3]` 等价处理（一致）；全角括号 `［1］` → 不匹配 candidate → 视为 prose（偏安全侧，不会误判为引用）；反斜杠转义 `\[1\]` → `escaped()` 按奇偶计数排除；未闭合 fence、`~~~` 与 ``` 混用、fence 长度不足、缩进代码块、list 上下文内的缩进行、`![1]` 图片、`[1](url)` 链接目标、`[1]:` 定义、`[label][1]` 引用式链接 → 均在 `protectedSpans` 内显式处理。unicode 换行（NEL/VT/U+2028）不被 `dateShape`/`candidate` 接受，落在 prose 侧（fail-safe）。

**C3 授权边界 ≠ 能力边界 — clear**
- 新增能力入口 `PUT /api/v1/summaries/:id/generation-config`（`router.go:102`）：`authorizeTaskAccess`（space 作用域 + 存在性统一 40008，不泄露跨空间存在性）→ creator 等值校验 → 行锁 → 状态门槛 → 绑定门槛。授权校验在**能力入口本身**，不是只在外层首入口。
- 追溯「谁能到达这个能力」：`generationRequirement` 的 `AgentMessage` 查询带 `user_id = task.CreatorID AND session_id = task.AgentSessionID` 且 `space_id = ? OR space_id = ''`；`agentMessageRequirement` 的 spec JOIN 带 `run.user_id = ?` → 不能借他人 run 恢复指令。`validateSources` 使用 `task.SpaceID`/`task.CreatorID`（服务端所有），不采信请求体里的身份。`ResolveSourceNameForActor` 以 `task.CreatorID` 为 actor，来源名服务端所有。
- 新增诊断能力（`tool_diagnostics.go`）：暴露面为「精确 runID 与 env 相等」才开启，输出仅 sha256/字节数/JSON 根类型/语法偏移/finish_reason，模型自造工具名统一降级为 `unregistered`，不落 prompt、参数值、凭证、用户/频道名 → 能力扩大但授权边界未松动。

**C4 授权生命周期 / 容器-成员状态级联 — clear（一处 P2）**
- `scheduleTaskSources` 只继承 `derived = 0` 行，避免把 worker 自选频道升级为硬约束、也避免把 creator-only 频道暴露给全体参与者（`schedule_sources.go:14-22` 明确记录该理由）→ 容器（schedule）与成员（source 行）的授权范围未被放大。
- 掩码 ≠ 授权：`edit.go:299` 在 `callerPlainCitationsVisible` 为假时把 `basePlainCitations` 清空而不是照用，方向正确；但清空会把 `maxIndex` 变成 0，从而改变 `Canonicalize` 走 `outsideEvidenceWindow` 还是 `valid()` 的分支 → **同一 endpoint 的可观察行为依赖调用者引用可见性**，已并入 P1 说明。
- 状态级联：写路径现有「task 终态」+「personal `worker_status` 非生成中」双层门槛，成员生成中禁止写（正确方向）；反向级联（task 仍 Processing 但成员已 Completed）被一刀切拒绝 → P2（`personal.go:658`）。父容器 status 字段确实被查询/校验（`lockPersonalWriteTask` 以 `FOR UPDATE` 读 `summary_task` 后判 status，不是只 JOIN 不 filter）。

**C5 build/note 通过 ≠ 运行期路径正确 — HIT（作为 P1 的判定依据，非新增缺陷）**
- 本 leg **未执行** build/test（skill 明令禁止），因此不以「构建通过/测试绿」缓解任何结论；报告 Verification 段已如实写明 static analysis only。
- 反向也成立：PR 自带的绿灯**不等于**运行期正确。`refine_numeric_prose_test.go:22` 的 prose 样本（`[2024-2025]`、`[2026-09-14]`、`[100-120]`）全部选在所 seed 的 `{1,3}` 窗口**之上**，恰好绕过 in-window 展开与 hole 硬失败两条真实生产路径；`citation_test.go:117` 的 `CanonicalizeAdjacent` 用例覆盖到了这些形状，但只在生成端断言。
- 已按运行期推演（而非依赖 note/测试）给出 5 个具体字符串在两个函数下的实际结果对照表，见 Findings P1。文档/note 中的「fail closed」「preserved as prose」声明均回到代码分支逐一验证，未采信声明本身。

**C6 治理/策略/安全文档自洽性 — clear（一处文档-代码矛盾，已作为 P1 证据）**
- `docs/personal-summary-main-parity.md` 声称 Agent 保存与 Workflow「reuse the same citation builder and generated-content normalizer」——**与代码矛盾**：两端用的是 strictness 不同的两个函数（`Canonicalize` vs `CanonicalizeAdjacent`）。这是 P1 的文档自洽性证据，而非独立缺陷。
- 同文档声称「ISO-shaped bracketed dates are prose」——代码只对**单个** 4-2-2 日期豁免（`citation.go:44` `dateShape` 作用于整个 body）；`[2026-09]`（4-2）与 `[2024-01-01, 2024-06-30]`（含逗号的复合）均不豁免且硬失败。策略声明与实际行为不一致。
- 同文档的 schedule 来源策略（「Create/update may resend the same set」）与 `generation_config.go:62` 的 regenerate 行为不一致 → 已单列为 P2。
- 本 PR 不含 SECURITY.md / 披露流程 / label 策略 / CODEOWNERS / 发布流程类改动，无「新政策与安全义务冲突」风险；迁移文件与文档 Rollout 段的执行顺序（备份 → 迁移 → 后端 → 前端）自洽，且迁移为 additive nullable，先部署后端或先跑迁移均不会破坏旧代码。

## 跨轮 blocker 复检（R6）

**N/A — 本 PR 为首轮审查。** 只读确认：PR #250 `reviews = []`（无任何已提交 review），本 Octo-Q issue 评论历史为空 `[]`，故不存在上一轮未解决 blocker 可逐条复检。

- 审查对象为**当前 head** `f4657b95f13b2071c8ea72e1583cc5c4df7c6a38`；GitHub 报告的 `base_sha` = `635cb6ad4bd117f1eb01c1032399c19782fdd604`，本地 `git merge-base pr-base HEAD` 校验一致，diff 范围 `pr-base...HEAD` = 55 files / +4433 / -259，与 PR 元数据吻合。
- 本 PR 为 #248 的重建（自述「从最新 main 重新建立分支并重新触发三方 review；旧 PR 保留完整评审历史」），本轮为该 PR 第 **1** 轮，未触发「第 4 轮及以上 → 停止打地鼠、出整体方案重审」的升级建议。

⚠️ **交接给终审腿的一项未完成复检**：PR 自述「包含 #248 各轮 review 已完成的 derived source、引用终态和失败恢复修复」。#248 的 review 时间线不在本 leg 允许读取范围内（skill 与任务约束均禁查 GitHub 旧评论），因此 **#248 历史 blocker 是否全部真的落地，未经本 leg 对账**。代码中可见四处显式引用 #248 的修复点，本轮已逐一独立验证：
- `personal_processor.go:585`（#248 P1-1，恢复逻辑曾被 `participantCount<=1` 门控成死代码）→ **已修**：恢复块已提到分支之外，单人与多人都会执行，且 `pr.CurrentVersionID` 在 Regenerate 重置列中被刻意保留，数据流第 4 条已验证到底。
- `schedule_sources.go:20`（#248 P1，继承 derived 行会锁死 schedule 并越权暴露频道）→ **已修**：`derived = 0` 过滤 + `hadStored` 回退。
- `schedule.go:1288`（#248 P1-4，`req.Sources == nil` 时静默重写/清空已存配置）→ **已修**：仅在 `req.Sources != nil` 时改写，且零来源任务保留调用方集合。
- `citation.go:120` / `personal_processor.go:1206`（#248 B-1/B-2，prose 被改写成假引用、不可解析组硬失败）→ **仅部分修**：生成路径已用 `CanonicalizeAdjacent` 修好，**精修/保存路径未覆盖**，即本轮 P1。

建议终审腿在拍板前拉一次 #248 时间线做完整 blocker 对账，重点确认除 B-1/B-2 之外是否还有仅落在生成端、未同步到其他写表面的修复。

## 运行说明

- 工作流：`requesting-code-review` skill — `brief`（943 行索引）→ `coverage`（36 units / 35 `requires_inspect` / triggers: `brief-truncated`, `many-changed-files`, `unexpanded-review-units`）→ 读 `references/go-api-worker.md` + `references/output-format.md` → 逐 unit inspect（55 个变更文件全部覆盖，含 22 个测试文件）→ `finalize`：`validation_valid: true`、`invalid_comments: []`、`validation_errors: []`、`missing_high_risk_families: []`、`relocated: []`，2 条 `weak-failure-path` 提示（`personal_refine.go:131`、`edit.go:302`）已在正文补齐显式失败路径（HTTP 500 / code 50000）。
- 产物：`/tmp/review.final.json`（8 条结构化 finding）+ `/tmp/review.body.md`（本文）。
- 第二条审查腿的 fan-out 脚本未随 skill 安装（`skills/requesting-code-review/scripts/` 下仅 `pr_review_tool.py`，无 `glm_review.py` / `qwen_normalize.py` / `merge_reviews.py`，SKILL.md 内亦无 Dual-Model Fan-out / GATING 段），本轮按 skill 既定降级路径**单腿**产出，渲染源即 `/tmp/review.final.json`，不存在合并步骤可回退。
- 未执行 build / test / 包管理器（skill 明令禁止，Verification 段已如实标注 static analysis only）。
- 未做任何 GitHub 写操作：无 `gh pr review`、无 approve / request-changes、无 commit status；`gh` 仅用于只读拉取 PR head/base 与 diff。

yujiawei (changes requested):
# Code Review — PR #250 (octo-smart-summary) — Addendum (head f4657b95)

Follow-up to my earlier `CHANGES_REQUESTED` review. A second automated pass surfaced a sharper characterization of the citation defect; I re-verified it empirically (ran `Canonicalize` / `CanonicalizeAdjacent` against the stored-prose shapes below) and it holds. The verdict is unchanged; this addendum makes the blocking findings precise and adds three confirmed P2s. No new formal button is needed — `reviewDecision` is already `CHANGES_REQUESTED`.

## P1 (upgraded) — Two incompatible citation policies wired to the same shared package

`internal/service/edit.go:30` `NormalizeGeneratedCitations` → strict `citationtext.Canonicalize`, but the generation path (`internal/worker/personal_processor.go:~1214`) uses `citationtext.CanonicalizeAdjacent`, which **preserves** isolated bracketed-number prose and **never errors**. The refine/edit/save surfaces (`personal_refine.go:131`/`:381`, `edit.go:302`/`:487`, `agent_summary.go:763`) now feed generation-produced text back through the strict variant. Verified behavior on realistic stored content (real citations {1,3,7}, maxIndex=7):

```
input                               Canonicalize (consume)     CanonicalizeAdjacent (produce)
预算区间 [3-5] 万元。…[1]…[7]        ErrInvalid  -> HTTP 500     preserved (byte-identical)
依据 GB/T [50011-2010]…[1]          ErrInvalid  -> HTTP 500     preserved
排期 [2026-09]…[1]                  ErrInvalid  -> HTTP 500     preserved
编号 [7;9]…[1]                      ErrInvalid  -> HTTP 500     preserved
涉及 [1,3] 两类问题。               "[1][3]" (silent expand)   preserved
```

Consequence is either a **deterministic HTTP 500** that makes an affected summary permanently un-refinable (the refine prompt tells the model to preserve unmodified content, so it recurs every retry), or **silent expansion of prose ranges into persisted false citation markers** written to `Content` and a new version row. `outsideEvidenceWindow` does not rescue the unparseable groups (it returns false when `len(Indices)==0`). Diff scope is **new** — these call sites did not normalize before this PR, so both outcomes are strictly worse than base. `refine_numeric_prose_test.go` misses it because its prose samples sit wholly above the seeded `{1,3}` window. Fix: route the consume-side surfaces through the same prose-safe normalizer generation uses (or a shared "expand only resolvable adjacent clusters, else byte-identical" helper), reserving strict `Canonicalize` for the Agent draft writer, which is the only surface with a bounded repair loop.

## P1 (related) — Agent save rejects citation-free drafts containing ordinary bracketed prose

`internal/api/handler/agent_summary.go:759`: the `len(cits)==0` branch calls `citationtext.Canonicalize(content, nil, -1)`; with `valid==nil` every compound group errors and `maxIndex=-1` disables the out-of-window prose exemption, so a citation-free draft with `规划[2024-2025]` / `页码[100-120]` fails `errWorkspacePreviewCitationUnresolved` — even though the package's own tests assert those exact shapes are prose under a known window. The writer already gated citations upstream with two repair attempts; this second strict gate converts a savable run into a user-visible failure with no repair path. Fix: keep the strict check for markers resolving against a known window; treat unresolvable compound groups as preserved prose here.

## P2 — confirmed
1. `internal/api/handler/personal.go:658` — `lockPersonalWriteTask(tx,taskID,true)` admits only Completed/Failed/Cancelled, so an early-finishing member of a BY_PERSON multi-person task (task still `Processing`) now gets 409/40005 on PersonalEdit; the base handler had no such gate. The `lockedTask` it returns is also unused when `participantCount>1` (revive returns first). Keep the `FOR UPDATE` + baseline + `worker_status` guard; admit `StatusProcessing` (or gate on `pr.WorkerStatus` alone).
2. `internal/api/handler/generation_config.go:62` — `validateRegenerationConfigDB` rejects any non-nil `req.Sources` on a schedule-bound task, including a byte-identical re-send, while `schedule_sources.go` explicitly permits resending the same identifiers. If the paired frontend echoes current sources on regenerate, every regeneration of a schedule-bound summary 409s. Reuse the same equality check.
3. `internal/api/handler/task.go:~1505` — `streamHub.Prepare()` mutates in-memory stream state before COMMIT; if the commit then fails, persisted rows keep the old content while subscribers sit in `awaitingStart` (Publish drops non-start frames) → endless "generating" for a run that never starts. Move Prepare to just after a successful commit, or record and undo prepared keys on tx error.

## Nit
`internal/agent/history.go:275` `stripInvalidToolHistory` drops the whole assistant message when any call in the batch is malformed, discarding the assistant's prose too; the stated intent is to keep a later user-visible answer. Preserve `Content` with `ToolCalls` cleared.

## Note on the per-person regeneration restore (still a P1 from my first review)
The whole-task `Regenerate` path correctly preserves `current_version_id` while clearing content. The **per-person** `RegeneratePersonalSummary` (`personal_refine.go:~809`) still resets with `current_version_id: nil`, so `markPersonalFailed`'s restore (`personal_processor.go:~588`) falls to `Order("version DESC").First` = highest version — reverting a user's previously *restored* older version on terminal failure. Verified `personal_regenerate` dispatch reaches `markPersonalFailed`. This is a distinct code path from the whole-task one and remains blocking.

Jerry-Xin (approved):
# Code Review — PR #250 (octo-smart-summary) — Round 2, fix confirmation at head 249a4c11

This review supersedes my DISMISSED approval (5212090205) at the prior head f4657b95. The blocking findings raised there by yujiawei (5212225941, 5212600986) and mochashanyao (5212509095) drove this round and are fix-confirmed individually below. Delta f4657b95..249a4c11 = 1 commit, 17 files, +221/−101; every hunk maps to a peer finding (classification in §2). Merge-base is still current main (635cb6ad) and f4657b95 is a direct ancestor of this head.

## Verdict: APPROVE

All confirmed blocking findings are genuinely fixed (executable probes + assertion-red mutations); no new blocking defect was introduced by the delta.

## 1. Per-finding fix confirmation

### P1-A — Terminal regeneration failure reverts the user's restored version (yujiawei 5212225941) — FIXED
The fix shape is the "stop nulling" variant: the per-person regenerate reset no longer clears `current_version_id`, mirroring the whole-task path. Full lifecycle of the pinned field at this head:
- Restore writes pin = selected version (unchanged; re-verified by bytes).
- `RegeneratePersonalSummary` now takes the task lock (`lockPersonalWriteTask`) plus the `FOR UPDATE` row lock in one transaction, snapshots the baseline (`ensurePersonalVersionBaseline`) BEFORE clearing content, and leaves `current_version_id` untouched. Dispatch happens strictly after commit — no reset-vs-restore window.
- Worker reload sees the pin; the success path re-pins to the new version (`personal_processor.go:106/:153`); terminal failure resolves the pinned row exactly (`:591-602`) with `ErrRecordNotFound` tolerance preserved.
- Team path (`task.go`) is untouched — it never nulled, so no regression there. Backfill (`currentPersonalVersion`) resolves the pin first and only re-pins highest when the pin is NULL — coherent with the fix.

Evidence:
- Full-chain probe (throwaway, ran then removed): seeded the exact post-handler DB state this head produces (result reset, `content=""`, retry budget exhausted, `current_version_id=` the older selected version, with version rows v1-selected and v2-highest), then ran the real `markPersonalFailed` → restored v1 body AND v1 citation JSON, pin stayed v1, worker_status=Failed, multi-person decline path intact. A highest-version leak fails the probe.
- Shipped pins at both ends of the chain: `TestRegeneratePersonalSummaryPreservesSelectedBaselineIdentity` (handler keeps the pin) and `TestMarkPersonalFailedRestoresSelectedBaselineNotNewestVersion` + `TestMarkPersonalFailed_MultiPerson_RestoresBaseline` (worker honors the pin). The handler test's asserted end-state is exactly the worker test's input state, so the two compose into the full chain.
- Mutation M-A (re-add `"current_version_id": nil` to the reset map) → assertion-red on the handler test. KILLED.

### P1 citation-policy cluster — two strictness policies on one shared package (mochashanyao 5212509095 P1; yujiawei 5212600986 upgraded P1) — FIXED
Old-head claim adjudication first, honest in both directions: I verified mochashanyao's table empirically AT f4657b95 by executing the old `NormalizeGeneratedCitations` (a byte-for-byte passthrough to strict `Canonicalize`) on the quoted corpus. Results: `GB/T [50011-2010]`, `[2026-09]`, `[7;9]`, and `[3-5]` (hole vs the seeded window) all return ErrInvalid → deterministic HTTP 500 on refine, permanent across retries because the prompt instructs preserving unmodified content; `[3-5]` (all members resolvable) silently expands to `[3][4][5]` and `[1,3]` → `[1][3]`, persisted to content and a version row. Claim: TRUE on every row.

This also corrects my own prior grading: at f4657b95 I graded the dangling-single finding (yujiawei P1-B) ADVISORY on base-parity grounds. That framing was too narrow — base-parity covered singles, but it missed that the strict compound policy on consume surfaces made refine/save strictly worse than base, which performed no normalization there. mochashanyao's blocking framing was right, and yujiawei's addendum independently corroborated it. I own that grading miss; my dismissed approval remains the public correction.

Fix shape at this head (unification onto the prose-safe policy, with strictness reserved for the one surface that has a repair loop):
- `service.NormalizeGeneratedCitations` = `CanonicalizeAdjacent` (never errors; expands only fully-resolvable ADJACENT clusters) PLUS new in-window single validation: a single marker with `n <= maxIndex` whose index is not built returns `ErrInvalid` atomically (original content returned). This is exactly the "validate in-window singles too" ask from yujiawei 5212225941 P1-B.
- All six consume surfaces now route through the single helper: `personal_refine.go:131/:381`, `edit.go:302/:487`, `agent_summary.go:755`, and `agent_summary_citations.go:281` (`citationsValid`).
- The citation-free strict gate in `CreateAgentSummary` (`Canonicalize(content, nil, -1)`) is REMOVED — mochashanyao P1 #2 / yujiawei related P1.
- The Agent writer keeps strict `Canonicalize`/`Valid` (`summary_draft.go:237/:259/:278`, `tool_emit_summary_response.go:287`) — the only surface with the bounded repair loop; `summary_draft.go` is byte-identical to f4657b95, so the earlier salvage-gate fix carries over intact.
- Full call-site inventory of `citationtext` re-walked at this head: the strict variant exists only on writer paths; the lenient helper covers every consume surface.
- `docs/personal-summary-main-parity.md` is rewritten to match the actual policy, resolving the documented-vs-code contradictions (the "same normalizer" claim and the over-broad "ISO-shaped dates" claim) called out in mocha's C6.

Evidence — full two-sided probe battery re-run at this head (22 cases, throwaway, ran then removed):
- No-hard-fail shapes preserved byte-identical: `[50011-2010]`, `[7;9]`, `[2-1]`, `[9 73]`, `[2026-09]` (vs windows {1,3,7} / {9,73,88}).
- In-window isolated compounds with holes preserved: `[3-5]`, `[1,2]`, `[9,74]`; above-window `[12-14]` preserved.
- Real adjacent clusters still expand: `[1][3-5]` → `[1][3][4][5]` (all resolvable); `[88][9,73]` → `[88][9][73]`; an unresolved cluster member (`[1][3-5]` vs {1,3,7}) stays byte-identical.
- Singles: in-window dangling `[2]` vs {1,3} → ErrInvalid with original content returned; out-of-window `[9]` vs {1,3} preserved; degenerate `[0]` and `[123456]` are skipped by the parser's bounds rule (no `arr[0]`-shaped refine 500).
- Zero-citation draft `规划[2024-2025]，页码[100-120]` preserved.
- 2500-message-window protection: compound `[2024-2025]` vs maxIndex=2500 preserved (compounds never hard-fail under the lenient helper).
- `CleanUnreferencedCitations` on preserved `[9,130]` prose keeps exactly the resolvable metadata {9,88}.
- `citationsValid` probe (8 cases): citation-free prose accepted for a non-fetched run; a fetched run with no metadata and a surviving `[1]` still rejected; in-window dangling singles rejected at the finish gate.
- Mutations: M-B (revert helper to strict `Canonicalize`) → red on `TestNormalizeGeneratedCitations`; M-C (delete the single-hole validation) → red; M-E (revert `citationsValid` to the inline strict check) → red on `TestBuildAgentCompoundCitationsUsesWorkflowBuilder`. All KILLED.

New advisory (A-7, non-blocking, recorded trade-off): the in-window single-hole rejection is the strictest point of the unified policy. A stored body containing a bracketed single whose number is ≤ maxIndex with that index unbuilt — which requires the citation window to reach that number, e.g. a year-shaped `[2024]` inside a ~2500-message window with a hole at exactly 2024 — hard-fails refinement with a user-visible error and no repair loop. Reachability is narrow (window ≥ marker value + a hole at that index + a bare bracketed year as prose; compounds unaffected; generation unaffected), it fails closed without corruption, and it is the behavior yujiawei's P1-B explicitly requested. Recorded for awareness; no action needed for merge.

### Zero-citation gate removal — pinning gap (advisory)
Mutation M-D (re-adding the removed `Canonicalize(content, nil, -1)` gate) SURVIVED the shipped suite: no test drives `CreateAgentSummary` with a citation-free draft containing bracketed prose. The fix is probe-verified but the removal itself is unpinned. Prescription: one handler test — a workspace save with zero built citations and content `规划[2024-2025]` → 200, content preserved. Non-blocking.

### P2 cluster — all addressed
- **Regenerate lacks the task lock** (yujiawei 5212225941 P2): FIXED — `lockPersonalWriteTask` plus a mode re-check inside the persistence transaction; lock order (task → personal result) now matches edit/refine/restore. The ordering itself is structural and not separately mutation-tested; the reset-state pin is (M-A).
- **PersonalEdit locked out while a multi-person task is Processing** (mochashanyao P2-1 / yujiawei 5212600986 P2-1): FIXED via an `allowProcessing` parameter; single-person Processing still 409s. Pinned on both sides: M-F (revoke the admission) → red on `TestPersonalEditAllowsCompletedMemberWhileTeamStillProcessing`; M-F2 (drop the single-person 409) → red on `TestPersonalEditTerminalStatesAndRetention`. Both KILLED.
- **streamHub.Prepare inside the transaction** (mochashanyao P2-2 / yujiawei 5212600986 P2-3; also my carried #248-r2 item): FIXED — Prepare moved strictly after successful commit, with the roster query kept inside the transaction. Mutation M-G (move it back inside) SURVIVED: the ordering is observable only under commit failure, which no shipped test injects. Advisory prescription: a fault-injection test (commit failure → subscribers stay on the persisted run, no awaitingStart leak). Non-blocking: success-path coverage (`TestRegenerateResetsEveryClearedPersonalStream`) is intact and the fix strictly reduces the failure surface.
- **Same-set sources rejected on schedule-bound regenerate** (mochashanyao P2-3 / yujiawei 5212600986 P2-2): FIXED — `scheduleTaskSources` equality is reused both in `validateRegenerationConfigDB` and, under the transaction lock, in `saveGenerationScope`; server-owned names stay authoritative (a spoofed name in a re-sent set is not persisted — asserted by the new test). M-H and M-H2 (revert either leg to the unconditional 409) → red on `TestScheduleBoundRegenerateAcceptsSameSourceSet`. Both KILLED. The doc line now reads "Create/update and regeneration may resend the same set".
- **Nit — malformed tool batch discards assistant prose** (mochashanyao + yujiawei): FIXED exactly as prescribed (keep `Content`, clear `ToolCalls`, paired tool messages still dropped); M-I → red on `TestStripInvalidToolHistoryKeepsAssistantProse`. KILLED.

### Build + test execution (mochashanyao 🟡)
Done — see §6.

## 2. Delta classification (all 17 files, no unrelated work)
- P1-A + P2 lock: `internal/api/handler/personal_refine.go` (+ tests)
- Citation-policy unification: `internal/service/edit.go` (+ tests), `agent_summary.go`, `agent_summary_citations.go` (+2 test files), `refine_numeric_prose_test.go`, `internal/worker/citation_test.go`
- P2 gates: `personal.go`; P2 Prepare ordering: `task.go`; P2 same-set sources: `generation_config.go` (+ test)
- Nit: `internal/agent/history.go` (+ test)
- Policy doc alignment: `docs/personal-summary-main-parity.md`
The `lockPersonalWriteTask` signature extension touches all four call sites (build green proves exhaustiveness).

## 3. Mutation battery at this head (every mutant build-green first; all kills assertion-red)
| # | Mutant | Result |
|---|--------|--------|
| M-A | re-null `current_version_id` on per-person reset | KILLED |
| M-B | revert normalizer to strict `Canonicalize` | KILLED |
| M-C | delete in-window single-hole validation | KILLED |
| M-D | re-add zero-citation strict gate on Agent save | SURVIVED (advisory, §1) |
| M-E | revert `citationsValid` to inline strict | KILLED |
| M-F | revoke `allowProcessing` admission | KILLED |
| M-F2 | drop single-person Processing 409 guard | KILLED |
| M-G | move `Prepare` back inside the transaction | SURVIVED (advisory, §1) |
| M-H | revert same-set acceptance to unconditional 409 | KILLED |
| M-H2 | revert `saveGenerationScope` race re-check to 409 | KILLED |
| M-I | revert assistant-prose preservation | KILLED |
| M6 (A-1) | drop CreateSchedule `hadStored` guard | SURVIVED (carried, §4) |
| M7 (A-1) | drop UpdateSchedule nil-body guard | SURVIVED (carried, §4) |
| M8 (A-1) | drop `derived = 0` filter | KILLED by `TestScheduleSourcesIgnoreDerivedRows` (shipped at f4657b9 — this A-1 leg is now pinned; the earlier unpinned record predates the fix commit) |
| M4 (A-3) | drop resolved group-member check in `validateSources` | KILLED by `TestSummaryWorkspaceValidateSourcesEnforcesSpaceAndDMMembership` — A-3 resolved; the earlier "unpinned" record was stale (the test dates to the #238 workspace feature) |

11/15 killed; 4 survivors graded above, none blocking.

## 4. Carried advisory sweep
- **A-1** (schedule semantics unpinned): PARTIALLY RESOLVED — the derived-filter leg is now pinned (M8 killed). The `hadStored` guard (M6) and nil-body guard (M7) remain unpinned by shipped tests; prescription restated a third time: (1) a schedule created on a zero-source Agent task keeps the caller-supplied `sourceConfig`; (2) a schedule update with nil `Sources` does not rewrite `source_config`. Non-blocking: both guards are byte-intact at this head and were probe-verified in prior rounds.
- **A-2** (client `source_name` persists via zero-row fallback; display-only): unchanged, still open, non-blocking.
- **A-3** (resolved-member check unpinned): RESOLVED (M4 killed; stale record corrected).
- **A-4** (isolated compound staleness, cosmetic): unchanged; under the unified lenient policy more isolated compounds are deliberately preserved as prose, so the cosmetic concern is now the documented policy itself.
- **A-5** (residual cluster ambiguity): unchanged, recorded trade-off.
- **A-6** (`TestNumericProseDoesNotBecomeCitationCorruption` name-vs-coverage): RESOLVED BY EXPANSION — the consume surfaces it did not cover are now pinned by the rewritten `TestNormalizeGeneratedCitations`, both refine transports, and the numeric-prose cross-surface test.
- **A-7** (new): in-window single-hole refine 500 band — see §1, recorded trade-off.
- **Spec deviations 1b/1c**: the PR description was NOT updated this round (it still claims unchanged external response/SSE surfaces despite the `generation_requirement` field, the `EffectiveTopic`→`DisplayTopic` projection, and the added `prepare_summary_draft` progress step; the out-of-scope Workflow changes remain undeclared). The parity DOC was updated and is now accurate. Non-blocking; recommend the author align the PR description before merge.
- **yujiawei's #248-r2 carryovers**: hub-Prepare-in-tx → FIXED this round; single-slot `retiredRunID` → still open (acknowledged best-effort); worker-side maxIndex divergence → moot; `draftSource.Scope` unsanitized → unchanged; runner aggregate wall-clock (non-stream) → unchanged; consecutive user roles vs strict gateways → unchanged; `GetSummary` generation_requirement disclosure to participants → unchanged.
- **mocha's zero-citation save-gate concern (#248 round)**: resolved by removal — per his own prescription this round.
- **N-1** (regenerate lost-update fast-follow): MITIGATED this round by the in-tx task lock + baseline re-check on the per-person path. **N-2/N-3** and bind-gate alignment: regions untouched by this delta; carried unchanged.

## 5. #249 interaction
Merge-tree of this head against #249 (1b728aa8, 3/3 approved, unmerged) is CLEAN in both orders (combined tree dde7a5222782c27893f6a2c609d0a684e3377790). This delta touches no part of `personal_processor.go`, so #249's lease machinery there remains disjoint from #250's cumulative changes in that file. Either merge order works; whichever PR merges second should run the full suite on the combined tree as standard combined-regression hygiene, since both modify different regions of the same file.

## 6. Verification record
- `go build ./...` and `go vet ./...`: green at this head.
- gofmt: 8 flagged files are all untouched by this PR (byte-identical to main) — pre-existing, out of scope.
- Full suite `-race -count=1 ./...`: 24 packages ok, 2830 test executions including subtests (1645 top-level tests), 0 failures.
- Probes: 22-case citation battery, 8-case finish-gate battery, the full-chain P1-A handoff probe, and the old-head claim-table reproduction — all executed against real code; all throwaway files removed and the tree verified clean before submission.
- MySQL lane: CARRIED from the prior round — this delta contains zero raw-SQL and zero migration bytes (verified by diff scan); the SQLite blind-spot families (derived-table alias, collation, GROUP BY) are unaffected.
- CI at this head: Build / Vet / Test / Test (race, cgo) / Lint / Generate (no drift) / secret scan / dependency review / OSV all green; the `check-sprint` failure is pre-existing and non-required (identical at the prior head).
- Ancestry: merge-base = 635cb6ad = current main tip; f4657b95 is a direct ancestor of this head.
- Every claim above comes from a command run at this head or at f4657b95, as labelled.

## 7. Conclusion
**APPROVE.** The three blocking clusters at f4657b95 — wrong-version restore, the split citation strictness policies, and the citation-free save rejection — are each fixed in exactly the prescribed shape, pinned by shipped tests, and mutation-verified. The full P2 cluster and the nit landed too. My dismissed approval 5212090205 at the prior head stands as the public correction and is superseded by this review. The peers' REQUEST_CHANGES reviews remain live against the old head and are theirs to clear; recommend re-requesting review from yujiawei and mochashanyao at this head.

wanyiwang06:
按计划由干净历史的 PR #251 继续承载；本 PR 保留作为 review 历史参考。

PR #251, opened 2026-09-16
feat(summary): 统一个人总结能力并修复 Agent 工具提交

wanyiwang06:
## 基线与提交历史

- 从当前 `main` 创建干净分支。
- 将已确认版本 `0dcefce0aafa6da9b6120c7bb0914d6ddd513d56` 的净改动 squash 为单个提交。
- 本 PR 不包含 #248/#250 后续提交历史；#248 和 #250 保持原状。

## 改动范围

- 为 Agent 保存的个人总结复用现有编辑、版本历史、Workflow 重新生成和定时更新能力；独立保存实际生成需求，不把显示标题当作原提示词。
- 增加 generation-config 更新接口，校验来源权限并默认持久化补齐的配置；重新生成不改变定时周期。
- 修复继续优化的工具提交：正文通过请求内 `content_handle` 提交，后端序列化为原有 `preview.content`。
- 非法工具参数和草稿修复均有次数上限；保留授权、Map/Reduce、引用和超时约束。

## 范围边界

仅包含已确认 SHA 的功能净改动，不包含后续“总结完成后点击侧边栏自动进入新会话”相关迭代。

配套前端：Mininglamp-OSS/octo-web#1686。

## 验证

- `LIBRARY_PATH=/home/mlamp/.local/lib go test ./internal/agent ./internal/llmfallback ./internal/service ./internal/api/handler ./internal/worker ./internal/streaming`
- `LIBRARY_PATH=/home/mlamp/.local/lib go build -buildvcs=false ./...`
- `git diff --check`
- 相对 `main` 仅 1 个提交，55 files，+4112/-259。

## 上线顺序

先备份数据并应用 `migrations/sql/20260911-01-summary-generation-requirement.sql`，再部署后端，最后部署配套前端。


Jerry-Xin (dismissed):
✅ **APPROVE** at e4c65da5 — this head is byte-for-byte the fully fixed, previously verified state of #250 (249a4c11): identical tree hash, all confirmed fixes and their test pins present, full local verification green.

## 1. Successor arc and equivalence (byte-verified)

- First head 8fb0461d was a single squash commit whose full patch (237,598 bytes, 55 files, +4112/−259) was **byte-identical (`cmp` clean) to #248's first, pre-fix head 0dcefce0's patch**; tree content matched 0dcefce0 except main-side `.gitleaks.toml` (#247, already in the base). Audit at that head found both fix waves absent (delta vs #250@249a4c11: 27 files, +132/−573, every added line a reverted pre-fix form): pre-dispatch `"current_version_id": nil` restored, `markPersonalFailed` restore re-gated behind `participantCount <= 1` (dead code on the per-person path given the single-person 40005 rejection), `CanonicalizeAdjacent` deleted (strict normalization: silent prose corruption `[3-5]万元` → `[3][4][5]`; hard-fail ErrInvalid on `GB/T [50011-2010]` / `[2026-09]` / `[7;9]` — all reproduced with deterministic probes at that head), `derived = 0` schedule-source filter and `req.Sources != nil` guard absent, `streamHub.Prepare` inside the transaction, and ~360 net lines of test pins deleted (re-importing the deleted pins: 6 failed against 8fb0461d, one file failed to compile with `undefined: CanonicalizeAdjacent`). No review was posted on that head — it drifted before submission.
- e4c65da5 "address review parity follow-ups" (+573/−132, 27 files) is the exact inverse of that missing-fix delta. **`git diff 249a4c11 e4c65da5` is empty; both trees hash to `40e59f8be10669eec72b68f9c22f91d198d46813`.** Complete parity with the fixed state verified in the #250 round; identical tree hash leaves no room for smuggled deltas.
- Stats reconciliation: #251 now reports +4564/−270 — exactly #250's final numbers (same tree, same base blobs). Effective arithmetic across the two commits: naive addition gives 4112+573 = 4685 additions; 121 of the fix commit's 132 deletions removed lines the first commit had added (the known double-count trap), so effective additions = 4685−121 = **4564** ✓; the other 11 deleted lines were base lines, so effective deletions = 259+11 = **270** ✓. (+4112/−259 at 8fb0461d matched #248-at-creation exactly because the base-side blobs of all 55 files are unchanged between f3cf986e and 635cb6ad.)

## 2. Fix-presence audit at e4c65da5 (all byte-verified)

| Fix | Status | Evidence |
|---|---|---|
| P1-A stop-nulling + pin lifecycle | ✅ | zero `"current_version_id": nil` in the per-person regenerate Updates map; task lock re-acquired (personal_refine.go:789); row-lock reload; unconditional restore block in `markPersonalFailed` resolves the exact pinned row via `pr.CurrentVersionID` |
| PersonalEdit concurrency guards | ✅ | `lockPersonalWriteTask(..., allowFailed, allowProcessing)` 4-param form (personal_refine.go:860); `StatusProcessing → 40005` guard (personal.go:716) |
| Citation unification (all consume surfaces) | ✅ | `CanonicalizeAdjacent` present; wired in generation worker (personal_processor.go:1214) and `NormalizeGeneratedCitations` (edit.go:32); refine (personal_refine.go:131,:381), team edit (edit.go:302,:487), agent-save (agent_summary.go:755) all consume it; zero-citation gate removed; `citationsValid` unified (agent_summary_citations.go:281); strict `Canonicalize` writer surface untouched |
| Schedule sources derived filter | ✅ | `task_id = ? AND derived = 0` + `hadStored` fallback (schedule_sources.go:27; schedule.go ×4) |
| UpdateSchedule nil guard | ✅ | `if req.Sources != nil` only (schedule.go:1289); CreateSchedule hadStored guard (schedule.go:645 region) |
| Generation-config idempotent re-send | ✅ | validate-and-noop with locked re-check in `saveGenerationScope` (`writeSources` gate, generation_config.go:192-214) |
| hub Prepare post-commit | ✅ | task.go:1516-1523, after the transaction, before trigger goroutines are spawned |
| Tool-history content preservation + draft fallback | ✅ | history.go:281 strips ToolCalls while keeping non-empty content; `validatedMergedDraftFallback` present (summary_draft.go) |

## 3. Tests and CI

- Local at e4c65da5: `go build` ✅, `go vet` ✅, full `go test -race -count=1 ./...`: **24 packages ok, 0 FAIL**; `-v` run count **2830** — exactly matching the count verified on this identical tree in the #250 round.
- The restored pins are in-tree and green, including `TestMarkPersonalFailed_MultiPerson_RestoresBaseline`, `TestMarkPersonalFailedRestoresSelectedBaselineNotNewestVersion`, `TestRegeneratePersonalSummaryPreservesSelectedBaselineIdentity`, `TestPersonalEditAllowsCompletedMemberWhileTeamStillProcessing`, `TestScheduleBoundRegenerateAcceptsSameSourceSet`, `TestNormalizeGeneratedCitations`, `TestPersonalRefinePreservesIsolatedNumericGroupsBothTransports`, the `CanonicalizeAdjacent` battery, the schedule-sources derived-row tests, and `TestRegenerateResetsEveryClearedPersonalStream`.
- CI at e4c65da5: all checks pass except the known pre-existing environmental `check-sprint` failure (present across recent heads; not related to this diff).
- MySQL lane carried: the tree is byte-identical to 249a4c11 including `migrations/`, so the previously validated SQL lane result applies unchanged.
- Mutation status: this exact tree was mutation-verified in the #250 round (15 mutants, 11 killed; the survivors were graded advisory there). Grades carry at e4c65da5 since the bytes are identical.

## 4. Fresh-eyes re-walk at this head

- Pinned-version lifecycle re-traced end to end: handler reset points (no nulling anywhere pre-dispatch), dispatch strictly post-commit, worker reload resolving `pr.CurrentVersionID`, terminal-failure restore hitting the unconditional block, read path returning persisted content.
- Roster parity holds: results-clear set (`acceptedParticipantIDs`, task.go:1450) and stream-reset set (`resetParticipants`, :1452 region) share the same `status NOT IN (Pending, Declined)` predicate.
- Go value-semantics trap confirmed handled: `req.Sources = nil` in `validateRegenerationConfigDB` mutates a copy; the authoritative persistence guard is `saveGenerationScope`'s locked re-check (`writeSources`), pinned by tests.
- 🟡 **Advisory A-1 (pinned at e4c65da5)** — post-commit `Prepare` vs the independent worker poller: mechanism is real at hub level (`Prepare` retires whatever run is active, hub.go:53-70; a `start` equal to `retiredRunID` is dropped and non-start events are dropped while `awaitingStart`). For a new run's `start` to precede `Prepare`, the worker poller would need a tick inside the post-commit window and then to complete SELECT + claim UPDATE + reload + pipeline init + HTTP ingest before the handler executes ~10 straight-line statements after the commit ack — the racing path contains strictly more network round-trips than the handler's remaining path, so inversion requires an anomalous multi-ms stall of the handler goroutine at an exact instruction point. Impact is bounded and recoverable: the DB stays canonical and correct, subscribers miss one run's live events until refresh/reconnect, and the hub self-heals on the next `start` (fresh RunID). The alternative placement (Prepare pre-commit) fails deterministically on every commit rollback — the exact defect this fix removed — so the current trade is sound. Non-blocking follow-ups worth taking: a deterministic ordering test for `start(R_new) → Prepare → delta(R_new)`, and generation fencing (have `Prepare` retire only runs older than the requeue's generation token).
- 🟡 **Advisory A-2 (pinned at e4c65da5; feature-inherent, same shape since 0dcefce0 and at 249a4c11 — not a delta of this PR)** — editing a multi-person task in Failed/Cancelled state persists the member edit, but `reviveCompletedForRecompute` is a strict no-op there, so the team aggregate never absorbs it. Consider a conflict response for multi-person terminal tasks or documenting the limitation.

## 5. Merge mechanics

- `git merge-tree` vs open #249 is clean in both directions (shared files: `internal/worker/personal_processor.go`, `personal_processor_test.go`, `scheduler.go`; hunks disjoint — #249 touches lease/heartbeat paths, this PR touches restore/citation paths). Whichever merges second should re-run the worker package suite.
- #248 and #250 are both CLOSED; #251 is now the single live carrier and merging it lands exactly the #250@249a4c11 content (identical tree). Reviewers whose CRs went stale on #250's earlier head may re-review here for a fresh ticket.
- Deployment order per the PR description: back up and apply `migrations/sql/20260911-01-summary-generation-requirement.sql` → deploy backend → deploy companion frontend (octo-web#1686).

## Verdict

✅ APPROVE at e4c65da5 — tree-identical to the fully fixed and previously verified #250 state; all confirmed-finding fixes and pins present; suite, race run, and CI green (modulo the known environmental check); two advisories carried with prescriptions, neither blocking.

yujiawei (changes requested):
# Code Review — PR #251 (octo-smart-summary)

Reviewed at head `e4c65da534ea71d5a7de2389c8738e05786ba316` against merge-base `635cb6a`. Build (`go build ./...`) and the listed test packages pass locally. This is a large change (56 files, +4564/−270) so the review is split into a spec-compliance pass and a code-quality pass.

## 1. Spec compliance

The diff implements the stated scope well:
- `generation_requirement` is added as a nullable column separate from the display title, with legacy recovery only from an owned Agent run (`model.go`, `generation_config.go:generationRequirement`).
- `PUT /summaries/:id/generation-config` and optional source/time config on `regenerate` are wired with creator-only checks, `SELECT ... FOR UPDATE` status guards, and schedule-source immutability (`schedule_sources.go`).
- The `prepare_summary_draft` → `content_handle` → `emit_summary_response` protocol keeps the public response/SSE shape unchanged and moves the body out of tool-call JSON.
- Bounded argument repair, stream-lifecycle reset (`hub.Prepare`), and personal version baseline/snapshot are all present.

**One documentation/behavior divergence worth flagging** (see P1 below): `docs/personal-summary-main-parity.md:68` states "Agent saves, Workflow and refinement reuse the same prose-safe generated-content normalizer." In the code the Workflow (`personal_processor.go:1214`) and refinement (`service/edit.go:32`) paths use the prose-safe `CanonicalizeAdjacent`, but the Agent draft/emit path (`summary_draft.go:259`, `tool_emit_summary_response.go:287`) uses the strict `Canonicalize`/`Valid`. That is not the same normalizer, and the difference has functional consequences.

**Spec: ❌** — not because scope was exceeded, but because the citation-consistency behavior the doc promises ("prose-safe, byte-identical isolated ranges") is not what the Agent path actually does, and that gap produces the correctness defects below.

## 2. Code quality

**Quality: Changes-Requested.** Three P1 defects share one root cause: `internal/citationtext` implements both a strict validator (`Valid`/`Canonicalize`) and a prose-safe expander (`CanonicalizeAdjacent`), but the Agent-facing surfaces use the strict half in places where the input can contain ordinary bracketed-number prose. The strict half is only safe when every bracketed number in the text is genuinely a citation — which is not true of free-form summary bodies.

### P1-1 — Legitimate Agent drafts are hard-rejected, discarding the whole turn
`internal/agent/summary_draft.go:237` (`summaryDraftQualityFailure`) and `internal/agent/tool_emit_summary_response.go:287` both call `citationtext.Valid(..., requireMarker=true)`. `Valid` returns `false` for any malformed compound marker inside the evidence window (`citation.go` — a group whose members don't parse yields `Indices=nil`, and `outsideEvidenceWindow` cannot excuse an in-window/ambiguous group). The shapes that fail are exactly the ones this PR's own test pins as legitimate prose (`citation_test.go:122-128`): `GB/T [50011-2010]`, `[7;9]`, `[3-5]`, `[2026-09]`.

Verified empirically — a draft that contains a real citation plus one of these prose shapes:
```
Valid("依据 GB/T [50011-2010] 抗震规范。参考 [2]。", {1..5}, 5, true) = false
```
Consequence: `summary_draft.go` retries twice, the fallback re-runs the same failing check, and the run ends in `SummaryDraftError` → HTTP 502 / code 50003, `transient=false`, so the entire (successful) fetch/Map/Reduce turn is discarded. The pre-PR `\[(\d+)\]` marker regex could not match these shapes, so this is a new failure mode.

### P1-2 — Prose ranges are silently corrupted into false citations in the persisted body
`internal/agent/summary_draft.go:259` (`canonicalizeDraftCitations`) calls `citationtext.Canonicalize`, which has no adjacency rule and expands any compound group whose members all resolve. The Agent evidence window is effectively contiguous, so ordinary prose resolves. Verified empirically:
```
Canonicalize("预算区间 [3-5] 万元。", {1..5}, 5) = "预算区间 [3][4][5] 万元。"
```
This is the exact corruption `citation.go:117-118` documents `CanonicalizeAdjacent` as existing to prevent, and it lands in the saved summary body. The save-time normalizer is idempotent and cannot undo it.

### P1-3 — Worker harvests prose-range indices into false `model.Citation` rows
`internal/worker/citation.go:23` swapped the old `citationRe` regex for `citationtext.Scan` in `extractCitationIndexes`. `Scan` returns the expanded members of compound groups, so an isolated prose range now yields citation indices the old regex ignored. Verified empirically:
```
extractCitationIndexes("预算区间 [3-5] 万元，无其他引用。") = [3 4 5]   // old regex matched nothing
```
`buildCitations` only filters by `idx <= maxIdx` (authorized message count), so if those indices fall within range, false `model.Citation` rows are built and persisted. `stripOrphanCitations` (`citation.go:373`) still uses the old single-marker `\[(\d{1,5})\]` regex, so it cannot remove the `[3-5]` text that produced them. `CanonicalizeAdjacent` at `personal_processor.go:1214` leaves the isolated range byte-identical, so the mismatch survives.

**Suggested direction:** route the Agent draft/emit validation and the worker index extraction through the same prose-safe policy the Workflow path already uses (`CanonicalizeAdjacent` + a validator that ignores isolated non-citation groups), so isolated `[3-5]` / `[50011-2010]` / `[2026-09]` prose is neither rejected, rewritten, nor harvested. Align `stripOrphanCitations` with whatever `Scan` now considers a marker.

### P2 — `generation-config` source re-validation skip relies on a downstream guard
In `validateRegenerationConfigDB` (`generation_config.go:62-69`), when `task.ScheduleID != nil && req.Sources != nil` the function sets a local `req.Sources = nil` to skip permission re-validation, and when the task has no stored sources `scheduleTaskSources` returns `(nil,false,nil)` so the same skip happens. This does not create a real exposure: `saveGenerationScope` re-reads the locked task and sets `writeSources = false` for any scheduled task, so nothing unvalidated is persisted (a caller-supplied set on a scheduled task is dropped, not written). But the safety depends entirely on that second guard, and `req` is passed by value so the local `nil` is invisible to the caller — the intent in the comment ("needs no persistence") is enforced elsewhere, not here. Consider making the skip explicit / documented at the persistence boundary so a future refactor of `saveGenerationScope` cannot reopen the gap.

### Nits
- `docs/personal-summary-main-parity.md:68` overstates normalizer sharing (see Spec section) — tighten the wording once the P1s are addressed.
- `generation_config.go` comment "needs no permission re-validation or persistence" reads as if this function enforces it; the persistence guarantee actually lives in `saveGenerationScope`.

## 3. Overall verdict
**REQUEST_CHANGES** — Spec ❌ (documented citation behavior not met) AND Quality Changes-Requested (three confirmed P1 correctness defects: legitimate drafts discarded, prose corrupted in saved bodies, false citation rows persisted). The generation-config, handle-protocol, streaming-reset, and versioning work is otherwise solid; the blockers are localized to the citation-normalization boundary and share a single fix direction.

wanyiwang06:
已在 `24d0d20a` 逐条修复本次 review 的三个 P1 与两条 nit。思路与 Workflow 既有策略对齐(#248/#250 的 `CanonicalizeAdjacent`),不放宽 blocker-4 的两条硬保证(窗口内单标记逐一校验 + chat 摘要至少一个真实标记,仍由新校验器强制)。

### P1-1 合法草稿被弃(retry→502)
- 新增 `citationtext.ValidAdjacent`(与 `CanonicalizeAdjacent` 同一邻接规则的安全校验):单标记与「真实引用簇内」的复合组仍逐一 fail-closed;**孤立复合组按普通文本放行**(`GB/T [50011-2010]`、`[2026-09]` 不再整稿作废)。
- 接入点:`summary_draft.go` 质量门 + `tool_emit_summary_response.go` 终稿门,均由 `Valid` 换为 `ValidAdjacent`。
- 探针用例固化:`TestValidAdjacentProbes`(含原始复现输入、requireMarker 语义、簇内损坏成员拒绝、越窗单标记拒绝)。

### P1-2 存档正文被改写
- `canonicalizeDraftCitations` 两个分支(证据窗 / reference-only)均改用 `CanonicalizeAdjacent`:只有紧邻引用簇内的复合组才展开,`排期 [3-5] 周` 保持字节不变,不再被静默写成 `[3][4][5]`。对应断言更新于 `TestPreparedDraftCanonicalizesGroupsWithoutAnotherModelCall`。

### P1-3 幽灵引用行
- `worker.extractCitationIndexes` 只收割单标记,与 `stripOrphanCitations` 的单标记正则口径一致(Scan 本身保留,继续享受代码块/转义/日期/链接防护)。孤立复合组端到端保持 prose:不建行、自然也不会出现「抽得出但清不掉」。
- 同步更新 worker 与 handler 侧引用 fixture(`TestExtractCitationIndexes`、`TestWorkflowCompoundCitationFinalization`、`TestPersonalPipelineCitationFinalization`、`TestCreateAgentSummary_WorkspaceCompoundCitationSave`、`TestBuildAgentCompoundCitationsUsesWorkflowBuilder`),并新增「孤立复合组建 0 行」反向断言。

### Nits
- `docs/personal-summary-main-parity.md`:措辞收敛为「normalizer 为 `CanonicalizeAdjacent`,Agent 门为伴生校验器 `ValidAdjacent`」,不再笼统说 shared normalizer。
- `generation_config.go` 注释:改为明确「此处仅跳过重复校验;持久化保证在 `saveGenerationScope` 的 `writeSources=false`」;并加 pin 测试 `TestRegenerateScheduledTaskDropsCallerSources`(定时任务重发相同 sources → 校验通过但落库被丢弃、存量集合不变)。

### 验证
- `CGO_ENABLED=1 go test ./internal/...` 24 包全绿;`CGO_ENABLED=0` 全绿;改动文件 `gofmt -l` 干净。
- 三条 P1 的原始探针输入均已成为回归用例。

yujiawei (approved):
# Code Review — PR #251 (octo-smart-summary)

Re-review at head `24d0d20a0741cfb274740ca5f5804dacc51a02bd` against merge-base `635cb6a`. This head adds exactly one commit on top of the previously-reviewed `e4c65da` — `fix(citations): prose-safe validation for Agent paths` — which targets the three P1 correctness defects raised in the prior `CHANGES_REQUESTED`. Local `go build ./...` is green and the citation/agent/worker/handler test packages all pass. A second (adversarial) review perspective was incorporated; a third external perspective was attempted but returned no findings and is therefore absent from this pass.

## 1. Spec compliance

The follow-up commit implements precisely what was requested and nothing beyond it:

- **P1-1 (legitimate drafts discarded)** — fixed. `summary_draft.go:237` and `tool_emit_summary_response.go:287` now validate with the new prose-safe `citationtext.ValidAdjacent`. Verified empirically: `依据 GB/T [50011-2010] 抗震规范。参考 [2]。`, `预算区间 [3-5] 万元。参考 [1]。`, and `时间 [2026-09] 完成。见 [4]。` all now validate instead of triggering the retry→502 loop.
- **P1-2 (prose corrupted in saved bodies)** — fixed. `summary_draft.go:268/283` switched from strict `Canonicalize` to `CanonicalizeAdjacent`. Verified: `预算区间 [3-5] 万元。` is now left byte-identical rather than rewritten to `[3][4][5]`; a genuine cluster `[1][3-5]` still expands to `[1][3][4][5]`.
- **P1-3 (phantom citation rows)** — fixed. `worker/citation.go:26 extractCitationIndexes` now harvests single markers only, agreeing with `stripOrphanCitations`' single-marker regex. Verified: `extractCitationIndexes("[9,73] [93–95]")` now returns `nil` (was `[9 73 93 94 95]`).
- **P2 (generation-config persistence guarantee)** — addressed with a clarifying comment at `generation_config.go:62-72` and a pin test that a caller-supplied source set on a scheduled task is dropped, not written (`saveGenerationScope` forces `writeSources=false`).
- **Doc divergence** — `docs/personal-summary-main-parity.md:66-70` now names `CanonicalizeAdjacent`/`ValidAdjacent` explicitly, so the documented normalizer matches the code across Agent, Workflow, and refinement paths.

The test fixtures were tightened to pin the new contract (isolated compounds are prose end-to-end), not loosened to hide behavior. No scope was exceeded.

**Spec: ✅**

## 2. Code quality

The fix is well-scoped and consistent. All four generated-content write surfaces — Agent draft (`canonicalizeDraftCitations`), Agent emit (`ValidAdjacent`), Workflow (`personal_processor.go:1214`), and refinement/save (`service.NormalizeGeneratedCitations` → `citationsValid`) — now share one prose-safe policy. `ValidAdjacent` correctly keeps the blocker-4 guarantees (single markers and cluster members must resolve inside the evidence window; `requireMarker` still demands a real citation) while exempting isolated compound prose.

**Quality: Approved**, with the following non-blocking observations.

### P2 — Residual adjacency-heuristic corner case (shared with the already-shipped Workflow path)

`adjacentToMarker` (`citation.go:169-185`) treats a compound group as part of a citation cluster when only spaces/tabs separate it from another marker. Two consequences of the heuristic remain:

- A prose range written flush against a real citation — e.g. `预算区间 [3-5] [2]` with evidence `1..5` — passes `ValidAdjacent` and is then expanded by `CanonicalizeAdjacent` to `预算区间 [3][4][5] [2]`, after which the Agent save path (`buildCitationsForSession` → `BuildCitations`) harvests phantom rows for 3/4/5. Verified empirically end-to-end.
- Conversely a descending standard number adjacent to a marker, `依据 GB/T [50011-2010] [2]`, becomes a malformed in-cluster group and fails `ValidAdjacent`.

I am **not blocking on this** for three concrete reasons: (a) it is the *identical* behavior of `CanonicalizeAdjacent` already in production on the Workflow path (#248/#250) — this PR brings the Agent path to parity, not regression; (b) it is strictly better than the pre-PR Agent path, whose strict `Canonicalize` expanded *every* resolvable compound unconditionally and hard-rejected `GB/T`; (c) it requires an adversarial input shape (a prose range separated from a real marker by only whitespace), which the single-marker `OutputRule` actively steers the model away from. 

Suggested follow-up (cross-cutting, affects the shared normalizer — best tracked separately rather than folded into this PR): tighten `adjacentToMarker` to require *zero* separating characters, since genuine clusters are contiguous `[9][73]` per the OutputRule while a whitespace gap (`[3-5] [2]`) is the signature of prose followed by a citation. That would resolve both cases above without weakening real-cluster expansion.

### P2 (pre-existing, out of this diff) — raw-regex cleanup is not `Scan`-protected

`dedupCitations`/`stripOrphanCitations` (`worker/citation.go:240,377`) still use the raw `\[(\d+)\]` regex, so a bracketed integer inside inline code (`items[1]`) or a link can be mis-handled during cleanup. These functions are **not modified by this PR** and are **not on the Agent save path** (which only calls the `Scan`-protected `BuildCitations`); this is pre-existing Workflow-path tech debt worth a separate ticket, not a change request here.

## 3. Overall verdict

**APPROVE** — Spec ✅ AND Quality Approved. The three P1 correctness defects are fixed and independently verified, the P2 persistence concern is documented and pin-tested, and the doc wording now matches the code. The remaining items are P2 advisories: the residual adjacency-heuristic corner case is consistent with the already-shipped Workflow normalizer and is a candidate for a cross-cutting follow-up, and the raw-regex cleanup is pre-existing and outside this diff.

## 4. Suggestions

- Consider the `adjacentToMarker` tightening above as a follow-up on the shared `citationtext` normalizer (benefits both Agent and Workflow paths).
- Track the `dedupCitations`/`stripOrphanCitations` `Scan`-protection gap as separate tech debt.

## 5. Additional notes (coverage / blind spots)

- **Concurrency (unverified at runtime):** `Hub.Prepare` (`streaming/hub.go`) is invoked after the regenerate transaction commits (`task.go:1518-1523`) and retires the active run via `retiredRunID` + `awaitingStart`. The ordering against a worker publishing `EventStart` for a new run looks correct under the lock, but the commit→Prepare→worker-start window was not exercised under real concurrency.
- **Migration idempotency:** `20260911-01-summary-generation-requirement.sql` is a bare `ADD COLUMN` with no `IF NOT EXISTS`; behavior on a partially-upgraded database was not verified.
- **Third review perspective absent:** one external advisory perspective was attempted but returned no findings; this verdict rests on the primary review plus one adversarial pass. Absence is not treated as agreement.

Jerry-Xin (approved):
# Round-2 Review — Fix Confirmation + Peer Adjudication at 24d0d20a

**Verdict: APPROVE** at `24d0d20a0741cfb274740ca5f5804dacc51a02bd` (tree `9aa18eb740a3751c548bb6d9fc1ab66954e02cb6`, parent `e4c65da534ea`, merge-base `635cb6ad4bd1` unchanged). Delta `e4c65da5..24d0d20a` = 1 commit, 13 files, +213/−40. Every hunk maps 1:1 to a finding in yujiawei's review 5218520958; no unrelated changes ride along. This review supersedes my DISMISSED approval 5218277296 (auto-dismissed by branch protection on the new push — normal mechanics).

## 1. Delta hunk mapping (smuggle audit: clean)

| File | Maps to |
|---|---|
| `internal/citationtext/valid_adjacent.go` (new, 50 lines) | P1-1 fix: prose-safe validator |
| `internal/citationtext/valid_adjacent_test.go` (new) | pins the reviewer's empirical probes as tests |
| `internal/agent/summary_draft.go` :237 gate, :268/:283 canonicalize | P1-1 (draft gate `Valid`→`ValidAdjacent`), P1-2 (`Canonicalize`→`CanonicalizeAdjacent`, both evidence and no-fetch paths; the now-unneeded `maxIndex` derivation is removed with it) |
| `internal/agent/tool_emit_summary_response.go` :287 | P1-1 (emit/preview gate `Valid`→`ValidAdjacent`) |
| `internal/worker/citation.go` :31 | P1-3 (`extractCitationIndexes` skips compound markers) |
| `docs/personal-summary-main-parity.md` :66-74 | Spec/docs divergence: doc now names `CanonicalizeAdjacent` + `ValidAdjacent` and matches the code |
| `internal/api/handler/generation_config.go` :63-75 | P2: comment-only hunk; documents that `req` is a value copy and the persistence guarantee lives in `saveGenerationScope`. `req.Sources = nil` line and the `task.ScheduleID != nil && req.Sources != nil` guard are byte-unchanged — behavior-neutral, reviewed on its own merits |
| 6 test files (summary_draft_test, agent_summary_citations_test, …cgo_test, generation_config_remediation_test, worker citation_test, personal_citation_pipeline_test) | expectation updates tracking the policy change + one new P2 pin test |

## 2. Adjudication of yujiawei CR 5218520958 (at e4c65da5) — every claim CONFIRMED TRUE

Update since this round began: yujiawei has already re-reviewed at 24d0d20a and posted APPROVE 5219503992 (07:02:27Z), independently confirming all three P1 fixes with his own empirical probes; his stale-headed CR below is superseded by his own new vote (his to dispose of). His new review's two non-blocking P2 observations were themselves byte-verified here before adoption (see §7 A-3 cross-reference and §8 note). No claim of his conflicts with any finding below.

Each claim was re-derived from bytes and reproduced with executable Go probes at **both** heads (throwaway test files, ran then removed; results below).

| Claim | Verdict | Probe evidence at e4c65da5 |
|---|---|---|
| P1-1 legitimate Agent drafts hard-rejected, whole turn discarded | **CONFIRMED TRUE** | strict `Valid` returned `false` for `依据 GB/T [50011-2010] 抗震规范。参考 [2]。`, `编号 [7;9] 待确认。依据 [2]。`, `排期 [2026-09] 确认。结论见 [2] [3]。` (window 1..5) despite each carrying real citations. Consequence chain verified: `summaryDraftQualityFailure` → 2 bounded writer repairs → `validatedMergedDraftFallback` (summary_draft.go:214) re-runs the same failing checks → `SummaryDraftError{invalid_citations}` discards the turn. The emit gate (`tool_emit_summary_response.go:287`) shares the defect for previews. |
| P1-2 prose ranges silently corrupted in the persisted body | **CONFIRMED TRUE** | strict `Canonicalize("预算区间 [3-5] 万元。", 1..5, 5)` returned `"预算区间 [3][4][5] 万元。"` — the exact corruption `CanonicalizeAdjacent` exists to prevent, landing in the saved body via `canonicalizeDraftCitations` (summary_draft.go:259 at that head). |
| P1-3 worker harvests prose-range indices into false `model.Citation` rows | **CONFIRMED TRUE** | `extractCitationIndexes("预算区间 [3-5] 万元，无其他引用。")` returned `[3 4 5]`; `BuildCitations` with a 1..5 pool built 3 false rows. `stripOrphanCitations`' single-marker regex cannot remove the producing text, and `CanonicalizeAdjacent` at personal_processor.go:1214 leaves the isolated range byte-identical — the extract/strip disagreement survives to persistence, exactly as described. |
| Docs describe behavior the code doesn't implement (Spec ❌) | **CONFIRMED TRUE** | `docs/personal-summary-main-parity.md:68` at that head claimed "Agent saves, Workflow and refinement reuse the same prose-safe generated-content normalizer" while the Agent draft/emit surfaces were wired to strict `Valid`/`Canonicalize` (summary_draft.go:237/:259, tool_emit_summary_response.go:287). |
| P2 source re-validation skip relies on downstream guard | **CONFIRMED as described** (author chose documentation + a pin test; behavior unchanged) | `validateRegenerationConfigDB` clears a value copy; `saveGenerationScope` (generation_config.go:196-212) re-reads the locked task and forces `writeSources=false` for scheduled tasks. New pin `TestRegenerateScheduledTaskDropsCallerSources` locks the drop-not-write contract. |
| 🟡 unify Agent/Workflow normalizers | **ADOPTED** by the delta — see matrix below. |

### Under-call acknowledgment

My APPROVE 5218277296 at e4c65da5 audited the citation unification on the generation/refine/edit/save consume surfaces and graded "the Agent writer keeps strict `Canonicalize`/`Valid`" as a correct intact salvage boundary — because it is the only surface with a bounded repair loop. That grading was too narrow: the writer boundary's strictness is only safe if every bracketed number reaching it is genuinely citation syntax, and free-form summary bodies falsify that premise (P1-1/P1-2 above, both probe-confirmed). The surface census behind my approval missed the agent-draft/emit consume surface, so the APPROVE at e4c65da5 was an under-call. yujiawei's fresh adversarial pass at identical bytes found what my tree-identity equivalence argument could not: byte-identity with a twice-approved state guarantees no regression against that state, not absence of new findings. Full credit to review 5218520958; it drove this entire delta.

## 3. Fix verification at 24d0d20a

**Policy matrix, surface by surface (code × docs × tests now agree):**

| Surface | Validation | Normalization | Unresolvable/isolated compound |
|---|---|---|---|
| Agent draft gate (summary_draft.go:237) | `ValidAdjacent`, requireMarker | — | isolated → prose (skip); adjacent malformed/unresolved → fail closed into repair loop |
| Agent emit/preview gate (tool_emit_summary_response.go:287) | `ValidAdjacent`, requireMarker | — | same |
| Agent draft canonicalize (summary_draft.go:268, :283) | — | `CanonicalizeAdjacent` (never errors, both evidence and no-fetch paths) | isolated → byte-identical; adjacent fully-authorized → expanded to singles |
| Agent save harvest (agent_summary_citations.go:180 → worker/citation.go:31) | `citationsValid` (unchanged) | content saved byte-identical | compounds skipped by `extractCitationIndexes` → no rows; consistent with `stripOrphanCitations`' single-marker regex |
| Workflow generation (personal_processor.go:1214) | — | `CanonicalizeAdjacent` before `buildCitations` | adjacent authorized clusters expand then harvest (no row loss); isolated stay prose |
| Refine/edit/save normalization (service/edit.go:32, six consume sites) | in-window dangling singles → `ErrInvalid` | `CanonicalizeAdjacent` | unchanged from the twice-verified state |

`ValidAdjacent` keeps both blocker-4 guarantees (every single marker and every cluster member must resolve in the evidence window; requireMarker still demands ≥1 real marker) and adds exactly the `CanonicalizeAdjacent` adjacency rule. Docs (`personal-summary-main-parity.md:66-74`) now name both functions and match the wiring.

**Probe battery at 24d0d20a (throwaway, ran then removed; bidirectional where noted):**
- The reviewer's four prose shapes all pass the gate with a real citation present; prose-only content still fails requireMarker (5 shapes); corrupt adjacent clusters still fail closed (`[1][3-9]`, `[1][2,6]`, `[1][9,]`, `[1][50011-2010]`, `[1][7;9]`); singles stay strict both legs (`[9]` rejected, `[2]` accepted at window 1..5). Each of these shapes was probe-confirmed to behave oppositely (rejected/corrupted) at e4c65da5 where applicable — the two-sided demonstration holds on every leg.
- P1-2: `CanonicalizeAdjacent` preserves `预算区间 [3-5] 万元。` byte-identically; `[1][3-5]` → `[1][3][4][5]`; `[88][9,73]` → `[88][9][73]`; expansion is idempotent; gate↔expander agreement verified (every gate-accepted shape re-passes the gate after canonicalization).
- P1-3: `extractCitationIndexes` on the prose range returns `[]`, `BuildCitations` returns 0 rows; cluster parity holds end-to-end (normalize-then-extract on `见 [1][3-5] 的讨论。` harvests exactly 1,3,4,5 → 4 rows), so the compound skip loses no legitimate cluster rows; mixed prose+citation content yields exactly the one real row with content byte-identical.
- The 22-case citation battery from the prior round re-ran green at this head (all preserve/expand/ErrInvalid/degenerate/zero-citation/big-window cases), as did the 8-case finish-gate battery (`citationsValid`), including the compound-prose save shape and the zero-citation bracketed-prose non-fetched shape. This is expected: `service/edit.go` and `citationtext/citation.go` are byte-identical to the prior head — the rerun confirms no interaction with the new code.

## 4. Regression re-audit — all previously verified fixes byte-intact

`git diff e4c65da5..24d0d20a` touches **none** of the carrying files (stat empty for task.go, schedule_sources.go, personal_processor.go, service/edit.go, handler edit.go, personal_refine.go, agent_summary.go, agent_summary_save.go, agent_summary_citations.go, streaming/, hub.go, scheduler.go). Spot-quoted at this head:
- **Stop-nulling**: no `current_version_id` nil-write anywhere in personal_refine.go; only real-version writes (:204, :454, :680) and the `IS NULL` baseline backfills (:938, :998).
- **Unconditional restore**: personal_processor.go terminal-failure block restores content/citations/pin from `pr.CurrentVersionID` (or newest) **before** the `participantCount <= 1` branch — "must run for BOTH single- and multi-person tasks" comment and code intact.
- **CanonicalizeAdjacent unification**: personal_processor.go:1214 + service/edit.go:32 + six consume sites unchanged (battery above).
- **derived=0 filter**: schedule_sources.go:27 `WHERE task_id = ? AND derived = 0` intact.
- **Sources nil guard**: `saveGenerationScope` locked re-read + `writeSources=false` for scheduled tasks intact (generation_config.go:196-212); the delta hunk in this file is comment-only.
- **Prepare post-commit**: task.go:1516-1522 — Prepare strictly after the transaction closure with the "Do not mutate in-memory stream state until the database transition is committed" rationale intact.

## 5. Mutation battery on the new fix (9 mutants, assertion-red discipline)

Run against the shipped suites of citationtext/worker/service/agent/api-handler plus the probe battery; reverted and tree verified clean after each.

| Mutant | Result | Kill signal |
|---|---|---|
| M2′ `ValidAdjacent` requireMarker neutralized | **KILLED** | shipped `TestValidAdjacentProbes`, `TestPreparedDraftRepairsQualityUsingSameEvidenceAtFinalStep`, `TestEmitSummaryResponseRequiresCitationForChatBackedPreview` + probes |
| M3 adjacent malformed group accepted | **KILLED** | shipped `TestValidAdjacentProbes`, draft repair test (`[1][1,]` shape) + probes |
| M4 adjacent members not validated | **KILLED** | shipped `TestValidAdjacentProbes`, draft repair test (`[1][1,2]`, `[1][1–3]`) + probes |
| M5 extract re-harvests compounds (P1-3 revert) | **KILLED** | 5 shipped tests: `TestExtractCitationIndexes`, `TestWorkflowCompoundCitationFinalization`, `TestPersonalPipelineCitationFinalization`, `TestCreateAgentSummary_WorkspaceCompoundCitationSave`, `TestBuildAgentCompoundCitationsUsesWorkflowBuilder` + probe |
| M6 draft canonicalize reverted to strict (P1-2 revert) | **KILLED** | shipped `TestPreparedDraftCanonicalizesGroupsWithoutAnotherModelCall` |
| M7′ expander ignores adjacency | **KILLED** | red in all 5 packages (`TestCanonicalizeAdjacent`, pipeline isolated-compound, `TestNormalizeGeneratedCitations`, draft canonicalize, both refine transports + cross-surface refine) |
| M8 draft gate polarity flipped | **KILLED** | 6 shipped draft/runner tests |
| M1 draft gate rewired to strict `Valid` (P1-1 revert at the call site) | **SURVIVED** | — |
| M1b emit gate rewired to strict `Valid` | **SURVIVED** | — |

🟡 **Advisory A-5 (new, pinned at 24d0d20a) — accept-side gate wiring is unpinned.** The shipped reject-side tests (`[1][1,2]`-style clusters) behave identically under strict `Valid`, and the only wired draft content with compounds (`Budget [9,73]. …`, window max 95) resolves in-window so strict `Valid` also accepts it; the emit tests carry only single markers. Consequence: rewiring either gate from `ValidAdjacent` back to `Valid` — silently reintroducing P1-1 — keeps the entire suite green. The function itself is pinned (`TestValidAdjacentProbes`), and the regression direction is fail-closed (repair loop), so this is a pin gap, not a live defect. Prescription (either test kills M1/M1b): (a) a draft-path accepted-turn case driving `依据 GB/T [50011-2010] 抗震规范。参考 [1]。` through the real runner at window {1}; (b) an emit-gate case with preview `依据 GB/T [50011-2010]。见 [1]。` expected accepted. Non-blocking.

## 6. Test, CI, and merge evidence

- `go build ./...`, `go vet ./...` clean; `go test -race -count=1 ./...` green at this head: **24 packages, 2832 runs (1653 top-level + 1179 subtests), 0 FAIL**. Reconciles exactly with the prior-head baseline measured the same way in the same environment: e4c65da5 = 2830 runs (1651 + 1179), 0 FAIL — delta +2 = `TestValidAdjacentProbes` + `TestRegenerateScheduledTaskDropsCallerSources`, subtest count unchanged.
- CI at 24d0d20a: Build, Test, Test (race, cgo), Vet, Lint, Generate (no drift), dependency/osv/secret scans all **pass**; `check-sprint` fails in 3s — the known environmental failure, noted, not a blocker.
- MySQL lane carried: the delta contains zero SQL/schema/model bytes (file list above), so the prior round's lane result stands.
- `git merge-tree` vs open #249 (1b728aa8) is **clean in both orders with an identical combined tree** (`1bbae7e22c69`) — order-insensitive by tree identity. Shared files remain personal_processor.go/scheduler.go with disjoint hunks; whichever PR merges second should re-run the worker package suite on the combined tree as standard hygiene.

## 7. Advisories

- 🟡 **Advisory A-1 (carried verbatim from 5218277296; graded and pinned at e4c65da5 — the delta touches no mechanism byte: task.go, streaming/hub.go and the worker scheduler/processor surfaces are all unchanged between e4c65da5 and 24d0d20a)** — post-commit `Prepare` vs the independent worker poller: mechanism is real at hub level (`Prepare` retires whatever run is active, hub.go:53-70; a `start` equal to `retiredRunID` is dropped and non-start events are dropped while `awaitingStart`). For a new run's `start` to precede `Prepare`, the worker poller would need a tick inside the post-commit window and then to complete SELECT + claim UPDATE + reload + pipeline init + HTTP ingest before the handler executes ~10 straight-line statements after the commit ack — the racing path contains strictly more network round-trips than the handler's remaining path, so inversion requires an anomalous multi-ms stall of the handler goroutine at an exact instruction point. Impact is bounded and recoverable: the DB stays canonical and correct, subscribers miss one run's live events until refresh/reconnect, and the hub self-heals on the next `start` (fresh RunID). The alternative placement (Prepare pre-commit) fails deterministically on every commit rollback — the exact defect this fix removed — so the current trade is sound. Non-blocking follow-ups worth taking: a deterministic ordering test for `start(R_new) → Prepare → delta(R_new)`, and generation fencing (have `Prepare` retire only runs older than the requeue's generation token).
- 🟡 **Advisory A-2 (carried verbatim from 5218277296; pinned at e4c65da5, bytes unchanged at 24d0d20a; feature-inherent, same shape since 0dcefce0 and at 249a4c11 — not a delta of this PR)** — editing a multi-person task in Failed/Cancelled state persists the member edit, but `reviveCompletedForRecompute` is a strict no-op there, so the team aggregate never absorbs it. Consider a conflict response for multi-person terminal tasks or documenting the limitation.
- 🟡 **A-3 (new, pinned at 24d0d20a) — `ValidAdjacent` drops the above-window prose exemption for adjacent groups.** Strict `Valid` exempted well-formed compounds wholly above a known window (`outsideEvidenceWindow`) regardless of adjacency; `ValidAdjacent` has no maxIndex input, so an adjacent well-formed above-window group now fails closed: probe-verified `Valid("结论 [1] [900-901]。", 1..5, 5, true) = true` at e4c65da5 vs `ValidAdjacent(same, 1..5, true) = false` at this head, while `CanonicalizeAdjacent` preserves the same content byte-identically (validator/expander disagreement) — and the pre-PR single-marker regex never saw such shapes either, so this is a narrowing vs both the prior head and base. Realistic trigger: a citation immediately followed by a year/page range, e.g. `依据 [2] [2024-2025] 双年度数据` (two-part ranges are not dateShape-skipped). Graded advisory, not blocking: strict adjacency (whitespace-only gap) required; fail-closed with the bounded repair loop and validated merged fallback behind it; no corruption or false rows. If peers judge model output to hit this shape often, an upgrade is warranted — flagging openly. Prescription: give `ValidAdjacent` the window bound (maxIndex parameter or an "all members above window max" closure probe) and restore the exemption for well-formed wholly-above-window groups even when adjacent, matching `CanonicalizeAdjacent`'s preserve behavior.
  **Cross-reference — the mirror corner is documented in yujiawei 5219503992 §2 and was probe-verified here:** an adjacent well-formed *in-window* prose range (`预算区间 [3-5] [2]。`, window 1..5) passes `ValidAdjacent`, is expanded by `CanonicalizeAdjacent` to `[3][4][5] [2]`, and the save path harvests phantom rows for 3/4/5; and `依据 GB/T [50011-2010] [2]。` fails the gate as a malformed in-cluster group. Both reproductions confirmed. That direction is fail-open but strictly parity with the shipped Workflow normalizer (#248/#250) and strictly narrower than the pre-PR Agent path (which expanded every resolvable compound regardless of adjacency), so it is not a regression of this delta — his non-blocking grading is correct. Together the two corners bracket the adjacency heuristic: A-3 (above-window adjacent → validator rejects / expander preserves) and his (in-window adjacent prose → validator accepts / expander expands). One caution on his suggested follow-up (tighten `adjacentToMarker` to zero separating characters): the shipped `CanonicalizeAdjacent` doc comment defines clusters to include the whitespace-gap shape `"[1] [3-5]"`, and probe-verified `见 [1] [3-5] 讨论。` → `见 [1][3][4][5] 讨论。` today — a zero-gap rule would stop expanding that documented genuine-cluster shape, so the tightening is a real cross-cutting design change (validator + expander + docs together), best tracked in its own PR as he suggests.
- 🟡 **A-4 (new, pinned at 24d0d20a) — stale docs bullet contradicts the shipped Prepare placement.** `docs/personal-summary-main-parity.md` (Stream preparation bullet, ~:136-140) still says "Prepare stays **before** commit to prevent worker/poller races. A rare failed commit may invalidate only the in-memory stream" — that described the pre-fix placement at the PR's first head; the parity commit deliberately moved Prepare **after** commit (task.go:1516: "Do not mutate in-memory stream state until the database transition is committed"), and the doc bullet was not updated in that wave. Docs-only defect, no runtime effect; note the code direction is the correct one (see A-1: pre-commit Prepare fails deterministically on rollback), so the fix is to rewrite the bullet to describe post-commit placement + the bounded race window + the "generation-fencing redesign is separate work" note already present. This doc line is also what the automated pass cited as a "documented requirement" contradiction (see §8).
- 🟡 **A-5** — gate-wiring pin gap, with prescriptions, in §5.
- 🟡 **Carried from the prior round (unchanged)**: the zero-citation-gate removal on the save path remains unpinned by shipped tests (mutant re-adding a strict zero-citation gate survived there; this delta's handler-test updates all build ≥1 citation, so the gap persists). Prescription unchanged: one save-path test — zero built citations + content `规划[2024-2025]` → 200, content preserved. The in-window single-hole strictness (prior A-7) also stands as recorded: it is the strictest point of the unified policy, narrow, fail-closed, and was explicitly requested in review.

## 8. Automated-pass disclosure

The raw automated verdict at this head was REQUEST_CHANGES with a single Critical: the post-commit `Prepare` vs worker-poller race, citing the docs bullet as a contradicted "documented requirement". Overridden to APPROVE on two legs: (1) the mechanism leg is A-1, adjudicated advisory in the prior round at byte-identical code — the full reachability/boundedness/self-healing argument and the deterministic-failure counter-argument against pre-commit placement stand unchanged, and no delta byte touches it; (2) the docs leg is TRUE but is a stale-docs defect (A-4) whose correct resolution is a docs rewrite, not moving the code back — moving Prepare pre-commit would reintroduce the exact rollback defect the parity commit removed. Raw output archived per process.

Separately, yujiawei's second non-blocking observation (raw `\[(\d+)\]` regex in `dedupCitations`/`stripOrphanCitations` is not `Scan`-protected against inline-code/link brackets) is confirmed pre-existing base tech debt on the Workflow cleanup path, untouched by this diff and not on the Agent save harvest path — agreed: separate ticket, not a change request here.

## 9. Merge mechanics and next votes

- Merge order vs #249: either order, combined tree identical (`1bbae7e22c69`); second merger re-runs the worker suite. Deployment order per the PR description: back up and apply `migrations/sql/20260911-01-summary-generation-requirement.sql` → deploy backend → deploy companion frontend (octo-web #1686).
- Vote state at posting time: yujiawei APPROVE 5219503992 at this head (his e4c65da5 CR superseded by his own re-review); this APPROVE is the second confirming vote with the fix-verification, mutation, and advisory record above. Branch protection still reports REVIEW_REQUIRED, so further required approvals/code-owner reviews may remain outstanding.
- **mochashanyao**: you haven't voted on #251 yet — a first pass at 24d0d20a is welcome, especially on the A-3 adjacency corners (§7), where an independent read on trigger frequency would help grade them.

